PR #805, opened 2024-11-11
feat: (halo2) SNARK verifier

MonkeyKing-1:
Guest library for verifying halo2 proofs.

Integration test currently failing because:
- [x] ~~`snark-verifier` requires std library and in particular uses `std::collections::HashMap`. We need to implement `sys_rand` with a phantom instruction passing to host randomness (INT-2955)~~ Change `snark-verifier` to use `BTreeMap`. https://github.com/axiom-crypto/snark-verifier/pull/68


lispc:
i started to have a try of this branch. My program now can build and run till a final PairingCheckErr.

Things i fixed:

1. i find the halo2curves-axiom and snark-verifier::halo2curves-axiom is a version management nightmare, so in my local env, i unify to use same version.
2. snark-verifier/src/verifier/plonk/proof.rs uses HashMap, which is not support inside zkvm. So i switched to BTreeMap.
3. this PR lacks this trait, i added.

```


impl<const LIMBS: usize, const BITS: usize> AccumulatorEncoding<G1Affine, OpenVmLoader>
for LimbsEncoding<LIMBS, BITS>
{
type Accumulator = KzgAccumulator<G1Affine, OpenVmLoader>;

fn from_repr(limbs: &[&OpenVmScalar<Halo2Fr, Fr>]) -> Result<Self::Accumulator, Error> {
```

Next
1. i am going to compare a NativeLoader verification vs this PR, side by side, to check what is going wrong.

Other thoughts:
1. i feel the current API, Fp::from_bytes() will not panic for overflowed inputs.  It will pass along the wrong data a long way till err.   Not easy to debug.

extensions/verifier/src/verifier/loader.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Clone, Debug)]
+pub struct OpenVmLoader;
+
+impl EcPointLoader<G1Affine> for OpenVmLoader {
+    type LoadedEcPoint = OpenVmEcPoint<G1Affine, EcPoint>;
+
+    fn ec_point_load_const(&self, value: &G1Affine) -> Self::LoadedEcPoint {
+        let point = EcPoint {
+            x: Fp::from_be_bytes(&value.x().to_bytes()),
+            y: Fp::from_be_bytes(&value.y().to_bytes()),
```
lispc, line 40: should be from_le_bytes

extensions/verifier/src/verifier/loader.rs
```diff
@@ -0,0 +1,150 @@
...
+        }
+        OpenVmEcPoint(msm::<EcPoint, Fr>(&scalars, &base), PhantomData)
+    }
+}
+
+impl ScalarLoader<Halo2Fr> for OpenVmLoader {
+    type LoadedScalar = OpenVmScalar<Halo2Fr, Fr>;
+
+    fn load_const(&self, value: &Halo2Fr) -> Self::LoadedScalar {
+        let value = Fr::from_be_bytes(&value.to_bytes());
```
lispc, line 77: should be from_le_bytes

extensions/verifier/src/verifier/loader.rs
```diff
@@ -0,0 +1,150 @@
...
+            .then_some(())
+            .unwrap_or_else(|| panic!("{:?}", Error::AssertionFailure(annotation.to_string())))
+    }
+}
+
+impl ScalarLoader<Halo2Fp> for OpenVmLoader {
+    type LoadedScalar = OpenVmScalar<Halo2Fp, Fp>;
+
+    fn load_const(&self, value: &Halo2Fp) -> Self::LoadedScalar {
+        let value = Fp::from_be_bytes(&value.to_bytes());
```
lispc, line 92: should be from_le_bytes

extensions/verifier/src/verifier/transcript.rs
```diff
@@ -0,0 +1,119 @@
...
+        Ok(scalar)
+    }
+
+    fn read_ec_point(&mut self) -> Result<OpenVmEcPoint<G1Affine, EcPoint>, Error> {
+        let [mut x, mut y] = [[0; 32]; 2];
+        for repr in [&mut x, &mut y] {
+            self.stream
+                .read_exact(repr.as_mut())
+                .map_err(|err| Error::Transcript(err.kind(), err.to_string()))?;
+            repr.as_mut().reverse();
```
lispc, line 111: should not reverse

jonathanpwang:
I think there was some nightmare with `halo2curves-axiom`, but in current `main` I tried to make everything use the same crates.io version (I hope). I'll look into the other issues, thanks!

extensions/verifier/src/verifier/transcript.rs
```diff
@@ -0,0 +1,119 @@
...
+            .cloned()
+            .chain(if self.buf.len() == 0x20 {
+                Some(1)
+            } else {
+                None
+            })
+            .collect_vec();
+        let hash = keccak256(&data);
+        self.buf = hash.to_vec();
+        OpenVmScalar(Fr::from_be_bytes(&hash), PhantomData)
```
lispc, line 55: currently the openvm executor can accept any bytes32 (and do computation correctly).  But for prover, the Fr::from_be_bytes input must be within Fr range.   

Need `let value = value % modulus::<F>();`

extensions/verifier/src/verifier/loader.rs
```diff
@@ -0,0 +1,150 @@
...
+            .unwrap_or_else(|| panic!("{:?}", Error::AssertionFailure(annotation.to_string())))
+    }
+}
+
+impl Loader<G1Affine> for OpenVmLoader {}
+
+/// KZG accumulation scheme. The second generic `MOS` stands for different kind
+/// of multi-open scheme.
+#[derive(Clone, Debug)]
+pub struct OpenVmKzgAs<M, MOS>(PhantomData<(M, MOS)>);
```
lispc, line 108: `OpenVmKzgAs` is not used? KzgAs with OpenVmLoader will be enough.

lispc:
some high level thoughts:

the plonk verification can be generally split into 2 parts: (1) msm and pairing, heavy computation, may >90% of all "cycles" needed (2) light weight computation, like fr ops and other control flow glue codes.   Maybe just "patch" msm and pairing, we can get most of the speedup.

Though the PR works well, i feel there are a lot of "wrapper" codes in trait.rs, also in other files, the codes are quite similar to NativeLoader.   I don't have an alternative good solution now, just feel this design may not be easy to maintain long term, it has to be changed with snark-verfier::EvmTranscript every time, like endianness and how squeeze_challenge is computated.

jonathanpwang:
@lispc previous tests were only on host, not targeting zkvm. I added integration test targeting zkvm and discovered that snark-verifier uses std library HashMap somewhere. This hashmap thing is annoying enough I will try to add the `sys_rand` support tomorrow

lispc:
> @lispc previous tests were only on host, not targeting zkvm. I added integration test targeting zkvm and discovered that snark-verifier uses std library HashMap somewhere. This hashmap thing is annoying enough I will try to add the `sys_rand` support tomorrow

oh actually I tested in guest. I change hashmap to breed map locally.  I remember it is a "evaluation.rs"?

jonathanpwang:
> > @lispc previous tests were only on host, not targeting zkvm. I added integration test targeting zkvm and discovered that snark-verifier uses std library HashMap somewhere. This hashmap thing is annoying enough I will try to add the `sys_rand` support tomorrow
> 
> oh actually I tested in guest. （and even proved it "recursively") I changed hashmap to btree map locally. I remember it is a "evaluation.rs"?

Ah, maybe it's easier if I just update in `snark-verifier` then... thanks

PR #1101, opened 2024-12-16
perf: Use u32 for `address_space` and `address` internally

zlangley:


crates/toolchain/transpiler/src/util.rs
```diff
@@ -166,10 +166,7 @@ pub fn elf_memory_image_to_openvm_memory_image<F: PrimeField32>(
     let mut result = MemoryImage::new();
     for (addr, word) in memory_image {
         for (i, byte) in word.to_le_bytes().into_iter().enumerate() {
-            result.insert(
-                (F::TWO, F::from_canonical_u32(addr + i as u32)),
-                F::from_canonical_u8(byte),
-            );
+            result.insert((2, addr + i as u32), F::from_canonical_u8(byte));
```
jonathanpwang, line 169: I think we have a constant for memory address space somewhere

zlangley: There is something in `extensions::native::compiler::src::conversion`. We should probably create something analogous here. (In fact, the one there shouldn't have Register, as that is not really a concept for the eDSL transpiler.)

jonathanpwang (approved):
LGTM

PR #1102, opened 2024-12-16
feat: modularize toolchain tests

arayikhalatyan:
Resolves INT-2740

Moved the tests in `./crates/toolchain/tests` into respective `test` crates inside the corresponding `extensions`

### DIFF
--- a/.github/workflows/recursion.yml
+++ b/.github/workflows/recursion.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@nightly
+      - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
         with:
           cache-on-failure: true
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@nightly
+      - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
         with:
           cache-on-failure: true
@@ -35,9 +35,23 @@ jobs:
       - name: Install solc # svm should support arm64 linux
         run: (hash svm 2>/dev/null || cargo install --version 0.2.23 svm-rs) && svm install 0.8.19 && solc --version
 
-      - name: Install tools
+      - name: Install architecture specific tools
         run: |
-          rustup component add rust-src --toolchain nightly-2024-10-30
+          arch=$(uname -m)
+          case $arch in
+            arm64|aarch64)
+              S5CMD_BIN="s5cmd_2.2.2_linux_arm64.deb"
+              rustup component add rust-src --toolchain nightly-2024-10-30-aarch64-unknown-linux-gnu
+              ;;
+            x86_64|amd64)
+              S5CMD_BIN="s5cmd_2.2.2_linux_amd64.deb"
+              rustup component add rust-src --toolchain nightly-2024-10-30-x86_64-unknown-linux-gnu
+              ;;
+            *)
+              echo "Unsupported architecture: $arch"
+              exit 1
+              ;;
+          esac
 
           echo "Checking s5cmd"
           if type s5cmd &>/dev/null; then
--- a/ci/scripts/utils.sh
+++ b/ci/scripts/utils.sh
@@ -86,7 +86,21 @@ commit_and_push_benchmark_results() {
 }
 
 install_s5cmd() {
-    rustup component add rust-src --toolchain nightly-2024-10-30
+    arch=$(uname -m)
+    case $arch in
+    arm64|aarch64)
+        rustup component add rust-src --toolchain nightly-2024-10-30-aarch64-unknown-linux-gnu
+        S5CMD_BIN="s5cmd_2.2.2_linux_arm64.deb"
+        ;;
+    x86_64|amd64)
+        rustup component add rust-src --toolchain nightly-2024-10-30-x86_64-unknown-linux-gnu
+        S5CMD_BIN="s5cmd_2.2.2_linux_amd64.deb"
+        ;;
+    *)
+        echo "Unsupported architecture: $arch"
+        exit 1
+        ;;
+    esac
 
     echo "Checking s5cmd"
     if type s5cmd &>/dev/null; then


jonathanpwang (commented):
We should update the layout.md doc too

crates/toolchain/build/src/lib.rs
```diff
@@ -25,7 +25,17 @@ const RUSTUP_TOOLCHAIN_NAME: &str = "nightly-2024-10-30";
 /// Returns the given cargo Package from the metadata in the Cargo.toml manifest
 /// within the provided `manifest_dir`.
 pub fn get_package(manifest_dir: impl AsRef<Path>) -> Package {
+    println!(
```
jonathanpwang, line 28: eprintln

crates/toolchain/build/src/lib.rs
```diff
@@ -25,7 +25,17 @@ const RUSTUP_TOOLCHAIN_NAME: &str = "nightly-2024-10-30";
...
+        "manifest_dir: {:?}",
+        manifest_dir.as_ref().join("Cargo.toml")
+    );
     let manifest_path = fs::canonicalize(manifest_dir.as_ref().join("Cargo.toml")).unwrap();
+    println!(
```
jonathanpwang, line 33: eprintln

crates/toolchain/build/src/lib.rs
jonathanpwang: this is actually the manifest path

crates/toolchain/build/src/lib.rs
jonathanpwang: maybe just delete this if its only for debugging

crates/toolchain/tests/src/integration-test.md
jonathanpwang, line 1: This is nice! Maybe we should move it directly to the contributor `docs/`. I think `docs/crates` makes sense.

crates/toolchain/tests/src/lib.rs
```diff
@@ -1,10 +1,80 @@
...
+
+pub fn build_example_program_at_path_with_features<S: AsRef<str>>(
+    manifest_dir: PathBuf,
+    example_name: &str,
+    features: impl IntoIterator<Item = S>,
+) -> Result<Elf> {
+    let pkg = get_package(manifest_dir);
+    let target_dir = tempdir()?;
+    // Build guest with default features
+    let guest_opts = GuestOptions::default()
```
jonathanpwang, line 55: I think now some combination of https://github.com/openvm-org/openvm/blob/3ca85e83a2ab9e6a5fadfcb6f5047e171a817981/crates/toolchain/build/src/lib.rs#L352 and https://github.com/openvm-org/openvm/blob/3ca85e83a2ab9e6a5fadfcb6f5047e171a817981/crates/toolchain/build/src/lib.rs#L352 should be able to do this with less code duplication

extensions/algebra/tests/programs/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[workspace]
+[package]
+name = "openvm-algebra-integration-test-programs"
```
jonathanpwang, line 3: for the sake of some brevity, let's just call it `openvm-algebra-test-programs`

extensions/bigint/tests/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+        let openvm_exe = VmExe::from_elf(
+            elf,
+            Transpiler::<F>::default()
+                .with_extension(Rv32ITranspilerExtension)
+                .with_extension(Rv32MTranspilerExtension)
+                .with_extension(Rv32IoTranspilerExtension)
+                .with_extension(Int256TranspilerExtension),
+        )?;
+        let config = Int256Rv32Config::default();
+        let executor = VmExecutor::<F, _>::new(config);
```
jonathanpwang, line 29: for uniformity, switch to `air_test`

extensions/bigint/tests/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+        executor.execute(openvm_exe, vec![])?;
```
jonathanpwang, line 47: ditto

extensions/ecc/tests/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+            .concat()
+            .into_iter()
+            .map(AbstractField::from_canonical_u8)
+            .collect();
+        new_air_test_with_min_segments(config, openvm_exe, vec![coords], 1, false);
+        Ok(())
+    }
+
+    #[derive(Clone, Debug, VmConfig, Serialize, Deserialize)]
+    pub struct Rv32ModularKeccak256Config {
```
jonathanpwang, line 91: should we delete this and just use the `SdkVmConfig`?

extensions/keccak256/tests/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+        let openvm_exe = VmExe::from_elf(
+            elf,
+            Transpiler::<F>::default()
+                .with_extension(Keccak256TranspilerExtension)
+                .with_extension(Rv32ITranspilerExtension)
+                .with_extension(Rv32MTranspilerExtension)
+                .with_extension(Rv32IoTranspilerExtension),
+        )?;
+        let executor = VmExecutor::<F, Keccak256Rv32Config>::new(Keccak256Rv32Config::default());
+        executor.execute(openvm_exe, vec![])?;
```
jonathanpwang, line 29: air test

jonathanpwang (changes requested):
Overall refactor and modularization looks good. I had some inline comments but also we should:

1. Add the `cargo nextest` to each of the respective github workflows (e.g., ecc.yml), right now the tests won't actually be run.
2. Can we add for each of the examples linked from the book, the correct `openvm.toml` to the test programs dir?
3. After (2), test locally that the cli works, then let's add tests to cli.yml where we run the cli from the test guest program dir exactly as one should do it from the book. This way CI will check that the book is up to date without us having to manually check.

.github/workflows/algebra-extension.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches: ["**"]
     paths:
       - "extensions/algebra/**"
+      - "extensions/algebra/tests/**"
```
jonathanpwang, line 10: this is covered by the previous `/**`

.github/workflows/cli.yml
```diff
@@ -23,13 +24,13 @@ env:
...
     runs-on:
-      - runs-on
-      - run-id=${{ github.run_id }}
-      - family=m7
+      - runs-on=${{ github.run_id }}
+      - runner=32cpu-linux-arm64
 
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
```
jonathanpwang, line 32: delete

.github/workflows/cli.yml
```diff
@@ -57,6 +58,23 @@ jobs:
         run: |
           bash ./extensions/native/recursion/trusted_setup_s3.sh
```
jonathanpwang, line 60: delete the Setup halo2 because we don't use it

.github/workflows/algebra-extension.yml
```diff
@@ -34,3 +35,9 @@ jobs:
         working-directory: extensions/algebra/circuit
         run: |
           cargo nextest run --cargo-profile=fast
+
+      - name: Run algebra integration tests
+        working-directory: extensions/algebra/tests
+        run: |
+          rustup component add rust-src --toolchain nightly-2024-10-30-aarch64-unknown-linux-gnu
```
jonathanpwang, line 42:
```suggestion
          rustup component add rust-src --toolchain nightly-2024-10-30
```
apparently just this works and is not architecture specific

jonathanpwang (approved):
LGTM except some minor CI cleanup

PR #1107, opened 2024-12-17
[feat] support weiestrass curve, add p256

luffykai:
fix INT-2594
fix INT-2583

- Tested with proving. Since initially proving failed and execution worked, turning on always prove for this test
- There are two issues:
- 1. double now has "constant" factors and empty row doesn't satisfy -> we need to finalize the double chip
- 2. Finalizing needs a small update

Added explanation in README as it's complicated and I had to rethink everytime... maybe we should revisit if there is a simpler way later

- [x] constraints added for reading `a` from `p1.y` in setup
- [x] `sw_init!` macro needs to be updated to be able to access `CURVE_A`

Later:
- refactor mod-builder to setup doesn't require so much casework


extensions/ecc/guest/src/p256.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl IntrinsicCurve for P256 {
+    type Scalar = P256Scalar;
+    type Point = P256Point;
+
+    fn msm(coeffs: &[Self::Scalar], bases: &[Self::Point]) -> Self::Point
+    where
+        for<'a> &'a Self::Point: Add<&'a Self::Point, Output = Self::Point>,
+    {
+        // heuristic
```
luffykai, line 72: rm?

crates/toolchain/tests/src/ecc_tests.rs
```diff
@@ -130,6 +130,29 @@ fn test_ec_runtime() -> Result<()> {
...
+        elf,
+        Transpiler::<F>::default()
+            .with_extension(Rv32ITranspilerExtension)
+            .with_extension(Rv32MTranspilerExtension)
+            .with_extension(Rv32IoTranspilerExtension)
+            .with_extension(EccTranspilerExtension)
+            .with_extension(ModularTranspilerExtension),
+    )?;
+    use openvm_algebra_guest::IntMod;
+    let p = openvm_ecc_guest::p256::P256_MODULUS.clone();
```
luffykai, line 146: rm

yi-sun:
@luffykai -- this PR should also update the book accordingly

crates/circuits/mod-builder/src/README.md
```diff
@@ -1,4 +1,35 @@
-### Example usage
+## Notes on flags and setup
```
luffykai, line 1: https://linear.app/intrinsictech/issue/INT-2907/simplyunify-the-expr-builder-to-all-have-setup

extensions/ecc/circuit/Cargo.toml
```diff
@@ -39,4 +39,9 @@ openvm-rv32-adapters = { workspace = true, features = ["test-utils"] }
 lazy_static = { workspace = true }
 
 [target.'cfg(not(target_os = "zkvm"))'.dependencies]
-openvm-ecc-guest = { workspace = true, features = ["halo2curves", "k256"] }
+openvm-ecc-guest = { workspace = true, features = [
+    "halo2curves",
+    "k256",
```
jonathanpwang, line 44: actually I don't think it needs either `k256` or `p256`, it only needed `halo2curves`

extensions/ecc/circuit/src/weierstrass_extension.rs
```diff
@@ -49,6 +53,13 @@ pub static SECP256K1_CONFIG: Lazy<CurveConfig> = Lazy::new(|| CurveConfig {
     b: BigUint::from_u8(7u8).unwrap(),
 });
 
+pub static P256_CONFIG: Lazy<CurveConfig> = Lazy::new(|| CurveConfig {
```
jonathanpwang, line 56: huh actually its a little weird these were here instead of in their modules. if it's minimal refactor, move into `p256`

jonathanpwang: (actually where are we using either of these configs at all?)

extensions/ecc/guest/src/p256.rs
```diff
@@ -0,0 +1,69 @@
...
+            "b01cbd1c01e58065711814b583f061e9d431cca994cea1313449bf97c840ae0a"
+        )),
+    };
+}
+
+impl IntrinsicCurve for P256 {
+    type Scalar = P256Scalar;
+    type Point = P256Point;
+
+    // TODO: msm optimization if needed
```
jonathanpwang, line 68: p256 has prime order, so you can copy the same implementation stuff from k256

jonathanpwang (changes requested):
`SETUP_SW_DOUBLE` now needs to read `CURVE_A` from memory and constrain it for security.

Something to watch out for in the chip: since we only read two field elements from memory, we'll need to use the "y" coordinate to read `CURVE_A`. However the other constraints currently don't allow y == 0 (mod p), whereas we can have `a == 0`, so the constraints might need to be adjusted.

cc @Golovanov399

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -123,11 +123,8 @@ where
         let writes: Vec<AB::Expr> = self
             .output_indices()
             .iter()
-            .map(|&i| vars[i].clone())
-            .collect::<Vec<_>>()
-            .concat()
-            .iter()
-            .map(|x| (*x).into())
+            .flat_map(|&i| vars[i].clone())
```
luffykai, line 126: nice... is there tip for this kind of optimization or just experience? I think even with AI I still got to more verbose solutions a lot of time.

crates/circuits/mod-builder/src/README.md
```diff
@@ -1,4 +1,35 @@
...
+3. No chip is in the third type right now.
+4. For the fourth type, there is no setup needed and no flags for selecting operations. Only `is_valid` is needed.
+
+### Finalizing
+
+The STARK backend requires the trace height to be a power of 2. Usually we pad the trace with empty (all 0) rows to make the height a power of 2. Since `is_valid` is 0 for padded rows, the constraints including interaction with memory are satisfied.
+However, there are some cases that all-0 row doesn't satisfy the constraints: when the computation involves non-zero constant values:
+
+- Some pairing chips involves adding a small constant value (like `1`). But since pairing chips don't have any flags, we will pad the trace with the last valid row and set `is_valid` to 0.
+- EcDouble for Weierstrass curve (not short Weierstrass ones): `y^2 = x^3 + ax + b` where `a != 0`. Last valid row with `is_valid = 0` won't work as in that case `setup = is_valid - sum(all_flags) = 0 - 1 = -1` is not a bool (0/1). So additionally we need to set the flags to 0.
```
luffykai, line 30: update: `y^2 = x^3 + ax + b` is "short weierstrass"

extensions/ecc/circuit/src/weierstrass_chip/double.rs
```diff
@@ -1,27 +1,265 @@
...
+        offset: usize,
+    ) -> Self {
+        config.check_valid();
+        let builder = ExprBuilder::new(config, range_bus.range_max_bits);
+        let builder = Rc::new(RefCell::new(builder));
+
+        let mut x1 = ExprBuilder::new_input(builder.clone());
+        let mut y1 = ExprBuilder::new_input(builder.clone());
+        let a = ExprBuilder::new_const(builder.clone(), a_biguint.clone());
+        let is_double_flag = builder.borrow_mut().new_flag();
```
luffykai, line 52: i think even when it's setup, it still won't be 0 unless i misread it 🤔 so i think this is unnecessary. will check

luffykai: nvm i think we need it when setup is actually using a

crates/circuits/mod-builder/src/README.md
luffykai, line 30: i think this is just wrong, we can't set the flags to 0 arbitrarily as the constraint won'y satisfy. but we don't this in the corechip now that double is not using it

crates/circuits/mod-builder/src/core_chip.rs
jonathanpwang: I wouldn't call it an optimization heh. Just familiarity with the permutation of functions available

crates/circuits/mod-builder/src/README.md
jonathanpwang: good catch, please update

extensions/ecc/circuit/src/weierstrass_extension.rs
luffykai: it's here so guest doesn't dependent on circuit? The extension and `CurveConfig` are in circuit crate

extensions/ecc/sw-setup/src/lib.rs
```diff
@@ -62,7 +71,7 @@ pub fn sw_declare(input: TokenStream) -> TokenStream {
                     &format!(
                         "{}_{}",
                         stringify!($name),
-                        intmod_type
+                        struct_path
```
luffykai, line 74: the extern function name is changed from using coord type to point type. I think it's fine

jonathanpwang:
can you resolve merge conflict?

luffykai:
> can you resolve merge conflict?

done

jonathanpwang (approved):
LGTM

PR #1113, opened 2024-12-17
fix: Incorrect lazy initialization of memory data structure

zlangley:


crates/vm/src/system/memory/manager/memory.rs
```diff
@@ -310,15 +310,7 @@ impl<F: PrimeField32> Memory<F> {
             .block_data
             .get(&(address_space, pointer))
             .copied()
-            .unwrap_or_else(|| {
-                for i in 0..size {
```
zlangley, line 314: This was the main bug. If `size` is large (e.g., 32), then there might not be a block at `address`, but it's wrong to overwrite blocks at, say, `address + 24`. It seems better to just only write blocks as needed, which is what this PR aims to do.

luffykai (approved):

PR #1120, opened 2024-12-19
book: sdk usage correction

alv-around:
Hi, 

I was playing with the vm, and following the book, I realized that the fix 4250ce882f1290902de163efd279330c87ba163f did not updated the documentation for the sdk.

I leave here the repo to the sdk working example: https://github.com/openvm-org/openvm-example-fibonacci


jonathanpwang:
Thanks for this catch! Leaving @Golovanov399 to review

jonathanpwang (approved):

PR #1121, opened 2024-12-20
eDSL Kernel Macro Prototype

TlatoaniHJ:
[Discussion](https://docs.google.com/document/d/1_NMUff1OcJK8uf4RNgPs1xv55thyTMZ4OadqY2SFVrs/edit?usp=sharing)

INT-2278


crates/toolchain/edsl-deserialize/src/castf_extension.rs
```diff
@@ -0,0 +1,56 @@
...
+};
+use openvm_native_compiler::CastfOpcode;
+use p3_field::PrimeField32;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Copy, Debug, Default, Serialize, Deserialize)]
+pub struct CastF;
+
+#[derive(ChipUsageGetter, Chip, InstructionExecutor, From, AnyEnum)]
+pub enum CastFExecutor<F: PrimeField32> {
```
jonathanpwang, line 19: structure of this crate is weird: `CastF` should be in `openvm-native-circuit`. 

This crate should be renamed `openvm-native-transpiler`

crates/toolchain/edsl-deserialize/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+mod castf_extension;
+pub mod config;
+pub mod deserialize_instruction;
+
+const LONG_FORM_INSTRUCTION_INDICATOR: u32 = (1 << 31) + 115115115;
```
jonathanpwang, line 5: This should be some RISC-V instruction with a 6-bit opcode that's one of the designated custom opcodes

crates/toolchain/edsl-deserialize/src/lib.rs
```diff
@@ -0,0 +1,8 @@
...
+const GAP_INDICATOR: u32 = (1 << 31) + 113113113;
```
jonathanpwang, line 6: same as above

crates/toolchain/edsl-kernel/Cargo.toml
jonathanpwang, line 1: this should be moved to `extensions/native/guest`, I think we can call the crate `openvm-native-guest` since kernel functions are intended to be called from guest programs

crates/toolchain/edsl-kernel/src/generate_edsl.rs
```diff
@@ -0,0 +1,98 @@
...
+";
+
+pub const CONFIG: &str = "
+[package]
+name = \"root\"
+version = \"0.1.0\"
+edition = \"2021\"
+
+[dependencies]
+openvm-native-compiler = { path = \"../../../../extensions/native/compiler\" }
```
jonathanpwang, line 39: we need a way to have dependency versions consistent with the workspace without hardcoding

jonathanpwang: harcoding path is also not good

crates/toolchain/edsl-kernel/src/generate_edsl.rs
```diff
@@ -0,0 +1,98 @@
...
+    conversion::{convert_program, CompilerOptions},
+    ir::{Array, Ext, Felt},
+};
+use p3_baby_bear::BabyBear;
+use p3_field::{extension::BinomialExtensionField, AbstractField};
+
+type F = BabyBear;
+type EF = BinomialExtensionField<BabyBear, 4>;
+
+<===>
```
jonathanpwang, line 17: use `format!` to include the `OUTSIDE_PLACEHOLDER` by constant name

crates/toolchain/edsl-kernel/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+mod execute;
+mod generate_edsl;
+mod parse_compiler_output;
+mod parse_kernel;
+mod transportation;
+
+const IMMEDIATE_ADDRESS_SPACE: usize = 0;
+const RUST_REGISTER_ADDRESS_SPACE: usize = 1;
+const KERNEL_ADDRESS_SPACE: usize = 5;
+const LONG_FORM_INSTRUCTION_INDICATOR: u32 = (1 << 31) + 115115115;
```
jonathanpwang, line 14: these should be defined in a common crate that can be imported by both this crate and the transpiler crate

crates/toolchain/edsl-kernel/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+use crate::parse_compiler_output::CompiledKernel;
+
+mod execute;
+mod generate_edsl;
+mod parse_compiler_output;
+mod parse_kernel;
+mod transportation;
+
+const IMMEDIATE_ADDRESS_SPACE: usize = 0;
+const RUST_REGISTER_ADDRESS_SPACE: usize = 1;
```
jonathanpwang, line 12: we already have this constant in `openvm-instructions`

crates/toolchain/edsl-kernel/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+const KERNEL_ADDRESS_SPACE: usize = 5;
```
jonathanpwang, line 13: this should be moved to `openvm-instructions`

crates/toolchain/edsl-kernel/src/generate_edsl.rs
jonathanpwang, line 1: General thoughts:
- This is actually a bit messy and potentially shows the benefits of having a DSL instead of eDSL (then you can just take in a text file and compile it)
- For now I think we should not include the generation of the serialized openvm assembly file in the proc-macro, and just let the macro load it from some pre-existing file

crates/toolchain/edsl-kernel/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+const LONG_FORM_INSTRUCTION_INDICATOR: u32 = (1 << 31) + 115115115;
+const GAP_INDICATOR: u32 = (1 << 31) + 113113113;
+const UTILITY_CELL: usize = 0;
+const VARIABLE_REGISTER_INDICATOR: u32 = (1 << 31) + 116;
+const PC_STEP: usize = 4;
+const REGISTER_LIMBS: usize = 4;
+const REGISTER_LIMB_SIZE: usize = 256;
+
+#[proc_macro]
+pub fn edsl_kernel(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
jonathanpwang, line 23: `native_kernel` is a better name

crates/toolchain/edsl-kernel/src/transportation.rs
```diff
@@ -0,0 +1,309 @@
...
+    result.push_str(&asm_call);
+    result.push('\t');
+    result.push_str(&return_name);
+    result.push('\n');
+    result.push_str("}\n");
+
+    result
+}
+
+fn u32_to_directive(x: u32) -> String {
```
jonathanpwang, line 131: unnecessary, `.insn` supports a direct 32-bit version

crates/toolchain/edsl-kernel/src/transportation.rs
jonathanpwang, line 1: directioinally correct, but we should use `quote!` and `paste!` whenever possible. only the raw `asm!` block should be generated using string manipulation

crates/toolchain/transpiler/src/extension.rs
```diff
@@ -10,5 +10,33 @@ pub trait TranspilerExtension<F> {
     /// the next contiguous section of RISC-V instructions into an [`Instruction`].
     /// It returns `None` if it cannot transpile. Otherwise it returns `(instruction, how_many_u32s)` to indicate that
     /// `instruction_stream[..how_many_u32s]` should be transpiled into `instruction`.
```
jonathanpwang, line 12: update comment

crates/toolchain/transpiler/src/extension.rs
```diff
@@ -10,5 +10,33 @@ pub trait TranspilerExtension<F> {
...
-    fn process_custom(&self, instruction_stream: &[u32]) -> Option<(Instruction<F>, usize)>;
+    fn process_custom(&self, instruction_stream: &[u32]) -> Option<TranspilerOutput<F>>;
```
jonathanpwang, line 13: noting the return type is a breaking change, but is necessary because we want one `Option<Instruction<F>>` per 32-bit chunk

benchmarks/programs/test_kernel/macro_output.txt
jonathanpwang, line 1: I would prefer if we just had a file with 32-bits per line, and then the macro loaded that file and converted into `asm!` block

crates/toolchain/transpiler/src/extension.rs
```diff
@@ -10,5 +10,33 @@ pub trait TranspilerExtension<F> {
...
+}
+
+pub struct TranspilerOutput<F> {
+    pub instructions: Vec<Option<Instruction<F>>>,
+    pub used_u32s: usize,
+}
+
+impl<F> TranspilerOutput<F> {
+    pub fn one_to_one(instruction: Instruction<F>) -> Self {
```
jonathanpwang, line 22: I think we could implement `From<Instruction<F>> for TranspilerOutput<F>`

extensions/ecc/transpiler/src/lib.rs
```diff
@@ -59,15 +59,12 @@ impl<F: PrimeField32> TranspilerExtension<F> for EccTranspilerExtension {
...
-                return Some((
-                    Instruction::phantom(
-                        PhantomDiscriminant(EccPhantom::HintDecompress as u16),
-                        F::from_canonical_usize(RV32_REGISTER_NUM_LIMBS * dec_insn.rs1),
-                        F::from_canonical_usize(RV32_REGISTER_NUM_LIMBS * dec_insn.rs2),
-                        curve_idx as u16,
-                    ),
-                    1,
-                ));
+                return Some(TranspilerOutput::one_to_one(Instruction::phantom(
```
jonathanpwang, line 62: see my note above about using `into()`

jonathanpwang (changes requested):
I think the general mechanisms are in place, but a lot of refactoring / adjustments to be made.

The compilation of eDSL within proc-macro seems like it will have versioning issues, and I'm also concerned it is too slow for rust-analyzer. I think for this first version, we should focus on the case where you just compile the eDSL separately into serialized 32-bit instruction format, and then the kernel proc-macro is focused on inserting those serialized instructions into the ELF.

jonathanpwang:
replaced by https://github.com/openvm-org/openvm/pull/1191

PR #1122, opened 2024-12-20
feat: use Plonky3 Poseidon2Air, split aggregation and system chips

stephenh-axiom-xyz:
WIP


jonathanpwang (approved):
I tried to reduce the code in the p3-poseidon2-air wrapper a bit, but unfortunately there is just a bunch of overhead related to generics because our traits are always for generic `F`, but Poseidon2 AIR and implementations in Plonky3 are always implemented slightly differently for different `F` for optimization purposes (e.g., Monty31 and Mersenne31 are treated very differently).

PR #1123, opened 2024-12-23
book: build sdk example

alv-around:
Last week, I corrected some outdated documentation on the sdk usage (#1120). However, the two underlying problems were not address, which are:

1. There are currently no checkguards done on the sdk interface.

2. The code snippets in the documentation are hard-coded, which can lead to inconsistencies between the documentation and the actual code (as happened in #1120)

In this PR, I tackle both problem by:

a. created a bin crate (`sdk.rs`) in the [examples folder](https://doc.rust-lang.org/cargo/guide/project-layout.html). This way, the compiler will run it's check on the code, and if anything if the interface changes this will trigger a compiler error on the example. This has also the advantage that the code is never run, against running expensive e2e tests.

> I also rename the guest-example directory from `example` to `guest` to avoid any confusion with the generic cargo directory `examples`.

b. I linked in `book` the code snippets to the compiled code in a, to ensure that the code snippets are automatically updated so no extra effort is incurred in updating the docs.


jonathanpwang (commented):
LGTM! The idea to use snippets is really nice and an overall improvement.

Now that you've added the bin, should we change the `target_path` etc to the `guest` package so that if we wanted to, `cargo run --example sdk` would actually run?

I'll approve once I generate the docs locally to double check.

jonathanpwang (approved):
LGTM

jonathanpwang:
Please fix lint

alv-around:
> Please fix lint

Sorry, that went completely under my radar. Is now fixed :)

alv-around:
> Now that you've added the bin, should we change the `target_path` etc to the `guest` package so that if we wanted to, `cargo run --example sdk` would actually run?

I **purposely** wrote `examples/sdk.rs` to **not run out the box**, because the binary executes  very expensive operations. This way, I wanted to discourage accidental triggers of this binary.

 I could have shadowed the variable `target_path` after line 59: 
https://github.com/alv-around/openvm/blob/6ef5dab4ed7bdfdb5d8e1a268dac2f05b72e57e2/crates/sdk/examples/sdk.rs#L59
With: 
https://github.com/alv-around/openvm/blob/6ef5dab4ed7bdfdb5d8e1a268dac2f05b72e57e2/crates/sdk/examples/sdk.rs#L47-L51

This way it would run out of the box, however this would have made the imports in the book a bit more convoluted. And I am not sure whether that make things better.

I think like this we have get the best of both worlds:
1. we tell the compiler: "this is how we want the users to use our product, please run your checks". But at the same time we don't run any expensive computation.
2. we export in a simple manner and straight-forward manner the documentation. 

But am happy to change anything you consider most appropriate :)

jonathanpwang:
^This makes sense. I think we can make a separate example repo with a runnable version (and without the heavier computation part) to complement.

PR #1124, opened 2024-12-23
refactor(toolchain): consolidate env vars and cleanup code

keroro520:
Added constants for environment variable names to improve maintainability and prevent typos

### DIFF
--- a/crates/toolchain/build/src/lib.rs
+++ b/crates/toolchain/build/src/lib.rs
@@ -20,10 +20,10 @@ pub use self::config::GuestOptions;
 mod config;
 
 const RUSTUP_TOOLCHAIN_NAME: &str = "nightly-2024-10-30";
-const OPENVM_BUILD_LOCKED_ENV: &str = "OPENVM_BUILD_LOCKED";
-const OPENVM_BUILD_DEBUG_ENV: &str = "OPENVM_BUILD_DEBUG";
-const OPENVM_SKIP_BUILD_ENV: &str = "OPENVM_SKIP_BUILD";
-const OPENVM_GUEST_LOGFILE_ENV: &str = "OPENVM_GUEST_LOGFILE";
+const BUILD_LOCKED_ENV: &str = "OPENVM_BUILD_LOCKED";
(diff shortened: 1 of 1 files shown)


crates/toolchain/build/src/lib.rs
```diff
@@ -162,25 +165,27 @@ pub fn cargo_command(subcmd: &str, rust_flags: &[&str]) -> Command {
     println!("Using rustc: {rustc}");
 
     let mut cmd = sanitized_cmd("cargo");
-    // TODO[jpw]: remove +nightly
+    let toolchain = format!("+{RUSTUP_TOOLCHAIN_NAME}");
```
jonathanpwang, line 168: can you move this above L158 and reuse?

crates/toolchain/build/src/lib.rs
```diff
@@ -19,8 +19,11 @@ pub use self::config::GuestOptions;
 
 mod config;
 
-#[allow(dead_code)]
 const RUSTUP_TOOLCHAIN_NAME: &str = "nightly-2024-10-30";
+const OPENVM_BUILD_LOCKED_ENV: &str = "OPENVM_BUILD_LOCKED";
```
jonathanpwang, line 23: I think the OPENVM_ prefix is redundant in the constant names (since this is in our codebase already) -- a nice perk of adding these constants

jonathanpwang (changes requested):
Thanks! A few small naming changes, but otherwise LGTM

crates/toolchain/build/src/lib.rs
keroro520: @jonathanpwang Thanks for your review. Two review suggestions are applied at [bb41c53](https://github.com/openvm-org/openvm/pull/1124/commits/bb41c53bc3552de5c359473d721c8b14d6476da9) , PTAL

jonathanpwang (approved):
LGTM

jonathanpwang:
Benchmark workflow failing for unrelated reasons due to some fork permissions

PR #1131, opened 2024-12-25
fix: lints to guest programs

### LINKED ISSUE #1125: Add lints to guest programs
The crates for guest programs used in benchmarks and integration tests are not part of the Cargo workspace, which is intentional since the guest programs are designed to target RISC-V. These guest programs are usually in a `**/benchmarks/programs` or `**/test/programs` folders.

We should add a lint workflow step to run `cargo fmt` and `cargo clippy` on these crates, making sure to lint for all targets.

rkdud007:
target to resolve #1125 
q. cannot find `**/test/programs` so had with `examples/*`

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -40,13 +40,19 @@ jobs:
 
       - name: Run fmt, clippy for guest
         run: |
-          # List all guest program crate paths
-          for crate_path in benchmarks/programs/* examples/*; do
+          # Find all directories named "programs" and include additional static paths
+          for crate_path in $(find . -type d -name "programs" -exec find {} -mindepth 0 -maxdepth 0 -type d \;) benchmarks/programs/* examples/*; do
             # Check if Cargo.toml exists in the directory
             if [ -f "$crate_path/Cargo.toml" ]; then
               echo "Running cargo fmt, clippy for $crate_path"
-              cargo fmt --manifest-path "$crate_path/Cargo.toml" --all -- --check 
-              cargo clippy --manifest-path "$crate_path/Cargo.toml" --all-targets --features std -- -D warnings 
+              cargo fmt --manifest-path "$crate_path/Cargo.toml" --all -- --check
+              if [[ "$crate_path" == *"extensions/ecc/tests/programs"* ]]; then
+                echo "Running cargo clippy with k256 feature for $crate_path"
+                cargo clippy --manifest-path "$crate_path/Cargo.toml" --all-targets --features "std k256" -- -D warnings
+              else
+                echo "Running cargo clippy for $crate_path"
+                cargo clippy --manifest-path "$crate_path/Cargo.toml" --all-targets --features std -- -D warnings
+              fi
             else
               echo "Skipping $crate_path as it does not contain a Cargo.toml"
             fi
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -49,6 +49,9 @@ jobs:
               if [[ "$crate_path" == *"extensions/ecc/tests/programs"* ]]; then
                 echo "Running cargo clippy with k256 feature for $crate_path"
                 cargo clippy --manifest-path "$crate_path/Cargo.toml" --all-targets --features "std k256" -- -D warnings
+              elif [[ "$crate_path" == *"extensions/pairing/tests/programs"* ]]; then
+                echo "Running cargo clippy with openvm_pairing_guest::bn254 feature for $crate_path"
+                cargo clippy --manifest-path "$crate_path/Cargo.toml" --all-targets --features "std bn254" -- -D warnings
               else
                 echo "Running cargo clippy for $crate_path"
                 cargo clippy --manifest-path "$crate_path/Cargo.toml" --all-targets --features std -- -D warnings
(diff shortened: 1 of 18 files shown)


jonathanpwang:
By `**/test/programs` I was referring to https://github.com/openvm-org/openvm/tree/main/extensions/algebra/tests/programs (and analogous ones for other extensions)

.github/workflows/lints.yml
```diff
@@ -37,3 +37,29 @@ jobs:
...
+          done
+
+      - name: Run clippy for guest
+        run: |
+          # List all guest program crate paths
+          for crate_path in benchmarks/programs/* examples/*; do
+            # Check if Cargo.toml exists in the directory
+            if [ -f "$crate_path/Cargo.toml" ]; then
+              echo "Running cargo clippy for $crate_path"
+              cargo clippy --manifest-path "$crate_path/Cargo.toml" --all-targets --features std -- -D warnings
```
jonathanpwang, line 61: why not re-use the loop above and do fmt and clippy together?

examples/algebra/Cargo.toml
```diff
@@ -12,11 +12,8 @@ openvm-platform = { git = "https://github.com/openvm-org/openvm.git" }
 openvm-algebra-guest = { git = "https://github.com/openvm-org/openvm.git" }
 openvm-algebra-complex-macros = { git = "https://github.com/openvm-org/openvm.git" }
 serde = { version = "1.0.216", default-features = false }
+num-bigint-dig = { version = "0.8.4", features = ["std", "serde"] }
```
jonathanpwang, line 15: why was this needed?

examples/i256/src/main.rs
```diff
@@ -14,10 +15,10 @@ pub fn get_matrix(val: i32) -> Matrix {
 
 pub fn mult(a: &Matrix, b: &Matrix) -> Matrix {
     let mut c = get_matrix(0);
-    for i in 0..N {
-        for j in 0..N {
-            for k in 0..N {
-                c[i][j] += &a[i][k] * &b[k][j];
+    for (i, row_a) in a.iter().enumerate() {
```
jonathanpwang, line 18: I know clippy complains but in this case could we put an `allow` just for readability?

examples/i256/src/main.rs
```diff
@@ -26,8 +27,8 @@ pub fn mult(a: &Matrix, b: &Matrix) -> Matrix {
 
 pub fn get_identity_matrix() -> Matrix {
     let mut res = get_matrix(0);
-    for i in 0..N {
-        res[i][i] = I256::from_i32(1);
+    for (i, row) in res.iter_mut().enumerate() {
```
jonathanpwang, line 30: ditto

examples/u256/src/main.rs
```diff
@@ -14,9 +15,9 @@ pub fn get_matrix(val: i32) -> Matrix {
 
 pub fn mult(a: &Matrix, b: &Matrix) -> Matrix {
     let mut c = get_matrix(0);
-    for i in 0..N {
-        for j in 0..N {
-            for k in 0..N {
+    for (i, row_a) in a.iter().enumerate() {
```
jonathanpwang, line 18: ditto

examples/u256/src/main.rs
```diff
@@ -26,8 +27,8 @@ pub fn mult(a: &Matrix, b: &Matrix) -> Matrix {
 
 pub fn get_identity_matrix() -> Matrix {
     let mut res = get_matrix(0);
-    for i in 0..N {
-        res[i][i] = I256::from_i32(1);
+    for (i, row) in res.iter_mut().enumerate() {
```
jonathanpwang, line 30: ditto

rkdud007:
@jonathanpwang so linting `**/test/programs` and `examples/*` and `extensions/*/tests/*` these 3? or should I remove `examples/*` now

examples/algebra/Cargo.toml
rkdud007: clippy complains
```console
error[E0433]: failed to resolve: use of undeclared crate or module `num_bigint_dig`
  --> src/main.rs:10:1
   |
10 | / moduli_declare! {
11 | |     Mod1 { modulus = "998244353" },
12 | |     Mod2 { modulus = "1000000007" }
13 | | }
   | |_^ use of undeclared crate or module `num_bigint_dig`
   |
   ```

File: .github/workflows/lints.yml (hunk 1 above)
Comment (rkdud007) [reply .github/workflows/lints.yml:61]: good point missed

File: examples/i256/src/main.rs (hunk 3 above)
Comment (rkdud007) [reply examples/i256/src/main.rs]: yup

File: .github/workflows/lints.yml (hunk 1 above)
Comment (rkdud007) [reply .github/workflows/lints.yml:49]: done

**Comment by jonathanpwang** [comment]
I think we can keep `examples/*`. So that in addition to:
```
find . -type d -name "programs"

./crates/toolchain/tests/programs
./extensions/pairing/tests/programs
./extensions/bigint/tests/programs
./extensions/keccak256/tests/programs
./extensions/rv32im/tests/programs
./extensions/ecc/tests/programs
./extensions/algebra/tests/programs
./benchmarks/programs
```

examples/algebra/Cargo.toml
jonathanpwang: good catch!

jonathanpwang (approved):
LGTM after adding the other `**/programs` directories

rkdud007:
Personally imo instead of allow clippy for functions like mult, have guess util crate and ppl can import it sounds nicer - also if then as its abstracted under other crate can just use iterator for have perf - but not scope of this PR

jonathanpwang:
Some CI tests are failing; you may need to merge/rebase main

jonathanpwang:
Can you merge `main`?

rkdud007:
umm issue was prob not from main merge ( also did one) it from `extensions/pairing/tests/programs/*/` which by removed unused variable lead to err on integration test, I backed up imported variables and just `#![allow(unused_imports)]` my this guess correct prob work now

jonathanpwang:
I see, I think because pairing test require feature `bn254` or `bls12_381`. Let's hope it works now.

jonathanpwang:
> Personally imo instead of allow clippy for functions like mult, have guest util crate and ppl can import it sounds nicer - also if then as its abstracted under other crate can just use iterator for have perf - but mb not scope of this PR : )

Yea I agree this is best practice. In this case the functions in those examples were purely one-time tests, so we don't really expect to re-use and they're not perf critical.

PR #1137, opened 2024-12-26
fix halo2 verifier guest code

lispc:


jonathanpwang (approved):
LGTM
After we merge main I think workspace halo2curves-axiom will also be 0.5.3

extensions/verifier/src/verifier/transcript.rs
```diff
@@ -52,7 +64,10 @@ impl<S> Transcript<G1Affine, OpenVmLoader> for OpenVmTranscript<G1Affine, S, Vec
             .collect_vec();
         let hash = keccak256(&data);
         self.buf = hash.to_vec();
-        OpenVmScalar(Fr::from_be_bytes(&hash), PhantomData)
+        let fr: Halo2Fr = u256_to_fe(U256::from_be_bytes(hash));
+        use halo2curves_axiom::ff::PrimeField;
+        let bytes: [u8; 32] = fr.to_repr();
+        OpenVmScalar(Fr::from_le_bytes(&bytes), PhantomData)
```
jonathanpwang, line 70: I think you can do `Fr::from_be_bytes(&hash)` and then `.reduce()` to ensure its canonical

extensions/verifier/src/verifier/loader.rs
```diff
@@ -31,13 +33,62 @@ lazy_static! {
...
+}
+
 impl EcPointLoader<G1Affine> for OpenVmLoader {
     type LoadedEcPoint = OpenVmEcPoint<G1Affine, EcPoint>;
 
     fn ec_point_load_const(&self, value: &G1Affine) -> Self::LoadedEcPoint {
         let point = EcPoint {
-            x: Fp::from_be_bytes(&value.x().to_bytes()),
-            y: Fp::from_be_bytes(&value.y().to_bytes()),
+            x: Fp::from_le_bytes(&value.x.to_bytes()),
```
jonathanpwang, line 90: just noting the halo2curves `.to_bytes()` actually does some computation, but with the way snark-verifier traits work, this seems unavoidable

extensions/verifier/src/verifier/transcript.rs
lispc: oh good.. i did not notice that function

extensions/verifier/src/verifier/loader.rs
```diff
@@ -31,13 +33,62 @@ lazy_static! {
...
+
+        let [lhs_x, lhs_y, rhs_x, rhs_y]: [_; 4] = limbs
+            .chunks(LIMBS)
+            .map(|limbs| {
+                let v: [Halo2Fr; LIMBS] = limbs
+                    .iter()
+                    .map(|limb| {
+                        let mut buf = limb.0.to_be_bytes();
+                        buf.reverse();
+                        Halo2Fr::from_bytes(&buf).expect("Halo2Fr::from_bytes")
```
jonathanpwang, line 54: there should be a way to do it without going through `Halo2Fr` since its really just going from bytes to limbs, but we can just do it the lazy way for now

extensions/verifier/Cargo.toml
```diff
@@ -12,7 +12,7 @@ openvm-pairing-guest = { workspace = true, features = [
 ] }
 openvm-keccak256-guest = { workspace = true, default-features = false }
 openvm-ecc-guest = { workspace = true, features = ["halo2curves"] }
-snark-verifier = { git = "https://github.com/axiom-crypto/snark-verifier.git", tag = "v0.1.7-git", default-features = false, features = ["halo2-axiom"] }
+snark-verifier = { git = "https://github.com/axiom-crypto/snark-verifier.git", tag = "v0.1.7-git", default-features = false, features = ["halo2-axiom", "loader_evm"] }
```
jonathanpwang, line 15: I think we need to switch to the workspace version (its a branch where we loosened `ruint` version and upgrade rust version)

extensions/verifier/src/verifier/loader.rs
lispc: seems the alternative solution is to use bigint shift of limbs.   Considering usually there is only 1 accumulator inside halo2 proof, i think this time, current solution of less code line number may be better.

PR #1142, opened 2024-12-30
chore: refactor metrics collection

jonathanpwang:
closes INT-2924

The `collect_metrics` flag in `ExecutionSegment` has been renamed to `profiling` for accuracy because when `bench-metrics` feature is enabled, at least some minimal set of metrics is always collected.

**Turn off flamegraph profiling in benchmarks -- these will be made into a separate on-demand workflow in the near future.**


crates/vm/src/arch/segment.rs
```diff
@@ -60,11 +53,11 @@ impl<F: PrimeField32, VC: VmConfig<F>> ExecutionSegment<F, VC> {
         program: Program<F>,
         init_streams: Streams<F>,
         initial_memory: Option<Equipartition<F, CHUNK>>,
-        fn_bounds: FnBounds,
+        _fn_bounds: FnBounds,
```
zlangley, line 56: This is a bit weird since it's used for `bench-metrics` feature. Maybe just silence the clippy warning instead?

zlangley (approved):
lgtm

PR #1162, opened 2025-01-03
feat: Offline memory

zlangley:
Separating memory into (online) `Memory` during execution and `OfflineMemory` for trace generation. Motivation is to speed up execution time.


jonathanpwang:
I'm getting
```bash
cargo bench --bench regex_execute
regex/execute           time:   [790.34 ms 825.10 ms 868.93 ms]
                        change: [-58.376% -56.529% -54.197%] (p = 0.00 < 0.05)
                        Performance has improved.
```
locally. Not sure why the CI shows less.

crates/sdk/src/keygen/dummy.rs
```diff
@@ -55,7 +55,7 @@ pub(super) fn compute_root_proof_heights(
         public_values: vec![F::ZERO; num_user_public_values],
     };
     let vm = SingleSegmentVmExecutor::new(root_vm_config);
-    let res = vm.execute(root_exe, root_input.write()).unwrap();
+    let res = vm.execute(root_exe, root_input.write(), true).unwrap();
```
jonathanpwang, line 58: is this the only place the `compute_heights` bool is used? I'm not sure where else the heights are used in this single segment result; if we actually don't use it anywhere else, I'd slightly prefer if we removed the bool arg and just handled it here

crates/vm/src/arch/extensions.rs
```diff
@@ -457,6 +460,14 @@ impl<F: PrimeField32> SystemBase<F> {
         PROGRAM_BUS
     }
 
+    pub fn memory_bridge(&self) -> MemoryBridge {
+        self.memory_controller.memory_bridge()
+    }
+
+    pub fn offline_memory(&self) -> Arc<Mutex<OfflineMemory<F>>> {
```
jonathanpwang, line 467: add comment around how this should only be used when passing into a chip constructor

crates/vm/src/arch/extensions.rs
```diff
@@ -992,6 +1025,23 @@ impl<F: PrimeField32, E, P> VmChipComplex<F, E, P> {
 
         builder.build()
     }
+
+    #[cfg(feature = "bench-metrics")]
+    fn finalize_metrics(&mut self, metrics: &mut VmMetrics)
```
jonathanpwang, line 1030: nit: `&self`?

crates/vm/src/system/memory/adapter/mod.rs
```diff
@@ -75,6 +75,13 @@ impl<F> AccessAdapterInventory<F> {
         debug_assert!(chip.n() == n);
```
jonathanpwang, line 75: perf related: can we have `AccessAdapterRecord` store the `log2_len` so we don't need to do this `log2_strict_usize` above?

not related to this PR

jonathanpwang: I guess unless we could do some compile time this with const generic (seems unlikely), it's not possible

crates/sdk/src/keygen/dummy.rs
zlangley: We also call it in `execute_for_air_heights`. The rest of the calls (where we pass `false`) are in tests, so maybe we can just always compute heights in `SingleSegmentVmExecutor::execute`?

zlangley: We can't (easily) handle it here because `res` is of type `SingleSegmentVmExecutionResult`, so we don't have the `ExecutionSegment` to call `finalize_memory` on.

crates/vm/src/system/memory/adapter/mod.rs
zlangley: Hmm, this will just move the `log2_strict_usize` to the `AccessAdapterRecord` construction rather than here, so I'm not sure it gains anything?

We could make `chips` an `FxHashMap` keyed by `n`, but I'm not sure `hash(n)` will be much faster than `log2_strict_usize`. We can use `n.trailing_zeros()` instead of `log2_strict_usize` though, since we don't need the assertion that `n` is a power of two (that should have been checked earlier).

crates/sdk/src/keygen/dummy.rs
zlangley: I will rename `execute` to `execute_and_compute_heights` and always finalize.

crates/vm/src/system/memory/controller/memory.rs
```diff
@@ -406,17 +487,148 @@ impl<F: PrimeField32> Memory<F> {
...
+    // just for unit testing
+    #[cfg(test)]
+    fn last_record(&self) -> &MemoryRecord<F> {
+        self.log.last().unwrap().as_ref().unwrap()
+    }
+}
+
+pub type MemoryImage<F> = FxHashMap<Address, F>;
+
+/// A partition of data into blocks where each block has size a power of two.
```
jonathanpwang, line 509: is this comment outdated?

crates/vm/src/system/memory/controller/memory.rs
```diff
@@ -406,17 +487,148 @@ impl<F: PrimeField32> Memory<F> {
...
+    /// Writes an array of values to the memory at the specified address space and start index.
+    ///
+    /// Returns the `RecordId` for the memory record and the previous data.
+    pub fn write<const N: usize>(
+        &mut self,
+        address_space: u32,
+        pointer: u32,
+        values: [F; N],
+    ) -> (RecordId, [F; N]) {
+        assert!(N.is_power_of_two());
```
jonathanpwang, line 554: I hope this gets compiled away

crates/vm/src/system/memory/controller/memory.rs
```diff
@@ -406,17 +487,148 @@ impl<F: PrimeField32> Memory<F> {
...
+        self.timestamp += 1;
+        (self.last_record_id(), values)
+    }
+
+    pub fn increment_timestamp_by(&mut self, amount: u32) {
+        self.timestamp += amount;
+        self.log.push(MemoryLogEntry::IncrementTimestampBy(amount))
+    }
+
+    pub fn increase_timestamp_to(&mut self, amount: u32) {
```
jonathanpwang, line 597: note to self to look into keccak; as discussed, would be nicer if we didn't need timestamp in online memory

crates/vm/src/system/memory/controller/memory.rs
zlangley: yes thanks

crates/vm/src/system/memory/controller/memory.rs
zlangley: it should!

crates/vm/src/system/memory/controller/memory.rs
zlangley: I removed this function now.

crates/vm/src/system/memory/merkle/tests/mod.rs
```diff
@@ -155,6 +169,20 @@ fn test<const CHUNK: usize>(
     .expect("Verification failed");
 }
 
+fn memory_to_partition<F: Default + Copy, const N: usize>(
```
jonathanpwang, line 172: can we name `N` something more descriptive?

crates/vm/src/system/memory/tree/mod.rs
```diff
@@ -92,18 +92,22 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
...
         memory_dimensions: MemoryDimensions,
-        memory: &Equipartition<F, CHUNK>,
+        memory: &MemoryImage<F>,
         hasher: &impl Hasher<CHUNK, F>,
     ) -> MemoryNode<CHUNK, F> {
         // Construct a BTreeMap that includes the address space in the label calculation,
         // representing the entire memory tree.
-        let mut memory_modified = BTreeMap::new();
-        for (&label, &values) in memory {
+        let mut memory_partition = BTreeMap::new();
```
jonathanpwang, line 100: minor: we could do `BTreeMap::with_capacity(memory.len())`

extensions/algebra/circuit/src/modular_chip/tests.rs
```diff
@@ -73,10 +73,11 @@ fn test_addsub(opcode_offset: usize, modulus: BigUint) {
     let adapter = Rv32VecHeapAdapterChip::<F, 2, 1, 1, BLOCK_SIZE, BLOCK_SIZE>::new(
         tester.execution_bus(),
         tester.program_bus(),
-        tester.memory_controller(),
+        tester.memory_bridge(),
```
jonathanpwang, line 76: it's a bit unfortunate we keep needing to pass around `memory_bridge, address_bits`. for later, but should `memory_bridge` just contain` address_bits`?

extensions/keccak256/circuit/src/lib.rs
```diff
@@ -163,15 +167,18 @@ impl<F: PrimeField32> InstructionExecutor<F> for KeccakVmChip<F> {
...
             }
             let mut reads = Vec::with_capacity(KECCAK_RATE_BYTES);
 
             let mut partial_read_idx = None;
             let mut bytes = [0u8; KECCAK_RATE_BYTES];
             for i in (0..KECCAK_RATE_BYTES).step_by(KECCAK_WORD_SIZE) {
                 if i < remaining_len {
-                    let read = memory.read(e, F::from_canonical_usize(src + i));
-                    let chunk = read.data.map(|x| {
+                    let read =
```
jonathanpwang, line 178: let `(foo, bar) = ...`

extensions/algebra/circuit/src/modular_chip/tests.rs
jonathanpwang: edit: doesn't really make sense for `MemoryBridge` to contain `address_bits` because `MemoryBridge` only cares about `timestamp_bits`. It guess it is just a separate parameter.

extensions/native/circuit/src/fri/mod.rs
```diff
@@ -361,21 +362,19 @@ impl<F: PrimeField32> InstructionExecutor<F> for FriReducedOpeningChip<F> {
...
 
-        let mut memory = RefCell::borrow_mut(&self.memory);
-
         let alpha_read = memory.read(addr_space, alpha_ptr);
         let length_read = memory.read_cell(addr_space, length_ptr);
         let a_ptr_read = memory.read_cell(addr_space, a_ptr_ptr);
         let b_ptr_read = memory.read_cell(addr_space, b_ptr_ptr);
 
-        let alpha = alpha_read.data;
+        let alpha = alpha_read.1;
```
jonathanpwang, line 370: I am not a fan of the `.0, .1` prefer if we do `let (alpha_read_id, a_ptr) = ...`

extensions/sha256/circuit/src/sha256_chip/trace.rs
```diff
@@ -47,21 +47,24 @@ where
...
             self.bitwise_lookup_chip.request_range(
-                record.dst_read.data[RV32_REGISTER_NUM_LIMBS - 1].as_canonical_u32()
-                    * mem_ptr_shift,
-                record.src_read.data[RV32_REGISTER_NUM_LIMBS - 1].as_canonical_u32()
-                    * mem_ptr_shift,
+                dst_read.data[RV32_REGISTER_NUM_LIMBS - 1].as_canonical_u32() * mem_ptr_shift,
+                src_read.data[RV32_REGISTER_NUM_LIMBS - 1].as_canonical_u32() * mem_ptr_shift,
             );
-            let len = compose(record.len_read.data);
+            let len = compose(len_read.data.clone().try_into().unwrap());
```
jonathanpwang, line 67: this could be avoided if `compose` just took a slice instead

crates/vm/src/system/memory/controller/memory.rs
```diff
@@ -1,13 +1,53 @@
...
+pub enum MemoryLogEntry<T> {
+    Read {
+        address_space: u32,
+        pointer: u32,
+        len: usize,
+    },
+    Write {
+        address_space: u32,
+        pointer: u32,
+        data: Vec<T>,
```
jonathanpwang, line 46: we might want to use `ShortVec` if the allocation is an issue

crates/vm/src/system/memory/controller/memory.rs
```diff
@@ -406,17 +487,142 @@ impl<F: PrimeField32> Memory<F> {
...
+        let prev_data = array::from_fn(|i| {
+            self.data
+                .insert((address_space, pointer + i as u32), values[i])
+                .unwrap_or(F::ZERO)
+        });
+
+        self.log.push(MemoryLogEntry::Write {
+            address_space,
+            pointer,
+            data: values.to_vec(),
```
jonathanpwang, line 567: I'm a bit concerned about this `to_vec()` all the time. maybe we just need to make the enum have a bunch of powers of two variants

crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -709,16 +757,8 @@ impl<F: PrimeField32> MemoryAuxColsFactory<F> {
...
-) -> Equipartition<F, N> {
-    let mut result = Equipartition::new();
-    for ((addr_space, addr), word) in memory_image {
-        let shift = (addr % N as u32) as usize;
-        let key = (addr_space, addr / N as u32);
-        result.entry(key).or_insert([F::ZERO; N])[shift] = word;
-    }
-    result
+pub fn memory_image_to_memory<F: PrimeField32>(memory_image: MemoryImage<F>) -> Memory<F> {
+    Memory::from_image(memory_image.into_iter().collect())
```
jonathanpwang, line 761: shouldn't this just be `memory_image` without any collect

jonathanpwang (approved):
LGTM, there are some things below we should follow up on with additional profiling, but this is already a big PR so let's merge it first. I did not check all the extension chip refactors as I assume it's largely routine.

Some things discussed offline that we'll want to revisit later:
- `Memory` shouldn't need `timestamp` at all, but its still used in a bunch of places because integration api uses `ExecutionState` and now we could have it just pass `pc` alone.
- `address_bits` is passed as a constructor args separately now. perhaps it should be put into `SystemPort`.
- `MemoryLogEntry` stores `data` as `Vec` instead of array or `ShortVec`
- Using page table for online (and offline) memory instead of hashmap

crates/vm/src/system/memory/tree/mod.rs
jonathanpwang: nvm, that's only available for `HashMap`

crates/vm/src/system/memory/controller/mod.rs
jonathanpwang: update this function is actually unused, deleting

jonathanpwang:
Noting there is **performance degradation** because we have some overhead from the separation (double amount of hashmap lookups). This will be addressed in follow-ups.

extensions/native/circuit/src/fri/mod.rs
zlangley: same!

extensions/sha256/circuit/src/sha256_chip/trace.rs
zlangley: I agree—I had tried this but ran into some snags that I wanted to punt on.

extensions/native/circuit/src/fri/mod.rs
Golovanov399: same btw, but this was the easiest change atm

PR #1176, opened 2025-01-06
feat: add iterator-based for loops

yi-sun:
Closes INT-2930, INT-2946


extensions/native/compiler/src/ir/collections.rs
```diff
@@ -134,7 +133,7 @@ impl<C: Config, V: MemVariable<C>> Array<C, V> {
                 }
 
                 let slice_len = builder.eval(end - start);
-                let address = builder.eval(ptr.address + start);
+                let address = builder.eval(ptr.address + start * RVar::from(V::size_of()));
```
yi-sun, line 136: I believe this fixes a previous bug, please check.

extensions/native/compiler/src/ir/collections.rs
```diff
@@ -78,10 +78,9 @@ impl<C: Config, V: MemVariable<C>> Array<C, V> {
                 }
             }
             Self::Dyn(ptr, len) => {
-                assert_eq!(V::size_of(), 1, "only support variables of size 1");
```
yi-sun, line 81: I'm not sure why this restriction was previously present, as it seems easily removed on L83 below.

extensions/native/compiler/src/ir/utils.rs
```diff
@@ -141,31 +141,25 @@ impl<C: Config> Builder<C> {
...
+    /// Example: if power_bits = [1, 0, 1, 0], then the result should be x^8 * x^2 = x^10.
+    ///
     /// Reference: [`openvm_stark_backend::p3_util::reverse_bits_len`]
-    pub fn exp_reverse_bits_len<V>(
-        &mut self,
-        x: V,
-        power_bits: &Array<C, Var<C::N>>,
-        bit_len: impl Into<RVar<C::N>>,
-    ) -> V
+    pub fn exp_reverse_bits_len<V>(&mut self, x: V, power_bits: &Array<C, Var<C::N>>) -> V
```
zlangley, line 147: Maybe just rename to `exp_bits`? There is no reversing or len now that you are slicing the bit prefix before calling this.

extensions/native/compiler/src/asm/compiler.rs
```diff
@@ -825,6 +835,103 @@ impl<F: PrimeField32 + TwoAdicField, EF: ExtensionField<F> + TwoAdicField> IfCom
     }
 }
 
+// Zipped for loop -- loop extends over the first entry in starts and ends
+pub struct ZipForCompiler<'a, F: Field, EF> {
+    compiler: &'a mut AsmCompiler<F, EF>,
+    starts: Vec<RVar<F>>,
+    ends: Vec<RVar<F>>,
+    step_sizes: Vec<F>,
+    loop_vars: Vec<Var<F>>,
```
zlangley, line 844: Noting that there's a potential optimization here—not sure it's worth it—to only have one loop var per step size (rather than per variable).

extensions/native/compiler/src/ir/builder.rs
```diff
@@ -362,6 +363,75 @@ impl<C: Config> Builder<C> {
...
+                    end: end.into(),
+                    step_size: V::size_of(),
+                    builder: self,
+                    array,
+                }
+            }
+        }
+    }
+
+    pub fn zipped_iter<'a>(
```
zlangley, line 394: Can we just call this `zip`?

extensions/native/compiler/src/ir/collections.rs
```diff
@@ -490,3 +557,35 @@ pub fn unsafe_array_transmute<C: Config, S, T>(arr: Array<C, S>) -> Array<C, T>
...
+
+#[allow(clippy::len_without_is_empty)]
+pub trait ArrayLike<C: Config> {
+    fn len(&self) -> Usize<C::N>;
+
+    fn ptr(&self) -> Ptr<C::N>;
+
+    fn is_fixed(&self) -> bool;
+
+    fn size_of(&self) -> usize;
```
zlangley, line 569: This is a bit confusing; at the call-site I interpreted this as size of the array type itself. Maybe rename to `element_size_of` or `size_of_element`?

extensions/native/compiler/src/ir/collections.rs
```diff
@@ -274,6 +273,74 @@ impl<C: Config> Builder<C> {
         Ref::from_ptr(ptr)
     }
 
+    pub fn iter_ptr_get<V: MemVariable<C>>(&mut self, slice: &Array<C, V>, ptr: RVar<C::N>) -> V {
```
zlangley, line 276: Can you add doc comments for what this and `iter_ptr_set` do?

zlangley (approved):
lgtm

extensions/native/compiler/src/asm/compiler.rs
zlangley: Actually I might be confused—I was thinking the loop_vars were `i * size` but I guess they are `ptr + i * size`, so never mind.

extensions/native/compiler/src/ir/builder.rs
yi-sun: Good call.

extensions/native/compiler/src/asm/compiler.rs
```diff
@@ -825,6 +835,103 @@ impl<F: PrimeField32 + TwoAdicField, EF: ExtensionField<F> + TwoAdicField> IfCom
...
+            .zip(self.step_sizes.iter())
+            .for_each(|(loop_var, step_size)| {
+                self.compiler.push(
+                    AsmInstruction::AddFI(loop_var.fp(), loop_var.fp(), *step_size),
+                    debug_info.clone(),
+                );
+            });
+
+        self.compiler.basic_block();
+        let end = self.ends[0];
```
zlangley, line 894: We should document the behavior of this function, namely that it assumes that the first argument is the shortest and stops after iterating over the first argument.

zlangley: Related: Can we skip computing `ends[i]` for `i != 0`? It seems like we only need the final pointer for the first iterator.

extensions/native/compiler/src/asm/compiler.rs
```diff
@@ -825,6 +835,103 @@ impl<F: PrimeField32 + TwoAdicField, EF: ExtensionField<F> + TwoAdicField> IfCom
...
+                    debug_info.clone(),
+                );
+            }
+            RVar::Val(end) => {
+                self.compiler.push(
+                    AsmInstruction::Bne(loop_label, loop_var.fp(), end.fp()),
+                    debug_info.clone(),
+                );
+            }
+        };
```
zlangley, line 909: Small optimization: We can put the `Bne` after the first `AddFI` so that we don't spend cycles incrementing pointers we don't need to use.

extensions/native/compiler/src/ir/builder.rs
```diff
@@ -458,16 +530,23 @@ impl<C: Config> Builder<C> {
...
-            let index = MemIndex {
-                index: i,
-                offset: 0,
-                size: 1,
-            };
-            builder
-                .operations
-                .push(DslIr::StoreHintWord(arr.ptr(), index));
-        });
+        self.zip(&[Box::new(arr.clone()) as Box<dyn ArrayLike<C>>])
```
zlangley, line 533: Can't we just use iter here?

extensions/native/compiler/src/ir/builder.rs
```diff
@@ -840,6 +919,178 @@ impl<C: Config> IfBuilder<'_, C> {
...
+            self.start,
+            self.end,
+            step_size,
+            loop_variable,
+            loop_instructions,
+        );
+        self.builder.operations.push(op);
+    }
+}
+
```
zlangley, line 1093: We should write some tests for this stuff in `tests/for_loops.rs`.

extensions/native/compiler/src/ir/collections.rs
```diff
@@ -78,10 +78,9 @@ impl<C: Config, V: MemVariable<C>> Array<C, V> {
                 }
             }
             Self::Dyn(ptr, len) => {
-                assert_eq!(V::size_of(), 1, "only support variables of size 1");
                 let len = RVar::from(len.clone());
                 let shift = shift.into();
-                let new_ptr = builder.eval(*ptr + shift);
+                let new_ptr = builder.eval(*ptr + shift * RVar::from(V::size_of()));
```
zlangley, line 83: Is our symbolic expression builder smart enough to know that if an `RVar` is `RVar::Const(C::N::ONE)`, then it should not do a multiplication here? I'm worried that this is introducing some cycles/adds.

(I do believe we shouldn't have to worry about it here though, and the logic belongs in the `Mul` implementation for `RVar` or `SymbolicVar`.)

extensions/native/compiler/src/ir/collections.rs
zlangley: I agree.

extensions/native/recursion/src/fri/mod.rs
```diff
@@ -193,55 +194,60 @@ pub fn verify_batch<C: Config>(
...
-            |builder| {
-                builder.assign(&left, root_ptr);
-                builder.assign(&right, sibling);
-            },
-        );
+    builder
+        .zip(&[
+            Box::new(proof.clone()) as Box<dyn ArrayLike<C>>,
+            Box::new(index_bits.clone()) as Box<dyn ArrayLike<C>>,
+        ])
```
zlangley, line 201: I wonder if it's worth making a macro so that you can do
```
compile_zip!(builder, proof, index_bits).for_each(|ptr_vec, builder| { ... });
```
The `Box` stuff is pretty ugly.

extensions/native/compiler/src/ir/collections.rs
zlangley: It looks like @nyunyunyunyu made this change in `SymbolicVar` so this should not incur any extra cost.

extensions/native/compiler/src/ir/builder.rs
yi-sun: I think we need `zip` -- it gives access to pointers, whereas `iter` gives access to the values.  It might be worth consolidating, but I'd like to do that afterwards.

extensions/native/compiler/src/asm/compiler.rs
yi-sun: I think we need the first `Bne` to cover the case where `starts[0] == end0`.

PR #1177, opened 2025-01-06
chore: prof md doesn't need to display cycle_tracker_span

jonathanpwang:


crates/prof/src/lib.rs
```diff
@@ -135,6 +136,18 @@ impl MetricDb {
...
+                    .iter()
+                    .enumerate()
+                    .map(|(i, val)| {
+                        if label_keys[i] == "cycle_tracker_span" {
+                            ""
+                        } else {
+                            val.as_str()
+                        }
+                    })
+                    .collect_vec();
```
zlangley, line 150: won't this still have a row for each cycle tracker span?

jonathanpwang: yea - do we just not want those tables at all?

jonathanpwang: removed the table entirely

PR #1178, opened 2025-01-07
Fix: Ensure program trace for frequencies skips `None` instructions

TlatoaniHJ:
Adds a `ProgramChip` test with a program that has `None` instructions. Fixes trace generation to pass this test: previously, the frequencies array was just directly being converted into the trace, even though the cached trace skipped the instructions that were `None`. Now those are skipped in trace generation for the frequencies as well.

Adds a constructor for `Program` that takes in a `Vec` of `Option<Instruction>`. Also changed the name of `Program::instructions()` to `Program::non_none_instructions` since the old name seemed misleading, and added `Program::num_non_none_instructions()` since that was the only use of `Program::instructions()` other than for display. These names don't seem very good so someone should suggest better ones.

### DIFF
--- a/crates/toolchain/instructions/src/program.rs
+++ b/crates/toolchain/instructions/src/program.rs
@@ -58,6 +58,27 @@ impl<F: Field> Program<F> {
         }
     }
 
+    pub fn new_without_debug_infos_with_option(
+        instructions: &[Option<Instruction<F>>],
+        step: u32,
+        pc_base: u32,
+        max_num_public_values: usize,
+    ) -> Self {
+        assert!(
+            instructions.is_empty()
+                || pc_base + (instructions.len() as u32 - 1) * step <= MAX_ALLOWED_PC
+        );
+        Self {
+            instructions_and_debug_infos: instructions
+                .iter()
+                .map(|instruction| instruction.clone().map(|instruction| (instruction, None)))
+                .collect(),
+            step,
+            pc_base,
+            max_num_public_values,
+        }
+    }
+
     /// We assume that pc_start = pc_base = 0 everywhere except the RISC-V programs, until we need otherwise
     /// We use [DEFAULT_PC_STEP] for consistency with RISC-V
     pub fn from_instructions_and_debug_infos(
@@ -105,12 +126,16 @@ impl<F: Field> Program<F> {
         self.instructions_and_debug_infos.is_empty()
     }
 
-    pub fn instructions(&self) -> Vec<Instruction<F>> {
+    pub fn non_none_instructions(&self) -> Vec<Instruction<F>> {
         self.instructions_and_debug_infos
             .iter()
             .flatten()
             .map(|(instruction, _)| instruction.clone())
             .collect()
+    }
+
+    pub fn num_non_none_instructions(&self) -> usize {
+        self.non_none_instructions().len()
     }
 
     pub fn debug_infos(&self) -> Vec<Option<DebugInfo>> {
@@ -168,7 +193,7 @@ impl<F: Field> Program<F> {
 }
 impl<F: Field> Display for Program<F> {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-        for instruction in self.instructions().iter() {
+        for instruction in self.non_none_instructions().iter() {
             let Instruction {
                 opcode,
                 a,
@@ -179,7 +204,7 @@ impl<F: Field> Display for Program<F> {
                 f,
                 g,
             } = instruction;
-            write!(
+            writeln!(
                 formatter,
                 "{:?} {} {} {} {} {} {} {}",
                 opcode, a, b, c, d, e, f, g,
@@ -190,7 +215,7 @@ impl<F: Field> Display for Program<F> {
 }
 
 pub fn display_program_with_pc<F: Field>(program: &Program<F>) {
-    for (pc, instruction) in program.instructions().iter().enumerate() {
+    for (pc, instruction) in program.non_none_instructions().iter().enumerate() {
         let Instruction {
             opcode,
             a,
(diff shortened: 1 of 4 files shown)


crates/toolchain/instructions/src/program.rs
```diff
@@ -58,6 +58,27 @@ impl<F: Field> Program<F> {
         }
     }
 
+    pub fn new_without_debug_infos_with_option(
```
jonathanpwang, line 61: `new_sparse`?

I feel like we can drop `without_debug_infos` (also from the other constructor), since that's more like the default, and `with_debug_infos` is extra descriptor

crates/toolchain/instructions/src/program.rs
```diff
@@ -105,14 +126,18 @@ impl<F: Field> Program<F> {
         self.instructions_and_debug_infos.is_empty()
     }
 
-    pub fn instructions(&self) -> Vec<Instruction<F>> {
+    pub fn non_none_instructions(&self) -> Vec<Instruction<F>> {
```
jonathanpwang, line 129: `dense_instructions` or `instructions_without_gaps`?

crates/toolchain/instructions/src/program.rs
```diff
@@ -105,14 +126,18 @@ impl<F: Field> Program<F> {
...
         self.instructions_and_debug_infos
             .iter()
             .flatten()
             .map(|(instruction, _)| instruction.clone())
             .collect()
     }
 
+    pub fn num_non_none_instructions(&self) -> usize {
```
jonathanpwang, line 137: `num_instructions_without_gaps` or `num_actual_instructions` seem ok

crates/vm/src/system/program/tests/mod.rs
```diff
@@ -47,29 +46,34 @@ fn interaction_test(program: Program<BabyBear>, execution: Vec<u32>) {
...
-            instruction.b,
-            instruction.c,
-            instruction.d,
-            instruction.e,
-            instruction.f,
-            instruction.g,
-        ]);
+    for (index, frequency) in execution_frequencies.into_iter().enumerate() {
+        let option = program.get_instruction_and_debug_info(index);
+        if option.is_some() {
```
jonathanpwang, line 51: nit: use `if let Some(..) = option`

crates/vm/src/system/program/tests/mod.rs
```diff
@@ -47,29 +46,34 @@ fn interaction_test(program: Program<BabyBear>, execution: Vec<u32>) {
...
+            let instruction = option.unwrap().0;
+            program_cells.extend(vec![
+                BabyBear::from_canonical_usize(frequency), // hacky: we should switch execution_frequencies into hashmap
```
jonathanpwang, line 54: on the hackiness, do you mean just in test, or in general? the goal of using the gaps was so execution_frequencies could stay a lookup instead of a hashmap

crates/vm/src/system/program/trace.rs
```diff
@@ -62,7 +62,9 @@ impl<F: PrimeField64> ProgramChip<F> {
         let common_trace = RowMajorMatrix::new_col(
             self.execution_frequencies
                 .into_iter()
-                .map(|x| F::from_canonical_usize(x))
+                .enumerate()
+                .filter(|(i, _)| self.program.get_instruction_and_debug_info(*i).is_some())
```
jonathanpwang, line 66: can you zip `execution_frequences` with `program.instructions_and_debug_infos` instead? it feels slightly more canonical (and also prevent a `get` for every index)

jonathanpwang (approved):
LGTM, good catch. Gave a few naming suggestions that came to mind

crates/vm/src/system/program/tests/mod.rs
```diff
@@ -47,29 +46,34 @@ fn interaction_test(program: Program<BabyBear>, execution: Vec<u32>) {
...
-            instruction.c,
-            instruction.d,
-            instruction.e,
-            instruction.f,
-            instruction.g,
-        ]);
+    for (index, frequency) in execution_frequencies.into_iter().enumerate() {
+        let option = program.get_instruction_and_debug_info(index);
+        if option.is_some() {
+            let instruction = option.unwrap().0;
```
zlangley, line 52:
```suggestion
        if let Some((instruction, _) = program.get_instruction_and_debug_info(index) {
```

crates/vm/src/system/program/tests/mod.rs
```diff
@@ -47,29 +46,34 @@ fn interaction_test(program: Program<BabyBear>, execution: Vec<u32>) {
...
+            program_cells.extend(vec![
```
zlangley, line 53:
```suggestion
            program_cells.extend([
```

crates/vm/src/system/program/trace.rs
zlangley: You can also use `filter_map` (or `flat_map`) instead of `filter` and `map`.

crates/toolchain/instructions/src/program.rs
```diff
@@ -58,6 +58,27 @@ impl<F: Field> Program<F> {
         }
     }
 
+    pub fn new_without_debug_infos_with_option(
+        instructions: &[Option<Instruction<F>>],
```
zlangley, line 62: I wonder if we should change `Program` to have parallel `instructions` and `debug_info` rather than a vec of pairs?

crates/toolchain/instructions/src/program.rs
zlangley: `defined_instructions`?

crates/toolchain/instructions/src/program.rs
jonathanpwang: hm that could be good - `debug_info` is weird and only for dsl

crates/vm/src/system/program/tests/mod.rs
TlatoaniHJ: Done

PR #1185, opened 2025-01-07
perf: Preallocate records

zlangley:


crates/vm/src/arch/config.rs
```diff
@@ -54,11 +54,13 @@ pub struct MemoryConfig {
...
     /// Maximum N AccessAdapter AIR to support.
     pub max_access_adapter_n: usize,
+    /// An expected upper bound on the number of memory accesses.
+    pub access_capacity: usize,
 }
 
 impl Default for MemoryConfig {
     fn default() -> Self {
-        Self::new(29, 1, 29, 29, 17, 64)
+        Self::new(29, 1, 29, 29, 17, 64, 1 << 24)
```
jonathanpwang, line 63: 2**24 seems pretty high, maybe just 2**20?

crates/vm/src/arch/integration_api.rs
```diff
@@ -201,7 +201,7 @@ where
         Self {
             adapter,
             core,
-            records: vec![],
+            records: Vec::with_capacity(1 << 20),
```
jonathanpwang, line 204: this should use `access_capacity`

zlangley: Do you have any suggestions for how to access `config` from here?

(Also I think this is a different parameter than `access_capacity`; `access_capacity` is the number of record IDs, which should be several factors larger than any one chip's `records.len()`, both because every chip contributes to memory records and because each chip record corresponds to several memory records.)

crates/vm/src/arch/config.rs
zlangley: This is for memory. If a chip has up to 2^20 records and we have dozens of chips and each record does several accesses, then 2^24 might still be small?

crates/vm/src/arch/integration_api.rs
jonathanpwang: I guess we could add it as a field in `OfflineMemory`...

zlangley: It seems weird because this is more about the number of instructions executed by any one chip rather than anything memory-related.

crates/vm/src/arch/config.rs
jonathanpwang: can you lmk the execution performance diff (via criterion bench) between 2^20 and 2^24, just for a sense?

crates/vm/src/arch/integration_api.rs
jonathanpwang: ok we can address later then

jonathanpwang: at least make it a private constant

jonathanpwang (approved):
LGTM but leave some todos and also make all unavoidable constants in `const`

crates/vm/src/arch/config.rs
zlangley: from 1 << 20 to 1 << 24 (regex_execute bench):
```
regex/execute           time:   [377.90 ms 391.66 ms 405.92 ms]
                        change: [-11.309% -6.3323% -0.8434%] (p = 0.05 < 0.05)
                        Change within noise threshold.
```
on the border of statistical signficance

PR #1191, opened 2025-01-07
feat: native guest macro for kernel functions

TlatoaniHJ:
Revision of #1121. Introduces the `openvm_native_guest_macro::kernel!` macro. Its current behavior is as follows:

```
openvm_native_guest_macro::kernel! {
    fn function_name(arg1: RustType1 | EdslType1<F>, ..)-> RustReturnType | EdslReturnType<F> {
        compiler_output.txt
    }
}
```

Inside the macro, you provide a function declaration. Each type (argument types and return types) should be specified in the form `A | B`, where `A` is a Rust type and `B` is an eDSL type (which will generally have a generic `F` and possibly `EF`). Currently, only `usize | Felt<F>` is supported.

Inside the function body, you should provide a filename. The corresponding file should contain a series of `u32`s in plaintext format, which should be obtained from compiling an eDSL program. For each argument of the function, there should be one `u32` equal to its `fp` in the eDSL program. Then, there should be a single `u32` equal to the same for the return value. The remaining `u32s` should be the serialization of the compiled program.

The intention is that the eDSL program's inputs and outputs should correspond to the eDSL types specified in the macro invocation. However for now, although only `Felt<F>` is allowed by the macro, the eDSL program can use either `Felt<F>` or `Usize<F>` as their runtime behavior is the same.

An example of a Fibonacci program is provided in `benchmarks/programs/test_kernel`: `benchmarks/programs/test_kernel/src/main.rs` is the Rust program containing the macro invocation, and `benchmarks/programs/test_kernel/kernel/src/main.rs` is the program containing the eDSL code which is run to produce `benchmarks/programs/test_kernel/compiler_output.txt`. You can run `benchmarks/src/bin/test_kernel.rs` to compile and execute this program.


benchmarks/programs/test_kernel/macro_output.txt
```diff
@@ -0,0 +1,115 @@
+fn function_name(var_n : usize) -> usize
```
lispc, line 1: why not use u32 instead of usize? u32 is more consistent and has no ambiguity on any platform

extensions/native/guest-macro/src/transportation.rs
```diff
@@ -0,0 +1,321 @@
...
+}
+
+pub fn transport_edsl_to_rust<F: Field>(
+    rust_type: String,
+    edsl_type: String,
+    rust_name: String,
+    edsl_fp: usize,
+) -> Vec<MacroInstruction<F>> {
+    match (rust_type.as_str(), edsl_type.as_str()) {
+        ("usize", "Felt < F >") => transport_felt_to_usize(rust_name, edsl_fp),
```
lispc, line 299: i am going to develop some examples tomorrow where a const size array of Felts is converted to array of u32s.  That is needed in my case (verify root proof inside a native-rv32 hybrid program).

lispc:
I would suggest split this pr into multi parts:    
first merge the transpiler relared pr, to allow hybrid rv32 instructions and long form instructions.    
when other parts.

Cargo.toml
```diff
@@ -125,6 +128,8 @@ openvm-circuit = { path = "crates/vm", default-features = false }
 openvm-circuit-derive = { path = "crates/vm/derive", default-features = false }
 openvm-toolchain-tests = { path = "crates/toolchain/tests", default-features = false }
 
+# eDSL Kernel
```
jonathanpwang, line 131: delete?

benchmarks/programs/test_kernel/.cargo/config.toml
jonathanpwang, line 1: i think this is old

benchmarks/programs/test_kernel/kernel/src/main.rs
jonathanpwang, line 1: a nicer way: we can move this to `build.rs` in `test_kernel` and it will run as a build script to generate the `compiler_output.txt` before you build the guest program

benchmarks/programs/test_kernel/macro_output.txt
jonathanpwang, line 1: i think this was just a debugging file, can be deleted

benchmarks/programs/test_kernel/src/main.rs
```diff
@@ -0,0 +1,21 @@
+#![cfg_attr(not(feature = "std"), no_main)]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+openvm_native_guest_macro::edsl_kernel! {
+    fn function_name(n: usize | Felt<F>) -> usize | Felt<F> {
+        compiler_output.txt
```
jonathanpwang, line 6: we can address later, I would slightly prefer if this was a string with " " to match `include_bytes!`

benchmarks/src/bin/test_kernel.rs
jonathanpwang, line 1: discussed offline: this does not belong in `benchmarks`; make a separate `openvm-native-integration-test` crate

crates/toolchain/instructions/src/riscv.rs
```diff
@@ -5,3 +5,4 @@ pub const RV32_CELL_BITS: usize = 8;
 pub const RV32_IMM_AS: u32 = 0;
 pub const RV32_REGISTER_AS: u32 = 1;
 pub const RV32_MEMORY_AS: u32 = 2;
+pub const NATIVE_KERNEL_AS: u32 = 5;
```
jonathanpwang, line 8: it's (1) weird this is in riscv.rs, (2) we prefer not to have these constants in `openvm-instructions` since its more specific to a certain extension. Arguably this whole file should be removed from `openvm-instructions`, but I think it exists for some archaic reasons

crates/toolchain/transpiler/src/lib.rs
```diff
@@ -29,7 +29,7 @@ impl<F: PrimeField32> FromElf for VmExe<F> {
     type ElfContext = Transpiler<F>;
     fn from_elf(elf: Elf, transpiler: Self::ElfContext) -> Result<Self, TranspilerError> {
         let instructions = transpiler.transpile(&elf.instructions)?;
-        let program = Program::new_without_debug_infos(
+        let program = Program::new_without_debug_infos_with_option(
```
jonathanpwang, line 32: did we rename this and you haven't rebased?

crates/toolchain/transpiler/src/transpiler.rs
```diff
@@ -60,15 +60,22 @@ impl<F: PrimeField32> Transpiler<F> {
...
-            }
-            if options.len() > 1 {
-                return Err(TranspilerError::AmbiguousNextInstruction);
-            }
-            let (instruction, advance) = options.pop().unwrap().unwrap();
-            instructions.push(instruction);
-            ptr += advance;
+            assert!(
+                !options.is_empty(),
+                //"couldn't parse the next instruction: {:032b}",
```
jonathanpwang, line 65:
```suggestion
```

crates/toolchain/transpiler/src/transpiler.rs
```diff
@@ -60,15 +60,22 @@ impl<F: PrimeField32> Transpiler<F> {
...
-            if options.is_empty() {
-                return Err(TranspilerError::ParseError(instructions_u32[ptr]));
-            }
-            if options.len() > 1 {
-                return Err(TranspilerError::AmbiguousNextInstruction);
-            }
-            let (instruction, advance) = options.pop().unwrap().unwrap();
-            instructions.push(instruction);
-            ptr += advance;
+            assert!(
```
jonathanpwang, line 63: can we go back to returning error? that was @Golovanov399 's intention with the `TranspilerError` type

crates/toolchain/transpiler/src/transpiler.rs
```diff
@@ -60,15 +60,22 @@ impl<F: PrimeField32> Transpiler<F> {
...
+                !options.is_empty(),
+                //"couldn't parse the next instruction: {:032b}",
+                "couldn't parse the next instruction: {} = {:032b}",
+                instructions_u32[ptr],
+                instructions_u32[ptr],
+            );
+            assert!(
```
jonathanpwang, line 70: return error?

crates/toolchain/transpiler/src/transpiler.rs
jonathanpwang: you can make TranspilerError have additional information and format the error message via `thiserror`

crates/toolchain/transpiler/src/extension.rs
```diff
@@ -8,7 +8,35 @@ pub trait TranspilerExtension<F> {
...
+
+impl<F> TranspilerOutput<F> {
+    pub fn one_to_one(instruction: Instruction<F>) -> Self {
+        Self {
+            instructions: vec![Some(instruction)],
+            used_u32s: 1,
+        }
+    }
+
+    pub fn many_to_one(instruction: Instruction<F>, used_u32s: usize) -> Self {
```
jonathanpwang, line 29: do you ever use this? also I'm not unsure: should this be `vec![Some(instruction), None, ..., None]`?

In fact should `used_u32s = instructions.len()` always? if we could just return `Vec<Instruction<F>>` where empty vector means it couldn't find anything, that'd clean things up

extensions/native/guest-macro/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use crate::parse_compiler_output::CompiledKernel;
+
+mod parse_compiler_output;
+mod parse_kernel;
+mod transportation;
+const UTILITY_CELL: usize = 0;
+
+#[proc_macro]
+pub fn edsl_kernel(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
jonathanpwang, line 11: rename to `native_kernel`

extensions/native/guest-macro/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+    let input = proc_macro2::TokenStream::from(input);
+
+    let parsed_kernel = parse_kernel::parse_raw_kernel(input);
+    let compiler_output = std::fs::read_to_string(parsed_kernel.file_path.clone()).unwrap();
```
jonathanpwang, line 15: not a strong opinion, but we could store the `Vec<u32>` as a raw binary file of `Vec<u8>`. I'm not sure we ever care its readable as a string?

extensions/native/guest-macro/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+    let compiled_kernel: CompiledKernel<BabyBear> =
+        parse_compiler_output::parse_compiled_kernel(parsed_kernel, compiler_output);
+    let rust_function = transportation::compiled_kernel_to_function(compiled_kernel);
+
+    std::fs::write("macro_output.txt", rust_function.to_string()).unwrap();
```
jonathanpwang, line 20: is this just for debugging? I feel like we shouldn't always print this out, rust macro expansion will do this for you anyways (in your IDE -> rust-analyzer should have expand macro as an option)

extensions/native/guest-macro/src/transportation.rs
```diff
@@ -0,0 +1,321 @@
...
+        compiled_kernel.edsl_return_type,
+        return_name.clone(),
+        compiled_kernel.return_fp,
+    ));
+
+    let mut instructions_string = String::new();
+    for instruction in instructions.iter() {
+        instructions_string.push_str(&format!("{:?}\n", instruction));
+    }
+    std::fs::write("instructions.txt", instructions_string).expect("Failed to write file");
```
jonathanpwang, line 123: remove this, I'm not sure why we ever need to write the MacroInstruction vector

extensions/native/guest-macro/src/transportation.rs
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+}
+
+fn u32_to_directive(x: u32) -> String {
+    let opcode = x & 0b1111111;
+    let funct3 = (x >> 12) & 0b111;
+    let rd = (x >> 7) & 0b11111;
+    let rs1 = (x >> 15) & 0b11111;
+    let mut simm12 = (x >> 20) as i32;
```
jonathanpwang, line 144: the .insn takes a signed number?

extensions/native/guest-macro/src/parse_compiler_output.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub arguments: Vec<CompiledKernelArgument>,
+    pub body: Vec<Instruction<F>>,
+    pub rust_return_type: String,
+    pub edsl_return_type: String,
+    pub return_fp: usize,
+}
+
+pub fn parse_compiled_kernel<F: PrimeField32>(
+    parsed_kernel: ParsedKernel,
+    compiler_output: String,
```
jonathanpwang, line 27: already mentioned above, but it's a bit weird to go through String for `Vec<u32>`, if you just had `Vec<u8>`, you could directly be chunking `u32` or even using `read_exact`

extensions/native/guest-macro/src/transportation.rs
```diff
@@ -0,0 +1,321 @@
...
+impl<F: Field> Operand<F> {
+    pub fn usize(val: usize) -> Self {
+        Literal(F::from_canonical_usize(val))
+    }
+
+    pub fn u32(val: u32) -> Self {
+        Literal(F::from_canonical_u32(val))
+    }
+
+    pub fn whatever() -> Self {
```
jonathanpwang, line 38: while this is funny, let's choose a better name..

extensions/native/guest-macro/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+use p3_baby_bear::BabyBear;
+
+use crate::parse_compiler_output::CompiledKernel;
+
+mod parse_compiler_output;
+mod parse_kernel;
+mod transportation;
+const UTILITY_CELL: usize = 0;
```
jonathanpwang, line 8: this is not safe: you are making an assumption that ptr=0 is free to use, whereas we might be using it for another purpose. I think you should use `A0` from the native-compiler crate?

extensions/native/guest-macro/src/transportation.rs
```diff
@@ -0,0 +1,321 @@
...
+    for instruction in instructions {
+        let directives = instruction_to_directives(instruction);
+        pc_diff += directives.len() - 1;
+        add_directives(directives);
+    }
+
+    let mut jal_instruction: MacroInstruction<F> = MacroInstruction::new(
+        VmOpcode::with_default_offset(NativeJalOpcode::JAL),
+        [
+            Operand::usize(UTILITY_CELL),
```
jonathanpwang, line 203: see my comment above around utility cell

extensions/native/guest-macro/src/transportation.rs
```diff
@@ -0,0 +1,321 @@
...
+            Operand::whatever(),
+            Operand::whatever(),
+            Operand::u32(NATIVE_KERNEL_AS),
+        ],
+    );
+    let jal_example_directives = instruction_to_directives(jal_instruction.clone());
```
jonathanpwang, line 209: why is this called example_ ?

extensions/native/guest-macro/src/transportation.rs
```diff
@@ -0,0 +1,321 @@
...
+    pc_diff += jal_example_directives.len() - 1;
+
+    jal_instruction.operands[1] = Operand::usize(DEFAULT_PC_STEP as usize * (pc_diff + 1));
```
jonathanpwang, line 212: nit: if you could somehow put this in the constructor of `jal_instruction`, it's a little clearer to read

extensions/native/serialization/src/lib.rs
```diff
@@ -0,0 +1,54 @@
+use openvm_instructions::{instruction::Instruction, VmOpcode};
+use p3_field::PrimeField32;
+
+/*
+ * The indicators use
+ * - opcode = 0x2b (custom-1 as defined in RISC-V spec document)
```
jonathanpwang, line 6: I thought a little, maybe we should use 0x0b (custom-0) because custom-1 currently all has instructions where there's some customization in the opcode offsets, whereas this one is just going to be a single non-changing opcode.

extensions/native/transpiler/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+    GAP_INDICATOR, LONG_FORM_INSTRUCTION_INDICATOR, VARIABLE_REGISTER_INDICATOR,
+};
+use openvm_transpiler::{TranspilerExtension, TranspilerOutput};
+use p3_field::PrimeField32;
+
+pub struct LongFormTranspilerExtension;
+
+impl<F: PrimeField32> TranspilerExtension<F> for LongFormTranspilerExtension {
+    fn process_custom(&self, instruction_stream: &[u32]) -> Option<TranspilerOutput<F>> {
+        if instruction_stream[0] == LONG_FORM_INSTRUCTION_INDICATOR {
```
jonathanpwang, line 12: I am confused why this function does not use `deserialize_instructions`, shouldn't there be some relation between the two?

extensions/native/transpiler/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+            let num_operands = instruction_stream[1] as usize;
+            let opcode = VmOpcode::from_usize(instruction_stream[2] as usize);
+            let mut operands = vec![];
+            let mut j = 3;
+            for _ in 0..num_operands {
+                if instruction_stream[j] == VARIABLE_REGISTER_INDICATOR {
```
jonathanpwang, line 18: what is the purpose of VARIABLE_REGISTER_INDICATOR? it seems weird to have here in this general processing function

extensions/native/transpiler/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+                a: operands[0],
+                b: operands[1],
+                c: operands[2],
+                d: operands[3],
+                e: operands[4],
+                f: operands[5],
+                g: operands[6],
+            };
+            Some(TranspilerOutput::many_to_one(instruction, j))
+        } else if instruction_stream[0] == GAP_INDICATOR {
```
jonathanpwang, line 46: when would we add a gap in the `u32` stream? I thought gaps were only a concept in the `ProgramChip`

jonathanpwang (approved):
We can merge after addressing comments.

Will merge this version first, but noting that we already had some discussions that the "transport" layer should be moved more into something that can be done in the eDSL implementation itself, and we can layer macros on top later to help with more complicated struct transport later.

Can you add documentation around the different opcode indicators for LongForm, Gap, VariableRegister? It's hard to follow what their role is.

extensions/native/guest-macro/src/transportation.rs
lispc: most insn string will be meaningless actually, why not directly use ".4byte 0x00000293" like raw string?

Cargo.toml
TlatoaniHJ: Done

benchmarks/programs/test_kernel/.cargo/config.toml
TlatoaniHJ: Removed

extensions/native/guest-macro/src/lib.rs
TlatoaniHJ: Removed

extensions/native/guest-macro/src/transportation.rs
TlatoaniHJ: Removed

extensions/native/serialization/src/lib.rs
TlatoaniHJ: Should we make funct3 = 0 as well, just to kind of make it all 0?

extensions/native/transpiler/src/lib.rs
TlatoaniHJ: There should, I just wasn't sure how to organize it; I was thinking that `LongFormTranspilerExtension` could remain as it is while `deserialize_instructions` calls it, does that seem reasonable to you?

crates/toolchain/transpiler/src/extension.rs
TlatoaniHJ: This is used by `LongFormTranspilerExtension`. It shouldn't be `vec![Some(instruction), None, ..., None]` because for a kernel, all the defined instructions should be contiguous, and the undefined instructions for padding should occur at the end of the entire kernel.

crates/toolchain/transpiler/src/transpiler.rs
TlatoaniHJ: This comment made redundant

crates/toolchain/transpiler/src/transpiler.rs
TlatoaniHJ: Done

crates/toolchain/transpiler/src/transpiler.rs
TlatoaniHJ: Done

extensions/native/guest-macro/src/transportation.rs
TlatoaniHJ: Renamed to `arbitrary`

extensions/native/guest-macro/src/transportation.rs
TlatoaniHJ: It's called example because it's not the actual JAL instruction that will be used, just a placeholder one to determine how many RISC-V instructions will be generated by JAL.

extensions/native/guest-macro/src/transportation.rs
TlatoaniHJ: I could just construct a new JAL instruction instead of reusing `jal_instruction` (maybe rename that one to `jal_example_instruction` and make the new one `jal_instruction`), would that be better?

extensions/native/transpiler/src/lib.rs
TlatoaniHJ: How could there be gaps in the `Program` later on if they aren't introduced in the transpilation process?

crates/toolchain/transpiler/src/lib.rs
TlatoaniHJ: No, `new_without_debug_infos_with_option` is a new constructor distinct from `new_without_debug_infos` that takes `Option<Instruction<F>>` instead of `Instruction<F>`.

extensions/native/transpiler/src/lib.rs
jonathanpwang: I somehow missed this doc :O

benchmarks/src/bin/test_kernel.rs
TlatoaniHJ: Done

extensions/native/transpiler/src/lib.rs
jonathanpwang: when do we use `deserialize_instructions` outside of `LongFormTranspilerExtension`?

extensions/native/serialization/src/lib.rs
jonathanpwang: funct3 can't be 0 because we want to make sure it does not conflict with https://github.com/openvm-org/openvm/blob/main/docs/specs/RISCV.md#system

for `0x0b` I think the only funct3 we have left is `0b111`

extensions/native/transpiler/src/lib.rs
TlatoaniHJ: The macro uses it to deserialize the compiler output

extensions/native/guest-macro/src/transportation.rs
jonathanpwang: according to https://sourceware.org/binutils/docs/as/RISC_002dV_002dDirectives.html
> But in fact, the instruction formats are difficult to use for some users, so most of them are using ‘.word’ to encode the instruction directly, rather than using ‘.insn’. It is fine for now, but will be wrong when the mapping symbols are supported, since ‘.word’ will not be shown as an instruction, it should be shown as data. Therefore, we also support two more formats of the ‘.insn’, the instruction ‘add a0, a1, a2’ could also be written as ‘.insn 0x4, 0xc58533’ or ‘.insn 0xc58533’. When the insn_length is set, then assembler will check if the value is a valid insn_length bytes instruction.

@TlatoaniHJ we can try, I'm not sure the rust `asm!` macro supports these variations of the directive

jonathanpwang: I checked the RISC-V ELF spec somewhere, and it seems mapping symbols are added, but I'm not sure it is really used yet

crates/toolchain/transpiler/src/extension.rs
jonathanpwang: my bad, forgot the transpilation process

crates/toolchain/transpiler/src/lib.rs
jonathanpwang: ah I was in favor of making `without_debug_infos` the default and hence remove it from function names in general

extensions/native/guest-macro/src/transportation.rs
jonathanpwang: ahhh

extensions/native/guest-macro/src/transportation.rs
jonathanpwang: mm let's just leave it

extensions/native/transpiler/src/lib.rs
jonathanpwang: I see, right. Yes please do as you said (one calls the other)

extensions/native/transpiler/src/lib.rs
jonathanpwang: discussed offline

benchmarks/programs/test_kernel/kernel/src/main.rs
TlatoaniHJ: Done

extensions/native/serialization/src/lib.rs
TlatoaniHJ: Done

extensions/native/guest-macro/src/transportation.rs
jonathanpwang: it seems rust does not support these nice .insn directives

PR #1193, opened 2025-01-08
Normal format for `custom_insn_*` macros we have

Golovanov399:
This resolves INT-2667.

The following is done here:
- `custom_insn_r` and `custom_insn_i` macros (are now procedural and) now have a clearer signature, where all arguments must be named, and the register types also have to be explicitly specified,
- technically, these macros generate the code under the `#[cfg(target_os = "zkvm")]` block, which would help the developer to catch the wrong macro signature things instantly if they have the language server running, but in reality we also want to `let uninit = ...` under the same block anyway, so now this change is not really represented. Maybe we will want to make `uninit` an additional valid macro argument at some point, but this would overcomplicate things imo.


crates/toolchain/platform/custom_insn/Cargo.toml
jonathanpwang, line 1: I prefer we don't have these nested crate within another crate folder as it gets hard to find stuff. let's just put it in `toolchain/`.
Also whenever adding a new crate, reminder to update `docs/repo/layout.md`

crates/toolchain/platform/custom_insn/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+            while !input.is_empty() && !input.peek(Token![,]) {
+                tokens.extend(TokenStream::from(input.parse::<proc_macro2::TokenTree>()?));
+            }
+            match key.to_string().as_str() {
+                "opcode" => opcode = Some(tokens),
+                "funct3" => funct3 = Some(tokens),
+                _ => unreachable!(),
+            }
+            None
+        } else if key == "rd" || key == "rs1" {
```
jonathanpwang, line 62: rs2 is not special enough?

jonathanpwang: nvm didn't see it says `common` fields

crates/toolchain/platform/custom_insn/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+struct CustomInsnI {
+    pub rd: AsmArg,
+    pub rs1: AsmArg,
+    pub imm: AsmArg,
+    pub opcode: TokenStream,
+    pub funct3: TokenStream,
+}
+
+// Helper function to parse common fields
+#[allow(clippy::type_complexity)]
```
jonathanpwang, line 33:
```suggestion
/// Returns `(rd, rs1, opcode, funct3)`.
#[allow(clippy::type_complexity)]
```

crates/toolchain/platform/custom_insn/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+    handle_reg_arg(&mut template, &mut args, &rd, "rd");
+    handle_reg_arg(&mut template, &mut args, &rs1, "rs1");
+    handle_reg_arg(&mut template, &mut args, &rs2, "rs2");
+
+    let expanded = quote::quote! {
+        #[cfg(target_os = "zkvm")]
+        unsafe {
+            core::arch::asm!(
+                #template,
+                opcode = const #opcode,
```
jonathanpwang, line 325: I find this weird: why do we do `= const` instead of just generating the whole line as #opcode?

crates/toolchain/platform/src/rust_rt.rs
```diff
@@ -17,7 +17,13 @@ extern crate alloc;
 #[inline(always)]
 pub fn terminate<const EXIT_CODE: u8>() {
     #[cfg(target_os = "zkvm")]
-    crate::custom_insn_i!(SYSTEM_OPCODE, 0, "x0", "x0", EXIT_CODE);
+    crate::custom_insn_i!(
+        opcode = SYSTEM_OPCODE,
+        funct3 = 0,
+        rd = Const "x0",
+        rs1 = Const "x0",
+        imm = Const EXIT_CODE
```
jonathanpwang, line 25: does Const work here instead of `sym`?

jonathanpwang: best to add a test to check

jonathanpwang (approved):
Overall this is a big improvement!
LGTM, but just the macro use of `opcode = const #opcode` to insert assembly is a little weird to me.
There is one `const vs sym` question.

Also, does anything in the book/docs need to be updated, or do we not even mention this macro in the book / documentation?

Heads up I will likely merge https://github.com/openvm-org/openvm/pull/1107 first, so there will be a small(hopefully) rebase necessary.

crates/toolchain/platform/src/rust_rt.rs
Golovanov399: `const` works as "use this constant expression". `sym` works as "use this address". I don't really understand your point

Golovanov399:
Btw yes, I think we don't mention these macros anywhere in the docs

crates/toolchain/platform/src/rust_rt.rs
jonathanpwang: sorry got confused because `STACK_TOP` was static and we used `sym STACK_TOP`

PR #1195, opened 2025-01-08
[chore] Make all chip records serde

nyunyunyunyu:
closes INT-2989


crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -161,7 +163,10 @@ where
     }
 }
 
+#[serde_as]
+#[derive(Serialize, Deserialize)]
 pub struct FieldExpressionRecord {
+    #[serde_as(as = "Vec<DisplayFromStr>")]
```
jonathanpwang, line 169: just wondering: is this just to help debugging or we actually need it as string somewhere?

Cargo.toml
```diff
@@ -225,6 +225,7 @@ getset = "0.1.3"
 rrs-lib = "0.1.0"
 rand = { version = "0.8.5", default-features = false }
 hex = { version = "0.4.3", default-features = false }
+serde_arrays = "0.1.0"
```
jonathanpwang, line 228: is this different from https://crates.io/crates/serde-big-array ? 
we were already using that one in some macros, and it seems more maintained (and originally suggested by dtolnay which is a rust expert)

extensions/rv32-adapters/src/vec_heap.rs
```diff
@@ -96,27 +98,32 @@ impl<
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(bound = "F: Field")]
 pub struct Rv32VecHeapReadRecord<
     F: Field,
     const NUM_READS: usize,
     const BLOCKS_PER_READ: usize,
     const READ_SIZE: usize,
 > {
     /// Read register value from address space e=1
+    #[serde_as(as = "[_; NUM_READS]")]
```
jonathanpwang, line 111: it seems there's some inconsistency between using serde_as and serde_array, I don't have a strong opinion, just noting

jonathanpwang (approved):
LGTM except `serde-arrays` should be replaced with either `serde-big-array` or `serde_as`

crates/circuits/mod-builder/src/core_chip.rs
nyunyunyunyu: I'm following the convention(e.g https://github.com/openvm-org/openvm/blob/3a1d8ec837e6a7f25093817d37e9037586f0bcfc/extensions/algebra/circuit/src/modular_extension.rs#L32). @luffykai any consideration for this choice?

extensions/rv32-adapters/src/vec_heap.rs
nyunyunyunyu: I have to use `serde_as` here because `reads` is a 2-d array, which `serde_arrays`/`serde-big-arry` doesn't support.

jonathanpwang: yea i just meant it seems `serde_as` actually seems to also support 1d array too

PR #1198, opened 2025-01-09
feat: parse native asm src

lispc:


jonathanpwang (commented):
This is useful, but `AsmInstruction` is an intermediate representation that is not quite pure OpenVM `Instruction` so I'm not sure we want to keep `AsmInstruction` at all / encourage its usage.

cc @yi-sun @TlatoaniHJ @zlangley

yi-sun:
We want to treat `AsmInstruction` as more of an IR, i.e. we will commit to stabilizing the native extension ISA opcodes relatively soon, but we may add / remove `AsmInstruction` opcodes at a faster pace.

lispc:
yes later i also have same feeling.  seems using rust codes to manipulate "Program<F>" is a better option.

PR #1199, opened 2025-01-09
[chore] Implement Stateful for all Chips

nyunyunyunyu:


crates/circuits/primitives/src/bitwise_op_lookup/mod.rs
```diff
@@ -196,3 +246,34 @@ impl<const NUM_BITS: usize> ChipUsageGetter for BitwiseOperationLookupChip<NUM_B
...
+
+    fn trace_width(&self) -> usize {
+        self.0.trace_width()
+    }
+}
+
+impl<const NUM_BITS: usize> Stateful<Vec<u8>> for SharedBitwiseOperationLookupChip<NUM_BITS> {
+    fn load_state(&mut self, state: Vec<u8>) {
+        // AtomicU32 can be deserialized as u32
+        let (count_range, count_xor): (Vec<u32>, Vec<u32>) = bitcode::deserialize(&state).unwrap();
```
jonathanpwang, line 267: maybe we should resolve the choice of bitcode vs bincode before/during this PR?

crates/circuits/primitives/src/range_tuple/mod.rs
```diff
@@ -100,6 +104,9 @@ pub struct RangeTupleCheckerChip<const N: usize> {
     count: Vec<Arc<AtomicU32>>,
 }
 
+#[derive(Debug, Clone)]
+pub struct SharedRangeTupleCheckerChip<const N: usize>(Arc<RangeTupleCheckerChip<N>>);
```
jonathanpwang, line 108: did you make a newtype for protected type purposes or because of some practical reason around traits?

crates/circuits/sha256-air/src/trace.rs
```diff
@@ -50,7 +50,7 @@ impl Sha256Air {
         trace_width: usize,
         trace_start_col: usize,
         input: &[u32; SHA256_BLOCK_WORDS],
-        bitwise_lookup_chip: &BitwiseOperationLookupChip<8>,
+        bitwise_lookup_chip: SharedBitwiseOperationLookupChip<8>,
```
jonathanpwang, line 53: why are we doing this...? it doesn't really matter, but cloning `Arc` is worse than `&`. 
You could also make `SharedBitwiseOperationLookupChip` implement `Borrow` I think?

jonathanpwang (changes requested):
Wasn't clear why we needed `Stateful...Chip`: does rust prevent you from implementing `Stateful` on `Arc<T>` even when you own `T`?

The fact that you need to specify `bitcode` explicitly in every `load_state` seems bad. Let's discuss if we should update `Stateful` to either pass in a deserializer / serializer (https://github.com/openvm-org/stark-backend/pull/20#issuecomment-2581225266) or you could do `impl Stateful<(Deserializer, Vec<u8>)> for ...` maybe?

nit: You can implement `T` on `StatefulXChip` to prevent some clone.

jonathanpwang (approved):
Approved after responding to comments and resolving choice of `bitcode` vs `bincode`.

PR #1202, opened 2025-01-10
[fix] turn on first row check for setup

luffykai:


jonathanpwang (approved):
going to merge now but we have some separate discussions around cleaning up `setup` more generally and need to figure out how to handle pairing chips

jonathanpwang:
note: this does not count as a security bug in the chips because you can inspect the program to see if setup was ever called. We were previously enforcing setup at the level of program correctness, but of course it's better to be able to enforce it at the chip level.

jonathanpwang:
@luffykai for some reason this failed after pushing to `main` but not in PR

yi-sun:
Looks like the algebra extension test didn't run on this PR -- we should add `crates/circuits/mod-builder/**` to the triggering paths in the workflow.

jonathanpwang:
https://github.com/openvm-org/openvm/pull/1204

PR #1203, opened 2025-01-10
feat: E2E benchmark halo2 generate flamegraphs

stephenh-axiom-xyz:
Implements the following:

- `bench-metrics` feature properly propagates to `openvm-native-compiler` module when set in the `openvm-benchmarks` and/or `openvm-sdk` modules.
- SDK supports profiling for halo2 proof generation.
- Flamegraph generation script a) generates halo2 flamegraphs and b) doesn't attempt to generate flamegraphs if the benchmark doesn't exist

Latest `fib_e2e` run results [here](https://github.com/openvm-org/openvm/blob/benchmark-results/benchmarks-dispatch/refs/heads/feat/bench-halo2-flamegraph/fib_e2e-c421044c1aa6768c3351023c88c186f8857a07c0.md) (note `fixed_cells` is zero throughout and thus doesn't show up).

### DIFF
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -55,7 +55,7 @@ pprof = { version = "0.13", features = [
 
 [features]
 default = ["parallel", "mimalloc", "bench-metrics"]
-bench-metrics = ["openvm-native-recursion/bench-metrics"]
+bench-metrics = ["openvm-native-recursion/bench-metrics", "openvm-native-compiler/bench-metrics"]
 profiling = ["openvm-sdk/profiling"]
 aggregation = []
 static-verifier = ["openvm-native-recursion/static-verifier"]
--- a/ci/scripts/metric_unify/flamegraph.py
+++ b/ci/scripts/metric_unify/flamegraph.py
@@ -50,6 +50,8 @@ def get_stack_lines(metrics_dict, group_by_kvs, stack_keys, metric_name):
 
 def create_flamegraph(fname, metrics_dict, group_by_kvs, stack_keys, metric_name, reverse=False):
     lines = get_stack_lines(metrics_dict, group_by_kvs, stack_keys, metric_name)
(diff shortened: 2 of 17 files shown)


jonathanpwang (commented):
Can we make a single flamegraph to consolidate `lookup_advice_cells + simple_advice_cells`? they should both be counted towards total cell count

benchmarks/Cargo.toml
```diff
@@ -55,7 +55,7 @@ pprof = { version = "0.13", features = [
 
 [features]
 default = ["parallel", "mimalloc", "bench-metrics"]
-bench-metrics = ["openvm-native-recursion/bench-metrics"]
+bench-metrics = ["openvm-native-recursion/bench-metrics", "openvm-native-compiler/bench-metrics"]
```
jonathanpwang, line 58: we should make `bench-metrics` feature in `native-recursion` also turn on `native-compiler/bench-metrics`

ci/scripts/metric_unify/flamegraph.py
```diff
@@ -101,6 +107,12 @@ def create_custom_flamegraphs(metrics_file, group_by=["group"]):
                            reverse=reverse)
         create_flamegraphs(metrics_file, group_by, ["cycle_tracker_span", "dsl_ir", "opcode", "air_name"], "cells_used",
                            reverse=reverse)
+        create_flamegraphs(metrics_file, group_by, ["cell_tracker_span"], "simple_advice_cells",
```
jonathanpwang, line 110: no need to have 3 separate ones, let's just add them all up into one flamegraph

jonathanpwang (approved):
LGTM after addressing comments

benchmarks/Cargo.toml
jonathanpwang: @stephenh-axiom-xyz ok to merge after this

PR #1206, opened 2025-01-11
fix(castf): handle range check correctly for padding rows

lispc:
Fixes the bug below.  The root cause was that the range check interaction in CASTF did not condition the send on row validity.

```
 $ cargo test --package openvm-native-circuit --lib -- castf::tests::castf_rand_test --exact --show-output

running 1 test
test castf::tests::castf_rand_test ... FAILED

successes:

successes:

failures:

---- castf::tests::castf_rand_test stdout ----
Bus 4 failed to balance the multiplicities for fields=[0, 8]. The bus connections for this were:
   Air idx: 2, Air name: VmAirWrapper<ConvertAdapterAir<1, 4>, CastFCoreAir>, interaction type: Send, count: 1
   Air idx: 2, Air name: VmAirWrapper<ConvertAdapterAir<1, 4>, CastFCoreAir>, interaction type: Send, count: 1
   Air idx: 2, Air name: VmAirWrapper<ConvertAdapterAir<1, 4>, CastFCoreAir>, interaction type: Send, count: 1
Bus 4 failed to balance the multiplicities for fields=[0, 6]. The bus connections for this were:
   Air idx: 2, Air name: VmAirWrapper<ConvertAdapterAir<1, 4>, CastFCoreAir>, interaction type: Send, count: 1
thread 'castf::tests::castf_rand_test' panicked at /Users/zhangzhuo/.cargo/git/checkouts/stark-backend-bcfa8b09bc28ca79/8c9f94b/crates/stark-backend/src/air_builders/debug/check_constraints.rs:164:9:
LogUp multiset equality check failed.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    castf::tests::castf_rand_test

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 15 filtered out; finished in 0.13s

error: test failed, to rerun pass `-p openvm-native-circuit --lib`
```

I  found this bug during debuging kernel codes, some simple codes like
```
VmOpcode(293) 44 16775749 0 1 5 0 0 // castf native_as[16775749] to x11 register
VmOpcode(293) 44 16775748 0 1 5 0 0 // repeat 
VmOpcode(293) 44 16775741 0 1 5 0 0 // repeat 
```
will fail.   So i managed to reproduce it using provided unittest.

(repeating castf twice will be ok, 3 times will trigger logup err)

PR #1210, opened 2025-01-12
chore: make CASTF unit test match

jonathanpwang:
https://github.com/openvm-org/openvm/pull/1206#issuecomment-2585623395

It doesn't matter for unit test because unit tester doesn't check against program code, but fixing the unit test so the usage matches how it should be used in practice.

We will more thoroughly address a better handling of offsets soon.


jonathanpwang:
cc @lispc

jonathanpwang:
closing to address in general offset cleanup because other tests just set `offset=0`

PR #1211, opened 2025-01-13
[feat] Implement Stateful for VmChipComplex

nyunyunyunyu:
1st commit:
- Implement `Stateful` for system chips.
- Implement `SharedVariableRangeCheckerChip`.

2st commit:
- Implement `Stateful` for `VmChipComplex`.
- Add `ExecutionSegment::new_for_proving` which constructs a `ExecutionSegment` just for proving.
- Add `ExecutionSegment::store_chip_complex_state` to store states.

Note: some memory states need to clone. We should to try to avoid it in the future.

closes INT-2951

### DIFF
--- a/crates/vm/src/arch/extensions.rs
+++ b/crates/vm/src/arch/extensions.rs
@@ -10,12 +10,12 @@ use getset::Getters;
 use itertools::Itertools;
 #[cfg(feature = "bench-metrics")]
 use metrics::counter;
-use openvm_circuit_derive::{AnyEnum, InstructionExecutor, Stateful};
+use openvm_circuit_derive::{AnyEnum, InstructionExecutor};
 use openvm_circuit_primitives::{
     utils::next_power_of_two_or_zero,
     var_range::{SharedVariableRangeCheckerChip, VariableRangeCheckerBus},
 };
-use openvm_circuit_primitives_derive::{Chip, ChipUsageGetter};
+use openvm_circuit_primitives_derive::{BytesStateful, Chip, ChipUsageGetter};
 use openvm_instructions::{
     program::Program, PhantomDiscriminant, PublishOpcode, SystemOpcode, UsizeOpcode, VmOpcode,
 };
@@ -558,13 +558,13 @@ impl<F: PrimeField32> Stateful<SystemBaseState<F>> for SystemBase<F> {
     }
 }
 
-#[derive(ChipUsageGetter, Chip, AnyEnum, From, InstructionExecutor, Stateful)]
+#[derive(ChipUsageGetter, Chip, AnyEnum, From, InstructionExecutor, BytesStateful)]
 pub enum SystemExecutor<F: PrimeField32> {
     PublicValues(PublicValuesChip<F>),
     Phantom(RefCell<PhantomChip<F>>),
 }
 
-#[derive(ChipUsageGetter, Chip, AnyEnum, From, Stateful)]
+#[derive(ChipUsageGetter, Chip, AnyEnum, From, BytesStateful)]
 pub enum SystemPeriphery<F: PrimeField32> {
     /// Poseidon2 chip with direct compression interactions
     Poseidon2(Poseidon2PeripheryChip<F>),
--- a/crates/vm/src/arch/extensions.rs
+++ b/crates/vm/src/arch/extensions.rs
@@ -12,7 +12,7 @@ use metrics::counter;
 use openvm_circuit_derive::{AnyEnum, InstructionExecutor, Stateful};
 use openvm_circuit_primitives::{
     utils::next_power_of_two_or_zero,
-    var_range::{VariableRangeCheckerBus, VariableRangeCheckerChip},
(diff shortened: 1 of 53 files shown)


crates/vm/src/arch/extensions.rs
```diff
@@ -489,6 +534,30 @@ impl<F: PrimeField32> SystemBase<F> {
     }
 }
 
+impl<F: PrimeField32> Stateful<SystemBaseState<F>> for SystemBase<F> {
```
luffykai, line 537: semi-related but not directly for this PR: shouldn't trait `Stateful` specify the state S should be `Serialize + Deserialize`

crates/vm/tests/integration_test.rs
```diff
@@ -509,6 +512,44 @@ fn test_vm_continuations() {
...
+    }
+    .with_continuations();
+    let engine = BabyBearPoseidon2Engine::new(FriParameters::standard_fast());
+    let vm = VirtualMachine::new(engine, config.clone());
+    let pk = vm.keygen();
+    let segments = vm
+        .executor
+        .execute_segments(program.clone(), Streams::default())
+        .unwrap();
+    let states: Vec<_> = segments
```
luffykai, line 531: for completeness, can you serialize the states and dump to a temp file, and the next line load it back? want to make sure every step works

crates/vm/src/arch/extensions.rs
```diff
@@ -380,6 +391,27 @@ impl<E, P> VmInventory<E, P> {
...
+        for (e, s) in self.executors.iter_mut().zip_eq(state.executors) {
+            e.load_state(s)
+        }
+        for (p, s) in self.periphery.iter_mut().zip_eq(state.periphery) {
+            p.load_state(s)
+        }
+    }
+
+    fn store_state(&self) -> VmInventoryState {
+        // TODO: parallelize this.
```
jonathanpwang, line 405: what's stopping you?

crates/vm/src/arch/extensions.rs
```diff
@@ -431,6 +463,12 @@ pub struct VmChipComplex<F: PrimeField32, E, P> {
     bus_idx_max: usize,
 }
 
+#[derive(Clone, Serialize, Deserialize)]
+pub struct VmChipComplexState<F> {
+    base: SystemBaseState<F>,
```
jonathanpwang, line 468: this is just a matter of where you invoke bitcode, but wouldn't it match the other chip's pattern more if you bitcode deserialized to `SystemBaseState` without load state? otherwise you will need to specify this state in protobuf

crates/vm/src/arch/segment.rs
```diff
@@ -85,6 +85,31 @@ impl<F: PrimeField32, VC: VmConfig<F>> ExecutionSegment<F, VC> {
...
+        program: Program<F>,
+        vm_chip_complex_state: VmChipComplexState<F>,
+    ) -> Self {
+        let mut chip_complex = config.create_chip_complex().unwrap();
+        let program = if !config.system().profiling {
+            program.strip_debug_infos()
+        } else {
+            program
+        };
+        chip_complex.set_program(program);
```
jonathanpwang, line 100: just checking: `chip_complex.load_state` won't overwrite the program will it?

crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -694,6 +694,13 @@ impl<F: PrimeField32> MemoryController<F> {
     pub fn offline_memory(&self) -> Arc<Mutex<OfflineMemory<F>>> {
         self.offline_memory.clone()
     }
+    pub fn get_memory_logs(&self) -> Vec<MemoryLogEntry<F>> {
+        // TODO: can we avoid clone?
```
jonathanpwang, line 698: you could make a separate `take_memory_logs` fn

jonathanpwang (approved):
LGTM after addressing @luffykai 's comments

crates/vm/src/arch/extensions.rs
nyunyunyunyu: I intend to avoid unnecessary serialization because `bitcode` is not zero-copy. The existing implementation only serialize one more time when sending the struct. I wanted to mention this but didn't find a good place for comments.

crates/vm/src/arch/segment.rs
nyunyunyunyu: No. `program` is relatively fixed so I expect the prover server might already have a copy. We can send `program` outside `State` if necessary

crates/vm/src/arch/extensions.rs
nyunyunyunyu: In the trait-level, It seems not a must because `Stateful` can be used in memory(maybe?).

crates/vm/src/system/memory/controller/mod.rs
jonathanpwang: huh why is it not `store_state(&mut self)`: some sync thing?

PR #1212, opened 2025-01-13
[chore] use expr builder for modular chips

luffykai:
follow up from https://github.com/openvm-org/openvm/pull/1204

- modular tests already call `setup_all_moduli`


jonathanpwang (approved):
Sorry I approved this a while ago but lost internet.
LGTM and saves a nice amount of code.

Please fix lint

PR #1216, opened 2025-01-14
draft: Minimal aggregation

ytham:
- Draft of minimal aggregation
- Currently does not remove leaf agg as a separate step


ytham:
Closing for now

PR #1221, opened 2025-01-16
[chore] sdk refactor: separate leaf running + scheduling

luffykai:
ref INT-3046
ref INT-3001

the binary is just skeleton, @Jocn2020 will take over

### DIFF
--- a/benchmarks/src/utils.rs
+++ b/benchmarks/src/utils.rs
@@ -16,7 +16,7 @@ use openvm_sdk::{
         DEFAULT_INTERNAL_LOG_BLOWUP, DEFAULT_LEAF_LOG_BLOWUP, DEFAULT_ROOT_LOG_BLOWUP,
     },
     keygen::{leaf_keygen, AppProvingKey},
-    prover::{AppProver, LeafProver},
+    prover::{vm::local::VmLocalProver, AppProver, LeafProvingController},
     StdIn,
 };
 use openvm_stark_backend::utils::metrics_span;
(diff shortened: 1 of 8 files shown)


crates/sdk/src/bin/app_prover.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
luffykai, line 1: @Jocn2020 will start building this one

benchmarks/src/utils.rs
```diff
@@ -234,8 +234,12 @@ where
         .expect("Verification failed");
     if let Some(leaf_vm_config) = leaf_vm_config {
         let leaf_vm_pk = leaf_keygen(app_config.leaf_fri_params.fri_params, leaf_vm_config);
-        let leaf_prover = LeafProver::new(leaf_vm_pk, app_pk.leaf_committed_exe);
-        leaf_prover.generate_proof(&app_proof);
+        let leaf_prover = VmLocalProver::<SC, NativeConfig, BabyBearPoseidon2Engine>::new(
+            leaf_vm_pk,
+            app_pk.leaf_committed_exe,
+        );
+        let leaf_controller = LeafProvingController { num_children: 1 };
```
jonathanpwang, line 241: a little weird this is being hardcoded here: previously it was from some constructor, or at least using the DEFAULT_* constant

jonathanpwang: oh this is in benchmark: I think there might be a parameter in `args`

jonathanpwang (approved):
LGTM except use benchmark arg for leaf num children if it exists

PR #1224, opened 2025-01-16
[feat] Make online and offline memory use vectors of pages instead of hashmaps

Golovanov399:
This resolves INT-2985 and INT-2984.

Instead of `FxHashMap` for memory, we now use the new structure `PageVec` which is `Vec<Option<Vec<T>>>`, where memory is split into pages of fixed size and we create the whole page when the element is not touched yet, and therefore it should give us faster access to elements.

In reality, this depends on the page size, on the queries locality and probably something else. In particular, for regex, execute time and trace gen time reduced by 12%, but for other benchmarks the results are different.

Among other dependent changes, the default `as_height` is now not 29 but 3, otherwise we would have to create half a billion `PageVec`s every time (and we don't need more than 8 address spaces anyway).


crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -465,6 +469,10 @@ impl<F: PrimeField32> MemoryController<F> {
                 &mut self.access_adapters,
             );
         }
+        eprintln!(
```
Golovanov399, line 472: forgot to remove

crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -241,13 +244,14 @@ impl<F: PrimeField32> MemoryController<F> {
...
                 ),
             },
-            memory: Memory::new(mem_config.access_capacity),
+            memory: Memory::new(&mem_config),
             offline_memory: Arc::new(Mutex::new(OfflineMemory::new(
                 initial_memory,
                 1,
                 memory_bus,
                 range_checker.clone(),
                 mem_config.clk_max_bits,
```
zlangley, line 253: We don't need to pass `clk_max_bits` separately if we are giving `mem_config`.

crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -346,14 +351,14 @@ impl<F: PrimeField32> MemoryController<F> {
...
-        offline_memory.set_initial_memory(memory.clone());
+        offline_memory.set_initial_memory(memory.clone(), self.mem_config);
 
         self.memory = Memory::from_image(memory.clone(), self.mem_config.access_capacity);
 
         match &mut self.interface_chip {
             MemoryInterface::Volatile { .. } => {
                 assert!(
-                    memory.is_empty(),
+                    memory.items().collect::<Vec<_>>().is_empty(),
```
zlangley, line 361: Let's add an `is_empty` function?

Also we should probably do `memory.items().next().is_some()` or something rather than collecting the entire memory into a Vec.

crates/vm/src/system/memory/merkle/tests/mod.rs
```diff
@@ -48,21 +49,24 @@ fn test<const CHUNK: usize>(
...
 
     // checking validity of test data
-    for (&(address_space, pointer), value) in final_memory {
+    for ((address_space, pointer), value) in final_memory.items() {
         let label = pointer / CHUNK as u32;
         assert!(address_space - as_offset < (1 << as_height));
-        assert!(label < (1 << address_height));
-        if initial_memory.get(&(address_space, pointer)) != Some(value) {
+        assert!(pointer < ((CHUNK << address_height).div_ceil(PAGE_SIZE) * PAGE_SIZE) as u32);
+        if initial_memory.get(&(address_space, pointer)).cloned() != Some(value) {
```
zlangley, line 56: nit
```suggestion
        if initial_memory.get(&(address_space, pointer)) != Some(&value) {
```

crates/vm/src/system/memory/merkle/tests/mod.rs
```diff
@@ -48,21 +49,24 @@ fn test<const CHUNK: usize>(
     let merkle_bus = MemoryMerkleBus(MEMORY_MERKLE_BUS);
 
     // checking validity of test data
-    for (&(address_space, pointer), value) in final_memory {
+    for ((address_space, pointer), value) in final_memory.items() {
         let label = pointer / CHUNK as u32;
         assert!(address_space - as_offset < (1 << as_height));
-        assert!(label < (1 << address_height));
-        if initial_memory.get(&(address_space, pointer)) != Some(value) {
+        assert!(pointer < ((CHUNK << address_height).div_ceil(PAGE_SIZE) * PAGE_SIZE) as u32);
```
zlangley, line 55: Isn't the bound just `pointer < (CHUNK << address_height)`? The boundary chip won't support `pointer` outside this range.

crates/vm/src/system/memory/mod.rs
```diff
@@ -6,6 +6,7 @@ pub mod merkle;
 mod offline;
 pub mod offline_checker;
 pub mod online;
+pub mod paged_vec;
```
zlangley, line 9: Does this need to be `pub`?

crates/vm/src/system/memory/offline.rs
```diff
@@ -4,18 +4,21 @@ use openvm_circuit_primitives::{
...
+        adapter::{AccessAdapterRecord, AccessAdapterRecordKind},
+        offline_checker::{MemoryBridge, MemoryBus},
+        MemoryAuxColsFactory, MemoryImage, RecordId, TimestampedEquipartition, TimestampedValues,
+    },
 };
 
 pub const INITIAL_TIMESTAMP: u32 = 0;
 
-#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[derive(Clone, Default, PartialEq, Eq, Debug)]
```
zlangley, line 21: Why remove `Copy`? I believe `Copy` tells the compiler than it can just do `memcpy` when cloning this, which is valid here.

crates/vm/src/system/memory/offline.rs
```diff
@@ -56,12 +60,14 @@ impl<F: PrimeField32> OfflineMemory<F> {
         memory_bus: MemoryBus,
         range_checker: SharedVariableRangeCheckerChip,
         timestamp_max_bits: usize,
+        config: MemoryConfig,
     ) -> Self {
```
zlangley, line 64: note: We should probably stop passing `initial_memory` here since I think it's only used with `set_initial_memory`.

crates/vm/src/system/memory/offline.rs
```diff
@@ -306,10 +333,17 @@ impl<F: PrimeField32> OfflineMemory<F> {
...
+                self.block_data.insert(
+                    &(address_space, pointer + i),
+                    Self::initial_block_data(pointer + i, self.initial_block_size),
+                );
+            }
             let block = self
                 .block_data
-                .entry((address_space, pointer + i))
-                .or_insert_with(|| Self::initial_block_data(pointer + i, self.initial_block_size));
+                .get_mut(&(address_space, pointer + i))
```
zlangley, line 345: Ideally we would do `get_mut` on line 336 so we don't need to do it here, but I guess this be problematic for the borrow checker. We should find a way to do this without so many accesses though.

crates/vm/src/system/memory/offline.rs
```diff
@@ -4,18 +4,21 @@ use openvm_circuit_primitives::{
...
+    },
 };
 
 pub const INITIAL_TIMESTAMP: u32 = 0;
 
-#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[derive(Clone, Default, PartialEq, Eq, Debug)]
 struct BlockData {
     pointer: u32,
     size: usize,
```
zlangley, line 24: Can you explain a bit how you are now using `size == 0`?

crates/vm/src/system/memory/paged_vec.rs
```diff
@@ -0,0 +1,368 @@
...
+                self.pages[start_page_idx].get_or_insert_with(|| vec![T::default(); PAGE_SIZE]);
+            let page_start = range.start - range.start % PAGE_SIZE;
+            let result = page[range.start - page_start..range.end - page_start].to_vec();
+            for (j, value) in range.zip(values.into_iter()) {
+                page[j - page_start] = value.clone();
+            }
+            result
+        } else {
+            // TODO: This can be more efficient by copying into two slices (but most queries should
+            // not be cross-page).
```
zlangley, line 90: Didn't you have an implementation of this in your other branch?

crates/vm/src/system/memory/paged_vec.rs
```diff
@@ -0,0 +1,368 @@
...
+        let end_page_idx = range.end / PAGE_SIZE;
+
+        if start_page_idx == end_page_idx {
+            let page =
+                self.pages[start_page_idx].get_or_insert_with(|| vec![T::default(); PAGE_SIZE]);
+            let page_start = range.start - range.start % PAGE_SIZE;
+            let result = page[range.start - page_start..range.end - page_start].to_vec();
+            for (j, value) in range.zip(values.into_iter()) {
+                page[j - page_start] = value.clone();
+            }
```
zlangley, line 86: Can't we do `copy_from_slice` or something?

crates/vm/src/system/memory/paged_vec.rs
```diff
@@ -0,0 +1,368 @@
...
+    ) -> Vec<T>
+    where
+        T: 'a,
+    {
+        let start_page_idx = range.start / PAGE_SIZE;
+        let end_page_idx = range.end / PAGE_SIZE;
+
+        if start_page_idx == end_page_idx {
+            let page =
+                self.pages[start_page_idx].get_or_insert_with(|| vec![T::default(); PAGE_SIZE]);
```
zlangley, line 81: Since we're going to copy the whole thing, we don't need to spend time initializing to default. We should optimize this.

crates/vm/src/system/memory/offline.rs
Golovanov399: `size == 0` is equivalent to "this block was created by default and should be replaced with `Self::initial_block_data`"

crates/vm/src/system/memory/paged_vec.rs
Golovanov399: Yes, but when I tried to apply it yesterday, it became slower. But maybe now it will make no difference

Golovanov399: The time deltas are inconsistent and make no sense

crates/vm/src/system/memory/mod.rs
Golovanov399: We need `Address`, `AddressMap` and `PAGE_SIZE` in other places. I can expose them I guess, but won't it be less structured?

crates/vm/src/system/memory/offline.rs
Golovanov399: I don't like the idea of implicit copying of eg `BlockData`s. I think we just need to add `copy_from_slice` or something wherever we want fast data copying

Golovanov399: Argh `copy_from_slice` requires `Copy`

zlangley: There is `clone_from_slice` I think. Still, I don't see the issue with making this type `Copy`. It is up to the compiler whether it wants to actually copy or not, and it's only 3 words if it wants to copy.

crates/vm/src/system/memory/offline.rs
zlangley: Why do we create with `size = 0` rather than `initial_block_data`?

crates/vm/src/system/memory/merkle/tests/mod.rs
zlangley: But that would still be an invalid `pointer`?

crates/vm/src/system/memory/offline.rs
Golovanov399: Because I don't know how to tell `PagedVec` "fill it with ermmm different values based on where it is"

crates/vm/src/system/memory/merkle/tests/mod.rs
Golovanov399: Why?

crates/vm/src/system/memory/merkle/tests/mod.rs
Golovanov399: Hehe you tricked me, this is wrong because equal `BabyBear`s can have different references (I thought about it when I wrote this `cloned()` but not now)

crates/vm/src/system/memory/merkle/tests/mod.rs
zlangley: If address_height is 27 and CHUNK is 8, then the max addressable cell is 2^30 - 1. If I understand correctly, this assertion might allow me to address into 2^30 or higher, if the `PAGE_SIZE` doesn't evenly divide 2^30. From the perspective of PagedVec, that's a fine index. But from the perspective of memory, that's not a valid pointer.

Or what am I missing?

crates/vm/src/system/memory/offline.rs
zlangley: Oh, PagedVec requires default. Got it.

crates/vm/src/system/memory/paged_vec.rs
zlangley: oh sorry, this is initializing the entire page, so we don't copy the whole thing.

crates/vm/src/system/memory/merkle/tests/mod.rs
Golovanov399: Yes, but how does it break anything here?

zlangley: This is a test, so I'm not really that concerned. But the point of the assertion is to assert that `pointer` is a valid pointer. And valid pointers are in the range `0..CHUNK << address_height` and, in particular, have nothing to do with the `PAGE_SIZE` of the underlying `PagedVec`.

crates/vm/src/system/memory/online.rs
```diff
@@ -117,7 +113,7 @@ impl<F: PrimeField32> Memory<F> {
     }
 
     fn range_array<const N: usize>(&self, address_space: u32, pointer: u32) -> [F; N] {
```
jonathanpwang, line 115: idk if it matters, do `#[inline(always)]` ?

crates/vm/src/system/memory/paged_vec.rs
```diff
@@ -0,0 +1,354 @@
...
+        if let Some(page) = self.pages[page_idx].as_mut() {
+            Some(std::mem::replace(&mut page[index % PAGE_SIZE], value))
+        } else {
+            let page = self.pages[page_idx].get_or_insert_with(|| vec![T::default(); PAGE_SIZE]);
+            page[index % PAGE_SIZE] = value;
+            None
+        }
+    }
+
+    pub fn get_range(&self, range: Range<usize>) -> Vec<T> {
```
jonathanpwang, line 48: idk if it matters, do `#[inline(always)]` ?

jonathanpwang (approved):
LGTM, but I'd like us to make `get_range` more performant (or what I feel like is the more performant implementation) before merging

crates/vm/src/system/memory/paged_vec.rs
zlangley: The slice `page[in_page_start..in_page_end]` can be made `unsafe`.

PR #1225, opened 2025-01-16
chore: replace num-bigint-dig with num-bigint throughout

stephenh-axiom-xyz:
Replaces usage of `num-bigint-dig` with the more standard `num-bigint`.


stephenh-axiom-xyz:
Note that I included patches in `.cargo` in order to pass the lint, as main OpenVM git branch still uses `num-bigint-dig`. It needs to be removed before merging.

jonathanpwang (approved):
LGTM

PR #1226, opened 2025-01-17
feat: Customizable static verifier public values handler

ytham:
Adds a way to optionally customize the static verifier's public values handler

### DIFF
--- a/crates/sdk/src/static_verifier/mod.rs
+++ b/crates/sdk/src/static_verifier/mod.rs
@@ -9,6 +9,7 @@ use openvm_native_recursion::{
     hints::Hintable,
     stark::StarkVerifier,
     utils::const_fri_config,
+    vars::StarkProofVariable,
     witness::Witnessable,
 };
 use openvm_stark_sdk::{
@@ -24,7 +25,9 @@ use crate::{
     keygen::RootVerifierProvingKey,
     prover::{vm::SingleSegmentVmProver, RootVerifierLocalProver},
     verifier::{
-        common::assert_single_segment_vm_exit_successfully_with_connector_air_id,
+        common::{
+            assert_single_segment_vm_exit_successfully_with_connector_air_id, types::SpecialAirIds,
+        },
         root::types::{RootVmVerifierInput, RootVmVerifierPvs},
     },
     RootSC, F, SC,
@@ -39,7 +42,7 @@ impl RootVerifierProvingKey {
     ) -> Halo2VerifierProvingKey {
         let mut witness = Witness::default();
         root_proof.write(&mut witness);
-        let dsl_operations = build_static_verifier_operations(self, &root_proof);
+        let dsl_operations = Self::build_static_verifier_operations(self, &root_proof);
         Halo2VerifierProvingKey {
             pinning: Halo2Prover::keygen(params, dsl_operations.clone(), witness),
             dsl_ops: dsl_operations,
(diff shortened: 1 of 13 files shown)


jonathanpwang:
@nyunyunyunyu please review

nyunyunyunyu:
> @nyunyunyunyu please review

kk

crates/sdk/src/static_verifier/mod.rs
```diff
@@ -135,16 +142,36 @@ fn build_static_verifier_operations(
...
+            builder.get(&commit_arr, 0)
+        } else {
+            unreachable!()
+        };
+        let expected_program_commit: [Bn254Fr; 1] = root_verifier_pk
+            .root_committed_exe
+            .get_program_commit()
+            .into();
+        builder.assert_var_eq(commit, expected_program_commit[0]);
+    }
```
nyunyunyunyu, line 171: what about the app exe commitment? Has it already been asserted inside root program?

nyunyunyunyu (commented):
Can we have a test case?

extensions/native/recursion/src/halo2/wrapper.rs
```diff
@@ -127,8 +127,7 @@ impl Halo2WrapperProvingKey {
         );
         assert_eq!(
             self.pinning.metadata.num_pvs[0],
-            // 12 is the number of public values for the accumulator
-            snark_to_verify.instances[0].len() + 12
+            snark_to_verify.instances[0].len() + 12,
```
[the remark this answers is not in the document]
ytham: sorry, i had added and removed something and the comment went with it. will restore the comment

crates/sdk/src/static_verifier/mod.rs
ytham: so there's now this `StaticVerifierPvHandler` trait that provides a default implementation (which exposes the app exe commit and leaf commit, same as before): https://github.com/openvm-org/openvm/pull/1226/files#diff-462d15dd85872c0c3ba4c3c0e8e1acb56421a28f32ab6503d85e8dddc21f1f73R97

And the PR in the repo that uses this trait to override the default implementation and assert the commits and split the other PVs to hi-lo

ytham:
> Can we have a test case?

yes, will add one

crates/sdk/src/static_verifier/mod.rs
ytham: hm, but doesn't this need to support both minimal aggregation and standard aggregation? I've added this:
https://github.com/openvm-org/openvm/pull/1226/files#diff-462d15dd85872c0c3ba4c3c0e8e1acb56421a28f32ab6503d85e8dddc21f1f73R39

where we can use it for standard aggregation if set to `None` and use the custom handler if it is `Some`: https://github.com/openvm-org/openvm/pull/1226/files#diff-462d15dd85872c0c3ba4c3c0e8e1acb56421a28f32ab6503d85e8dddc21f1f73R122

adding some comments as well

nyunyunyunyu: Ah I see, the new `StaticVerifierPvHandler` hasn't been implemented yet. NVM then

ytham:
@nyunyunyunyu added test that shows how it's used (same pattern in the other repo PR): https://github.com/openvm-org/openvm/pull/1226/files#diff-e8080e6de56d5f572a1f800ab4ebba8d622c988ae8e5bce485c5067ca5528bfaR267

nyunyunyunyu (approved):

PR #1227, opened 2025-01-17
fix: WeierstrassConfig zip moduli

jonathanpwang:
WeierstrassConfig (only used for testing) auto creates the moduli. It is more natural to do `coordinate,scalar` per curve instead of doing all coordinate first.


jonathanpwang:
@luffykai this one is fixed, can you take a look and then merge

extensions/ecc/circuit/src/weierstrass_chip/double.rs
```diff
@@ -205,6 +205,12 @@ where
         let limb_bits = self.air.expr.canonical_limb_bits();
         let Instruction { opcode, .. } = instruction.clone();
         let local_opcode_idx = opcode.local_opcode_idx(self.air.offset);
+        println!(
```
luffykai, line 208: rm

luffykai (approved):

PR #1228, opened 2025-01-17
range check: add debug_assert

lispc:
is this a good idea?


jonathanpwang:
I think we have negative tests that you'd need to deal with somehow. maybe you can use some static variable when #[cfg(debug_assertions)]?

lispc:
ok negative tests..

lispc:
i feel in future we may design a systematic way for these checks.  Either env var, or another feature flag, or a boolean const generic.   I'd like to close it now.

PR #1234, opened 2025-01-18
chore: clean up for loops and breaks

yi-sun:
- Removes breaks
- Removes RangeBuilder, old IteratorBuilder, `builder.iter()`
- Renames `compile_zip!` to `iter_zip!`

Closes INT-3015


extensions/native/compiler/src/asm/compiler.rs
```diff
@@ -815,7 +781,8 @@ impl<F: PrimeField32 + TwoAdicField, EF: ExtensionField<F> + TwoAdicField> IfCom
     }
 }
 
-// Zipped for loop -- loop extends over the first entry in starts and ends
+// Zipped for loop -- loop extends over the first entry in starts and end0
+// ATTENTION: starting with starts[0] > end0 will lead to undefined behavior.
```
jonathanpwang, line 785: This is a bit scary in the VM case: is there any way we could prevent this at compile time? 
Is it basically only used to iterate through arrays and in the `range(0, ?)` cases (where it's safe)?

jonathanpwang (approved):
LGTM, truly was a cleanup, no behavior changes.

extensions/native/compiler/src/asm/compiler.rs
yi-sun: Correct, only via `builder.zip()` and `builder.range()`

PR #1239, opened 2025-01-19
fix: EcDoubleChip dummy row

jonathanpwang:
We need the dummy row to have `is_valid=0, is_double=0`. Previously we did this by duplicating the first row. However when the same chip is used across continuation segments, the first row might not be setup. So now we explicitly generate the dummy row to have the above properties.

### DIFF
--- a/extensions/ecc/circuit/src/weierstrass_chip/double.rs
+++ b/extensions/ecc/circuit/src/weierstrass_chip/double.rs
@@ -2,7 +2,7 @@ use std::{cell::RefCell, iter, rc::Rc};
 
 use itertools::{zip_eq, Itertools};
 use num_bigint::BigUint;
-use num_traits::One;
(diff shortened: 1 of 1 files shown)


extensions/ecc/circuit/src/weierstrass_chip/double.rs
```diff
@@ -277,15 +277,40 @@ where
...
+    ) -> Vec<F> {
+        let record = EcDoubleCoreRecord {
+            x: BigUint::zero(),
+            y: BigUint::zero(),
+            is_double_flag: false,
+        };
+        let mut row = vec![F::ZERO; adapter_width + core_width];
+        let core_row = &mut row[adapter_width..];
+        // We **do not** want this trace row to update the range checker
+        // so we must create a temporary range checker
```
luffykai, line 303: 🧠

PR #1242, opened 2025-01-20
[feat] opcode offsets redesign

Golovanov399:
This resolves INT-2355 to some extent.

While this doesn't eliminate the need for the offsets completely, this change presents an API for opcode mapping at the cost of need to define the new `VmCoreAir` trait function `start_offset(&self) -> usize`. After that one can use `opcode_to_global_expr` or `expr_to_global_expr` to map from local opcode to global in the air.

In most of the chips though the `offset` is no more stored and the default offset is used instead.

To convert from global to local, we can still use the `VmOpcode::local_opcode_idx(offset)` method.

Among other things:
- `UsizeOpcode` is renamed to `LocalOpcode`
- `UsizeOpcode::default_offset()` is now `LocalOpcode::CLASS_OFFSET`,
- `UsizeOpcode::as_usize()` is now `LocalOpcode::local_usize()` to avoid confusion,
- `UsizeOpcode::with_default_offset()` is now `LocalOpcode::global_opcode()`,
- All of the above, including the `LocalOpcode` trait itself, can be renamed easily.


extensions/algebra/circuit/src/modular_extension.rs
```diff
@@ -88,8 +88,8 @@ impl<F: PrimeField32> VmExtension<F> for ModularExtension {
         for (i, modulus) in self.supported_modulus.iter().enumerate() {
             // determine the number of bytes needed to represent a prime field element
             let bytes = modulus.bits().div_ceil(8);
-            let class_offset = Rv32ModularArithmeticOpcode::default_offset()
-                + i * Rv32ModularArithmeticOpcode::COUNT;
+            let class_offset =
```
jonathanpwang, line 91: should we rename this `start_offset`?

extensions/algebra/circuit/src/fp2_extension.rs
```diff
@@ -79,7 +79,7 @@ impl<F: PrimeField32> VmExtension<F> for Fp2Extension {
         for (i, modulus) in self.supported_modulus.iter().enumerate() {
             // determine the number of bytes needed to represent a prime field element
             let bytes = modulus.bits().div_ceil(8);
-            let class_offset = Fp2Opcode::default_offset() + i * Fp2Opcode::COUNT;
+            let class_offset = Fp2Opcode::CLASS_OFFSET + i * Fp2Opcode::COUNT;
```
jonathanpwang, line 82: `start_offset` ?

extensions/rv32im/circuit/src/loadstore/core.rs
```diff
@@ -70,19 +73,25 @@ pub struct LoadStoreCoreRecord<F, const NUM_CELLS: usize> {
...
 }
 
-impl<F: Field, const NUM_CELLS: usize> BaseAirWithPublicValues<F> for LoadStoreCoreAir<NUM_CELLS> {}
+impl<F: Field, OpcodeClass: LocalOpcode + Sync, const NUM_CELLS: usize> BaseAirWithPublicValues<F>
+    for LoadStoreCoreAir<OpcodeClass, NUM_CELLS>
+{
+}
 
-impl<AB, I, const NUM_CELLS: usize> VmCoreAir<AB, I> for LoadStoreCoreAir<NUM_CELLS>
+impl<AB, I, OpcodeClass: LocalOpcode + Sync, const NUM_CELLS: usize> VmCoreAir<AB, I>
```
jonathanpwang, line 93: this `OpcodeClass` as generic seems to differ from how we handle this in other chips, was there a reason for this?

jonathanpwang (approved):
LGTM except:
- some variables named `class_offset` should be `start_offset`
- the loadstore core chip's usage of `LocalOpcode` as a generic differs from how it's handled in e.g., ALU. is there a reason or is this some leftover?

extensions/rv32im/circuit/src/loadstore/core.rs
Golovanov399: This is really redundant, it's the first such change and back then I thought that there are only a few such cases

PR #1243, opened 2025-01-21
feat: expose aggregation stark proof inside sdk

lispc:


jonathanpwang:
@luffykai how does this align with your plans to separate out the sdk?

crates/sdk/src/prover/agg.rs
```diff
@@ -87,6 +87,20 @@ impl AggStarkProver {
         })
     }
 
+    /// Generate a root stark proof to aggregate app proofs.
+    pub fn generate_root_proof(
```
luffykai, line 91: I think this shouldn't be called root_proof, as it sounds like the output of root verifier?
Should just call generate_root_verifier_input to match the struct type.

luffykai:
> @luffykai how does this align with your plans to separate out the sdk?

not directly related. This is still "coupled", but just exposing one of the layers. But also not make it harder.
But i think it will be easier or not needed after the refactor.

jonathanpwang:
Let's revisit this after our SDK refactor.

lispc:
sounds  ok

jonathanpwang:
@luffykai can you take a look again and see if we should merge this?

luffykai:
yea we can merge after naming fixed https://github.com/openvm-org/openvm/pull/1243#discussion_r1924217127

luffykai:
i think your branch is not in this repo so i can't make changes directly, so redoing here https://github.com/openvm-org/openvm/pull/1272

PR #1244, opened 2025-01-21
[chore] Optimize `alpha_pow` computation

nyunyunyunyu:
- `alpha_pow` is only computed once for each matrix.
- Let's say `w_i` is the width of i-th matrix. Before we have `max(w_i)` multiplication for `alpha_pow` and now we only need `sum(log(w_i))`. 
- `range_with_step` and `step_sizes` are added for a reverted change. It's not necessary but good to have. 
- Fixed a bug in `for_each_unrolled`.
- Based on metrics, we might revert the change on the static verifier part.

### DIFF
--- a/extensions/native/recursion/src/fri/two_adic_pcs.rs
+++ b/extensions/native/recursion/src/fri/two_adic_pcs.rs
@@ -66,17 +66,14 @@ pub fn verify_two_adic_pcs<C: Config>(
         challenger.observe_slice(builder, final_poly_elem_felts);
     });
 
-    let num_query_proofs = proof.query_proofs.len().clone();
-    builder
-        .if_ne(num_query_proofs, RVar::from(config.num_queries))
-        .then(|builder| {
-            builder.error();
-        });
+    // **ATTENTION**: always check shape of user inputs.
+    builder.assert_eq::<Usize<_>>(proof.query_proofs.len(), RVar::from(config.num_queries));
 
     challenger.check_witness(builder, config.proof_of_work_bits, proof.pow_witness);
 
     let log_max_height =
         builder.eval_expr(proof.commit_phase_commits.len() + RVar::from(log_blowup));
+    let round_alpha_pows = compute_round_alpha_pows(builder, rounds.clone(), alpha);
 
     iter_zip!(builder, proof.query_proofs).for_each(|ptr_vec, builder| {
         let query_proof = builder.iter_ptr_get(&proof.query_proofs, ptr_vec[0]);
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -4462,6 +4462,7 @@ dependencies = [
 [[package]]
 name = "openvm-stark-backend"
 version = "0.2.0-alpha"
+source = "git+https://github.com/openvm-org/stark-backend.git?rev=d2788c#d2788c71e68aa386fc8c0d8f258c25557a1bbb09"
```
zlangley, line 4465: Is this just pinning the version because we weren't before?

extensions/native/recursion/src/fri/two_adic_pcs.rs
```diff
@@ -99,178 +96,167 @@ pub fn verify_two_adic_pcs<C: Config>(
...
+                        let ps_at_z = builder.iter_ptr_get(&mat_values, ptr_vec[1]);
+
+                        builder.cycle_tracker_start("single-reduced-opening-eval");
+
+                        if builder.flags.static_only {
+                            let n: Ext<C::F, C::EF> = builder.constant(C::EF::ZERO);
+                            let width = ps_at_z.len().value();
+                            for i in 0..width {
+                                // In the reverse order
+                                let t = width - i - 1;
```
zlangley, line 230:
```suggestion
                            for t in (0..width).rev() {
```

nyunyunyunyu:
> LGTM, but isn't it easier (and more efficient) to just keep one alpha_pow that accumulates over all matrices, rather than factoring to alpha powers per matrix?

Let `w_i` be the width of i-th matrix. "accumulates over all matrices" is `max(w_i)` and "factoring to alpha powers per matrix" is `sum(log(w_i))`. The latter should be more efficient since Poseidon2 has 300+ columns

extensions/native/recursion/src/fri/two_adic_pcs.rs
```diff
@@ -382,6 +368,61 @@ where
...
+                builder.eval(ret)
+            } else {
+                let width = width.get_var();
+                // This is dynamic only so safe to cast.
+                let width_f = builder.unsafe_cast_var_to_felt(width);
+                let bits = builder.num2bits_f(width_f, MAX_LOG_WIDTH as u32);
+                let ret: Ext<_, _> = builder.eval(C::EF::ONE.cons());
+                for i in 0..MAX_LOG_WIDTH {
+                    let bit = builder.get(&bits, i);
+                    builder.if_eq(bit, RVar::one()).then(|builder| {
```
yi-sun, line 413: We should check if it is better to use `if_eq` + conditional multiply or `select` and unconditional multiply.

zlangley:
> > LGTM, but isn't it easier (and more efficient) to just keep one alpha_pow that accumulates over all matrices, rather than factoring to alpha powers per matrix?
> 
> Let `w_i` be the width of i-th matrix. "accumulates over all matrices" is `max(w_i)` and "factoring to alpha powers per matrix" is `sum(log(w_i))`. The latter should be more efficient since Poseidon2 has 300+ columns

Hmm, I feel like you are only counting the work in computing the matrix alpha powers? My understanding is that we're computing (correct me if I'm wrong)
```
a^0 x0 + a^1 x1 + a^2 x2 + ...
```
via
```
a^0 (a^0 x0 + ... + a^{w0 - 1} x_{w0 - 1}) + a^{w_0} (a^0 y0 + ... + a^{w1 - 1} y_{w1 - 1}) + a^{w0 + w1} * ( ... ) + ...
```
In the latter form, we can precompute the powers of `a` so that each `a^w_i` is a table lookup. Then we compute a^{w0 + ... } by multiplying one of the lookups by the previous alpha power. But can't we just do a Horner evaluation of the entire sum `a^0 x0 + a^1 x1 + a^2 x2 + ...` and skip the preprocessing entirely?

yi-sun:
> ```
> a^0 (a^0 x0 + ... + a^{w0 - 1} x_{w0 - 1}) + a^{w_0} (a^0 y0 + ... + a^{w1 - 1} y_{w1 - 1}) + a^{w0 + w1} * ( ... ) + ...
> ```

I believe it is actually 

> ```
> a^0 (a^0 x0 + ... + a^{w0 - 1} x_{w0 - 1}) / b_0 + a^{w_0} (a^0 y0 + ... + a^{w1 - 1} y_{w1 - 1}) / b_1 + a^{w0 + w1} * ( ... ) / b_2 + ...
> ```

So we need to do separate Horner evaluations for each chunk.

extensions/native/recursion/src/fri/two_adic_pcs.rs
nyunyunyunyu: Will benchmark in a follow-up PR

yi-sun (approved):
LGTM, we can address the perf items in a followup to not block the FriReducedOpening PR

PR #1245, opened 2025-01-21
[feat] support custom segmentation strategy

luffykai:
based on discussion from https://github.com/openvm-org/openvm/pull/1217

fix INT-3005

### DIFF
--- a/crates/sdk/src/prover/mod.rs
+++ b/crates/sdk/src/prover/mod.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use openvm_circuit::arch::VmConfig;
+use openvm_circuit::arch::{DefaultSegmentationStrategy, VmConfig};
 use openvm_native_recursion::halo2::EvmProof;
 use openvm_stark_sdk::openvm_stark_backend::Chip;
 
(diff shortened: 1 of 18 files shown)


crates/sdk/src/prover/mod.rs
```diff
@@ -59,6 +59,16 @@ impl<VC> ContinuationProver<VC> {
         self
     }
 
+    pub fn set_max_segment_len(&mut self, max_segment_len: usize) -> &mut Self {
+        self.stark_prover
```
luffykai, line 63: should refactor somehow

extensions/native/circuit/src/extension.rs
```diff
@@ -67,8 +67,7 @@ impl NativeConfig {
                     ..Default::default()
                 },
                 num_public_values,
-            )
-            .with_max_segment_len((1 << 24) - 100),
```
luffykai, line 71: hmm i am not sure where should put this.
It might be more ideal for this segment strategy to be vm_config so all the "config values" are in one place... but couldn't not make it serde

luffykai:
for #1254

PR #1248, opened 2025-01-22
[chore] Optimize FriReducedOpeningChip

nyunyunyunyu:


extensions/native/circuit/src/utils.rs
```diff
@@ -7,7 +7,8 @@ use crate::{Native, NativeConfig};
 pub fn execute_program(program: Program<BabyBear>, input_stream: impl Into<Streams<BabyBear>>) {
     let system_config = SystemConfig::default()
         .with_public_values(4)
-        .with_max_segment_len((1 << 25) - 100);
+        .with_max_segment_len((1 << 25) - 100)
+        .with_profiling();
```
yi-sun, line 11: Should this be on by default?

extensions/native/compiler/src/asm/instruction.rs
```diff
@@ -116,8 +116,8 @@ pub enum AsmInstruction<F, EF> {
     /// (a, b, c) are memory pointers to (dst, lhs, rhs)
     Poseidon2Compress(i32, i32, i32),
 
-    /// (a, b, res, len, alpha, alpha_pow)
-    FriReducedOpening(i32, i32, i32, i32, i32, i32),
+    /// (a, b, res, len, alpha)
+    FriReducedOpening(i32, i32, i32, i32, i32),
```
yi-sun, line 120: Can we remove the 7th operand globally now?

extensions/native/compiler/tests/fri_ro_eval.rs
```diff
@@ -64,19 +63,15 @@ fn test_single_reduced_opening_eval() {
     let ext_1210 = builder.constant(EF::from_base_slice(&[F::ONE, F::TWO, F::ONE, F::ZERO]));
     builder.print_e(ext_1210);*/
```
yi-sun, line 64: above seems extraneous

extensions/native/compiler/tests/fri_ro_eval.rs
```diff
@@ -64,19 +63,15 @@ fn test_single_reduced_opening_eval() {
...
     builder.assign(&cur_alpha_pow, initial_alpha_pow);
-    let single_ro_eval_res =
-        builder.fri_single_reduced_opening_eval(alpha, cur_alpha_pow, &mat_opening, &ps_at_z);
-    let actual_final_alpha_pow = cur_alpha_pow;
+    let single_ro_eval_res = builder.fri_single_reduced_opening_eval(alpha, &mat_opening, &ps_at_z);
     let actual_result: Ext<_, _> = builder.uninit();
-    builder.assign(&actual_result, single_ro_eval_res / (z - x));
+    builder.assign(&actual_result, single_ro_eval_res * cur_alpha_pow / (z - x));
 
     //builder.print_e(actual_result);
```
yi-sun, line 71: remove extra print

yi-sun:
Can you also update the ISA documentation?

extensions/native/compiler/src/asm/instruction.rs
nyunyunyunyu: No - now the new `VerifyBatch` chip is using it.

docs/specs/ISA.md
```diff
@@ -492,20 +572,25 @@ size `CHUNK`, depending on the output size of the corresponding opcode.
 
 ### Verification
 
-We have the following special opcode tailored to optimize verification. Due to already having a large number of operands,
+We have the following special opcode tailored to optimize verification. Due to already having a large number of
+operands,
```
yi-sun, line 576: This is only for `VERIFY_BATCH`, language needs to be changed for `FRI_REDUCED_OPENING`

extensions/native/circuit/src/fri/mod.rs
```diff
@@ -36,262 +33,398 @@ use openvm_stark_backend::{
...
+
+    fn eval_instruction2_row<AB: InteractionBuilder>(
+        &self,
+        builder: &mut AB,
+        local_slice: &[AB::Var],
+        next_slice: &[AB::Var],
+    ) {
+        let local: &Instruction2Cols<AB::Var> = local_slice[..INS_2_WIDTH].borrow();
+        let next: &WorkloadCols<AB::Var> = next_slice[..WL_WIDTH].borrow();
+        {
```
TlatoaniHJ, line 385: Not sure why this is necessary, shouldn't it already be covered by lines 302-303?

nyunyunyunyu: For example, the last row is a `WorkloadCols` with idx = -1

nyunyunyunyu: Maybe this case will be avoided by other constraints. But this sets an explicit boundary condition just for more safety

extensions/native/circuit/src/fri/mod.rs
```diff
@@ -36,262 +33,398 @@ use openvm_stark_backend::{
...
+        self.memory_bridge
+            .read(
+                MemoryAddress::new(local_data.addr_space, next.data.b_ptr),
+                local.b,
+                start_timestamp + ptr_reads + AB::Expr::ONE,
+                &local.b_aux,
+            )
+            .eval(builder, multiplicity);
+        {
+            let mut when_transition = builder.when_transition();
```
TlatoaniHJ, line 229: Not sure why it's necessary to restrict to `when_transition` here -- didn't we already restrict that first row has to be the first workload row of a FriReducedOpening, so last row necessary cannot be?

extensions/native/circuit/src/fri/tests.rs
```diff
@@ -10,17 +10,15 @@ use openvm_stark_backend::{
...
+use crate::OVERALL_WIDTH;
 
 fn compute_fri_mat_opening<F: Field>(
     alpha: [F; EXT_DEG],
-    mut alpha_pow: [F; EXT_DEG],
     a: &[F],
     b: &[[F; EXT_DEG]],
-) -> ([F; EXT_DEG], [F; EXT_DEG]) {
+) -> [F; EXT_DEG] {
+    let mut alpha_pow: [F; EXT_DEG] = elem_to_ext(F::ONE);
```
zlangley, line 21: nit: Shouldn't we be using `EF` in general here rather than `[F; EXT_DEG]`?

extensions/native/circuit/src/fri/mod.rs
nyunyunyunyu: Ah you are right

extensions/native/circuit/src/fri/tests.rs
nyunyunyunyu: What is `EF`?

extensions/native/circuit/src/fri/mod.rs
TlatoaniHJ: I think the case that the last row is a workload row with index -1 is prevented by the fact that an instruction1 row is always followed by an instruction2 row (lines 302-303), an instruction2 row is always followed by a workload row with index 0 (line 403), and a disabled row other than the last is always followed by another disabled row (line 193), so a workload row with index -1 can only be preceded by a workload row with index -2, and so on.

The condition you have here is anyway not preventing workload rows with index -1 that appear in places other than the last row. I guess we can keep it just in case but if you can verify that it's not needed I think it's better to remove to make it less confusing later

TlatoaniHJ:
I looked through the AIR constraints and it seems to be sound -- I noted two places above where I think there are unnecessary constraints but they are not harmful.

extensions/native/circuit/src/fri/tests.rs
nyunyunyunyu: Hmm how should I change? `[F; EXT_DEG]` is not a generic

zlangley: We don't need to change it here. But we would make `FriReducedOpeningChip` also generic over `EF`, which would have the trait bound `EF: ExtensionField<F>`. Anyway, I'm not sure we want to do it; I suppose the AIR stills needs to constrain field extension multiplication for a very specific field extension, not a generic one.

TlatoaniHJ (approved):
Will merge, above comments can be addressed in future

PR #1251, opened 2025-01-22
[feat] support execute until segment

luffykai:
fix INT-3004

### DIFF
--- a/crates/vm/src/arch/vm.rs
+++ b/crates/vm/src/arch/vm.rs
@@ -74,6 +74,18 @@ pub struct VmExecutorResult<SC: StarkGenericConfig> {
     pub final_memory: Option<VmMemoryState<Val<SC>>>,
 }
 
+pub struct VmExecutorNextSegmentState<F: PrimeField32> {
+    pub memory: MemoryImage<F>,
+    pub input: Streams<F>,
+    pub pc: u32,
+    pub segment_idx: usize,
+}
+
+pub struct VmExecutorOneSegmentResult<F: PrimeField32, VC: VmConfig<F>> {
+    pub segment: ExecutionSegment<F, VC>,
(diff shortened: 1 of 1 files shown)


crates/vm/src/arch/vm.rs
```diff
@@ -110,70 +122,92 @@ where
...
+            1 << mem_config.pointer_max_bits,
+            exe.init_memory.clone(),
+        );
+        let mut pc = exe.pc_start;
+        let mut segment_idx = 0;
+
+        loop {
+            let mut one_segment_result =
+                self.execute_until_segment(exe.clone(), memory, streams, segment_idx, pc)?;
+            streams = one_segment_result.segment.chip_complex.take_streams();
```
jonathanpwang, line 141: this is unrelated, but could you see if you can take the cycle tracker from the previous segment and pass it to the next segment? currently it's causing multi-segment flamegraphs to break

jonathanpwang (approved):
LGTM but @nyunyunyunyu please also take a look

crates/vm/src/arch/vm.rs
```diff
@@ -110,70 +122,92 @@ where
...
+
+    /// Executes a program until a segmentation happens.
+    /// Returns the last segment and the vm state for next segment.
+    /// This is so that the tracegen and proving of this segment can be immediately started (on a separate machine).
+    pub fn execute_until_segment(
+        &self,
+        exe: impl Into<VmExe<F>>,
+        memory: MemoryImage<F>,
+        input: impl Into<Streams<F>>,
+        segment_idx: usize,
```
nyunyunyunyu, line 164: Why do we need `segment_idx` as an argument? Can we have a span to wrap `execute_until_segment` in `execute_segments`?

luffykai: oh it was for the span. I can move it to execute and remove the idx as argument

PR #1252, opened 2025-01-22
[fix] modular setup macro padding

luffykai:
Currently, if the modulus has 40 bytes, the blocksize is 8 and final number of limbs is still 40.
But we can't really do 40, as the adapter is either 32 or 48. Change the macro to match that logic

### DIFF
--- a/extensions/algebra/moduli-setup/src/lib.rs
+++ b/extensions/algebra/moduli-setup/src/lib.rs
@@ -72,11 +72,6 @@ pub fn moduli_declare(input: TokenStream) -> TokenStream {
             panic!("limbs must be at most 48");
         }
 
-        // The largest power of two so that at most 10% of all space is wasted
-        let block_size = 1usize << ((limbs - 1) ^ (limbs + limbs / 9)).ilog2();
-        let limbs = limbs.next_multiple_of(block_size);
-        println!("block_size: {}", block_size);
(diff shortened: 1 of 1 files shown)


extensions/algebra/moduli-setup/src/lib.rs
```diff
@@ -63,15 +63,17 @@ pub fn moduli_declare(input: TokenStream) -> TokenStream {
...
-            proc_macro::Diagnostic::new(proc_macro::Level::Warning, "`limbs` has been set to 32 because it was too small; this is going to be changed once we support more flexible reads").emit();
+        } else if limbs <= 48 {
+            limbs = 48;
+            block_size = 16;
+        } else {
+            panic!("limbs must be at most 48");
         }
 
-        // The largest power of two so that at most 10% of all space is wasted
-        let block_size = 1usize << ((limbs - 1) ^ (limbs + limbs / 9)).ilog2();
```
luffykai, line 73: i am not sure if i understand this correctly, but since the limbs is either 32 or 48, i just make the block 32 or 16 for them (like the memory read block). lmk if this doesn't work

jonathanpwang:
> Seems ok for now, but maybe at some point we need to do something to the adapters or at least to how many limbs we are using exactly in the operations

Yes will revisit. @manh9203 can help track this issue.

PR #1254, opened 2025-01-22
[feat] support custom segmentation

luffykai:
redo #1245

fix INT-3005

### DIFF
--- a/crates/vm/src/arch/config.rs
+++ b/crates/vm/src/arch/config.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use derive_new::new;
 use openvm_circuit::system::memory::MemoryTraceHeights;
 use openvm_instructions::program::DEFAULT_MAX_NUM_PUBLIC_VALUES;
@@ -13,12 +15,12 @@ pub use super::testing::{
     POSEIDON2_DIRECT_BUS, RANGE_TUPLE_CHECKER_BUS, READ_INSTRUCTION_BUS,
 };
 use super::{
+    segment::{DefaultSegmentationStrategy, SegmentationStrategy},
     AnyEnum, InstructionExecutor, SystemComplex, SystemExecutor, SystemPeriphery, VmChipComplex,
     VmInventoryError, PUBLIC_VALUES_AIR_ID,
 };
 use crate::system::memory::BOUNDARY_AIR_OFFSET;
 
-const DEFAULT_MAX_SEGMENT_LEN: usize = (1 << 22) - 100;
 // sbox is decomposed to have this max degree for Poseidon2. We set to 3 so quotient_degree = 2
 // allows log_blowup = 1
 const DEFAULT_POSEIDON2_MAX_CONSTRAINT_DEGREE: usize = 3;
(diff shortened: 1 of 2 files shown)


crates/vm/src/arch/config.rs
```diff
@@ -64,6 +66,15 @@ impl Default for MemoryConfig {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct SegmentationStrategyArc(pub Arc<dyn SegmentationStrategy>);
```
jonathanpwang, line 70: Let's call this `SegmentationStrategyRef` to be consistent with other namings

jonathanpwang: actually the type might not be needed, see below

crates/vm/src/arch/config.rs
```diff
@@ -86,11 +97,14 @@ pub struct SystemConfig {
...
     pub num_public_values: usize,
-    /// When continuations are enabled, a heuristic used to determine when to segment execution.
-    pub max_segment_len: usize,
     /// Whether to collect detailed profiling metrics.
     /// **Warning**: this slows down the runtime.
     pub profiling: bool,
+    /// Segmentation strategy
+    /// This field is skipped in serde as it's only used in execution and
+    /// not needed after any serialize/deserialize.
+    #[serde(skip)]
```
jonathanpwang, line 106: in serde [docs](https://serde.rs/field-attrs.html#skip), you can provide an attribute `default = ...` to specify the function to use for deserializing. This can save having to make a newtype

crates/vm/src/arch/segment.rs
```diff
@@ -27,6 +27,84 @@ use crate::{
...
+//    its trace width is 36 and its after challenge trace width is 80.
+const DEFAULT_MAX_CELLS_PER_CHIP_IN_SEGMENT: usize = DEFAULT_MAX_SEGMENT_LEN * 120;
+
+pub trait SegmentationStrategy:
+    std::fmt::Debug + Send + Sync + std::panic::UnwindSafe + std::panic::RefUnwindSafe
+{
+    fn should_segment(
+        &self,
+        air_names: &[String],
+        trace_heights: &[usize],
```
jonathanpwang, line 45: can we just pass in all of `chip_complex`? that seems more flexible - but maybe there's an issue with `VC` generic?

crates/vm/src/arch/segment.rs
```diff
@@ -276,19 +354,15 @@ impl<F: PrimeField32, VC: VmConfig<F>> ExecutionSegment<F, VC> {
...
-
-        false
+        let segmentation_strategy = self.system_config().segmentation_strategy.0.clone();
+        segmentation_strategy.should_segment(
+            &self.air_names,
+            &self
+                .chip_complex
+                .dynamic_trace_heights()
+                .collect::<Vec<_>>(),
+            &self.chip_complex.current_trace_cells(),
```
jonathanpwang, line 364: can you check if `current_trace_cells` includes the preprocessed traces? This function is also slightly performance sensitive, it'd be better if we could get both the heights and trace cells using a single loop function

crates/vm/src/arch/config.rs
```diff
@@ -136,7 +151,9 @@ impl SystemConfig {
     }
```
jonathanpwang, line 151: provide a utility function like `set_segmentation_strategy<S: SegmentationStrategy>(strategy: S) { .. }`.

jonathanpwang (approved):
LGTM except some comments. Much cleaner than the other PR!

crates/vm/src/arch/config.rs
luffykai: 🧠

crates/vm/src/arch/segment.rs
luffykai: yea I don't know if possible to do generic in that trait to work with the `Arc<dyn>` 
https://github.com/openvm-org/openvm/pull/1245#discussion_r1924428035

crates/vm/src/arch/segment.rs
luffykai: discussed offline, fine for now

PR #1255, opened 2025-01-22
[feat] Add support for Twisted Edwards Curves into the elliptic curve VM extension

Avaneesh-axiom:
Related changes:
- Updated the `mod-builder` framework to handle setup rows that verify more than one constant
- Updated the `mod-builder` framework to the new method of padding rows (using a temporary range checker)
- Changed the `CurveConfig` struct to accomodate for curves in Twisted Edwards curve form

Closes INT-2999


extensions/ecc/circuit/Cargo.toml
```diff
@@ -21,6 +21,7 @@ openvm-rv32-adapters = { workspace = true }
 openvm-ecc-transpiler = { workspace = true }
 
 num-bigint = { workspace = true }
+num-bigint-dig = { workspace = true }
```
jonathanpwang, line 24: we got rid of num-bigint-dig, let's only use num-bigint from now on

extensions/ecc/circuit/src/ecc_extension.rs
```diff
@@ -0,0 +1,421 @@
...
+
+#[derive(Clone, Debug, derive_new::new, Serialize, Deserialize)]
+pub struct EccExtension {
+    pub supported_curves: Vec<CurveConfig>,
+}
+
+#[derive(Chip, ChipUsageGetter, InstructionExecutor, AnyEnum, BytesStateful)]
+pub enum EccExtensionExecutor<F: PrimeField32> {
+    // 32 limbs prime
+    SwEcAddNeRv32_32(EcAddNeChip<F, 2, 32>),
```
jonathanpwang, line 98: hmm we should probably rename `EcAddNeChip` to `SwAddNeChip`

extensions/ecc/circuit/Cargo.toml
Avaneesh-axiom: num-bigint-dig has jacobi symbol computation while num-bigint doesn't. I'll try to replicate it with num-bigint

Avaneesh-axiom: I need jacobi symbol for checking if a is QR and d is not QR for completeness

Cargo.toml
```diff
@@ -248,6 +250,7 @@ k256 = { version = "0.13.3", default-features = false }
 elliptic-curve = { version = "0.13.8", default-features = false }
 ecdsa = { version = "0.16.9", default-features = false }
 num-bigint = { version = "0.4.6", default-features = false }
+num-bigint-dig = { version = "0.8.4", default-features = false }
```
jonathanpwang, line 253: we are no longer using `num-bigint-dig` as its not as maintained as `num-bigint`. please switch over to using `num-bigint` and remove this dep

book/src/custom-extensions/ecc.md
```diff
@@ -29,37 +37,45 @@ The OpenVM Elliptic Curve Cryptography Extension provides support for elliptic c
 
 For elliptic curve cryptography, the `openvm-ecc-guest` crate provides macros similar to those in [`openvm-algebra-guest`](./algebra.md):
 
-1. **Declare**: Use `sw_declare!` to define elliptic curves over the previously declared moduli. For example:
+1. **Declare**: Use `sw_declare!` or `te_declare!` to define weierstrass or twisted edwards elliptic curves, respectively,over the previously declared moduli. For example:
```
jonathanpwang, line 40:
```suggestion
1. **Declare**: Use `sw_declare!` or `te_declare!` to define short Weierstrass or twisted Edwards elliptic curves, respectively, over the previously declared moduli. For example:
```

book/src/custom-extensions/ecc.md
```diff
@@ -29,37 +37,45 @@ The OpenVM Elliptic Curve Cryptography Extension provides support for elliptic c
...
 
 **Summary**:
 
-- `sw_declare!`: Declares elliptic curve structures.
-- `sw_init!`: Initializes them once, linking them to the underlying moduli.
-- `setup_sw_<i>()`/`setup_all_curves()`: Secures runtime correctness.
+- `sw_declare!`: Declares short Weierstrass elliptic curve structures.
+- `te_declare!`: Declares twisted Edwards elliptic curve structures.
+- `sw_init!`: Initializes the short Weierstrass elliptic curves once, linking them to the underlying moduli.
+- `te_init!`: Initializes the twisted Edwards elliptic curves once, linking it to the underlying moduli.
```
jonathanpwang, line 73:
```suggestion
- `te_init!`: Initializes the twisted Edwards elliptic curves once, linking them to the underlying moduli.
```

Avaneesh-axiom:
FYI: I added some more changes that fix a bug that I found (weierstrass and edwards opcodes would overlap in some cases)

I also added decompression hints (just like for weierstrass) to the edwards curves since I will use it in eddsa

Avaneesh-axiom:
Update: rebased onto develop

extensions/ecc/circuit/Cargo.toml
jonathanpwang: hm... that is really unfortunate (I don't want two bigint crates, and `num-bigint-dig` is less maintained than `num-bigint`) let me investigate

Avaneesh-axiom: I forgot to update, but I have removed the dependency on `num-bigint-dig`. I copied over the jacobi symbol code into `extensions/ecc/circuit/src/edwards_chip/utils.rs`

examples/ecc/src/main.rs
```diff
@@ -1,16 +1,67 @@
...
+    edwards::TwistedEdwardsPoint,
     k256::{Secp256k1Coord, Secp256k1Point},
     weierstrass::WeierstrassPoint,
+    Group,
 };
+extern crate alloc;
+
 // ANCHOR_END: imports
+openvm_algebra_guest::moduli_macros::moduli_declare! {
+    // The Secp256k1 modulus and scalar field modulus are already declared in the k256 module
```
jonathanpwang, line 14: oudated comment

extensions/ecc/circuit/src/ecc_extension.rs
```diff
@@ -44,44 +52,73 @@ pub struct CurveConfig {
...
+    coeffs: TeCurveCoeffs {
+        a: BigUint::from_bytes_le(ED25519_A.as_le_bytes()),
+        d: BigUint::from_bytes_le(ED25519_D.as_le_bytes()),
+    },
 });
 
 #[derive(Clone, Debug, derive_new::new, Serialize, Deserialize)]
-pub struct WeierstrassExtension {
-    pub supported_curves: Vec<CurveConfig>,
+pub struct EccExtension {
```
jonathanpwang, line 94: I can see it's easier to say "ECC Extension" but here I feel like we could have done `WeierstrassExtension` and `TwistedEdwardsExtension` as two separate structs. I don't have strong feelings either way

extensions/ecc/circuit/src/ecc_extension.rs
```diff
@@ -44,44 +52,73 @@ pub struct CurveConfig {
...
+    pub supported_sw_curves: Vec<CurveConfig<SwCurveCoeffs>>,
```
jonathanpwang, line 95: another option is to keep a single vector but use enum to specify different curve types within `CurveConfig`. That seems like a might be a little nicer ux wise?

jonathanpwang:
Closing this in favor of https://github.com/openvm-org/openvm/pull/1858 

Can we force reset `feat/ed25519` to branch off of `feat/new-execution`?

extensions/ecc/circuit/src/ecc_extension.rs
jonathanpwang, line 1: how come github can't detect this is a `git mv weierstrass_extension.rs`

Avaneesh-axiom: I think the content is different enough that it can't detect it as a rename

Avaneesh-axiom: Screenshot 2025-08-09 at 8 26 19 PM
Yeah, it looks like it has 46% similarity

PR #1270, opened 2025-01-23
[docs] Document specification and implementation of `VERIFY_BATCH`

TlatoaniHJ:
Add documentation as per INT-3070. I've added what I could think needs to be documented so far, let me know what else should be added.

Closes INT-3070


jonathanpwang (approved):
I'll fix a few typos before merging but this is really nice!

PR #1271, opened 2025-01-23
[fix] support execute until segment 

luffykai:
redo #1254 
the fix is in second commit (I double-take the streams) -> we do have test with continuation, but not actually using input

fix INT-3004


luffykai:
reth: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/12938455876

jonathanpwang (approved):

PR #1274, opened 2025-01-24
feat: config inheritance in the VmConfig derive macro

stephenh-axiom-xyz:
Implements the following:
- `VmConfig` macro implements config inheritance (via the `#[config]` attribute in place of `#[system]`)
- `Rv32ImConfig` inherits `base` and `io` from `Rv32IConfig`
- `impl VmExtension` for the option of an arbitrary `VmExtension`


jonathanpwang:
@Golovanov399 if you get a chance can you review

Golovanov399 (approved):
Seems good
Although I didn't really understand the point of `impl VmExtension for Option`

PR #1278, opened 2025-01-24
[fix] complex macro

Golovanov399:
This resolves INT-3061.

It also adds a test that used to fail but passes now. So now running reth benchmark is not the only way to see that it fails sometimes.

But just to be sure, here is the workflow: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/12956655766


jonathanpwang (commented):
Is there a corresponding part of the docs to be updated?

Golovanov399:
I will check, but I think I wrote the docs thinking that it was already like after this change

jonathanpwang (approved):
LGTM if there's no doc changes.

Just to make sure I understand: the `complex_init` uses `mod_idx` but then the complex extern functions themselves should be using `complex_idx`, and so this is an internal typo?

Golovanov399:
The only usage of `mod_idx` is that setup instruction compares against the `mod_idx`th modulus from the `.openvm` section

PR #1286, opened 2025-01-25
[feat] Add HintLoad phantom opcode

nyunyunyunyu:
closes INT-3096

### DIFF
--- a/extensions/native/circuit/src/extension.rs
+++ b/extensions/native/circuit/src/extension.rs
@@ -224,6 +224,11 @@ impl<F: PrimeField32> VmExtension<F> for Native {
             PhantomDiscriminant(NativePhantom::Print as u16),
         )?;
 
+        builder.add_phantom_sub_executor(
+            NativeHintLoadSubEx,
+            PhantomDiscriminant(NativePhantom::HintLoad as u16),
+        )?;
+
         Ok(inventory)
     }
 }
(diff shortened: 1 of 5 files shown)


extensions/native/circuit/src/extension.rs
```diff
@@ -308,6 +314,29 @@ pub(crate) mod phantom {
...
+        ) -> eyre::Result<()> {
+            let payload = match streams.input_stream.pop_front() {
+                Some(hint) => hint,
+                None => {
+                    bail!("EndOfInputStream");
+                }
+            };
+            let id = streams.hint_space.len();
+            streams.hint_space.push(payload);
+            streams.hint_stream.push_back(F::from_canonical_usize(id));
```
jonathanpwang, line 336: should we just reset hint stream to `vec![id]` or assert its empty?

jonathanpwang: nvm I see above you explicitly wanted to be able to have multiple id added to hint stream

jonathanpwang: ok this works well with `HINTSTORE_W` since that uses `hint_stream.pop_front`. However, right now we need to inspect the program to ensure the hint stream is clear beforehand. I'm just going to add a `hint_stream.clear()` defensively.

jonathanpwang (approved):
LGTM

PR #1289, opened 2025-01-26
[fix] expr builder eval inconsistency

luffykai:
Issue: the way we do `evaluate_overflow_isize` and  `evaluate_overflow_expr` were inconsistent for `Select`
Since we had flag available for isize we directly using it to choose left or right, but that's not viable in expr. But this inconsistency MIGHT result in different `max_overflow_bit` in some condition (=> ended up in range checker error), which we only discovered with BLS curve double setup...


crates/circuits/mod-builder/src/builder.rs
```diff
@@ -102,11 +105,24 @@ impl ExprBuilder {
...
         }
     }
 
+    // This can be used to debug, when we only want to print something in a specific chip.
+    pub fn set_debug(&mut self) {
+        self.debug = true;
+    }
+
+    #[allow(unused)]
+    fn debug_print(&self, msg: &str) {
```
luffykai, line 120: i used this to only print something in double chip. should be useful to keep

extensions/ecc/tests/Cargo.toml
```diff
@@ -18,11 +18,13 @@ openvm-algebra-transpiler.workspace = true
 openvm-ecc-transpiler.workspace = true
 openvm-ecc-circuit.workspace = true
 openvm-ecc-guest.workspace = true
+openvm-pairing-guest = { workspace = true, features = ["bls12_381"] }
```
jonathanpwang, line 21: it's a little weird that pairing is dependency in ecc, even if for test. can we move this test to `pairing/tests`?

crates/circuits/mod-builder/src/builder.rs
```diff
@@ -102,11 +105,24 @@ impl ExprBuilder {
...
+        if self.debug {
+            println!("{}", msg);
```
jonathanpwang, line 122: how is this different than when you just use `dbg!`?

jonathanpwang (approved):
LGTM but can you move the test to pairing tests crate instead

crates/circuits/mod-builder/src/builder.rs
luffykai: when adding print to mod-builder, it applies to all chips using it. But usually when debugging i only care about one chip. So the way I use this is to call set_debug only at the specific chip

PR #1291, opened 2025-01-26
chore: update docs

yi-sun:
Closes INT-2938, INT-3109, INT-3123

PR #1294, opened 2025-01-26
[feat] Implement `sys_rand`

Golovanov399:
This resolves INT-2955.

`sys_rand` now:
- calls a new phantom instruction that adds a hint stream with some random numbers, 
- then reads this many numbers from this hint stream to the required memory location.

The integration test is a simple one and just shows that the hashmap works somehow (it didn't use to because of "`sys_rand` unimplemented", but maybe we could also check the statistical randomness of our calls somehow.


extensions/native/tests/programs/Cargo.toml
```diff
@@ -18,3 +19,7 @@ openvm-native-transpiler = { path = "../../transpiler", default-features = false
 openvm-instructions = { path = "../../../../crates/toolchain/instructions", default-features = false }
 openvm-stark-sdk = { git = "https://github.com/openvm-org/stark-backend.git", rev = "8c9f94", default-features = false }
 openvm-stark-backend = { git = "https://github.com/openvm-org/stark-backend.git", rev = "8c9f94", default-features = false }
+
+[[example]]
```
jonathanpwang, line 23: test should not be in `extensions/native`. it should be in `extensions/rv32im` or `crate/toolchain/tests`

extensions/native/tests/programs/src/examples/hashmap.rs
```diff
@@ -0,0 +1,18 @@
...
+openvm::entry!(main);
+
+fn main() {
+    let mut map = HashMap::new();
+    map.insert(1, 2);
+    map.insert(2, 8);
+    assert!(map.get(&1) == Some(&2));
+    assert!(map.get(&2) == Some(&8));
+    assert!(!map.contains_key(&3));
+    openvm::io::println(format!("{:?}", map.get(&1).unwrap()));
```
jonathanpwang, line 17: when feature = "std", `println!` just works :)

extensions/rv32im/circuit/src/extension.rs
```diff
@@ -508,6 +521,25 @@ mod phantom {
         }
     }
 
+    impl<F: PrimeField32> PhantomSubExecutor<F> for Rv32HintRandomSubEx {
+        fn phantom_execute(
+            &mut self,
+            _: &MemoryController<F>,
+            streams: &mut Streams<F>,
+            _: PhantomDiscriminant,
+            len: F,
```
jonathanpwang, line 530: this seems wrong? when you do `= In len` in the macro, that means here what's passed is register index of `len`. You need to unsafe read from address space 1 to get the actual len.

extensions/rv32im/guest/src/lib.rs
```diff
@@ -26,5 +26,6 @@ pub const CSRRW_FUNCT3: u8 = 0b001;
 #[repr(u16)]
 pub enum PhantomImm {
     HintInput = 0,
+    HintRandom,
```
jonathanpwang, line 29: put this after PrintStr so we don't need to change the numbering in ISA spec

extensions/rv32im/transpiler/src/instructions.rs
```diff
@@ -275,6 +275,8 @@ pub enum Rv32HintStoreOpcode {
 pub enum Rv32Phantom {
     /// Prepare the next input vector for hinting, but prepend it with a 4-byte decomposition of its length instead of one field element.
     HintInput = 0x20,
+    /// Prepare given amount of random numbers for hinting.
+    HintRandom,
```
jonathanpwang, line 279: again put enum after PrintStr

jonathanpwang (changes requested):
Move integration test to `rv32im/test`.

Currently the phantom instruction is wrong: it needs to read len from memory address space 1.

Update ISA doc: it's easier to make a PR into https://github.com/openvm-org/openvm/pull/1291

jonathanpwang (approved):
LGTM

PR #1306, opened 2025-01-27
fix: remove incorrect nop transpilation for intrinsics

yi-sun:
Closes INT-3149


extensions/algebra/transpiler/src/lib.rs
```diff
@@ -119,7 +119,11 @@ impl<F: PrimeField32> TranspilerExtension<F> for ModularTranspilerExtension {
                     _ => unimplemented!(),
                 };
                 let global_opcode = global_opcode + mod_idx_shift;
-                Some(from_r_type(global_opcode, 2, &dec_insn))
+                // The only opcode in this extension which can write to rd is `IsEqMod`
```
jonathanpwang, line 122: | SETUP_ISEQMOD_RV32\<N\> | `a,b,c,1,2` | `assert([r32{0}(b): N::NUM_LIMBS]_2 == N)` in the chip that handles modular equality. For the sake of implementation convenience it also writes something (can be anything) into register value of `[a:4]_1`                                                                                        |

This is on L84 above. We should just return a transpiler error if `rd == 0` in this case.

jonathanpwang (commented):
We need to also handle `SETUP_ISEQ_MOD`. 
I checked the ISA again and to the best of my knowledge that is the only other case where address space 1 is written to.

jonathanpwang (approved):

PR #1308, opened 2025-01-27
fix: update integration tests for new native address space restrictions

yi-sun:
Some tests relied on now-invalid behavior from the native extension and have been removed.

### DIFF
--- a/crates/vm/tests/integration_test.rs
+++ b/crates/vm/tests/integration_test.rs
@@ -75,34 +75,34 @@ where
 fn test_vm_1() {
     let n = 6;
     /*
-    Instruction 0 assigns word[0]_1 to n.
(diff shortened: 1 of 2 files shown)


crates/vm/tests/integration_test.rs
```diff
@@ -298,14 +282,14 @@ fn test_vm_public_values() {
 
 #[test]
 fn test_vm_initial_memory() {
-    // Program that fails if mem[(1, 0)] != 101.
+    // Program that fails if mem[(1, 7)] != 101.
```
jonathanpwang, line 285:
```suggestion
    // Program that fails if mem[(4, 7)] != 101.
```

jonathanpwang (approved):

PR #1312, opened 2025-01-27
Memory profiler

alv-around:
Hi, 

Here is the PR for #1029 which implements metrics recorder for Prometheus. Previous discussion can be found in #1201. 

The PR implements a recoder that:
1. **Ttarts a thread which collects process-metrics for a given Duration**. It is important that the [process metrics](https://docs.rs/metrics-process/latest/metrics_process/) (memory/cpu usage.. etc) in it's own separate thread spawned is, otherwise, in an async task will be block anytime the VM performs any of it's cpu-intesive computations. 
2. **`[metrics_context_tracing](https://docs.rs/metrics-tracing-context/latest/metrics_tracing_context/)` enabled**( as stated in the [docs](https://github.com/openvm-org/openvm/blob/main/docs/crates/metrics.md)). so any metric recorded within a span, the span is automatically added as a metric label. However, this does not apply to the process metrics metrics regularly collected in the previous point, because they run in a separate thread. 

**Usage:** In this [repo](https://github.com/alv-around/openvm-memory-profiler-examples), I made two examples how the memory profiler could work with both (pull/push) prometheus exporters.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -364,6 +364,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-waker"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
 name = "aurora-engine-modexp"
 version = "1.1.0"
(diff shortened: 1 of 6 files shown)


jonathanpwang:
Sorry for delay, I'll try to get to this soon.

alv-around:
> Sorry for delay, I'll try to get to this soon.

Sure no problem, I rebased the branched to latest `main`.

PR #1313, opened 2025-01-27
[feat] Make the `openvm::entry!` macro a proc-macro attribute

Golovanov399:
This resolves INT-2784. Now, instead of `openvm::entry!(main)`, we just write `#[openvm::entry]` before `fn main`.

### DIFF
--- a/crates/toolchain/openvm/entry/src/lib.rs
+++ b/crates/toolchain/openvm/entry/src/lib.rs
@@ -17,11 +17,11 @@ use syn::{parse_macro_input, ItemFn};
 /// #![no_main]
 /// #![no_std]
 ///
-/// #[openvm::entry]
+/// #[openvm::main]
 /// fn main() { }
 /// ```
 #[proc_macro_attribute]
-pub fn entry(_attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn main(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let input = parse_macro_input!(item as ItemFn);
 
     let fn_name = &input.sig.ident;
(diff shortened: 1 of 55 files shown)


crates/toolchain/openvm/entry/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+///
+/// ```ignore
+/// #![no_main]
+/// #![no_std]
+///
+/// #[openvm::entry]
+/// fn main() { }
+/// ```
+#[proc_macro_attribute]
+pub fn entry(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
jonathanpwang, line 24: I think we can call it `main` now so it's `#[openvm::main]`

crates/toolchain/openvm/entry/src/lib.rs
```diff
@@ -0,0 +1,58 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, ItemFn};
+
+/// Used for defining the guest's entrypoint and main function.
+///
+/// When `#![no_main]` is used, the programs entrypoint and main function is left undefined. The
```
jonathanpwang, line 7: I was wondering, do you think it's safe to add the `#![no_main]` within the macro, where it figures out based on whether std is enabled etc? This has the benefit that on host we can make it always `std` and allows `cargo clippy` to work etc

jonathanpwang (approved):
Rename suggestion, and wanted to ask/discuss if we can handle `no_main` within the macro for better UX. Otherwise LGTM

jonathanpwang:
@Golovanov399 FYI I am going to drop this from `develop` in the rebase for now because I'm afraid it will introduce too many rebase conflicts. We should remember this PR and cherry-pick it in when we're ready to merge to `main`.

PR #1317, opened 2025-01-28
perf: Write directly into access adapter chips rather than via buffer

zlangley:


crates/vm/src/system/memory/offline.rs
```diff
@@ -532,23 +527,39 @@ mod tests {
...
 
+    fn setup_test(
+        initial_memory: MemoryImage<BabyBear>,
+        initial_block_size: usize,
+    ) -> (OfflineMemory<BabyBear>, AccessAdapterInventory<BabyBear>) {
+        let memory_bus = MemoryBus(0);
+        let range_checker =
+            SharedVariableRangeCheckerChip::new(VariableRangeCheckerBus::new(1, 29));
+        let mem_config = MemoryConfig {
+            as_offset: 0,
```
jonathanpwang, line 538: can we set this to 1 to be more realistic?

jonathanpwang (approved):
lgtm but realized the tests were using `as_offset=0` but we have fixed `as_offset=1` in the spec so we should test using that

PR #1323, opened 2025-01-29
perf: Remove touch address for VolatileMemory

zlangley:
VolatileMemory does not need to internally maintain touched addresses as it is given `final_memory` in `finalize`. This does mean that `current_trace_height` is no longer known until finalization.

### DIFF
--- a/crates/vm/src/system/memory/volatile/mod.rs
+++ b/crates/vm/src/system/memory/volatile/mod.rs
@@ -23,7 +23,6 @@ use openvm_stark_backend::{
     rap::{BaseAirWithPublicValues, PartitionedBaseAir},
     AirRef, Chip, ChipUsageGetter,
 };
-use rustc_hash::FxHashSet;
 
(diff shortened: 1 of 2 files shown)


crates/vm/src/system/memory/volatile/mod.rs
```diff
@@ -269,7 +258,11 @@ impl<F: PrimeField32> ChipUsageGetter for VolatileBoundaryChip<F> {
     }
 
     fn current_trace_height(&self) -> usize {
```
jonathanpwang, line 260: actually VolatileBoundary = no continuations, so `current_trace_height` is never going to matter because we don't segment

jonathanpwang (approved):

crates/vm/src/system/memory/volatile/mod.rs
zlangley: good point

PR #1324, opened 2025-01-30
add new field implementation

MonkeyKing-1:
Babybear implementation with better multiplication;
I had to do some weirdish things to make traits work but I guess it worked out in the end

Closes INT-3081


extensions/native/custom-ef/Cargo.toml
```diff
@@ -0,0 +1,50 @@
+# Forked from sp1-recursion-compiler under MIT license.
```
jonathanpwang, line 1: this has a lot of copy paste; clean it up and use cargo machete

jonathanpwang:
This should be in https://github.com/openvm-org/stark-backend

PR #1327, opened 2025-01-30
perf: Improve replay iteration speed

zlangley:
Previously 18% of finalize was spent just calling `Iterator::next()` (locally for reth block 2100000).

### DIFF
--- a/crates/vm/src/system/memory/offline.rs
+++ b/crates/vm/src/system/memory/offline.rs
@@ -146,7 +146,7 @@ impl<F: PrimeField32> OfflineMemory<F> {
         &mut self,
         address_space: u32,
         pointer: u32,
-        values: Vec<F>,
+        values: &[F],
         records: &mut AccessAdapterInventory<F>,
     ) {
         let len = values.len();
@@ -159,14 +159,14 @@ impl<F: PrimeField32> OfflineMemory<F> {
 
         let pointer = pointer as usize;
         let prev_data = self.data[(address_space - self.as_offset) as usize]
-            .set_range(pointer..pointer + len, &values);
+            .set_range(pointer..pointer + len, values);
 
         let record = MemoryRecord {
             address_space: F::from_canonical_u32(address_space),
             pointer: F::from_canonical_usize(pointer),
             timestamp: self.timestamp,
             prev_timestamp,
-            data: values,
+            data: values.to_vec(),
             prev_data: Some(prev_data),
         };
         self.log.push(Some(record));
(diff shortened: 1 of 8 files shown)


zlangley:
before: https://github.com/axiom-crypto/openvm-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/main/reth-3f1272a788e6ba792dc1530586a4bfc197d574bee81c8a6d159701d2b2f1c499.md
after: https://github.com/axiom-crypto/openvm-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/main/reth-f86a17c2994fe425a6ffa66380929124e00074302979580ef6da6792d207b358.md

crates/vm/src/system/memory/smallvec.rs
```diff
@@ -0,0 +1,81 @@
+use std::{
+    fmt::Debug,
+    ops::{Deref, DerefMut},
+};
+
+use serde::{Deserialize, Serialize};
+use smallvec::{Array, SmallVec};
+
+pub struct SerdeSmallVec<A: Array>(SmallVec<A>);
```
jonathanpwang, line 9: is this purely for serde? can https://crates.io/crates/serde_with help with this

jonathanpwang:
- m7a.24xlarge: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13091437663 
- m7a.16xlarge: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13086454610
- c7a.48xlarge: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13086450020

On x86 it seems to not be able to replicate any perf gain.

jonathanpwang:
I tried on an ARM machine and got the same results as above

crates/vm/src/system/memory/smallvec.rs
zlangley: Yes. I had tried using `serde_with`, but ran into issues because `SmallVec` cannot be directly converted into `Vec`.

zlangley:
Okay, let's close then. This was based on samply output from my local machine, which we have determined to be quite unreliable for this.

crates/vm/src/system/memory/offline.rs
```diff
@@ -159,14 +159,14 @@ impl<F: PrimeField32> OfflineMemory<F> {
...
-            .set_range(pointer..pointer + len, &values);
+            .set_range(pointer..pointer + len, values);
 
         let record = MemoryRecord {
             address_space: F::from_canonical_u32(address_space),
             pointer: F::from_canonical_usize(pointer),
             timestamp: self.timestamp,
             prev_timestamp,
-            data: values,
+            data: values.to_vec(),
```
Golovanov399, line 169: Does this converting stuff really improve performance?

jonathanpwang: apparently not

PR #1330, opened 2025-02-01
[chore] Optimize hint for Var/Felt

nyunyunyunyu:


extensions/native/circuit/src/extension.rs
```diff
@@ -273,6 +279,29 @@ pub(crate) mod phantom {
...
+            _: F,
+            _: u16,
+        ) -> eyre::Result<()> {
+            let hint = match streams.input_stream.pop_front() {
+                Some(hint) => hint,
+                None => {
+                    bail!("EndOfInputStream");
+                }
+            };
+            streams.hint_stream.clear();
```
jonathanpwang, line 298: you changed the rest to assert empty so this should be as well?

jonathanpwang (approved):
LGTM but you need to update ISA.md with this new phantom instruction.

PR #1332, opened 2025-02-03
[chore] Add permute-opened-values span

nyunyunyunyu:


extensions/native/circuit/src/poseidon2/columns.rs
```diff
@@ -90,12 +91,14 @@ pub struct TopLevelSpecificCols<T> {
     pub commit_pointer: T,
     pub commit_read: MemoryReadAuxCols<T>,
 }
+const_assert_eq!(TopLevelSpecificCols::<usize>::width(), 69);
```
jonathanpwang, line 94: do you need this? the width is already statically known: you could use `size_of` to define it with `T = u8`

nyunyunyunyu: It's for reference. Otherwise people need to run something in order to know the width

jonathanpwang: that is what I meant, if you defined the constant using size_of::<u8>, the IDE will show the constant number when you hover

nyunyunyunyu: Well RustRover doesn't have this feature. I guess it's a feature of Zed

PR #1335, opened 2025-02-03
[perf] Store blocks in a potentially smarter way

Golovanov399:
**Metric:** tracegen
**Before:** 21246 ms
**After:** 19288 ms
**Percentage gain:** -9.2%

Source: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13120710102

---
This resolves INT-3235.

_Quote from the ticket._
Currently one actual block corresponds to several memory addresses. Whenever we modify the block itself (e.g. change the timestamp), we have to change it for every memory address it covers.
We could instead store the blocks we constructed in a separate storage (Vec), and in the positions we will store a simple usize denoting the block index in the storage. This also kind of simplifies the concept of the "default" block (we assign 0 to be a fictious index meaning "not initialized yet"), and generally makes the block data memory take less space and probably makes constructing it easier. Also, we can now only change the block of size 4 once and not touch 4 of its copies.
One drawback is that we don't remove the erased blocks from the storage, which means that we can potentially run out of memory because of this storage. On the other hand, this will probably happen about when we run out of timestamps anyway.
_End of quote._

Side note: I also tried to merge a block with the next in a smarter way: instead of creating a completely new block, just modify one of the existing ones. This is worse for some reason: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13122033497

### DIFF
--- a/crates/vm/src/system/memory/offline.rs
+++ b/crates/vm/src/system/memory/offline.rs
@@ -108,27 +108,6 @@ impl BlockMap {
             .filter(|(_, idx)| *idx > 0)
             .map(|(address, idx)| (address, &self.storage[idx - 1]))
     }
-
-    /// Merges the block starting at `address` and the one immediately after it. The blocks must have the same size.
-    pub fn merge_block_with_next(&mut self, address: &(u32, u32)) {
-        let &(address_space, pointer) = address;
-        let left = *self.id.get(address).unwrap();
-        let size = self.storage[left - 1].size;
-        let timestamp = self.storage[left - 1].timestamp.max(
-            self.storage[self
-                .id
-                .get(&(address_space, pointer + size as u32))
-                .unwrap()
-                - 1]
-            .timestamp,
-        );
-        self.storage[left - 1].size += size;
-        self.storage[left - 1].timestamp = timestamp;
-        for i in 0..size {
-            self.id
-                .insert(&(address_space, pointer + size as u32 + i as u32), left);
-        }
-    }
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -406,7 +385,7 @@ impl<F: PrimeField32> OfflineMemory<F> {
                 self.block_data
                     .set_range(&(address_space, mid_ptr), half_size, block);
             }
-            if query >= mid_ptr {
+            if query >= cur_ptr + half_size_u32 {
                 // The left is finalized; add it to the partition.
                 let block = BlockData {
                     pointer: cur_ptr,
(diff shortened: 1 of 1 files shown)


crates/vm/src/system/memory/offline.rs
```diff
@@ -342,21 +416,13 @@ impl<F: PrimeField32> OfflineMemory<F> {
...
-            if block.is_none() || block.unwrap().size == 0 {
-                self.block_data.insert(
-                    &(address_space, pointer + i),
-                    Self::initial_block_data(pointer + i, self.initial_block_size),
-                );
-            }
-            let block = self
-                .block_data
-                .get_mut(&(address_space, pointer + i))
-                .unwrap();
```
zlangley, line 356: This was the issue I had mentioned; seems like your PR does indeed address it.

crates/vm/src/system/memory/offline.rs
```diff
@@ -25,6 +25,91 @@ struct BlockData {
...
+        } else {
+            self.storage[idx - 1].clone()
+        }
+    }
+
+    pub fn get(&mut self, address: &(u32, u32)) -> &BlockData {
+        let (address_space, pointer) = *address;
+        let idx = self.id.get(&(address_space, pointer)).unwrap_or(&0);
+        if idx == &0 {
+            let pointer = pointer - pointer % self.initial_block_size as u32;
```
zlangley, line 68: Is this faster than `pointer / N * N`, or is it just stylistic?

I wonder if we should do `pointer & ~(N - 1)` (assuming `N` is a power of 2). If we make things generic over const `N` I think the compiler can compile this to one instruction.

crates/vm/src/system/memory/offline.rs
```diff
@@ -25,6 +25,91 @@ struct BlockData {
...
+        BlockData {
+            pointer: aligned_pointer,
+            size: initial_block_size,
+            timestamp: INITIAL_TIMESTAMP,
+        }
+    }
+
+    pub fn get_without_adding(&self, address: &(u32, u32)) -> BlockData {
+        let idx = self.id.get(address).unwrap_or(&0);
+        if idx == &0 {
```
zlangley, line 57: Can this be
```
match self.id.get(address) {
    None => ...
    Some(idx) => ...
}
```
or do we use `0` as a value in the mapping as well?

zlangley (approved):

crates/vm/src/system/memory/offline.rs
Golovanov399: `0` means "uninitialized", so `Some(&0)` is the same as `None` for us

jonathanpwang (approved):
Nice

PR #1336, opened 2025-02-03
[fix] Add serde as a dependency with "std" feature enabled

Avaneesh-axiom:
Fix a bug found in `openvm-circuit-primitives` where it wouldn't compile because the "std" feature of `serde` was not enabled. `openvm-circuit-primitives` indirectly depended on `serde` but the version that was used did not have the "std" feature enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3424,6 +3424,7 @@ dependencies = [
  "hex-literal",
  "num-bigint 0.4.6",
  "num-traits",
+ "openvm-entry",
  "openvm-platform",
  "openvm-rv32im-guest",
  "serde",
@@ -3859,6 +3860,15 @@ dependencies = [
  "openvm-transpiler",
  "rrs-lib",
  "strum",
+]
+
+[[package]]
+name = "openvm-entry"
+version = "0.1.0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.96",
 ]
 
 [[package]]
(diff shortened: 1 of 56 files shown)


jonathanpwang:
@Avaneesh-axiom why is lint failing

jonathanpwang (approved):
I think we should probably fix this on `main`, and then I'll rebase

Avaneesh-axiom:
> @Avaneesh-axiom why is lint failing

I think the examples fail lint because openvm::entry! was turned into a proc macro by a recent commit on `develop` (I based by branch on develop) but the Cargo.toml in the examples pulls the `openvm` package from the git repo which gives it the `main` branch version.

It passes locally when I change `openvm = { git = "https://github.com/openvm-org/openvm.git" }` to `openvm = { path = "../../crates/toolchain/openvm" }`

Should be fixed after rebasing onto `main` I think

Avaneesh-axiom:
> I think we should probably fix this on `main`, and then I'll rebase

do you mean I should rebase this PR onto `main` and then merge it into `main`?

jonathanpwang:
> > I think we should probably fix this on `main`, and then I'll rebase
> 
> do you mean I should rebase this PR onto `main` and then merge it into `main`?

yea

Avaneesh-axiom:
lint is still failing. I think it's because the commit that changes the `openvm::entry!` macro into a proc macro is not main. That commit was on develop, and it's part of this PR so it will be in main after this PR gets merged. However, cargo is pulling the head of main as a dependency, which doesn't contain the commit.

should I cherry pick the commit from develop onto main and make a PR for it? After that gets merged, this lint should pass

jonathanpwang:
We can only address the lint issue by changing the example guest program Cargo dependencies to use this repo on `develop` branch. Let's do that in a separate PR if it becomes a persistent issue. Merging this one for now.

jonathanpwang:
replaced by https://github.com/openvm-org/openvm/pull/1343

PR #1338, opened 2025-02-04
[perf] Use `Vec` instead of `BTreeMap` in `tree_from_memory`

Golovanov399:
**Metric:** tracegen
**Before:** 10,901.18 ms
**After:** 10,131.55 ms
**Percentage gain:** -7.1%

Workflow: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13160394734

---
This resolves INT-3236.

_Quote from the ticket._
Our current `AddressMap.items()` returns something sorted by address (more specifically by `(address_space, pointer)` lexicographically), therefore no need to push it into a `BTreeMap`, we can push it into `Vec` and then modify the recursion a little bit.
We can also save on creating the "all-zero" nodes by building them once and then taking the corresponding node every time we need such node for some height. This should in theory always be strictly better, but this is worse now [...].
_End of quote._

I also tried to calculate the capacity once and then reserve the exact capacity needed once instead of `Vec::new()` and `.push()` every time, but this was slightly slower or at least the same.

### DIFF
--- a/crates/vm/src/system/memory/tree/mod.rs
+++ b/crates/vm/src/system/memory/tree/mod.rs
@@ -47,9 +47,26 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
         }
     }
 
+    /// Returns a tree of height `height` with all leaves set to `leaf_value`.
+    pub fn construct_uniform(
+        height: usize,
+        leaf_value: [F; CHUNK],
+        hasher: &impl Hasher<CHUNK, F>,
+    ) -> MemoryNode<CHUNK, F> {
+        if height == 0 {
+            Self::new_leaf(leaf_value)
+        } else {
+            let child = Arc::new(Self::construct_uniform(height - 1, leaf_value, hasher));
+            NonLeaf {
+                hash: hasher.compress(&child.hash(), &child.hash()),
+                left: child.clone(),
+                right: child,
+            }
+        }
+    }
+
     fn from_memory(
         memory: &Vec<(u64, [F; CHUNK])>,
-        zero_nodes: &Vec<MemoryNode<CHUNK, F>>,
         lookup_range: Range<usize>,
         height: usize,
         from: u64,
@@ -57,14 +74,15 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
     ) -> MemoryNode<CHUNK, F> {
         if height == 0 {
             if lookup_range.is_empty() {
-                zero_nodes[height].clone()
+                MemoryNode::new_leaf(hasher.hash(&[F::ZERO; CHUNK]))
             } else {
                 debug_assert_eq!(memory[lookup_range.start].0, from);
                 debug_assert_eq!(lookup_range.end - lookup_range.start, 1);
                 MemoryNode::new_leaf(hasher.hash(&memory[lookup_range.start].1))
             }
         } else if lookup_range.is_empty() {
-            zero_nodes[height].clone()
+            let leaf_value = hasher.hash(&[F::ZERO; CHUNK]);
+            MemoryNode::construct_uniform(height, leaf_value, hasher)
         } else {
             let midpoint = from + (1 << (height - 1));
             let mid = {
@@ -84,22 +102,9 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
                     right
                 }
             };
-            let left = Self::from_memory(
-                memory,
-                zero_nodes,
-                lookup_range.start..mid,
-                height - 1,
-                from,
-                hasher,
-            );
-            let right = Self::from_memory(
-                memory,
-                zero_nodes,
-                mid..lookup_range.end,
-                height - 1,
-                midpoint,
-                hasher,
-            );
+            let left = Self::from_memory(memory, lookup_range.start..mid, height - 1, from, hasher);
+            let right =
+                Self::from_memory(memory, mid..lookup_range.end, height - 1, midpoint, hasher);
             NonLeaf {
                 hash: hasher.compress(&left.hash(), &right.hash()),
                 left: Arc::new(left),
@@ -138,20 +143,8 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
             memory_partition.last().map_or(0, |(addr, _)| *addr)
                 < (1 << memory_dimensions.overall_height())
         );
-
-        let mut zero_nodes = vec![Self::new_leaf([F::ZERO; CHUNK])];
-        for _ in 0..memory_dimensions.overall_height() {
-            let child = Arc::new(zero_nodes.last().unwrap().clone());
-            zero_nodes.push(NonLeaf {
-                hash: hasher.compress(&child.hash(), &child.hash()),
-                left: child.clone(),
-                right: child,
-            });
-        }
-
         Self::from_memory(
             &memory_partition,
-            &zero_nodes,
             0..memory_partition.len(),
             memory_dimensions.overall_height(),
             0,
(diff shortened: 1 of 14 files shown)


Golovanov399:
@zlangley I couldn't think of a simple way to avoid collecting items

crates/vm/src/system/memory/tree/mod.rs
```diff
@@ -95,19 +118,34 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
         memory: &MemoryImage<F>,
         hasher: &impl Hasher<CHUNK, F>,
     ) -> MemoryNode<CHUNK, F> {
-        // Construct a BTreeMap that includes the address space in the label calculation,
+        // Construct a Vec that includes the address space in the label calculation,
         // representing the entire memory tree.
-        let mut memory_partition = BTreeMap::new();
+        let mut memory_partition: Vec<(u64, [F; CHUNK])> = Vec::new();
         for ((address_space, pointer), value) in memory.items() {
```
zlangley, line 124: Note: It's a bit unfortunate that `MemoryImage` is `(address_space, pointer) -> F` rather than `(address_space, label) -> [F; 8]`. I feel like this means we have to do a lot of extra work here and at the end of a segment to convert it into both forms.

zlangley (approved):

crates/vm/src/system/memory/tree/mod.rs
jonathanpwang, line 124: pondering: `memory_partition` is only needed for `from_memory` and currently it's non-parallelized because of the `CHUNK` partitioning, whereas `.items()` could be made into a parallel iterator (still sorted). Is it even worth still having `memory_partition`, or can the chunking just be incorporated into your new `from_memory` function somehow?

crates/vm/src/system/memory/tree/mod.rs
```diff
@@ -66,22 +66,45 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
...
+                        if memory[mid].0 < midpoint {
+                            left = mid;
+                        } else {
+                            right = mid;
+                        }
+                    }
+                    right
+                }
+            };
+            let left = Self::from_memory(memory, lookup_range.start..mid, height - 1, from, hasher);
```
jonathanpwang, line 105: I think in principle you could do left, right in different threads

jonathanpwang (approved):
LGTM, added some future ideas

crates/vm/src/system/memory/tree/mod.rs
```diff
@@ -95,19 +118,34 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
         memory: &MemoryImage<F>,
         hasher: &impl Hasher<CHUNK, F>,
     ) -> MemoryNode<CHUNK, F> {
-        // Construct a BTreeMap that includes the address space in the label calculation,
+        // Construct a Vec that includes the address space in the label calculation,
         // representing the entire memory tree.
-        let mut memory_partition = BTreeMap::new();
+        let mut memory_partition: Vec<(u64, [F; CHUNK])> = Vec::new();
```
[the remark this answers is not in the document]
jonathanpwang: see PR description, he tried and there was no perf diff

crates/vm/src/system/memory/tree/mod.rs
```diff
@@ -66,22 +66,45 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
...
         from: u64,
         hasher: &impl Hasher<CHUNK, F>,
     ) -> MemoryNode<CHUNK, F> {
-        let mut range = memory.range(from..from + (1 << height));
         if height == 0 {
-            let values = *memory.get(&from).unwrap_or(&[F::ZERO; CHUNK]);
-            MemoryNode::new_leaf(hasher.hash(&values))
-        } else if range.next().is_none() {
+            if lookup_range.is_empty() {
+                MemoryNode::new_leaf(hasher.hash(&[F::ZERO; CHUNK]))
```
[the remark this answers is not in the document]
zlangley: I think this is different than the PR description? PR description seems to be about calculating all zeroes for each height, while my suggestion was to do it just for height = 0.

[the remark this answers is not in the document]
Golovanov399: I'll try

crates/vm/src/system/memory/tree/mod.rs
Golovanov399: It most probably can

crates/vm/src/system/memory/tree/mod.rs
```diff
@@ -66,22 +66,59 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
...
+                zero_leaf,
+            );
+            let right = Self::from_memory(
+                memory,
+                mid..lookup_range.end,
+                height - 1,
+                midpoint,
+                hasher,
+                zero_leaf,
+            );
```
zlangley, line 121: Parallelizing this makes tracgen 36% faster: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13145453065

crates/vm/src/system/memory/tree/mod.rs
zlangley: https://github.com/openvm-org/openvm/pull/1341

crates/vm/src/system/memory/tree/mod.rs
Golovanov399: Now a little bit faster

crates/vm/src/system/memory/tree/mod.rs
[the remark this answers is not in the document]
Golovanov399: With the other changes I want to preserve this seems to increase gain, before that changes it didn't change the gain

crates/vm/src/system/memory/tree/mod.rs
```diff
@@ -95,19 +118,34 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
         memory: &MemoryImage<F>,
         hasher: &impl Hasher<CHUNK, F>,
     ) -> MemoryNode<CHUNK, F> {
-        // Construct a BTreeMap that includes the address space in the label calculation,
+        // Construct a Vec that includes the address space in the label calculation,
         // representing the entire memory tree.
-        let mut memory_partition = BTreeMap::new();
+        let mut memory_partition: Vec<(u64, [F; CHUNK])> = Vec::new();
         for ((address_space, pointer), value) in memory.items() {
+            if pointer as usize / CHUNK >= (1 << memory_dimensions.address_height) {
```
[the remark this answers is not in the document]
Golovanov399: This will happen when memory size is not divisible by page size, which, if both of these are powers of two, is equivalent to memory size being less than page size. We can assert that this is not the case and modify the tests correspondingly, in principle

crates/vm/src/system/memory/tree/mod.rs
```diff
@@ -140,34 +147,28 @@ impl<const CHUNK: usize, F: PrimeField32> MemoryNode<CHUNK, F> {
...
-            let index = memory_dimensions.label_to_index(label);
-            if memory_partition.is_empty() || memory_partition.last().unwrap().0 != index {
-                memory_partition.push((index, [F::ZERO; CHUNK]));
-            }
-            let chunk = memory_partition.last_mut().unwrap();
-            chunk.1[(pointer % CHUNK as u32) as usize] = value;
-        }
-        debug_assert!(memory_partition.is_sorted_by_key(|(addr, _)| addr));
+        let memory_items = memory
+            .items()
```
jonathanpwang, line 151: change `items()` to use `par_iter` now!

jonathanpwang: actually maybe it's too trivial, let's try separately of this PR

PR #1339, opened 2025-02-04
[perf] Opened Value Hint Direct

nyunyunyunyu:


nyunyunyunyu:
Some notes for the remaining changes:
- In instruction, the pointer of `hint_id` is provided. But memory read constraints of `hint_id` is unnecessary.
For `FriReducedOpening`
- In my implementation, `opened_values[hint_id][i]` is actually sent as (`hint_id`, `length-i-1`, `opened_values[hint_id][i]`) due to some internal details. It's possible to send `i` instead of `length-i-1` with 1 extra column.
For `VerifyBatch`
- Like `FriReducedOpening`, `VerifyBatch` also needs `OpenedValueBus` for interaction. 
- `VerifyBatch` is used for both `Felt` and `Ext`. Only `Felt` version use direct hint. `opened_values` in `Ext` version is based on a challenge. We need 2 different implementations. 
- Most changes should be around `InsideRow` to reduce memory reads. We might need 2 variants - `Felt` version uses direct hint. `Ext` version needs to read memory but it's always 2 `Ext`s(`Felt` version is variable - # of columns * `Felt`).
- Not 100% sure. Maybe `opened_element_size_inv` and `g` in instruction can be removed.

jonathanpwang:
The direct lookup approach has some problems, so we likely will not go with this approach.

PR #1340, opened 2025-02-04
[perf] Make `set_range` and `range_vec` faster via `unsafe`

Golovanov399:
**Metric:** tracegen
**Before:** 10,956.73 ms
**After:** 10,673.18 ms
**Percentage gain:** -2.6%

Workflow: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13162579475

---
Here we do some unsafe stuff similar to what we did in `range_array` when implementing pagevecs, but back then I somehow missed the same thing with vec-analogs of these functions.

### DIFF
--- a/crates/vm/src/system/memory/paged_vec.rs
+++ b/crates/vm/src/system/memory/paged_vec.rs
@@ -52,44 +52,45 @@ impl<T: Default + Clone, const PAGE_SIZE: usize> PagedVec<T, PAGE_SIZE> {
         // SAFETY: We set the length and then initialize every element.
         unsafe {
             result.set_len(len);
-            let dst = result.as_mut_ptr() as *mut T;
-            let start_page = range.start / PAGE_SIZE;
-            let end_page = (range.start + len - 1) / PAGE_SIZE;
-            if start_page == end_page {
-                let offset = range.start % PAGE_SIZE;
+        }
+        let dst = result.as_mut_ptr() as *mut T;
+        let start_page = range.start / PAGE_SIZE;
+        let end_page = (range.start + len - 1) / PAGE_SIZE;
+        if start_page == end_page {
+            let offset = range.start % PAGE_SIZE;
+            unsafe {
                 if let Some(page) = self.pages[start_page].as_ref() {
                     let src = page.as_ptr().add(offset);
                     std::ptr::copy_nonoverlapping(src, dst, len);
                 } else {
-                    for i in 0..len {
-                        std::ptr::write(dst.add(i), T::default());
-                    }
+                    std::slice::from_raw_parts_mut(dst, len).fill(T::default());
                 }
-            } else {
-                debug_assert_eq!(start_page + 1, end_page);
-                let offset = range.start % PAGE_SIZE;
-                let first_part = PAGE_SIZE - offset;
+            }
+        } else {
+            debug_assert_eq!(start_page + 1, end_page);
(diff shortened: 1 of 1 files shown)


crates/vm/src/system/memory/paged_vec.rs
```diff
@@ -44,36 +44,109 @@ impl<T: Default + Clone, const PAGE_SIZE: usize> PagedVec<T, PAGE_SIZE> {
...
-        for page_idx in (range.start / PAGE_SIZE)..range.end.div_ceil(PAGE_SIZE) {
-            let in_page_start = range.start.saturating_sub(page_idx * PAGE_SIZE);
-            let in_page_end = (range.end - page_idx * PAGE_SIZE).min(PAGE_SIZE);
-            if let Some(page) = self.pages[page_idx].as_ref() {
-                result.extend(page[in_page_start..in_page_end].iter().cloned());
+        let len = range.end - range.start;
+        // Create a vector for uninitialized values.
+        let mut result: Vec<MaybeUninit<T>> = Vec::with_capacity(len);
+        // SAFETY: We set the length and then initialize every element.
+        unsafe {
```
zlangley, line 53: Can we put the `unsafe` blocks *only* around the parts that use unsafe APIs? This will make it much easier to reason about safety.

jonathanpwang: it does seem like nearly the whole block is unsafe

jonathanpwang: maybe it can be split up into a few more blocks though

crates/vm/src/system/memory/paged_vec.rs
```diff
@@ -44,36 +44,109 @@ impl<T: Default + Clone, const PAGE_SIZE: usize> PagedVec<T, PAGE_SIZE> {
...
+                let dst_old = result.as_mut_ptr() as *mut T;
+                let src_old = page.as_ptr().add(offset);
+                std::ptr::copy_nonoverlapping(src_old, dst_old, len);
+
+                let dst_page = page.as_mut_ptr().add(offset);
+                let src_new = values.as_ptr();
+                std::ptr::copy_nonoverlapping(src_new, dst_page, len);
+            }
+        } else {
+            debug_assert_eq!(start_page + 1, end_page);
```
jonathanpwang, line 118: I think indeed it doesn't fit here, but just checking: do we have any higher level api check that block_size is less than PAGE_SIZE?

jonathanpwang (approved):
LGTM

A bit unfortunate the amount of code increased, but indeed one needs more logic.

crates/vm/src/system/memory/paged_vec.rs
```diff
@@ -44,36 +44,109 @@ impl<T: Default + Clone, const PAGE_SIZE: usize> PagedVec<T, PAGE_SIZE> {
...
+                let offset = range.start % PAGE_SIZE;
+                let first_part = PAGE_SIZE - offset;
+                if let Some(page) = self.pages[start_page].as_ref() {
+                    let src = page.as_ptr().add(offset);
+                    std::ptr::copy_nonoverlapping(src, dst, first_part);
+                } else {
+                    for i in 0..first_part {
+                        std::ptr::write(dst.add(i), T::default());
+                    }
+                }
```
zlangley, line 79: I feel like we can extract some of this repeated logic into a function for readability? (Rust should compile it the same.)

PR #1341, opened 2025-02-04
perf: Compute initial memory Merkle tree using multiple threads

zlangley:
**Metric**: tracegen
**Before**: 18,057.45 ms
**After**: 11,425.45 ms
**Percentage gain**: -36.7%

Workflow: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13145453065


jonathanpwang (approved):
LGTM!

PR #1345, opened 2025-02-06
[perf] direct hint opened values in recursion program

jonathanpwang:


jonathanpwang:
replaced by https://github.com/openvm-org/openvm/pull/1347

PR #1346, opened 2025-02-06
chore: update primitive crate to enable serde `std`

Avaneesh-axiom:
Duplicating fix #1343 on `develop`


jonathanpwang:
let me rebase `develop` against `main`, better to avoid these duplicates

PR #1349, opened 2025-02-08
[perf] merkle memory poseidon parallel

Golovanov399:
Towards INT-3256

Reth bench compare: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13213906613

### DIFF
--- a/crates/vm/src/system/memory/merkle/trace.rs
+++ b/crates/vm/src/system/memory/merkle/trace.rs
@@ -344,3 +344,26 @@ impl<F: PrimeField32> SerialReceiver<[F; PERIPHERY_POSEIDON2_WIDTH]> for Poseido
         }
     }
 }
+
+impl<'a, F: PrimeField32, const SBOX_REGISTERS: usize> SerialReceiver<&'a [F]>
+    for Poseidon2PeripheryBaseChip<F, SBOX_REGISTERS>
+{
+    /// Receives a permutation preimage, pads with zeros to the permutation width, and records.
+    /// The permutation preimage must have length at most the permutation width (panics otherwise).
+    fn receive(&mut self, perm_preimage: &'a [F]) {
+        assert!(perm_preimage.len() <= PERIPHERY_POSEIDON2_WIDTH);
+        let mut state = [F::ZERO; PERIPHERY_POSEIDON2_WIDTH];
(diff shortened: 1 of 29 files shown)


crates/vm/src/system/memory/merkle/trace.rs
```diff
@@ -117,33 +122,32 @@ struct TreeHelper<'a, const CHUNK: usize, F: PrimeField32> {
...
-        &mut self,
+        &self,
         height: usize,
         initial_node: &MemoryNode<CHUNK, F>,
         as_label: u32,
         address_label: u32,
-        hasher: &mut impl HasherChip<CHUNK, F>,
-    ) -> MemoryNode<CHUNK, F> {
+        hasher: &(impl HasherChip<CHUNK, F> + Send + Sync),
+    ) -> (MemoryNode<CHUNK, F>, Vec<MemoryMerkleCols<F, CHUNK>>) {
```
jonathanpwang, line 135: I think it's better if we just allocate a big 2^height array first, and then pass in pointers. can do this after we get everything working

crates/vm/src/system/memory/merkle/trace.rs
```diff
@@ -117,33 +122,32 @@ struct TreeHelper<'a, const CHUNK: usize, F: PrimeField32> {
...
         if height == 0 {
             let address_space = as_label + self.memory_dimensions.as_offset;
             let leaf_values = *self
                 .final_memory
                 .get(&(address_space, address_label))
                 .unwrap_or(&[F::ZERO; CHUNK]);
-            MemoryNode::new_leaf(hasher.hash(&leaf_values))
+            (MemoryNode::new_leaf(hasher.hash(&leaf_values)), vec![])
```
jonathanpwang, line 142: note to self: beware that `new_leaf` no longer has record

crates/vm/src/system/memory/merkle/trace.rs
jonathanpwang: nevermind: (1) 2^height is too big, (2) I used `crossbeam_channel` instead

jonathanpwang:
The lesson is that channels are bad for these lightweight thread operations due to potentially locking. Or the thread memory bandwidth is bad.

We will need to redesign the merkle tree memory layout to do better parallelism. Closing now.

PR #1350, opened 2025-02-08
fix: Check shape of inputs

zlangley:


extensions/native/recursion/src/stark/mod.rs
```diff
@@ -145,10 +145,29 @@ where
             air_perm_by_height,
         } = proof;
 
-        let num_airs = RVar::from(air_proofs.len());
+        let num_airs = RVar::from(m_advice.per_air.len());
```
zlangley, line 148: This doesn't account for optional AIRs.

jonathanpwang: I think it should be the count without optional. `m_advice_per_air` does have the optional. 
@nyunyunyunyu please help

extensions/native/recursion/src/stark/mod.rs
```diff
@@ -468,56 +498,60 @@ where
...
-                                values,
-                                points: trace_points,
-                            };
-                            builder.set_value(&mats, mat_idx.clone(), after_challenge_mat);
-                            builder.inc(&mat_idx);
-                        });
-                });
-                builder.assert_eq::<Usize<_>>(mat_idx, values_per_mat.len());
+        builder.assert_eq::<Usize<_>>(opening.values.after_challenge.len(), num_phases);
+        builder.if_eq(num_phases, RVar::one()).then(|builder| {
```
jonathanpwang, line 502: what is this doing? are you requiring there to be one challenge phase?

extensions/native/recursion/src/stark/mod.rs
nyunyunyunyu: Right, we had already discussed offline. This pr made some wrong assertion due to optional airs

extensions/native/recursion/src/stark/mod.rs
zlangley: We already require there to be at most one challenge phase in many places. This is just making it more explicit.

extensions/native/recursion/src/fri/two_adic_pcs.rs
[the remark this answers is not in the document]
zlangley: There is a comment somewhere that our implementation only works for log_final_poly_len = 0, and we assert that.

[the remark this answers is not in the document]
nyunyunyunyu: @zlangley mentioned this assertion:
```
builder.assert_usize_eq(proof.final_poly.len(), RVar::from(config.blowup));
```

This seems weird to me because we only use the first element in `final_poly` here: https://github.com/openvm-org/openvm/blob/2330fd6392b2e0cbf6aea7227bd1d2d780e40b72/extensions/native/recursion/src/fri/two_adic_pcs.rs#L284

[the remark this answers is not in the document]
zlangley: Our verifier runs FRI until the polynomial is constant, so final_poly should be an RS-code of a constant polynomial.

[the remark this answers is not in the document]
nyunyunyunyu: You mean we don't need this loop? https://github.com/Plonky3/Plonky3/blob/b0591e9b82d58d10f86359875b5d5fa96433b4cf/fri/src/verifier.rs#L97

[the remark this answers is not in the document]
jonathanpwang: Plonky3 added early stopping https://github.com/Plonky3/Plonky3/pull/579 
We do not support this as it has unclear optimization benefits.

Our implementation matches the plonky3 behavior before this PR.
I think the confusion is that `proof.final_poly.len()` should always equal `2^(log_blowup + log_final_poly_len)`, due to bad naming on the plonky3 side.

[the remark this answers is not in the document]
jonathanpwang: We do not need the referenced loop, because we expect `final_poly` to be `constant, 0, 0, ..., 0`. Therefore the `folded_eval = constant`, which is what we are checking. This is what the final FRI low degree test reduces to. We do sample the additional terms of `final_poly` but I don't think this leads to any soundness issue.

We could provide an additional assertion that the other terms in `final_poly` are all zero, just to be defensive.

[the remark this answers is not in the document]
jonathanpwang: Equivalently, let us only sample `final_poly[0]`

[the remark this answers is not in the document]
jonathanpwang: Since plonky3 samples all `blowup` number of final_poly, and technically we should match their implementation and inputs exactly, I did https://github.com/openvm-org/openvm/pull/1350/commits/71f184908845b821023888e695ea9e322199e0de

Alternatively we could change our final poly to go back to the old implementation. Since this is negligible in performance, let's just do the simplest thing for now (which is the commit above)

extensions/native/recursion/src/stark/mod.rs
```diff
@@ -216,60 +225,65 @@ where
                 builder.unsafe_cast_var_to_felt(air_proof.log_degree.get_var())
             };
             challenger.observe(builder, log_degree);
+
+            // Constrain that degree_bits is in [0, MAX_TWO_ADICITY].
+            // If we add less-than chip, then we can do this a less-than instruction.
+            for i in 0..=MAX_TWO_ADICITY {
+                let diff = builder.eval(air_proof.log_degree.clone() - RVar::from(i));
+                builder.assert_nonzero(&diff);
+            }
```
zlangley, line 234: Brain fart--fixing.

extensions/native/recursion/src/view.rs
```diff
@@ -35,6 +35,8 @@ pub fn get_advice_per_air<C: Config>(
             builder.inc(&idx);
         });
     }
+    // Make sure we didn't index out of bounds; final idx should be <= air_ids.len().
+    builder.assert_less_than_slow(idx.clone(), air_ids.len() + RVar::one());
```
nyunyunyunyu, line 39: This cannot avoid out-of-bound access. It seems better to add `builder.if_ne(idx, air_ids.len())` in L31.

extensions/native/recursion/src/stark/mod.rs
```diff
@@ -145,17 +145,48 @@ where
...
             let num_airs = num_airs.value();
             let perm = (0..num_airs).map(|i| builder.eval(RVar::from(i))).collect();
             &builder.vec(perm)
         } else {
+            builder.assert_usize_eq(air_perm_by_height.len(), num_airs);
+            // Assert that each index in `air_perm_by_height` is unique and in range [0, num_airs).
+            let mask: Array<_, Usize<_>> = builder.dyn_array(num_airs);
+            let one: Usize<_> = builder.eval(C::N::ONE);
+            iter_zip!(builder, air_perm_by_height).for_each(|ptr_vec, builder| {
+                let perm_i = builder.iter_ptr_get(air_perm_by_height, ptr_vec[0]);
```
nyunyunyunyu, line 169:
```suggestion
                let perm_i = builder.iter_ptr_get(air_perm_by_height, ptr_vec[0]);
                builder.assert_less_than_slow(perm_i, num_airs);
```

nyunyunyunyu: To avoid arbitrary memory write

nyunyunyunyu (approved):

PR #1352, opened 2025-02-08
perf: trivial opt to reduce alloc in recursion

jonathanpwang:
Trivial optimization to save a few allocations.

Reth benchmark comparison: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13212925789
Metric: reth root circuit main cells used
After: (-309252 [-0.9%]) 34,194,443

### DIFF
--- a/extensions/native/recursion/src/fri/two_adic_pcs.rs
+++ b/extensions/native/recursion/src/fri/two_adic_pcs.rs
@@ -99,12 +99,14 @@ pub fn verify_two_adic_pcs<C: Config>(
     };
     builder.cycle_tracker_end("pre-compute-rounds-context");
 
+    // Working variables for reduced opening, reset per query. Both arrays indexed by log_height.
+    let ro: Array<C, Ext<C::F, C::EF>> = builder.array(32);
+    let alpha_pow: Array<C, Ext<C::F, C::EF>> = builder.array(32);
+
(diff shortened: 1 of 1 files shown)


zlangley (approved):

extensions/native/recursion/src/fri/two_adic_pcs.rs
```diff
@@ -99,12 +99,14 @@ pub fn verify_two_adic_pcs<C: Config>(
     };
     builder.cycle_tracker_end("pre-compute-rounds-context");
 
+    // Working variables for reduced opening, reset per query. Both arrays indexed by log_height.
+    let ro: Array<C, Ext<C::F, C::EF>> = builder.array(32);
```
yi-sun, line 103: Could we use 31? or even a few bits lower?  

We should make this a generic `MAX_LOG_HEIGHT` (and replace all other instances of this in the code).

yi-sun (changes requested):
Document source of magic number 32 and make into generic constant.

extensions/native/recursion/src/fri/two_adic_pcs.rs
jonathanpwang: done in https://github.com/openvm-org/openvm/pull/1352/commits/89511e377092f8c5d37edb36a9fe29d7347a8068

PR #1353, opened 2025-02-08
perf: better halo2-base `select` implementation

MonkeyKing-1:
Not really that important, just patched select to be better (you can reduce from 8 cells to 6 cells). It's not even used that much anyway.


jonathanpwang:
Unfortunately `halo2-base` doesn't support gates `2` apart. The optimization seems valid, but it seems we can't merge this unless we make some changes to `halo2-base`.

jonathanpwang:
Will address in `halo2-base` later

PR #1354, opened 2025-02-08
clippy: add more clippy rules for vm crate

tcoratger:
A list of already proven clippy lints taken from https://github.com/paradigmxyz/reth/blob/46d3b6a32eac58485dd7f4d217060aeea50feadc/Cargo.toml#L170-L255

In my experience this can help with some tricky stuffs so if you think this is useful, we can propagate these rules in the other crates in follow up PRs for ex. What do you thunk @jonathanpwang ?


crates/vm/src/system/memory/offline_checker/columns.rs
```diff
@@ -83,3 +87,12 @@ pub struct MemoryReadOrImmediateAuxCols<T> {
     pub(crate) is_immediate: T,
     pub(crate) is_zero_aux: T,
 }
+
+impl<T, const N: usize> AsRef<MemoryReadAuxCols<T>> for MemoryWriteAuxCols<T, N> {
```
jonathanpwang, line 91: was this added in this PR? or is it just the github diff is weird?

docs/specs/ISA.md
jonathanpwang, line 1: I think you need to rebase this file again

Cargo.toml
```diff
@@ -67,6 +67,92 @@ members = [
 exclude = ["crates/sdk/example"]
 resolver = "2"
 
+[workspace.lints.clippy]
+# These are some of clippy's nursery (i.e., experimental) lints that we like.
```
jonathanpwang, line 71: Add a link to reth so we know where we got it from

jonathanpwang (commented):
You'll need to rebase one (last) time, and then I think it'll be good

Cargo.toml
tcoratger: done

PR #1355, opened 2025-02-09
perf: precompute alpha pow accumulators per matrix

jonathanpwang:
All `alpha` power calculations in the reduced opening computation in 2-adic FRI verification only depend on the general shape of the proot input (number of matrices, matrix widths, number of out-of-domain points to open at). The key observation is that **none of this depends on the FRI query** so it should be precomputed and cached as part of the `RoundContext`.

Also, some alpha power calculations in the static verifier can be memoized even further, which is now done in a `static_alpha_pows` vector.

Reth comparison benchmark: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13229825060

## Metrics
Reth root circuit main cells used: 
- before: 33,353,572
- after: 32,952,509 (-1.2%)

Reth halo2_outer main cells used: 
- before: 58,181,721
- after: 57,429,149 (-1.2%)

### DIFF
--- a/extensions/native/recursion/src/fri/two_adic_pcs.rs
+++ b/extensions/native/recursion/src/fri/two_adic_pcs.rs
@@ -156,7 +156,7 @@ pub fn verify_two_adic_pcs<C: Config>(
                     ov_ptrs,
                     perm_ov_ptrs,
                     batch_dims,
-                    alpha_pow_acc_per_mat,
+                    alpha_pow_acc,
                     log_batch_max_height,
                 } = round_context;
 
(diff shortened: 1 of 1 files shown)


extensions/native/recursion/src/fri/two_adic_pcs.rs
```diff
@@ -87,27 +87,16 @@ pub fn verify_two_adic_pcs<C: Config>(
...
-        }
-        ret
-    } else {
-        vec![]
-    };
+    let rounds_context =
+        compute_rounds_context(builder, &rounds, log_blowup, alpha, &mut static_alpha_pows);
     builder.cycle_tracker_end("pre-compute-rounds-context");
 
     // Working variables for reduced opening, reset per query. Both arrays indexed by log_height.
```
yi-sun, line 98:
```suggestion
    // Accumulators of the reduced opening sums, reset per query. The array `ro` is indexed by log_height.
```

extensions/native/recursion/src/fri/two_adic_pcs.rs
```diff
@@ -453,38 +446,58 @@ fn compute_rounds_context<C: Config>(
         let ov_ptrs: Array<C, Array<C, Felt<C::F>>> = builder.array(round.mats.len());
         let perm_ov_ptrs: Array<C, Array<C, Felt<C::F>>> = builder.array(round.mats.len());
         let batch_dims: Array<C, DimensionsVariable<C>> = builder.array(round.mats.len());
-        let mat_alpha_pows: Array<C, Ext<_, _>> = builder.array(round.mats.len());
+        let alpha_pow_acc: Array<C, Array<C, Ext<_, _>>> = builder.array(round.mats.len());
```
yi-sun, line 449: Add comment that this is indexed by `alpha_pow_acc[mat_idx][ood_point_idx]`

jonathanpwang:
Since the perf gain is not as big as expected, and the diff is large, we will close for now but possibly revisit.

PR #1363, opened 2025-02-12
clippy: add more clippy rules for vm crate

tcoratger:
Superseed https://github.com/openvm-org/openvm/pull/1354 to avoid strange rebasing behaviour

@jonathanpwang I think we should merge main into develop because I suspect that develop is not up to date with some of the latest changes


jonathanpwang (approved):
LGTM

jonathanpwang:
@tcoratger on second thought, I'm going to close this for now because I'm afraid of merge/rebase conflicts with `main`. I do still thinking adding these is a good idea, but will try to do it after we've finalized a release on `main`. Please remind me if I haven't done it for a while.

PR #1366, opened 2025-02-12
[chore] Fix openvm links for example guest programs

manh9203:
- Change the OpenVM links in the example guest programs to the `develop` branch so that the lint check passes


jonathanpwang (approved):
wow there were a lot, thanks

PR #1367, opened 2025-02-13
[chore] Add public `const` constructors

Avaneesh-axiom:
- changed visibility of `from_const_bytes` in `moduli_declare` to public
- Added a public const constructor for structs created by `impl_sw_affine`

Required by INT-3315

### DIFF
--- a/extensions/ecc/guest/src/weierstrass.rs
+++ b/extensions/ecc/guest/src/weierstrass.rs
@@ -243,6 +243,12 @@ macro_rules! impl_sw_affine {
         #[repr(transparent)]
         pub struct $struct_name(AffinePoint<$field>);
 
+        impl $struct_name {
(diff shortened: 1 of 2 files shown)


extensions/ecc/guest/src/weierstrass.rs
```diff
@@ -243,6 +243,12 @@ macro_rules! impl_sw_affine {
         #[repr(transparent)]
         pub struct $struct_name(AffinePoint<$field>);
 
+        impl $struct_name {
+            pub const fn new(x: $field, y: $field) -> Self {
```
jonathanpwang, line 247: there's a `From` impl below so I think we can skip this

jonathanpwang: nvm `From` is not const...

jonathanpwang (approved):
LGTM

PR #1369, opened 2025-02-14
feat: Add support for SHA-512 into the existing SHA-256 VM chip

Avaneesh-axiom:
WIP

Note: need to update openvm commit in `examples/sha2/Cargo.toml` to `develop` after PR is merged.

Closes INT-3213

PR #1370, opened 2025-02-14
[feat] Modify existing SHA-256 subair to support SHA-512

Avaneesh-axiom:
CI won't pass because the sha vm chip is not updated (that's in #1369) but the tests in `sha-air` pass

- Introduced the `sha-macros` crate that provides derive macros to generate reference-versions of the sha subair columns structs.
- The `ShaConfig` trait holds associated constants that determine the sha2 variant (i.e. SHA256 or SHA512).

Closes INT-3212


Avaneesh-axiom:
Closing because I'll update both the subair and vm chip in a single PR.

PR #1371, opened 2025-02-15
feat: Memory access instrumentation

stephenh-axiom-xyz:
Collect and emit memory-related metrics for analysis.
- Emits number of reads and writes per opcode
- TODO: Visualize gaps in memory due to alignment issues

### DIFF
--- a/crates/vm/src/metrics/mod.rs
+++ b/crates/vm/src/metrics/mod.rs
@@ -64,6 +64,10 @@ where
 
             #[cfg(feature = "function-span")]
             self.metrics.update_current_fn(pc);
+        } else {
+            let executor = self.chip_complex.inventory.get_executor(opcode).unwrap();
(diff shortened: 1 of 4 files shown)


crates/vm/src/metrics/mod.rs
```diff
@@ -28,6 +28,8 @@ pub struct VmMetrics {
     #[allow(dead_code)]
     pub(crate) current_fn: FnBound,
     pub(crate) current_trace_cells: Vec<usize>,
+    // Maps opcode to number of (instances, reads, writes)
+    pub opcode_mem_accesses: BTreeMap<String, (usize, usize, usize)>,
```
jonathanpwang, line 32: I think what we likely care more about is where these reads and write are located (which addresses), and what their alignment and block sizes look like.

jonathanpwang:
closing to be revisited later

PR #1372, opened 2025-02-16
chore: fix macro usage order to prevent potential compilation errors

famouswizard:
I noticed that `openvm::entry!(main);` was placed before the `use` statements. This could lead to compilation issues since the macro might reference types or functions that haven't been declared yet.

P.S. In Rust, `use` statements should always come before their usage. I've moved the macro after the imports to ensure proper compilation.


jonathanpwang (approved):

jonathanpwang:
This will not lead to compilation errors. It is purely cosmetic.

PR #1374, opened 2025-02-18
manual serde

luffykai:


crates/vm/src/serde/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+        let num_pages = u64::from_le_bytes(data[0..8].try_into()?);
+        let mut current_pos = 8;
+
+        // Create vector to store pages
+        let mut pages = Vec::with_capacity(num_pages as usize);
+
+        // Read each page
+        for _ in 0..num_pages {
+            if current_pos >= data.len() {
+                return Err(eyre::eyre!("Unexpected end of data while reading pages"));
```
jonathanpwang, line 163: it's bad practice to use `eyre` in a library; use thiserror

PR #1376, opened 2025-02-18
chore: remove all `Stateful` implementations

Jocn2020:
Sync with https://github.com/openvm-org/stark-backend/pull/45
Reverts https://github.com/openvm-org/openvm/pull/1199 and https://github.com/openvm-org/openvm/pull/1211

Closes INT-3335

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4478,7 +4478,7 @@ dependencies = [
 [[package]]
 name = "openvm-stark-backend"
 version = "1.0.0-rc.0"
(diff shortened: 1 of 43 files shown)


crates/circuits/primitives/derive/src/lib.rs
```diff
@@ -315,9 +315,9 @@ pub fn chip_usage_getter_derive(input: TokenStream) -> TokenStream {
 pub fn bytes_stateful_derive(input: TokenStream) -> TokenStream {
```
luffykai, line 315: don't change this file. just leave the macro here and don't use it

Jocn2020: I need to remove it, otherwise it gives compiler issue of `.load_state is not implemented`

jonathanpwang (approved):
LGTM

PR #1380, opened 2025-02-23
Fix redundant "Read" in Rv32VecHeapTwoReadsReadRecord struct name

crStiv:
Renamed Rv32VecHeapTwoReadsReadRecord to Rv32VecHeapTwoReadsRecord to eliminate redundancy in the struct name, as "Read" was already present in "TwoReads" part. Updated all related references.


jonathanpwang:
While the naming is funny, I prefer to keep the current naming because it's intended to parallel `Rv32VecHeapTwoReadsWriteRecord`

PR #1382, opened 2025-02-24
fix: typos

crStiv:
Corrects the misspelling of "lastest" to "latest" in variable names and function parameters across SHA256-related files:
- extensions/sha256/circuit/src/sha256_chip/air.rs
- crates/circuits/sha256-air/src/air.rs


jonathanpwang:
`is_latest` is I believe intentional, because it really means last of the last. Not going to make changes to security critical code (the AIR constraints) for now.

PR #1348, opened 2025-02-06
[feat] Add constructors for Ecdsa verifying key

manh9203:
- Add constructors for `VerifyingKey<C>` that takes a public key or an affine point
- Add methods to encode and decode in SEC1 format

PR #1384, opened 2025-02-25
[feat] Add constructors for ECDSA verifying key on `main`

manh9203:
Add constructors for ECDSA verifying key on `main` (mirror commit of https://github.com/openvm-org/openvm/pull/1348).


jonathanpwang (approved):
LGTM

PR #1387, opened 2025-02-27
fix(bench): fix the issue segment cycle counters being [0,0,0,..,$value]

lispc:
current codes will emit cycle counter metrics as `[0,0,0,0, ..., $count_of_last_segment]`.

The reason is when `finalize_metric` is called, the metric inside a segment has already be move out (to the new segment, and then the cycle_counter is cleared before new segment execution).

This PR will keep local metrics inside local segment to fix the issue.

I also want to make `finalize_metrics` public.  This can help with outside metric recorder. 

a further question is, what about make segment.metrics public? then caller can access metric directly without need of MetricRecorder?


jonathanpwang:
I think we did the same thing. Can you review https://github.com/openvm-org/openvm/pull/1388 instead?

PR #1388, opened 2025-02-27
fix(metrics): cycle count emission across segments

jonathanpwang:
Previously when proving with multiple segments, we execute all segments, but we take the `metrics` from each `segment` before trace generation. When we do trace generation, we finalize memory and then finalize metrics and emit, but at this point the metrics in `segment` are partially empty. This led to the cycle count emission to always be those of the last segment.

The `take` from `metrics` was only needed to get the cycle tracker and function symbol information, so now we do a `partial_take` to avoid cloning.

Old reth benchmark: https://github.com/axiom-crypto/openvm-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/main/reth-bb4d597ec9b5edfc486d022fa2ae29202f7e6281-e2fa8f8fba28e8315642c3ddeb11fcd29a7e53d06ecdeba44aae28ee1aef837e.md
New reth benchmark: https://github.com/axiom-crypto/openvm-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/main/reth-bb4d597ec9b5edfc486d022fa2ae29202f7e6281-e2fa8f8fba28e8315642c3ddeb11fcd29a7e53d06ecdeba44aae28ee1aef837e.md

closes INT-3496

### DIFF
--- a/crates/vm/src/arch/segment.rs
+++ b/crates/vm/src/arch/segment.rs
@@ -126,7 +126,7 @@ where
     pub(crate) air_names: Vec<String>,
     /// Metrics collected for this execution segment alone.
     #[cfg(feature = "bench-metrics")]
-    pub(crate) metrics: VmMetrics,
+    pub metrics: VmMetrics,
 }
 
 pub struct ExecutionSegmentState {
--- a/crates/vm/src/arch/vm.rs
+++ b/crates/vm/src/arch/vm.rs
@@ -187,7 +187,6 @@ where
         #[cfg(feature = "bench-metrics")]
         {
             segment.metrics = from_state.metrics;
-            segment.metrics.clear();
         }
         if let Some(overridden_heights) = self.overridden_heights.as_ref() {
             segment.set_override_trace_heights(overridden_heights.clone());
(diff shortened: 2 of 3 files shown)


lispc (approved):
i think this pr will fix the problem.

other changes in my PR you can either choose to pick or not:

1. i remove the Metric.clear() at the begin of segment execution.   With the change,  it is the Metric only being prepared once.  Current code (build a new Metric at the end of last segment, and ALSO clear it in the begin of next) is "defensive" but redundant.  
2. i also expose Metrics as public.  So sdk users can count cycles without MetricRecorder. wdyt?

jonathanpwang:
> i think this pr will fix the problem.
> 
> other changes in my PR you can either choose to pick or not:
> 
> 1. i remove the Metric.clear() at the begin of segment execution.   With the change,  it is the Metric only being prepared once.  Current code (build a new Metric at the end of last segment, and ALSO clear it in the begin of next) is "defensive" but redundant.
> 2. i also expose Metrics as public.  So sdk users can count cycles without MetricRecorder. wdyt?

Both seem good to me; applied them in https://github.com/openvm-org/openvm/pull/1388/commits/1c4238a8b2140aaf8e4ed98259d7ae359e2ee933

Issue #1386, opened 2025-02-26
Bug: `cargo openvm bench  --profile ..` is not working


alv-around:
Hi!

I noticed while testing things in [openvm-example-fibonacci](https://github.com/openvm-org/openvm-example-fibonacci) in the following manner:

1. When running `cargo openvm bench` it requires you to provide a profile. error message:
```console
$ cargo openvm bench
error: The following required argument was not provided: profile

Usage: cargo <COMMAND>

For more information, try '--help'.
```

2. When defining a profile such as `release`/ `debug` / `profiling` (user defined in `Cargo.toml`) the profile is not recognised. See error:
```console
$ cargo openvm bench --profile debug
error: unexpected argument 'debug' found

Usage: cargo openvm bench [OPTIONS]

For more information, try '--help'.
```

I can fix provide a fix, but I wanted it to flag if first 🎏

jonathanpwang:
Specifically for `debug` I believe the profile is `dev` in cargo.
Does profile=release also not work?

alv-around:
> Specifically for `debug` I believe the profile is `dev` in cargo. Does profile=release also not work?

yes, you are right the `debug` example was not most appropriate. But the problem still remains `release`, `dev` or a user specific profile like `profiling`.

The problem is that the argument `profile` is present twice. One is used as a flag enabling tracing, and the other (the right one inherited from the `build` command which is the actual compiler profile. I'll submit a PR fixing that and another issue, I encountered with `bench`.

![Image](https://github.com/user-attachments/assets/7180ba07-dd66-4b94-846b-5f3dc396ff5c)

PR #1392, opened 2025-02-28
Bugfix cli bench profile option

alv-around:
This PR fixes the broken cli command `bench`.

1. It renames the argument which enables tracing to `tracing`. The previous name was collisioning with another argument by the name of `profile` in charge of setting the compiler profile. See more details in #1386.
2. After the rename the command was still not working. This is because, the command was calling some functions which are not currently used in the comands `build` and `run`. I rewrite that part so that `bench` reuses the code from `build` and `run` to avoid the splitting of functionality. **The command does should now work again.** Howerver, I would suggest to apply further this approach, so that bench reuses as much funcitonality from the other commands (such as keygen, prove and verify).

### DIFF
--- a/crates/cli/src/commands/bench.rs
+++ b/crates/cli/src/commands/bench.rs
@@ -46,8 +46,12 @@ impl BenchCmd {
             setup_tracing();
         }
 
-        let elf_path = build(&self.build_args)?.unwrap();
-        let (exe, _vm_config) = execute(elf_path, &self.build_args.config, &self.input).unwrap();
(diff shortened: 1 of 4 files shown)


jonathanpwang:
I am more inclined to just delete the `bench` command, so it is more of an amalgamation of the other commands but as you pointed out not implemented in a maintanable way.

jonathanpwang:
replacing with https://github.com/openvm-org/openvm/pull/1442

PR #1394, opened 2025-03-07
fix: hint offset between segments for FriReducedOpeningChip

lispc:
the offsets is set to 0 when segment transition, which is not correct.


lispc:
what is the meaning of this deployment failed CI err?..

jonathanpwang:
> what is the meaning of this deployment failed CI err?..

It's Vercel from the mdbook, but it shouldn't trigger -- I have been trying but unable to figure out why.

jonathanpwang (approved):

PR #1398, opened 2025-03-10
chore: fix matrix multiplication by using values

0xbryer:
I changed the matrix multiplication to use values directly instead of references.
This fixes errors with the arithmetic operations and ensures the code works properly.


jonathanpwang:
The reference is intentional as our implementation of multiplication works on references to prevent clone

PR #1402, opened 2025-03-10
fix: fix unused result of function call

ozrg:
I spotted that the result of the function `f` wasn’t being used, even though it seems like the intention was to add it to `test_cases`. I’ve updated the code to push the result properly into the array.


nyunyunyunyu (approved):
LGTM, thanks for the fix!

PR #1403, opened 2025-03-10
fix: Poseidon2 verify_batch `start_top_level` is not constrained only at start

jonathanpwang:
**Finding Link:** https://cantina.xyz/code/c486d600-bed0-4fc6-aed1-de759fd29fa2/findings/167

### Description of Fix

While the variable was called `start_top_level`, we only constrained that the top level has `start_top_level = 1` but never that it must be equal to `0` on non-top levels. We change the existing constraint so that it enforces `start_top_level = 1` iff its the start of a top level (meaning previous row is `end` and current row is `incorporate_row`).

Closes INT-3567

### DIFF
--- a/extensions/native/circuit/src/poseidon2/air.rs
+++ b/extensions/native/circuit/src/poseidon2/air.rs
@@ -417,7 +417,6 @@ impl<AB: InteractionBuilder, const SBOX_REGISTERS: usize> Air<AB>
 
         let mut when_top_level_not_end =
             builder.when(incorporate_row + incorporate_sibling - end_top_level);
-        when_top_level_not_end.assert_zero(next.start_top_level);
         when_top_level_not_end
             .assert_eq(next_top_level_specific.dim_base_pointer, dim_base_pointer);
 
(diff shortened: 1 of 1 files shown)


TlatoaniHJ (approved):
Looks good to me. I should note that when the row is not top level at all we are not constraining start_top_level. We don't need to because the value of start_top_level is not relevant for those rows (setting it to 1 on those rows neither allows an exploit nor causes a constraint to fail), but this maybe should be documented (or we could just constrain it to be 0 on those rows).

jonathanpwang:
> Looks good to me. I should note that when the row is not top level at all we are not constraining start_top_level. We don't need to because the value of start_top_level is not relevant for those rows (setting it to 1 on those rows neither allows an exploit nor causes a constraint to fail), but this maybe should be documented (or we could just constrain it to be 0 on those rows).

I'll constrain it's zero just for clarity.

zlangley:
Can't we just do
```
    builder.assert_eq(end.clone() * next.incorporate_row, next.start_top_level);
```
instead of
```
    builder
        .when(end.clone())
        .when(next.incorporate_row)
        .assert_one(next.start_top_level);
```

jonathanpwang:
> Can't we just do
> 
> ```
>     builder.assert_eq(end.clone() * next.incorporate_row, next.start_top_level);
> ```
> 
> instead of
> 
> ```
>     builder
>         .when(end.clone())
>         .when(next.incorporate_row)
>         .assert_one(next.start_top_level);
> ```

I think so, will add it.

PR #1406, opened 2025-03-11
[chore] Remove unnecessary constraints in RV32IM extension

manh9203:
Removed unnecessary or duplicated constraints in the RV32IM extension.

### DIFF
--- a/extensions/rv32im/circuit/src/adapters/loadstore.rs
+++ b/extensions/rv32im/circuit/src/adapters/loadstore.rs
@@ -211,12 +211,6 @@ impl<AB: InteractionBuilder> VmAdapterAir<AB> for Rv32LoadStoreAdapterAir {
         // constrain is_load == 1 and rd == x0 when write_count == 0
         builder.assert_bool(write_count);
         builder.when(write_count).assert_one(is_valid.clone());
-        builder
-            .when(is_valid.clone() - write_count)
-            .assert_one(is_load.clone());
-        builder
-            .when(is_valid.clone() - write_count)
-            .assert_zero(local_cols.rd_rs2_ptr);
 
         // read rs1
         self.memory_bridge
--- a/extensions/rv32im/circuit/src/divrem/core.rs
+++ b/extensions/rv32im/circuit/src/divrem/core.rs
@@ -239,10 +239,6 @@ where
             .when(nonzero_q)
             .when(not(cols.zero_divisor))
             .assert_eq(cols.q_sign, cols.sign_xor);
-        builder
-            .when_ne(cols.q_sign, cols.sign_xor)
-            .when(not(cols.zero_divisor))
-            .assert_zero(cols.q_sign);
 
         // Check that the signs of b and c are correct.
         let sign_mask = AB::F::from_canonical_u32(1 << (LIMB_BITS - 1));
--- a/extensions/rv32im/circuit/src/adapters/alu.rs
+++ b/extensions/rv32im/circuit/src/adapters/alu.rs
@@ -153,10 +153,7 @@ impl<AB: InteractionBuilder> VmAdapterAir<AB> for Rv32BaseAluAdapterAir {
         );
         self.bitwise_lookup_bus
             .send_range(rs2_limbs[0].clone(), rs2_limbs[1].clone())
-            .eval(
-                builder,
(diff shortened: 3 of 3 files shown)


extensions/rv32im/circuit/src/adapters/loadstore.rs
```diff
@@ -211,12 +211,6 @@ impl<AB: InteractionBuilder> VmAdapterAir<AB> for Rv32LoadStoreAdapterAir {
         // constrain is_load == 1 and rd == x0 when write_count == 0
         builder.assert_bool(write_count);
         builder.when(write_count).assert_one(is_valid.clone());
-        builder
-            .when(is_valid.clone() - write_count)
-            .assert_one(is_load.clone());
-        builder
-            .when(is_valid.clone() - write_count)
-            .assert_zero(local_cols.rd_rs2_ptr);
```
manh9203, line 219: I think we don’t need to constrain write_count since it’s an instruction operand and is already constrained by the execution bus.

extensions/rv32im/circuit/src/divrem/core.rs
```diff
@@ -239,10 +239,6 @@ where
             .when(nonzero_q)
             .when(not(cols.zero_divisor))
             .assert_eq(cols.q_sign, cols.sign_xor);
-        builder
-            .when_ne(cols.q_sign, cols.sign_xor)
-            .when(not(cols.zero_divisor))
-            .assert_zero(cols.q_sign);
```
manh9203, line 245: This constraint duplicates the one above it, so I removed it and added a comment clarifying the case when q is zero in the specs PR.

extensions/rv32im/circuit/src/adapters/loadstore.rs
jonathanpwang: this isn't about constraining write count but instead about asserting `is_load=1, rd_rs2_ptr=0` when its not a write? 

either explain in much more detail why we can remove, or just keep it

extensions/rv32im/circuit/src/divrem/core.rs
jonathanpwang: it is pretty hard to understand even the comment without knowing the details of this chip. for example I don't even know how q_ext is supposed to related to `b = c * q + r`.
please explain further

jonathanpwang (changes requested):
The loadstore and divrem removals require more explanation (or just don't remove them)

extensions/rv32im/circuit/src/adapters/loadstore.rs
manh9203: I think it is the transpiler’s job to ensure that assertions is true—similar to the [RdWrite adapter](https://github.com/openvm-org/openvm/blob/main/extensions/rv32im/circuit/src/adapters/rdwrite.rs#L122-L129), where we didn’t impose any constraints on `need_write`. If `need_write` doesn't meet the assertion, then the instruction fetched from the program ROM would be incorrect, causing the execution bus to fail.

extensions/rv32im/circuit/src/divrem/core.rs
manh9203: `q_ext`, `b_ext`, and `c_ext` use `q_sign`, `b_sign`, and `c_sign` to sign-extend `b`, `c`, and `q`, respectively. The sign-extended values are then used to verify that: `b = c * q + r` and `|r| < |c|`.

Since `q` and `r` are uniquely determined by these constraints, an incorrect `q_sign` would lead to an incorrect `q_ext`, causing one of these constraints to fail.

jonathanpwang: do you mean `b_ext = c_ext * q_ext + r_ext`?

manh9203: yes

jonathanpwang: Please leave this constraint, but you can leave a comment that you think it is already implied

extensions/rv32im/circuit/src/adapters/loadstore.rs
jonathanpwang: please leave a code comment, but leave the constraints intact

jonathanpwang (approved):

PR #1409, opened 2025-03-11
fix: Fix bug in setup row of modular equality chip (`ModularIsEqualCoreChip`)

Avaneesh-axiom:
This PR fixes a bug in the algebra extension that was found during the internal security review.

**Description:** The setup row of the `ModularIsEqualCoreChip` is unsound. It does not correctly constrain that the modulus is correct. The problem is that the casework on `prefix_sum` that allows constraining `b[i] = N[i]` for the correct indices (when `i > b_diff_idx`) fails on the setup row. The intended behavior on the setup row is that `c_lt_mark = 2` and `lt_marker` has exactly one 2 (at `c_diff_idx`) and the remaining entries are 0s. 

There are three ways in which the setup row is underconstrained:
1. `c_lt_mark` is not constrained to be 2 
2. Even if `c_lt_mark = 2`, `sum_i lt_marker[i]` is not constrained to be 2 (which would ensure that `lt_marker` has exactly one 2 and that the remaining entries are 0s).
3. Even if the above two constraints hold, `b[i] = N[i]` is only constrained for `i > c_diff_idx` while it should be constrained for all `i`.

See the failing tests added in the first commit of this PR for a proof of concept.

Closes INT-3599

### DIFF
--- a/extensions/algebra/circuit/src/modular_chip/is_eq.rs
+++ b/extensions/algebra/circuit/src/modular_chip/is_eq.rs
@@ -47,8 +47,11 @@ pub struct ModularIsEqualCoreCols<T, const READ_LIMBS: usize> {
     // Define c_diff_idx analogously. Then let b_lt_diff = N[b_diff_idx] - b[b_diff_idx] and
     // c_lt_diff = N[c_diff_idx] - c[c_diff_idx], where both must be in [0, 2^LIMB_BITS).
     //
-    // Additionally, set lt_marker[b_diff_idx] = 1, lt_marker[c_diff_idx] = c_lt_mark, and 0 everywhere
-    // else. If b_diff_idx == c_diff_idx then c_lt_mark = 1, else c_lt_mark = 2.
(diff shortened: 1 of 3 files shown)


jonathanpwang (approved):
I believe these constraints are necessary, but it was not easy to understand what is going on (what exactly are `c_lt_mark` and `lt_marker_sum` supposed to be?) Some additional explanations (can be inline docs) would be helpful

Avaneesh-axiom:
I rebased onto `specs/algebra` and added some additional comments.

See also the proof of soundness [here](https://github.com/axiom-crypto/openvm-v1-internal-audit/blob/1bb489a15d6616bb2e006c2dce7df29d0ad02dba/extension-algebra.md#31-soundness-of-is_eq-chip)

jonathanpwang (approved):
LGTM, the proof is great.

nit: it was easier to see the diff without all the comments; ideally comment additions should be done in a separate PR. I'll leave it for this one.

PR #1416, opened 2025-03-11
[fix] Native ASM compiler checks stack overflow

nyunyunyunyu:
All frame pointers are known at compile time. So we can check stack overflow at compile time.


zlangley (approved):
lgtm

PR #1417, opened 2025-03-12
docs: Add documentation for NativePoseidon2 columns

zlangley:


extensions/native/circuit/src/poseidon2/columns.rs
```diff
@@ -4,34 +4,59 @@ use openvm_poseidon2_air::Poseidon2SubCols;
 
 use crate::{poseidon2::CHUNK, utils::const_max};
 
+/// A valid trace is composed four types of contiguous blocks:
```
jonathanpwang, line 7:
```suggestion
/// A valid trace is composed of four types of contiguous blocks:
```

jonathanpwang (approved):

PR #1421, opened 2025-03-12
fix: always filter out build script target

### LINKED ISSUE #1420: Bug: build script cause `cargo openvm build` failed
```
Error: Expected 1 target, got 2: [
    Target {
        name: "openvm-test-secp256r1",
        kind: [
            "bin",
        ],
        crate_types: [
            "bin",
        ],
        required_features: [],
        src_path: "/Users/hhq/workspace/openvm-tests/src/main.rs",
        edition: E2021,
        doctest: false,
        test: true,
        doc: true,
    },
    Target {
        name: "build-script-build",
        kind: [
            "custom-build",
        ],
        crate_types: [
            "bin",
        ],
        required_features: [],
        src_path: "/Users/hhq/workspace/openvm-tests/build.rs",
        edition: E2021,
        doctest: false,
        test: false,
        doc: false,
    },
]

Location:
    /Users/hhq/.cargo/git/checkouts/openvm-e4fe658744a82733/30576cc/crates/toolchain/build/src/lib.rs:388:13
```

lightsing:
fix #1420

### DIFF
--- a/crates/toolchain/build/src/lib.rs
+++ b/crates/toolchain/build/src/lib.rs
@@ -377,6 +377,12 @@ pub fn find_unique_executable<P: AsRef<Path>, Q: AsRef<Path>>(
         .targets
         .into_iter()
         .filter(move |target| {
+            // always filter out build script target
+            if target.kind.iter().any(|k| k == "custom-build")
+                && target.name == "build-script-build"
+            {
+                return false;
+            }
             if let Some(target_filter) = target_filter {
                 return target.kind.iter().any(|k| k == &target_filter.kind)
                     && target.name == target_filter.name;


crates/toolchain/build/src/lib.rs
```diff
@@ -377,6 +377,12 @@ pub fn find_unique_executable<P: AsRef<Path>, Q: AsRef<Path>>(
         .targets
         .into_iter()
         .filter(move |target| {
+            // always filter out build script target
+            if target.kind.iter().any(|k| k == "custom-build")
+                && target.name == "build-script-build"
```
jonathanpwang, line 382: isn't this custom to the name of your specific crate/package?

jonathanpwang:
I think the fix is good if you remove the filter on `target.name` (which is only for your specific crate)

crates/toolchain/build/src/lib.rs
lightsing: I don't think so, all build script should have this name.

jonathanpwang: ok I'm convinced: https://github.com/rust-lang/cargo/blob/611b7c4ee40b63d0818040f7290a8e14424a9519/src/cargo/core/compiler/layout.rs#L70

jonathanpwang (approved):
LGTM! Good find

PR #1423, opened 2025-03-12
[fix] more checks in `verify_segments`

Golovanov399:
Wip

### DIFF
--- a/crates/vm/src/arch/vm.rs
+++ b/crates/vm/src/arch/vm.rs
@@ -7,6 +7,7 @@ use openvm_stark_backend::{
     keygen::types::{MultiStarkProvingKey, MultiStarkVerifyingKey},
     p3_commit::PolynomialSpace,
     p3_field::PrimeField32,
+    p3_util::log2_strict_usize,
     proof::Proof,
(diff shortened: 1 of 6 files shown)


crates/vm/src/arch/vm.rs
```diff
@@ -801,8 +804,18 @@ where
...
+        Com<SC>: Into<[F; 8]>,
+    {
+        let app_program_commit = proof.commitments.main_trace[PROGRAM_CACHED_TRACE_INDEX].clone();
+        let app_program_commit = app_program_commit.into();
         let init_memory_commit = {
-            let pvs: &MemoryMerklePvs<F, CHUNK> = proofs[0]
+            let pvs: &MemoryMerklePvs<F, CHUNK> = proof
                 .per_air
                 .iter()
                 .find(|air| air.air_id == MERKLE_AIR_ID)
```
jonathanpwang, line 821: unwrap won't work for `verify_single`

jonathanpwang:
I have moved the relevant logic into https://github.com/openvm-org/openvm/pull/1438

PR #1413, opened 2025-03-11
feat: thinner pairing circuit

jonathanpwang:
### Minor changes to benchmarking unrelated to main PR

Changes the SDK default `app_log_blowup = 1, leaf_log_blowup = 1`.
For benchmarks, changes default to jemalloc with additional jemalloc config settings in environment variables.

# Pairing circuit with less columns

This PR is about reducing the number of chips the Pairing VM Extension uses because we have found that the width of the chips makes the leaf recursion circuit's size increase not worth the speedup from the additional chip.

Before this PR, here's the main trace widths of the chips using mod-builder:
```
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=ModularAddSub, main_width=132
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=ModularMulDiv, main_width=194
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=ModularAddSub, main_width=132
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=ModularMulDiv, main_width=194
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=Fp2AddSub, main_width=261
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=Fp2MulDiv, main_width=385
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=MillerDoubleStep, main_width=1785
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=MillerDoubleAndAddStep, main_width=3063
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=EvaluateLine, main_width=701
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=Mul013By013, main_width=1471
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=MulBy01234, main_width=4529
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=Fp12Mul, main_width=4978
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=EcAddNe, main_width=511
INFO     ┝━ ｉ [info]: FieldExpressionCoreChip: opcode=EcDouble, main_width=447
```

[...]

PR #1425, opened 2025-03-12
chore: change benchmarks to jemalloc

yi-sun:


jonathanpwang:
not sure why github isn't picking it up, but this was already done in https://github.com/openvm-org/openvm/pull/1413

PR #1426, opened 2025-03-12
docs: Algebra VM Extension Specification

Avaneesh-axiom:
Added comments.

I didn't add any README documentation since the circuit is easy enough to understand (it uses Mod Builder) and the moduli macros and complex macros already have READMEs.

Note: this PR doesn't have all the comments that I added. This is because I rebased [another PR](https://github.com/openvm-org/openvm/pull/1409) onto this branch and that PR got merged in. So now the diff of this PR doesn't include those comments.


jonathanpwang (approved):
LGTM after adding proof of soundness for `iseq`

PR #1427, opened 2025-03-12
[chore] Remove unused write_a_x_is_first column in FriReducedOpeningChip

nyunyunyunyu:
`write_a_x_is_first` is not used meaningfully because now we assert `write_a` boolean in Workload rows instead.


nyunyunyunyu:
duplicated #1428

PR #1436, opened 2025-03-13
fix: prevent starting with is_exhausted in NativePoseidon

zlangley:
Could have also been fixed more simply with `builder.assert_zero(is_exhausted[0])`, but this feels wasteful then we are just adding an all zero column and a bunch of unnecessary constraints.

### DIFF
--- a/extensions/native/circuit/src/poseidon2/air.rs
+++ b/extensions/native/circuit/src/poseidon2/air.rs
@@ -201,14 +201,15 @@ impl<AB: InteractionBuilder, const SBOX_REGISTERS: usize> Air<AB>
 
         // handle exhausted cells on next row
 
-        for i in 0..CHUNK {
+        // Can skip i = 0 since first cell is never exhausted.
+        for i in 1..CHUNK {
             builder
                 .when(inside_row - end_inside_row)
-                .when(next.is_exhausted[i])
+                .when(next.is_exhausted[i - 1])
                 .assert_eq(next_left_input[i], left_output[i]);
             builder
                 .when(end.clone())
-                .when(next.is_exhausted[i])
+                .when(next.is_exhausted[i - 1])
                 .assert_zero(next_left_input[i]);
         }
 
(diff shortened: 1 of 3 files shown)


jonathanpwang (commented):
What prevents `cells` from being empty at the ISA level?

zlangley:
> What prevents `cells` from being empty at the ISA level?

I think it's part of the definition of a row hash. I.e., we are supposed to apply the Poseidon2 hash function on a sequence of field elements.

But the previous implementation allows the prover to compute something else: after computing the hash, the prover can effectively keep applying the Poseidon2 permutation to the state after all the row values have been absorbed.

jonathanpwang:
Can you check that the case `is_first_in_row = 0 && is_exhausted = 0` doesn't happen?

zlangley:
> Can you check that the case `is_first_in_row = 0 && is_exhausted = 0` doesn't happen?

This case is allowed (it is called "normal" in the AIR); it's for absorbing a non-initial element into the sponge. The possible states for a cell are:
- `is_first_in_row = 1 && is_exhausted = 0`: This is the initial cell of some matrix row we are absorbing.
- `is_first_in_row = 0 && is_exhausted = 0`: This is a non-initial cell of some matrix row we are absorbing.
- `is_first_in_row = 0 && is_exhausted = 1`: We drained the row so this cell should be the same as the corresponding cell in the previous output state.

Maybe you meant the case `is_first_in_row = 1 && is_exhausted = 1`, which is an invalid state because a new matrix must have at least one element to absorb (i.e., it cannot be empty). This is prohibited by the following existing constraints:
```
            builder.when(inside_row).assert_bool(cell.is_first_in_row);
            builder.assert_bool(is_exhausted.clone());
            builder
                .when(inside_row)
                .assert_bool(cell.is_first_in_row + is_exhausted.clone());
```

jonathanpwang (approved):

PR #1438, opened 2025-03-13
fix: complete `verify_segments` functionality

nyunyunyunyu:
**Finding Links:**
- https://cantina.xyz/code/c486d600-bed0-4fc6-aed1-de759fd29fa2/findings/39
- https://cantina.xyz/code/c486d600-bed0-4fc6-aed1-de759fd29fa2/findings/56

### Description of Fix

- Move the logic to compute exe commit into `openvm_circuit` crate.
- Finish `verify_segments` function but where exe commitment and final memory root are returned.
- Add `UserPublicValuesProof::verify` function for verifying user public values with respect to the final memory root. Remove the left/right indicators from the proof because they **must** be determined from the address space + memory dimensions
- Add `MemoryDimensions` to `AppVerifyingKey` for the above reason
- The combination of `verify_segments` and user public values verify is equivalent in functionality to the full recursive aggregation tree down to the root proof verification, which is now what the SDK `verify_app_proof` does.
- To get some stuff to work with generics, `verify_segments` and `verify_single` are now standalone functions not part of `VirtualMachine`.

Closes INT-3286


crates/vm/src/system/memory/tree/public_values.rs
```diff
@@ -56,14 +69,70 @@ impl<const CHUNK: usize, F: PrimeField32> UserPublicValuesProof<CHUNK, F> {
...
+        // Verify user public values Merkle proof:
+        // 0. Get correct indices for Merkle proof based on memory dimensions
+        // 1. Verify user public values commitment with respect to the final memory root.
+        // 2. Compare user public values commitment with Merkle root of user public values.
+        let pv_commit = self.public_values_commit;
+        // 0.
+        let pv_as = PUBLIC_VALUES_ADDRESS_SPACE_OFFSET + memory_dimensions.as_offset;
+        let pv_start_idx = memory_dimensions.label_to_index((pv_as, 0));
+        let pvs = &self.public_values;
+        if pvs.len() % CHUNK != 0 || !(pvs.len() / CHUNK).is_power_of_two() {
```
Golovanov399, line 88: Do we never call `verify` when there are zero public values?

crates/vm/src/system/memory/tree/public_values.rs
```diff
@@ -56,14 +69,70 @@ impl<const CHUNK: usize, F: PrimeField32> UserPublicValuesProof<CHUNK, F> {
...
+            curr_root = if idx_prefix & (1 << i) != 0 {
+                hasher.compress(sibling_hash, &curr_root)
+            } else {
+                hasher.compress(&curr_root, sibling_hash)
+            }
+        }
+        if curr_root != final_memory_root {
+            return Err(UserPublicValuesProofError::FinalMemoryRootMismatch);
+        }
+        // 2. Compute merkle root of public values
```
Golovanov399, line 112: Is this `hasher.merkle_root(pvs)`?

crates/vm/src/arch/vm.rs
```diff
@@ -561,115 +577,204 @@ where
...
-                            actual: pvs.exit_code.as_canonical_u32(),
+        let mut program_air_present = false;
+        let mut connector_air_present = false;
+        let mut merkle_air_present = false;
+
+        // Check public values.
+        for air_proof_data in proof.per_air.iter() {
+            let pvs = &air_proof_data.public_values;
+            let air_vk = &vk.per_air[air_proof_data.air_id];
+            if air_proof_data.air_id == PROGRAM_AIR_ID {
```
Golovanov399, line 684: Should we, in general, also check here that all `air_id`s are different in each segment? That the sets of `air_id`s are same for every segment? That the _sequences_ of `air_id`s are same for every segment?

Golovanov399 (approved):
Seems good

I also lack intuition of when to use `vm_poseidon2_hasher` and when to accept `hasher` as the argument, but probably not important so far

Golovanov399:
nit: the description says "only final memory root is returned", but we actually return the committed exe as well?

crates/vm/src/system/memory/tree/public_values.rs
jonathanpwang: The docs say must be `8 * pow_of_two`. @nyunyunyunyu can num public values be 0?

crates/vm/src/system/memory/tree/public_values.rs
jonathanpwang: wait is that a function

crates/vm/src/arch/vm.rs
jonathanpwang: they will not be the same because of optional airs

crates/vm/src/system/memory/tree/public_values.rs
jonathanpwang: confirmed `num_public_values` must be nonzero.

docs/specs/ISA.md
```diff
@@ -34,7 +34,7 @@ OpenVM depends on the following parameters, some of which are fixed and some of
 | `as_offset`         | The index of the first writable address space.                     | Fixed to 1.                                                           |
 | `as_height`         | The base 2 log of the number of writable address spaces supported. | Configurable, must satisfy `as_height <= F::bits() - 2`               |
 | `pointer_max_bits`  | The maximum number of bits in a pointer.                           | Configurable, must satisfy `pointer_max_bits <= F::bits() - 2`        |
-| `num_public_values` | The number of user public values.                                  | Configurable.                                                         |
+| `num_public_values` | The number of user public values.                                  | Configurable, must equal `8` times a power of two. In particular, it must be nonzero. |
```
nyunyunyunyu, line 37: This is only required in the continuation mode

nyunyunyunyu:
LGTM

PR #1445, opened 2025-03-13
fix: Don't segment if continuation is disabled

nyunyunyunyu:
If continuation is disabled, VM runtime doesn't segment. This might cause a trace cannot be proven because of its height is too large but this is out of scope of VM runtime.

### DIFF
--- a/crates/vm/src/arch/segment.rs
+++ b/crates/vm/src/arch/segment.rs
@@ -331,6 +331,9 @@ impl<F: PrimeField32, VC: VmConfig<F>> ExecutionSegment<F, VC> {
     ///
     /// Default config: switch if any runtime chip height exceeds 1<<20 - 100
     fn should_segment(&mut self) -> bool {
+        if !self.system_config().continuation_enabled {
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


crates/vm/src/arch/segment.rs
```diff
@@ -287,7 +287,7 @@ impl<F: PrimeField32, VC: VmConfig<F>> ExecutionSegment<F, VC> {
             #[cfg(feature = "bench-metrics")]
             self.update_instruction_metrics(pc, opcode, dsl_instr);
 
-            if self.should_segment() {
+            if self.system_config().continuation_enabled && self.should_segment() {
```
zlangley, line 290: Should we move this condition into `should_segment` to keep the logic simpler here?

zlangley (approved):
LGTM, one minor suggestion

crates/vm/src/arch/segment.rs
nyunyunyunyu: Done

jonathanpwang (approved):

PR #1452, opened 2025-03-14
fix: Fix out-of-bound accesses to pows_of_two

nyunyunyunyu:
In `check_trace_height_constraints`, [here](https://github.com/openvm-org/openvm/blob/18096f4194b76b743463c0fb39955f24d010e9bf/extensions/native/recursion/src/stark/mod.rs#L812) uses `air_proof.log_degree` as an index to access `pows_of_two`, which length is `MAX_TWO_ADICITY`. But before this access,
only [here](https://github.com/openvm-org/openvm/blob/18096f4194b76b743463c0fb39955f24d010e9bf/extensions/native/recursion/src/stark/mod.rs#L177) asserts that `air_proof.log_degree < MAX_TWO_ADICITY + 1`. So it's possible to access an out-of-bound address when `air_proof.log_degree == MAX_TWO_ADICITY`.

closes INT-3635


extensions/native/recursion/src/stark/mod.rs
```diff
@@ -798,8 +804,8 @@ where
             }
         };
 
-        // Will index by log_air_height. Max value is assumed to be MAX_TWO_ADICITY - 1.
-        let pows_of_two: Array<C, Var<C::N>> = builder.array(MAX_TWO_ADICITY);
+        // Will index by log_air_height. Max value is assumed to be MAX_TWO_ADICITY.
+        let pows_of_two: Array<C, Var<C::N>> = builder.array(MAX_TWO_ADICITY + 1);
```
zlangley, line 808: Given the change above, we now know that `air_proofs[i].log_degree < MAX_TWO_ADICITY - pcs.config.log_blowup + 1 <= MAX_TWO_ADICITY`. So `air_proofs[i].log_degree <= MAX_TWO_ADICITY - 1`, and I don't think this can stay as it was?

nyunyunyunyu: OK fixed

extensions/native/recursion/src/stark/mod.rs
```diff
@@ -798,7 +804,8 @@ where
             }
         };
 
-        // Will index by log_air_height. Max value is assumed to be MAX_TWO_ADICITY - 1.
+        // Will index by log_air_height. Max value is assumed to be MAX_TWO_ADICITY - 1 because
+        // `log_blowup > 1`.
```
zlangley, line 808:
```suggestion
        // `log_blowup >= 1`.
```

zlangley (approved):

PR #1455, opened 2025-03-14
fix: recursion max trace height check

nyunyunyunyu:
[Here](https://github.com/openvm-org/openvm/blob/18096f4194b76b743463c0fb39955f24d010e9bf/extensions/native/recursion/src/view.rs#L91) sets `height_max` to `C::F::ORDER_U32 / max_coefficient`. 
Because we check `height < height_max`, `C::F::ORDER_U32 / max_coefficient` is a valid value for `height`.
`C::F::ORDER_U32 / max_coefficient * max_coefficient < C::F::ORDER_U32` because `C::F::ORDER_U32` is a prime and
`max_coefficient > 1`.

closes INT-3636

### DIFF
--- a/extensions/native/recursion/src/view.rs
+++ b/extensions/native/recursion/src/view.rs
@@ -87,7 +87,7 @@ pub fn get_advice_per_air<C: Config>(
         } else {
             OptionalVar {
                 is_some: Usize::from(1),
-                // Because `C::F::ORDER_U32 / max_coefficient * max_coefficient < C::F::ORDER_U32`,
+                // Because `C::F::ORDER_U32` is prime and `max_coefficient > 1`, `floor(C::F::ORDER_U32 / max_coefficient) * max_coefficient < C::F::ORDER_U32`,
                 // `height * max_coefficient` cannot overflow `C::F`.
                 value: builder.constant(C::N::from_canonical_u32(
                     C::F::ORDER_U32 / max_coefficient + 1,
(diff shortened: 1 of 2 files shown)


extensions/native/recursion/src/view.rs
```diff
@@ -87,8 +87,11 @@ pub fn get_advice_per_air<C: Config>(
         } else {
             OptionalVar {
                 is_some: Usize::from(1),
-                value: builder
-                    .constant(C::N::from_canonical_u32(C::F::ORDER_U32 / max_coefficient)),
+                // Because `C::F::ORDER_U32 / max_coefficient * max_coefficient < C::F::ORDER_U32`,
```
jonathanpwang, line 90: This is not true if `max_coefficient` is 1?

jonathanpwang, line 90:
```suggestion
                // Because `C::F::ORDER_U32` is prime and `max_coefficient > 1`, `floor(C::F::ORDER_U32 / max_coefficient) * max_coefficient < C::F::ORDER_U32`,
```

jonathanpwang (approved):
Confirmed that we want `height <= ORDER / max_coefficient` and we do
```
            builder
                .if_eq(height_max.is_some, C::N::ONE)
                .then(|builder| {
                    assert_less_than(builder, height, height_max.value);
                });
```
so `height_max = ORDER / max_coefficient + 1` is correct

PR #1456, opened 2025-03-15
docs: ECC Extension specs

manh9203:
Fixed ECC extension specs to add `HintNonQr` and updated code comments.

Resolves INT-3639


docs/specs/RISCV.md
```diff
@@ -145,7 +145,8 @@ The elliptic curve extension supports arithmetic over short Weierstrass curves,
...
-| hint_decompress | R   | 0101011     | 001    | `idx*8+3` | Read `x: C::Fp` from `[rs1: C::COORD_SIZE]_2` and `rec_id: u8` from `[rs2]_2`. Reset the hint stream to equal the unique `y: C::Fp` such that `(x, y)` is a point on `C` and `y` has the same parity as `rec_id`, if it exists. Otherwise reset hint stream to arbitrary `C::Fp`. `rd` should be `x0`.                                                                                                                                                                                                                                                         |
+| hint_decompress | R   | 0101011     | 001    | `idx*8+3` | Read `x: C::Fp` from `[rs1: C::COORD_SIZE]_2` and `rec_id: u8` from `[rs2]_2`. If there exists an unique `y: C::Fp` such that `(x, y)` is a point on `C` and `y` has the same parity as `rec_id`, then reset the hint stream to a byte set to `1` followed by 3 zero bytes and the limbs represent `y` in little-endian order; otherwise, reset it to 4 zero bytes followed by the limbs represent `sqrt` where `sqrt * sqrt == (x^3 + ax + b) * non_qr` (`non_qr` is a quadratic nonresidue of `C::Fp`). `rd` should be `x0`. |
```
jonathanpwang, line 148:
```suggestion
| hint_decompress | R   | 0101011     | 001    | `idx*8+3` | Read `x: C::Fp` from `[rs1: C::COORD_SIZE]_2` and `rec_id: u8` from `[rs2]_2`. If there exists a unique `y: C::Fp` such that `(x, y)` is a point on `C` and `y` has the same parity as `rec_id`, then reset the hint stream to a byte set to `1` followed by 3 zero bytes and the limbs represent `y` in little-endian order; otherwise, reset it to 4 zero bytes followed by the limbs represent `sqrt` where `sqrt * sqrt == (x^3 + ax + b) * non_qr` (`non_qr` is a quadratic nonresidue of `C::Fp`). `rd` should be `x0`. |
```

extensions/ecc/circuit/src/weierstrass_extension.rs
```diff
@@ -359,6 +365,8 @@ pub(crate) mod phantom {
         }
     }
 
+    /// Find the square root of `x` modulo `modulus` with `non_qr` is a
```
jonathanpwang, line 368:
```suggestion
    /// Find the square root of `x` modulo `modulus` with `non_qr` a
```

jonathanpwang (approved):

PR #1457, opened 2025-03-15
specs: circuit architecture, memory, continuations

Golovanov399:
Write the specs for the circuit and memory


docs/specs/circuit.md
```diff
@@ -53,228 +53,219 @@ The chips that get to execute instructions are _instruction executors_. These ch
...
+}
+```
 
-The initial and final memory accesses are constrained different when the VM has continuations.
-See [Continuations](./continuations.md) for full details. In summary, because the initial and final memory states are
-committed to in a **trie**, the uniqueness of the addresses is constrained by the trie, so the arguments of the previous
-paragraph are not used.
+Here, `Reads` is some type that describes the memory reads required for the chip, and `Writes` is similar but about writes. The `VmAdapterChip` defines what the adapter needs to be able to do:
+
+```rust
```
jonathanpwang, line 80: this kind of section with more specific code-level usage should go in `docs/crate` or in a README.md next to `integration_api.rs`

The "specs" markdown should be kept at a level above the code specifics, except in cases where you really need the code for precision. 

Since the integration API is only one of several ways to achieve what is laid out in our circuit architecture, I feel that it falls into the "crate docs" category and not part of the architecture spec.

docs/specs/circuit.md
```diff
@@ -3,210 +3,67 @@
 We build our virtual machines in a STARK proving system with a multi-matrix commitment scheme and shared verifier
 randomness between AIR matrices to enable permutation arguments such as log-up.
 
-In the following, we will refer to a circuit as a collection of AIR matrices (also referred to as chips) of possibly
-different heights, which may communicate with one another over buses using a log-up permutation argument. We refer to
-messages sent to such a bus as [interactions](https://github.com/openvm-org/stark-backend/tree/main/crates/stark-backend/src/interaction).
+In the following, we will refer to a circuit as a collection of chips that communicate with one another over buses using a LogUp permutation argument. We refer to messages sent to such a bus as [interactions](https://github.com/openvm-org/stark-backend/blob/main/docs/interactions.md). Every _chip_ is an entity responsible for a certain operation (or set of operations), and it has an AIR to check the correctness of its execution.
```
jonathanpwang, line 6: technically, in the backend we used "interaction" to refer to the "for every row, send ..." constraint/rule. But we do also colloquially use interaction to kind of refer to the message. I couldn't find a good way to really distinguish this, and here when you say refer to mesage**s** -> interaction, it is technically still correct in a sense. so no change required

jonathanpwang (approved):
LGTM

PR #1463, opened 2025-03-17
chore: allow configurability of `manifest_dir` so external crates can reuse `openvm-benchmarks`

HrikB:


benchmarks/src/utils.rs
```diff
@@ -171,9 +171,14 @@ impl BenchmarkCli {
 }
 
 fn get_programs_dir() -> PathBuf {
-    let mut dir = PathBuf::from(env!("CARGO_MANIFEST_DIR")).to_path_buf();
-    dir.push("programs");
-    dir
+    match option_env!("BENCH_DIR") {
```
jonathanpwang, line 174: can you make it an arg in `BenchArgs` instead of env var? prefer not to have env var floating around

HrikB: Do you mean in the `BenchmarkCli` struct?

benchmarks/benches/fibonacci_execute.rs
```diff
@@ -10,7 +10,7 @@ use openvm_stark_sdk::p3_baby_bear::BabyBear;
 use openvm_transpiler::{transpiler::Transpiler, FromElf};
 
 fn benchmark_function(c: &mut Criterion) {
-    let elf = build_bench_program("fibonacci", "release").unwrap();
+    let elf = build_bench(get_programs_dir().join("fibonacci"), "release").unwrap();
```
jonathanpwang, line 13: nit: you can revert this

benchmarks/benches/regex_execute.rs
```diff
@@ -11,7 +11,7 @@ use openvm_stark_sdk::p3_baby_bear::BabyBear;
 use openvm_transpiler::{transpiler::Transpiler, FromElf};
 
 fn benchmark_function(c: &mut Criterion) {
-    let elf = build_bench_program("regex", "release").unwrap();
+    let elf = build_bench(get_programs_dir().join("regex"), "release").unwrap();
```
jonathanpwang, line 14: nit: you can revert this

benchmarks/benches/fibonacci_execute.rs
HrikB: Actually I don't think these can be reverted. There is no `build_bench_program` anymore.

PR #1465, opened 2025-03-17
perf: Reduce betas

nyunyunyunyu:


nyunyunyunyu:
No much difference. Close it for now: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13906659135

PR #1468, opened 2025-03-17
fix: Validate length of proof.commit_phase_commits

nyunyunyunyu:
The length of `proof.commit_phase_commits` is not validated and [here](https://github.com/openvm-org/openvm/blob/18096f4194b76b743463c0fb39955f24d010e9bf/extensions/native/recursion/src/fri/two_adic_pcs.rs#L125) we directly use it as the maximum log trace height.

close INT-3646

### DIFF
--- a/extensions/native/recursion/src/fri/two_adic_pcs.rs
+++ b/extensions/native/recursion/src/fri/two_adic_pcs.rs
@@ -25,7 +25,7 @@ pub const MAX_TWO_ADICITY: usize = 27;
 /// 1. FieldMerkleTreeMMCS sorts traces by height in descending order when committing data.
 /// 2. **Required** that `C::F` has two-adicity <= [MAX_TWO_ADICITY]. In particular this implies that all LDE matrices have
 ///    `log2(lde_height) <= MAX_TWO_ADICITY`.
-/// 3. **Required** that all trace heights are less than `2^log_max_height`
+/// 3. **Required** that the maximum trace height is `2^log_max_height - 1`.
 ///
 /// Reference:
 /// <https://github.com/Plonky3/Plonky3/blob/27b3127dab047e07145c38143379edec2960b3e1/merkle-tree/src/merkle_tree.rs#L53>
--- a/extensions/native/recursion/src/fri/two_adic_pcs.rs
+++ b/extensions/native/recursion/src/fri/two_adic_pcs.rs
@@ -124,10 +124,10 @@ pub fn verify_two_adic_pcs<C: Config>(
 
     challenger.check_witness(builder, config.proof_of_work_bits, proof.pow_witness);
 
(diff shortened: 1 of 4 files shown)


extensions/native/recursion/src/fri/two_adic_pcs.rs
```diff
@@ -25,6 +25,7 @@ pub const MAX_TWO_ADICITY: usize = 27;
 /// 1. FieldMerkleTreeMMCS sorts traces by height in descending order when committing data.
 /// 2. **Required** that `C::F` has two-adicity <= [MAX_TWO_ADICITY]. In particular this implies that all LDE matrices have
 ///    `log2(lde_height) <= MAX_TWO_ADICITY`.
+/// 3. **Required** that all trace heights are less than `2^log_max_height`
```
jonathanpwang, line 28: since you require `commit_phase_commits.len()` to equal `log_max_height`, you are saying `log_max_height` has to be strictly equal to the max of the trace heights?

nyunyunyunyu: Yes, it comes from `log_degree`

extensions/native/recursion/src/fri/two_adic_pcs.rs
```diff
@@ -117,16 +122,12 @@ pub fn verify_two_adic_pcs<C: Config>(
...
     });
 
-    // **ATTENTION**: always check shape of user inputs.
-    builder.assert_usize_eq(proof.query_proofs.len(), RVar::from(config.num_queries));
-
     challenger.check_witness(builder, config.proof_of_work_bits, proof.pow_witness);
 
-    let log_max_height =
-        builder.eval_expr(proof.commit_phase_commits.len() + RVar::from(log_blowup));
+    let log_max_height_after_blowup = builder.eval_expr(log_max_height + RVar::from(log_blowup));
```
jonathanpwang, line 127: nit: perhaps shorter name is `log_max_lde_height`

extensions/native/recursion/src/stark/mod.rs
```diff
@@ -200,6 +200,12 @@ where
         };
         // (T02a): `air_perm_by_height` is a valid permutation of `0..num_airs`.
         // (T02b): For all `i`, `air_proofs[i].log_degree <= MAX_TWO_ADICITY - log_blowup`.
+        // (T02c): For all `0<=i<num_air-1`, `air_proofs[air_perm_by_height[i]].log_degree <= air_proofs[air_perm_by_height[i+1]].log_degree`.
```
jonathanpwang, line 203: I think you mean `>=`?

jonathanpwang (approved):
lgtm if CI passes:
- clarify `log_max_height` needs to be precisely equal to the max, not just upper bound
- I think an inequality in reversed in a comment
- one naming nit

PR #1470, opened 2025-03-17
chore: Add tests for some constants in the pairing extension

Avaneesh-axiom:
As part of the pairing internal security review, I found some unvalidated constants so I'm adding tests for them here.


jonathanpwang (approved):
LGTM

PR #1471, opened 2025-03-17
bug: Added test for `FROBENIUS_COEFF_FQ6_C1` and fixed incorrect constant

Avaneesh-axiom:
Added a test for the `FROBENIUS_COEFF_FQ6_C1` constant in the pairing extension. The test checks that the values are the same as those in `halo2curves_shims` I found that `FREBENIUS_COEFF_FQ6_C1[0]` did not pass the test. This value is not currently being used in the code, but for completeness, I updated it to be correct.


jonathanpwang (approved):
Good catch

PR #1473, opened 2025-03-18
fix: completeness of ModBuilder trace generation

jonathanpwang:
Previously the trace generation for ModBuilder was not calculating the number of limbs necessary in a way consistent with the constraints. This was because there were some underestimates when the prime is much smaller than $`2^\textrm{num\_limbs * limb\_bits}`$. We fix it by updating all calculations to use the theoretical bound that an input or saved var have big integer value in $`[0, 2^\text{num\_limbs * limb\_bits})`$ rather than $[0, p)$.

This causes no soundness bugs, but it does affect completeness of tracegen.

closes INT-3641

### DIFF
--- a/crates/circuits/mod-builder/src/builder.rs
+++ b/crates/circuits/mod-builder/src/builder.rs
@@ -2,7 +2,7 @@ use std::{cell::RefCell, iter, ops::Deref, rc::Rc};
 
 use itertools::{zip_eq, Itertools};
 use num_bigint::{BigInt, BigUint, Sign};
-use num_traits::Zero;
+use num_traits::{One, Zero};
 use openvm_circuit_primitives::{
     bigint::{
         check_carry_mod_to_zero::{CheckCarryModToZeroCols, CheckCarryModToZeroSubAir},
@@ -56,6 +56,7 @@ pub struct ExprBuilder {
     pub limb_bits: usize,
     /// Number of limbs in canonical representation of the bigint field element.
     pub num_limbs: usize,
+    proper_max: BigUint,
     // The max bits to range check.
     pub range_checker_bits: usize,
 
@@ -89,6 +90,7 @@ pub struct ExprBuilder {
 impl ExprBuilder {
     pub fn new(config: ExprBuilderConfig, range_checker_bits: usize) -> Self {
         let prime_bigint = BigInt::from_biguint(Sign::Plus, config.modulus.clone());
+        let proper_max = (BigUint::one() << (config.num_limbs * config.limb_bits)) - BigUint::one();
         Self {
             prime: config.modulus.clone(),
             prime_bigint,
@@ -97,6 +99,7 @@ impl ExprBuilder {
             num_flags: 0,
             limb_bits: config.limb_bits,
             num_limbs: config.num_limbs,
+            proper_max,
             range_checker_bits,
             num_variables: 0,
             constants: vec![],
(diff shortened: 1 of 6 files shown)


crates/circuits/mod-builder/src/symbolic_expr.rs
```diff
@@ -327,23 +330,30 @@ impl SymbolicExpr {
...
+    /// The cached value `proper_max` should equal `2^{limb_bits * num_limbs} - 1`.
     pub fn constraint_limbs(
         &self,
         prime: &BigUint,
         limb_bits: usize,
         num_limbs: usize,
+        proper_max: &BigUint,
     ) -> (usize, usize) {
-        let (max_pos_abs, max_neg_abs) = self.max_abs(prime);
+        println!("{}", self);
```
Avaneesh-axiom, line 342: can we delete this debugging code?

crates/circuits/mod-builder/src/symbolic_expr.rs
```diff
@@ -327,23 +330,30 @@ impl SymbolicExpr {
...
         let max_q_abs = (&max_abs + prime - BigUint::one()) / prime;
         let q_bits = max_q_abs.bits() as usize;
         let p_bits = prime.bits() as usize;
         let q_limbs = q_bits.div_ceil(limb_bits);
+        // Attention! This must match with prime_overflow in `FieldExpr::generate_subrow`
         let p_limbs = p_bits.div_ceil(limb_bits);
         let qp_limbs = q_limbs + p_limbs - 1;
 
         let expr_limbs = self.expr_limbs(num_limbs);
+        println!("expr_limbs={}", expr_limbs);
```
Avaneesh-axiom, line 354: delete debug

crates/circuits/mod-builder/src/symbolic_expr.rs
```diff
@@ -327,23 +330,30 @@ impl SymbolicExpr {
...
         let carry_limbs = max(expr_limbs, qp_limbs);
+        println!("qp_limbs={}", qp_limbs);
```
Avaneesh-axiom, line 356: delete debug

extensions/algebra/tests/programs/examples/moduli_setup.rs
```diff
@@ -39,6 +39,10 @@ pub fn main() {
 
     let mut non_reduced = Mersenne61::from_le_bytes(&[0xff; 32]);
     assert!(!non_reduced.is_reduced());
+    let reduced = &non_reduced + &Mersenne61::ZERO;
+    reduced.assert_reduced();
+
+    // assert_eq!(&non_reduced + &non_reduced, reduced.double());
```
Avaneesh-axiom, line 45: delete or uncomment

Avaneesh-axiom (approved):
👍 after nits

crates/circuits/mod-builder/src/builder.rs
```diff
@@ -448,8 +462,8 @@ impl<F: PrimeField64> TraceSubRowGenerator<F> for FieldExpr {
             .collect::<Vec<_>>();
         let zero = OverflowInt::<isize>::from_canonical_unsigned_limbs(vec![0], limb_bits);
         let mut vars_overflow = vec![zero; self.num_variables];
-        let prime_overflow =
-            OverflowInt::<isize>::from_biguint(&self.prime, self.limb_bits, Some(self.num_limbs));
+        // Note: in cases where the prime fits in less limbs than `num_limbs`, we use the smaller number of limbs.
```
luffykai, line 465: I feel like i've implicitly made the assumption that prime has `num_limbs` of limbs. I think it's true so far, but i am not really sure if anything breaks when it's not the case?

jonathanpwang: at least this one is only used in the expression with carry, and I checked the calculation matches up (now) with `SymbolicExpr::constraint_limbs`

luffykai (approved):

PR #1474, opened 2025-03-18
docs: update specs on misaligned RV32 access not allowed

jonathanpwang:
RISC-V spec says that 
>  Loads and stores whose effective address is not naturally aligned to the referenced datatype (i.e., the effective address is not divisible by the size of the access in bytes) have behavior dependent on the EEI.

Our circuits already support arbitrary misaligned accesses, so this PR just updated execution (in the LoadStore Adapter `preprocess`) so trace generation actually handles misaligned accesses properly.

Reminder on how it works: 
If you want to access `ptr_val`, then the circuit will do block access of size 4 on `[ptr_val - shift_amt..ptr_val - shift_amt + 4]` where `shift_amt` can be `0,1,2,3` for byte accesses, `0,2` for half word, and `0` only for word. The block access will always work regardless of alignment because the memory access adapter will handle misalignment behind the scenes. The `shift_amt` could just be set to 0 always, but we try to set it advantageously to keep the block accesses aligned when possible.

### DIFF
--- a/extensions/rv32im/circuit/src/load_sign_extend/core.rs
+++ b/extensions/rv32im/circuit/src/load_sign_extend/core.rs
@@ -307,7 +307,9 @@ pub(super) fn run_write_data_sign_extend<
             }
             write_data[0] = read_data[shift];
         }
-        _ => unreachable!("memory block access shift miscalculation: {opcode:?}, shift: {shift}"),
+        _ => unreachable!(
+            "unaligned memory access not supported by this execution environment: {opcode:?}, shift: {shift}"
+        ),
     };
     write_data
 }
--- a/extensions/rv32im/circuit/src/load_sign_extend/core.rs
+++ b/extensions/rv32im/circuit/src/load_sign_extend/core.rs
@@ -307,6 +307,9 @@ pub(super) fn run_write_data_sign_extend<
             }
             write_data[0] = read_data[shift];
         }
+        // Currently the adapter AIR requires `ptr_val` to be aligned to the data size in bytes.
+        // The circuit requires that `shift = ptr_val % 4` so that `ptr_val - shift` is a multiple of 4.
+        // This requirement is non-trivial to remove, because we use it to ensure that `ptr_val - shift + 4 <= 2^pointer_max_bits`.
         _ => unreachable!(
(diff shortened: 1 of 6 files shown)


jonathanpwang:
Back to draft: 
this constraint and maybe others need to be changed: https://github.com/openvm-org/openvm/blob/a26742a499e7723c49ecd47d765e83ea34d23b23/extensions/rv32im/circuit/src/adapters/loadstore.rs#L262

arayikhalatyan:
> Back to draft: this constraint and maybe others need to be changed:
> 
> https://github.com/openvm-org/openvm/blob/a26742a499e7723c49ecd47d765e83ea34d23b23/extensions/rv32im/circuit/src/adapters/loadstore.rs#L262

Pushed the necessary changes. Made ready for review again

arayikhalatyan (approved):
LGTM. @jonathanpwang since we are now supporting missaligned accesses do you think it might be better to not `shift` the pointer at all? That way we could make the chip slightly cheaper (by ~3 columns).

jonathanpwang:
> LGTM. @jonathanpwang since we are now supporting misaligned accesses now do you think it might be better to not `shift` the pointer at all? That way we could make the chip slightly cheaper (by ~3 columns).

@arayikhalatyan I pushed your changes to a separate branch https://github.com/openvm-org/openvm/tree/feat/loadstore-misalign 
While trying to update the tests, I realized that with the current implementation, if you allow the misaligned, we do not range check that `mem_ptr = 2^29 - 3` with word size 4 should be invalid. It seems too complicated a change, so I'm going to just preserve the old behavior, and this PR will just be about updating docs

extensions/rv32im/circuit/src/load_sign_extend/core.rs
```diff
@@ -307,7 +307,9 @@ pub(super) fn run_write_data_sign_extend<
             }
             write_data[0] = read_data[shift];
         }
-        _ => unreachable!(),
+        _ => unreachable!(
```
Golovanov399, line 310: Maybe makes sense to leave a comment for future us here about why exactly we don't support it

Golovanov399 (approved):

PR #1475, opened 2025-03-18
chore: Range check carries to fixed value

Avaneesh-axiom:
`save_if_overflow` previously saved the expression if its carries could be larger than `range_checker_bits`. However, this means `range_checker_bits` must be small enough to not cause overflow. Really, the condition to do a save should be if the carries could be larger than `min(range_checker_bits, max_carry_bits)` for a `max_carry_bits` value that ensures no overflow.

I added this `max_carry_bits` value to be `31 - limb_bits - 2`. This ensures that the carries are in $\left(-\frac{p}{2}, \frac{p}{2}\right)$ where $p$ is the BabyBear modulus which ensures the carry constraints do not overflow. See the code comment in `check_carry_to_zero.rs` for a justification of this bound.

Closes INT-3533

### DIFF
--- a/crates/circuits/mod-builder/src/builder.rs
+++ b/crates/circuits/mod-builder/src/builder.rs
@@ -89,15 +89,17 @@ pub struct ExprBuilder {
     needs_setup: bool,
 }
 
+// Number of bits in BabyBear modulus
+const MODULUS_BITS: usize = 31;
+
 impl ExprBuilder {
     pub fn new(config: ExprBuilderConfig, range_checker_bits: usize) -> Self {
         let prime_bigint = BigInt::from_biguint(Sign::Plus, config.modulus.clone());
         let proper_max = (BigUint::one() << (config.num_limbs * config.limb_bits)) - BigUint::one();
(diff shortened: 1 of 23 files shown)


crates/circuits/mod-builder/src/builder.rs
```diff
@@ -91,6 +93,11 @@ impl ExprBuilder {
     pub fn new(config: ExprBuilderConfig, range_checker_bits: usize) -> Self {
         let prime_bigint = BigInt::from_biguint(Sign::Plus, config.modulus.clone());
         let proper_max = (BigUint::one() << (config.num_limbs * config.limb_bits)) - BigUint::one();
+        // modulus of the proof system (BabyBear)
+        // we hardcode this now, but we can make it configurable in the future
+        let modulus_bits = 31;
```
jonathanpwang, line 98: can you make it a private constant just to clarify?

jonathanpwang (approved):
LGTM, PR description is missing a minus in $-p/2$ I think

PR #1476, opened 2025-03-18
chore: Add `#[ignore]` to pairing fallback tests

Avaneesh-axiom:
The pairing hint fallback tests take forever to run (14 mins compared to 1-3 mins for other extensions), and they are triggered in the CI workflow.

This PR adds `#[ignore]` to these tests.


jonathanpwang (commented):
Can we have it run some subset or at least one overall test for the case where fallback is triggered? I agree we don't need them all

jonathanpwang (approved):
Now pairing test is 7m

PR #1479, opened 2025-03-18
refactor: Add type `EvmProof` into openvm-sdk

nyunyunyunyu:
- Rename the existing `EvmProof` into `RawEvmProof`
- Add a more semantic type `EvmProof` into `openvm-sdk`. 
- [Breaking change] Replace all `RawEvmProof` usages in the SDK interface with `EvmProof`
- Serialize `EvmProof` in json instead of `bitcode`. All fields are represented as bytes.

closes INT-3596

### DIFF
--- a/crates/sdk/src/fs.rs
+++ b/crates/sdk/src/fs.rs
@@ -10,7 +10,7 @@ use serde::{de::DeserializeOwned, Serialize};
 
 use crate::{
     keygen::{AggProvingKey, AppProvingKey, AppVerifyingKey},
-    types::EvmOpenvmProof,
+    types::EvmProof,
     F, SC,
 };
 
@@ -62,11 +62,11 @@ pub fn write_agg_pk_to_file<P: AsRef<Path>>(agg_pk: AggProvingKey, path: P) -> R
     write_to_file_bitcode(path, agg_pk)
 }
 
-pub fn read_evm_proof_from_file<P: AsRef<Path>>(path: P) -> Result<EvmOpenvmProof> {
+pub fn read_evm_proof_from_file<P: AsRef<Path>>(path: P) -> Result<EvmProof> {
     read_from_file_bitcode(path)
 }
 
-pub fn write_evm_proof_to_file<P: AsRef<Path>>(proof: EvmOpenvmProof, path: P) -> Result<()> {
+pub fn write_evm_proof_to_file<P: AsRef<Path>>(proof: EvmProof, path: P) -> Result<()> {
     write_to_file_bitcode(path, proof)
 }
 
--- a/crates/sdk/src/fs.rs
+++ b/crates/sdk/src/fs.rs
@@ -1,5 +1,5 @@
 use std::{
-    fs::{create_dir_all, read, write},
+    fs::{create_dir_all, read, write, File},
     path::Path,
 };
 
@@ -63,11 +63,13 @@ pub fn write_agg_pk_to_file<P: AsRef<Path>>(agg_pk: AggProvingKey, path: P) -> R
 }
 
 pub fn read_evm_proof_from_file<P: AsRef<Path>>(path: P) -> Result<EvmProof> {
-    read_from_file_bitcode(path)
+    let proof: EvmProof = serde_json::from_reader(File::open(path)?)?;
+    Ok(proof)
 }
 
 pub fn write_evm_proof_to_file<P: AsRef<Path>>(proof: EvmProof, path: P) -> Result<()> {
-    write_to_file_bitcode(path, proof)
+    serde_json::to_writer(File::create(path)?, &proof)?;
+    Ok(())
 }
 
 pub fn read_evm_verifier_from_file<P: AsRef<Path>>(path: P) -> Result<EvmVerifier> {
--- a/crates/sdk/src/fs.rs
+++ b/crates/sdk/src/fs.rs
@@ -5,11 +5,12 @@ use std::{
 
 use eyre::Result;
 use openvm_circuit::arch::{instructions::exe::VmExe, ContinuationVmProof, VmConfig};
-use openvm_native_recursion::halo2::{wrapper::EvmVerifier, EvmProof};
+use openvm_native_recursion::halo2::wrapper::EvmVerifier;
 use serde::{de::DeserializeOwned, Serialize};
 
 use crate::{
     keygen::{AggProvingKey, AppProvingKey, AppVerifyingKey},
+    types::EvmOpenvmProof,
     F, SC,
 };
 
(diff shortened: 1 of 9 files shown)


crates/sdk/src/fs.rs
```diff
@@ -61,11 +62,11 @@ pub fn write_agg_pk_to_file<P: AsRef<Path>>(agg_pk: AggProvingKey, path: P) -> R
     write_to_file_bitcode(path, agg_pk)
 }
 
-pub fn read_evm_proof_from_file<P: AsRef<Path>>(path: P) -> Result<EvmProof> {
+pub fn read_evm_proof_from_file<P: AsRef<Path>>(path: P) -> Result<EvmOpenvmProof> {
```
nyunyunyunyu, line 65: @luffykai do you want this to be serialized as json with hex strings or keep it as it is?

jonathanpwang (changes requested):
Is it possible to just delete the old `EvmProof` completely, and rename the new one to `EvmProof`?

The lowest level `prove_for_evm` should return the new `EvmProof` struct, doing the necessary parsing and conversions.

The struct can have functions `instances(&self) -> Vec<Fr>` and `partial_proof(&self) -> Vec<u8>` (check with @HrikB on naming of the latter). 
Then it should have a function `.encode_calldata(&self) -> Vec<u8>` to encode the whole thing.

I believe default serialization method for `EvmProof` should ban all usage of bitcode, and serialize the fields as hex string in JSON. There should also be some other option to just serialize the whole thing as one big hex string; again @HrikB may have thoughts here on which is preferrable.

You should also provide a function to clarify how to `decode` from `Vec<u8>` to `EvmProof` with `Result` handling.

nyunyunyunyu:
> Is it possible to just delete the old `EvmProof` completely, and rename the new one to `EvmProof`?
> 
> The lowest level `prove_for_evm` should return the new `EvmProof` struct, doing the necessary parsing and conversions.
> 
> The struct can have functions `instances(&self) -> Vec<Fr>` and `partial_proof(&self) -> Vec<u8>` (check with @HrikB on naming of the latter). Then it should have a function `.encode_calldata(&self) -> Vec<u8>` to encode the whole thing.
> 
> I believe default serialization method for `EvmProof` should ban all usage of bitcode, and serialize the fields as hex string in JSON. There should also be some other option to just serialize the whole thing as one big hex string; again @HrikB may have thoughts here on which is preferrable.
> 
> You should also provide a function to clarify how to `decode` from `Vec<u8>` to `EvmProof` with `Result` handling.

Discussed offline. We will keep the old `EvmProof` but rename it to `RawEvmProof`.

crates/sdk/src/lib.rs
```diff
@@ -255,9 +259,10 @@ impl Sdk {
     pub fn verify_evm_proof(
         &self,
         evm_verifier: &EvmVerifier,
-        evm_proof: &EvmProof,
+        evm_openvm_proof: &EvmProof,
```
jonathanpwang, line 262:
```suggestion
        evm_proof: &EvmProof,
```

crates/sdk/src/lib.rs
```diff
@@ -255,9 +259,10 @@ impl Sdk {
...
     ) -> Result<u64> {
-        let gas_cost = Halo2WrapperProvingKey::evm_verify(evm_verifier, evm_proof)
+        let evm_proof: RawEvmProof = evm_openvm_proof.clone().into();
```
jonathanpwang, line 264:
```suggestion
        let evm_proof: RawEvmProof = evm_proof.clone().into();
```

crates/sdk/src/types.rs
```diff
@@ -0,0 +1,87 @@
+use itertools::Itertools;
+use openvm_native_recursion::halo2::{Fr, RawEvmProof};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+const BN254_BYTES: usize = 32;
+const NUM_BN254_ACCUMULATORS: usize = 12;
+const NUM_BN254_PROOF: usize = 43;
```
jonathanpwang, line 8: not this would change if we changed the wrapper circuit or halo2 in some way

crates/sdk/src/types.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[serde_as]
+#[derive(Clone, Deserialize, Serialize)]
+pub struct EvmProof {
+    #[serde_as(as = "serde_with::hex::Hex")]
+    pub accumulators: Vec<u8>,
```
jonathanpwang, line 14: can you document the endian-ness of different fields? I remember there was something about instances being big but proof being little

crates/sdk/src/types.rs
```diff
@@ -0,0 +1,87 @@
...
+
+impl EvmProof {
+    /// Return bytes calldata to be passed to the verifier contract.
+    pub fn verifier_calldata(&self) -> Vec<u8> {
+        let evm_proof: RawEvmProof = self.clone().into();
+        evm_proof.verifier_calldata()
+    }
+}
+
+impl From<RawEvmProof> for EvmProof {
```
jonathanpwang, line 33: should we use TryFrom since there are many length asserts?

crates/sdk/src/types.rs
```diff
@@ -0,0 +1,87 @@
...
+            accumulators,
+            exe_commit,
+            leaf_commit,
+            user_public_values,
+            proof,
+        }
+    }
+}
+
+impl From<EvmProof> for RawEvmProof {
```
jonathanpwang, line 58: it seems more ok to use `From` here, but perhaps also consider `TryFrom`

extensions/native/recursion/src/halo2/wrapper.rs
```diff
@@ -99,18 +99,18 @@ impl Halo2WrapperProvingKey {
...
         ))
     }
-    pub fn prove_for_evm(&self, params: &Halo2Params, snark_to_verify: Snark) -> EvmProof {
+    pub fn prove_for_evm(&self, params: &Halo2Params, snark_to_verify: Snark) -> RawEvmProof {
         #[cfg(feature = "bench-metrics")]
         let start = std::time::Instant::now();
         let k = self.pinning.metadata.config_params.k;
         let prover_circuit = self.generate_circuit_object_for_proving(k, snark_to_verify);
-        let pvs = prover_circuit.instances();
+        let mut pvs = prover_circuit.instances();
```
jonathanpwang, line 107: assert `pvs.len() == 1` for safety

jonathanpwang (approved):
LGTM after comments, but @luffykai should also review

crates/sdk/src/types.rs
luffykai: what does this number mean? maybe add comments to explain

luffykai (approved):

PR #1480, opened 2025-03-18
fix: Add `bls12_381` feature to clippy command in lint CI

Avaneesh-axiom:
The lint workflow currently runs clippy on guest programs used as integration tests. It special cases the ecc and pairing integration test guest programs and runs clippy with certain features turned on.

I noticed that when clippy is run on the pairing guest programs, it misses the `bls12_381` feature which, due to conditional compilation used in the guest code, does not lint some of the code.

This PR adds the `bls12_381` feature to this clippy command. It also fixes the linter errors that this change reveals.


jonathanpwang (approved):
lgtm, funnily enough there are lint errors

PR #1481, opened 2025-03-19
chore: build also output committed exe as bytes

luffykai:


crates/cli/src/commands/build.rs
```diff
@@ -127,7 +144,12 @@ pub(crate) fn build(build_args: &BuildArgs) -> Result<Option<PathBuf>> {
         let data = read(elf_path.clone())?;
         let elf = Elf::decode(&data, MEM_SIZE as u32)?;
         let exe = Sdk.transpile(elf, transpiler)?;
+        let committed_exe = commit_app_exe(app_config.app_fri_params.fri_params, exe.clone());
```
jonathanpwang, line 147: should we document somewhere that `cargo openvm build` must have consistent app_fri_params with keygen/prove now?

jonathanpwang (approved):

crates/cli/src/commands/build.rs
luffykai: all steps are reading the same openvm.toml right? so unless someone modify it after build it should be the case

jonathanpwang: yea true

PR #1483, opened 2025-03-19
fix: check the heap size on memory initialization

Golovanov399:
This resolves INT-3659.

The problem was that we could potentially try to initialize about `usize::MAX` of heap memory due to unchecked subtraction.

### DIFF
--- a/crates/toolchain/platform/src/heap/embedded.rs
+++ b/crates/toolchain/platform/src/heap/embedded.rs
@@ -22,8 +22,10 @@ pub fn init() {
         static _end: u8;
     }
     let heap_pos: usize = unsafe { (&_end) as *const u8 as usize };
-    let heap_size: usize = crate::memory::GUEST_MAX_MEM
-        .checked_sub(heap_pos)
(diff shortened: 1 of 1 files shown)


crates/toolchain/platform/src/heap/embedded.rs
```diff
@@ -22,6 +22,8 @@ pub fn init() {
         static _end: u8;
     }
     let heap_pos: usize = unsafe { (&_end) as *const u8 as usize };
-    let heap_size: usize = crate::memory::GUEST_MAX_MEM - heap_pos;
+    let heap_size: usize = crate::memory::GUEST_MAX_MEM
+        .checked_sub(heap_pos)
+        .expect("Not enough memory for heap");
```
jonathanpwang, line 27: I am ever so slightly worried about re-entrancy from use of `expect`. can you follow what `sys_alloc_aligned` does and manually call `openvm::io::println`

jonathanpwang (approved):

PR #1485, opened 2025-03-19
feat: add `reveal_bytes32` io function

jonathanpwang:
closes INT-3642

### DIFF
--- a/crates/toolchain/openvm/src/io/mod.rs
+++ b/crates/toolchain/openvm/src/io/mod.rs
@@ -81,9 +81,26 @@ pub(crate) fn read_vec_by_len(len: usize) -> Vec<u8> {
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/toolchain/openvm/src/io/mod.rs
```diff
@@ -81,9 +81,26 @@ pub(crate) fn read_vec_by_len(len: usize) -> Vec<u8> {
     }
 }
 
+/// Publish `[u8; 32]` as the first 32 bytes of the user public output.
+/// In general, it is *recommended* that you reveal a single `[u8; 32]` which is
+/// the hash digest of all logical outputs.
+///
+/// Note: this will overwrite any previous data in the first 32 bytes of the user public
```
Golovanov399, line 88: Do we maybe want any check that we only do it at most once?

Golovanov399 (approved):

PR #1486, opened 2025-03-19
fix: alignment in unsafe rust

jonathanpwang:
Rust treats `[u8; LEN]` as having alignment of `1`, so if you unsafely transmute or convert a slice to `[T; LEN]` which expects alignment of `align_of<T>()`, then this may result in misaligned load/store instructions, often during operations like memcpy. OpenVM RV32 does not support misaligned access (see https://github.com/openvm-org/openvm/pull/1474), so we make sure all our use of unsafe Rust has proper alignment.

### DIFF
--- a/extensions/algebra/moduli-macros/src/lib.rs
+++ b/extensions/algebra/moduli-macros/src/lib.rs
@@ -751,14 +751,19 @@ pub fn moduli_init(input: TokenStream) -> TokenStream {
 
         let modulus_bytes = string_to_bytes(&modulus);
         let mut limbs = modulus_bytes.len();
+        let mut block_size = 32;
 
         if limbs <= 32 {
             limbs = 32;
         } else if limbs <= 48 {
             limbs = 48;
+            block_size = 16;
         } else {
             panic!("limbs must be at most 48");
         }
+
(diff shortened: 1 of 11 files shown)


extensions/algebra/moduli-macros/src/lib.rs
```diff
@@ -751,15 +751,20 @@ pub fn moduli_init(input: TokenStream) -> TokenStream {
...
         if limbs <= 32 {
             limbs = 32;
         } else if limbs <= 48 {
             limbs = 48;
+            block_size = 16;
         } else {
             panic!("limbs must be at most 48");
         }
 
+        let block_size = proc_macro::Literal::usize_unsuffixed(block_size);
```
Golovanov399, line 765: oh wow good to know this exists

Golovanov399 (approved):
Hope it's all and hope no major performance issues are introduced

extensions/algebra/moduli-macros/src/lib.rs
jonathanpwang: wait, I copied this from code you wrote!

Golovanov399: Hmm I thought I used some hacks for this, but maybe this was the hack...

PR #1487, opened 2025-03-19
feat: Serialize EvmVerifier as json & EvmVerifier includes sol codes

nyunyunyunyu:
- [Breaking change] Change struct `EvmVerifier` to include sol codes and compiler information.
- [Breaking change] Now `EvmVerifier` is serialized as json.

closes INT-3638


extensions/native/recursion/evm_verifier.example.json
jonathanpwang, line 1: why is it a json of the Solidity code, and not just a .sol file?

extensions/native/recursion/src/halo2/wrapper.rs
```diff
@@ -1,37 +1,47 @@
...
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct EvmVerifier(pub Vec<u8>);
-
-impl From<Vec<u8>> for EvmVerifier {
-    fn from(bytes: Vec<u8>) -> Self {
-        Self(bytes)
-    }
-}
+// #[derive(Debug, Clone, Serialize, Deserialize)]
```
jonathanpwang, line 34: delete

extensions/native/recursion/evm_verifier.example.json
jonathanpwang: nvm i missed it also has bytecode. @HrikB @luffykai does this format make sense?

extensions/native/recursion/src/halo2/wrapper.rs
```diff
@@ -197,3 +206,45 @@ fn emit_wrapper_circuit_metrics(agg_circuit: &AggregationCircuit) {
     let total_cell = total_advices + total_lookups + stats.gate.total_fixed;
     metrics::gauge!("halo2_total_cells").set(total_cell as f64);
 }
+
+fn gen_evm_verifier_sol_code(
```
jonathanpwang, line 210: comment and link that this is copied from `gen_evm_verifier` in `snark-verifier` and the only change is to return the sol code instead of writing to file.

We may also consider just moving this change into `snark-verifier` itself?

extensions/native/recursion/evm_verifier.example.json
yi-sun: I think we should have a separate `.sol` file -- it's very confusing to have a `.sol` file as a string in a JSON.

extensions/native/recursion/src/halo2/wrapper.rs
jonathanpwang: on second thought, let's move it into `snark-verifier`

extensions/native/recursion/evm_verifier.example.json
nyunyunyunyu: @HrikB mentioned we need the compiler version and options so I put everything into a json. Actually serializing as json is not a part of `recursion` crate. Maybe I can write these fields into different files in CLI?

yi-sun: I think the other fields in the `.json` seem fine, but because the `.sol` is already a file it's better to have it as file.

nyunyunyunyu: Changed the CLI. Now it outputs separated files

extensions/native/recursion/src/halo2/wrapper.rs
```diff
@@ -14,30 +12,31 @@ use snark_verifier_sdk::{
...
 use crate::halo2::{
     utils::{Halo2ParamsReader, KZG_PARAMS_FOR_SVK},
-    Fr, Halo2Params, Halo2ProvingMetadata, Halo2ProvingPinning, RawEvmProof,
+    Halo2Params, Halo2ProvingMetadata, Halo2ProvingPinning, RawEvmProof,
 };
 
-// #[derive(Debug, Clone, Serialize, Deserialize)]
-// pub struct EvmVerifier(pub Vec<u8>);
-
 #[serde_as]
```
jonathanpwang, line 30:
```suggestion
```

unused now?

jonathanpwang (approved):
LGTM if you check the book to make sure anything that needs updating is updated

PR #1489, opened 2025-03-20
docs: update specs on RISC-V execution environment

jonathanpwang:
Updates specs on the RISC-V execution environment for user IO.

Update internal docs to explain soundness of hint buffer. This relies on an invariant of the memory bus which is stated in https://github.com/openvm-org/openvm/pull/1490.

Also found that our `read_vec_with_len` implementation uses a bit of unsafe Rust that does not quite match the documented safety assumptions (though in practice currently does not have any problems). This is around a hidden assumption that the allocator always allocates on a 4-byte boundary.

### DIFF
--- a/docs/specs/ISA.md
+++ b/docs/specs/ISA.md
@@ -419,7 +419,7 @@ with user input-output.
 | Name             | Operands        | Description                                                                                                                                                                       |
 | ---------------- | --------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | HINT_STOREW_RV32 | `_,b,_,1,2`     | `[r32{0}(b):4]_2 = next 4 bytes from hint stream`. Only valid if next 4 values in hint stream are bytes.                                                                          |
-| HINT_BUFFER_RV32 | `a,b,_,1,2`     | `[r32{0}(b):4 * l]_2 = next 4 * l bytes from hint stream` where `l = r32{0}(a)`. Only valid if next `4 * l` values in hint stream are bytes. Very important: `l` should not be 0. |
+| HINT_BUFFER_RV32 | `a,b,_,1,2`     | `[r32{0}(b):4 * l]_2 = next 4 * l bytes from hint stream` where `l = r32{0}(a)`. Only valid if next `4 * l` values in hint stream are bytes. Very important: `l` should not be 0. The pointer address `r32{0}(b)` does not need to be a multiple of `4`. |
 | REVEAL_RV32      | `a,b,c,1,3,_,g` | Pseudo-instruction for `STOREW_RV32 a,b,c,1,3,_,g` writing to the user IO address space `3`. Only valid when continuations are enabled.                                           |
 
 #### Phantom Sub-Instructions
--- a/crates/toolchain/openvm/src/io/mod.rs
+++ b/crates/toolchain/openvm/src/io/mod.rs
@@ -72,7 +72,7 @@ pub(crate) fn read_vec_by_len(len: usize) -> Vec<u8> {
         // The heap-embedded-alloc uses linked list allocator, which has a minimum alignment of
         // `sizeof(usize) * 2 = 8` on 32-bit architectures: https://github.com/rust-osdev/linked-list-allocator/blob/b5caf3271259ddda60927752fa26527e0ccd2d56/src/hole.rs#L429
         let mut bytes = Vec::with_capacity(capacity);
-        hint_buffer_u32!(ptr_start, num_words);
+        hint_buffer_u32!(bytes.as_mut_ptr(), num_words);
         // SAFETY: We populate a `Vec<u8>` by hintstore-ing `num_words` 4 byte words. We set the length to `len` and don't care about the extra `capacity - len` bytes stored.
         unsafe {
             bytes.set_len(len);
--- a/crates/toolchain/openvm/src/io/mod.rs
+++ b/crates/toolchain/openvm/src/io/mod.rs
@@ -63,15 +63,21 @@ pub(crate) fn read_vec_by_len(len: usize) -> Vec<u8> {
 
     #[cfg(target_os = "zkvm")]
     {
-        // Allocate a buffer of the required length that is 4 byte aligned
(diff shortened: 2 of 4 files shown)


extensions/rv32im/circuit/src/hintstore/mod.rs
```diff
@@ -348,6 +365,10 @@ impl<F: PrimeField32> InstructionExecutor<F> for Rv32HintStoreChip<F> {
         let mem_ptr = compose(mem_ptr_limbs);
 
         debug_assert!(mem_ptr <= (1 << self.air.pointer_max_bits));
+        assert!(
+            mem_ptr % 4 == 0,
```
Golovanov399, line 369:
```suggestion
            mem_ptr % RV32_REGISTER_NUM_LIMBS == 0,
```

extensions/rv32im/circuit/src/hintstore/mod.rs
```diff
@@ -219,6 +222,20 @@ impl<AB: InteractionBuilder> Air<AB> for Rv32HintStoreAir {
...
+        // This implies that `mem_ptr + 4 <= 2^pointer_max_bits`
+        //
+        // Note[jpw]: we could have used BitwiseOperationLookupBus to range check
+        // `mem_ptr_limbs[0] / 4 < 2^8` instead,
+        // but there are an odd number of range checks so we just use VariableRangeCheckerBus
+        self.memory_bridge
+            .range_bus()
+            .range_check::<AB::Expr>(
+                local_cols.mem_ptr_limbs[0] * AB::F::from_canonical_u32(4).inverse(),
+                RV32_CELL_BITS - 2,
```
Golovanov399, line 236:
```suggestion
                local_cols.mem_ptr_limbs[0] * AB::F::from_canonical_usize(RV32_REGISTER_NUM_LIMBS).inverse(),
                RV32_CELL_BITS - log2_strict_usize(RV32_REGISTER_NUM_LIMBS),
```

extensions/rv32im/circuit/src/hintstore/mod.rs
```diff
@@ -457,6 +478,9 @@ impl<F: PrimeField32> Rv32HintStoreChip<F> {
             mem_ptr_msl << (RV32_REGISTER_NUM_LIMBS * RV32_CELL_BITS - pointer_max_bits),
             rem_words_msl << (RV32_REGISTER_NUM_LIMBS * RV32_CELL_BITS - pointer_max_bits),
         );
+        memory
+            .range_checker()
+            .add_count((mem_ptr & 0xff) >> 2, RV32_CELL_BITS - 2);
```
Golovanov399, line 483: also 2 -> something of `RV32_REGISTER_NUM_LIMBS`

extensions/rv32im/circuit/src/hintstore/mod.rs
```diff
@@ -219,6 +222,20 @@ impl<AB: InteractionBuilder> Air<AB> for Rv32HintStoreAir {
...
+        // Note[jpw]: we could have used BitwiseOperationLookupBus to range check
+        // `mem_ptr_limbs[0] / 4 < 2^8` instead,
+        // but there are an odd number of range checks so we just use VariableRangeCheckerBus
+        self.memory_bridge
+            .range_bus()
+            .range_check::<AB::Expr>(
+                local_cols.mem_ptr_limbs[0] * AB::F::from_canonical_u32(4).inverse(),
+                RV32_CELL_BITS - 2,
+            )
+            .eval(builder, is_start.clone());
```
Golovanov399, line 238: Why only `is_start`?

docs/specs/ISA.md
```diff
@@ -419,7 +419,7 @@ with user input-output.
 | Name             | Operands        | Description                                                                                                                                                                       |
 | ---------------- | --------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | HINT_STOREW_RV32 | `_,b,_,1,2`     | `[r32{0}(b):4]_2 = next 4 bytes from hint stream`. Only valid if next 4 values in hint stream are bytes.                                                                          |
-| HINT_BUFFER_RV32 | `a,b,_,1,2`     | `[r32{0}(b):4 * l]_2 = next 4 * l bytes from hint stream` where `l = r32{0}(a)`. Only valid if next `4 * l` values in hint stream are bytes. Very important: `l` should not be 0. |
+| HINT_BUFFER_RV32 | `a,b,_,1,2`     | `[r32{0}(b):4 * l]_2 = next 4 * l bytes from hint stream` where `l = r32{0}(a)`. Only valid if next `4 * l` values in hint stream are bytes. Very important: `l` should not be 0. The pointer address `r32{0}(b)` must be divisible by 4. |
```
Golovanov399, line 422: I don't konw if we want this to also hold for, e.g., `HINT_STOREW_RV32` (which I would assume from the PR name)

Golovanov399 (approved):
Seems good modulo nit comments

docs/specs/ISA.md
jonathanpwang: indeed

extensions/rv32im/circuit/src/hintstore/mod.rs
jonathanpwang: I guess it should be true always

jonathanpwang: actually it is auto implied for the other rows if you check it on start, and it prevents unnecessary calls to range checker chip to update counters so i'll keep it

extensions/rv32im/circuit/src/hintstore/mod.rs
jonathanpwang: I'm going to keep this one because I don't want a `log2_strict_usize` computation during tracegen

extensions/rv32im/circuit/src/hintstore/mod.rs
jonathanpwang: sorry I'm vastly mistaken, the `mem_ptr` increases, so we must keep checking it

PR #1490, opened 2025-03-20
fix: `VolatileBoundaryAir` range checks addresses

jonathanpwang:
VolatileBoundaryAir was using IsLessThanArray, which compares `x < y` **assuming** that `x, y` are both in a range `[0, max_bits)`. But for the volatile boundary case, we **must** range check this assumption, where `x = (addr_space, pointer)`.

We fix this by decomposing `addr_space, pointer` both into limbs and range checking.
I slightly optimized to only support `addr_space in [0, 2^range_max_bits)` and `pointer in [0, 2^{2*range_max_bits})` but did it with configurable constants so this could be changed to support a greater max bits in address space in the future.

Comparison Benchmark: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/13963291660 (no significant perf diff)

Closes INT-3664

### DIFF
--- a/docs/specs/memory.md
+++ b/docs/specs/memory.md
@@ -183,7 +183,7 @@ Assume that the MEMORY_BUS interactions and the constraints mentioned above are 
 
 ### Time goes forward
 
-In the connector chip, we constrain that the final timestamp is less than $`2^\text{timestamp\_max\_bits}`$. It is [guaranteed](https://github.com/openvm-org/stark-backend/blob/main/docs/interactions.md) that the total number of interaction messages is less than $p$. In our current circuit set, all chips increase timestamp [less than they do interactions](./circuit.md#inspection-of-vm-chip-timestamp-increments), which guarantees that the final timestamp cannot overflow: its actual (not mod $p$) value is less than $`2^\text{timestamp\_max\_bits}`$. Given that, our check that `timestamp - prev_timestamp - 1 < 2^timestamp\_max\_bits` guarantees that `prev_timestamp < timestamp` everywhere we check it.
+In the connector chip, we constrain that the final timestamp is less than $`2^\text{timestamp\_max\_bits}`$. It is [guaranteed](https://github.com/openvm-org/stark-backend/blob/main/docs/interactions.md) that the total number of interaction messages is less than $p$. In our current circuit set, all chips increase timestamp [less than they do interactions](./circuit.md#inspection-of-vm-chip-timestamp-increments), which guarantees that the final timestamp cannot overflow: its actual (not mod $p$) value is less than $`2^\text{timestamp\_max\_bits}`$. Given that, our check that `timestamp - prev_timestamp - 1 < 2^timestamp_max_bits` guarantees that `prev_timestamp < timestamp` everywhere we check it.
 
 ### Memory consistency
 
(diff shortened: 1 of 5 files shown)


nyunyunyunyu (approved):

Golovanov399 (approved):

docs/specs/memory.md
```diff
@@ -158,13 +159,31 @@ In **persistent** memory, we use `PersistentBoundaryChip` to handle the final me
...
 Both boundary chips perform, for every subsegment ever existed in our nice set, a receive interaction on birth and a send interaction on death.
 
+## Invariants
+
+The following invariants **must** be maintained by the memory architecture:
+1. In the MEMORY_BUS, the `timestamp` is always in range `[0, 2^timestamp_max_bits)` where `timestamp_max_bits <= F::bits() - 2` is a configuration constant.
+2. In the MEMORY_BUS, the `address_space` is always in range `[0, 2^as_height)` where `as_height <= F::bits() - 2` is a configuration constant.
+3. In the MEMORY_BUS, the `pointer` is always in range `[0, 2^pointer_max_bits)` where `pointer_max_bits <= F::bits() - 2` is a configuration constant.
+
+Invariant 1 is guaranteed by [time goes forward](#time-goes-forward) under the [assumption](./circuit.md#instruction-executors) that the timestamp increase during instruction execution is bounded by the number of AIR interactions.
```
zlangley, line 169: This also needs that timestamp starts at 0. We are actually currently missing a constraint that execution must start at timestamp zero (we only have constraints that memory is initialized at timestamp 0). Fix here: https://github.com/openvm-org/openvm/pull/1495

Golovanov399: Should we _not_ do this in persistent boundary then?

zlangley: In PersistentBoundaryAir we just directly send timestamp `0` in the memory bus so it's easier there to constrain it this way than allow an arbitrary initial timestamp. In Volatile we did add an initial timestamp constraint that I think would be implied by the ConnectorChip initial timestamp, but it seems easier to reason about if we keep the constraint.

Golovanov399: No I meant "what is the point of https://github.com/openvm-org/openvm/pull/1467 after 1495"

zlangley: Sorry, I think I confused "Volatile" and "Persistent" in my response. But that is what I thought you said. My answer is: I don't think the constraint is necessary anymore, but we should probably keep it for clarity's sake.

docs/specs/memory.md
```diff
@@ -158,13 +159,31 @@ In **persistent** memory, we use `PersistentBoundaryChip` to handle the final me
...
+
+Invariant 2 and 3 are guaranteed at timestamp `0` in the MEMORY_BUS by the boundary chips:
+- VolatileBoundaryChip constrains the range checks outright.
+- PersistentBoundaryChip populates the MEMORY_BUS at timestamp `0` from the initial memory state, which is committed to in the public value `initial_memory_root`. PersistentBoundaryChip upholds Invariants 2 and 3 **assuming** that the initial memory state only contains addresses in the required range. This assumption needs to be checked outside the scope of the circuit.
+
+> [!IMPORTANT]
+> At all later timestamps, it is the responsibility of each chip to ensure their memory accesses maintain Invariants 2 and 3.
```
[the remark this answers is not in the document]
jonathanpwang: updated explanation to clarify

PR #1491, opened 2025-03-20
docs: rust `std` vs `no-std`

jonathanpwang:
closes INT-2956


benchmarks/programs/revm_transfer/Cargo.toml
```diff
@@ -15,7 +15,6 @@ alloy-primitives = { version = "0.8.10", default-features = false, features = [
 
 [features]
 default = []
-std = ["openvm/std", "revm/std"]
```
Golovanov399, line 18: Is it intentional that we don't use `revm/std` by default now?

Golovanov399 (approved):

benchmarks/programs/revm_transfer/Cargo.toml
jonathanpwang: it still worked, so I didn't bother

PR #1494, opened 2025-03-20
fix: turn continuations off for native config

zlangley:


extensions/native/circuit/src/extension.rs
```diff
@@ -35,29 +35,15 @@ use crate::{
...
-    fn default() -> Self {
-        Self {
-            system: SystemConfig::default().with_continuations(),
-            native: Default::default(),
-        }
-    }
-}
-
 impl NativeConfig {
-    pub fn with_continuations(mut self) -> Self {
```
jonathanpwang, line 56: keep it for tests, but remove default

jonathanpwang (approved):

PR #1495, opened 2025-03-20
fix: Constrain initial timestamp to 0 in connector chip

zlangley:


crates/vm/src/system/connector/mod.rs
```diff
@@ -145,6 +145,8 @@ impl<AB: InteractionBuilder + PairBuilder + AirBuilderWithPublicValues> Air<AB>
             .when_transition()
             .assert_eq(end.is_terminate, is_terminate);
 
+        builder.assert_zero(begin.timestamp);
```
zlangley, line 148: We could of course instead just send `0` in the interaction below, but I think this is clearer. 
Otherwise `begin.timestamp` would be unconstrained and potentially confusing.

crates/vm/src/system/connector/mod.rs
```diff
@@ -145,6 +145,8 @@ impl<AB: InteractionBuilder + PairBuilder + AirBuilderWithPublicValues> Air<AB>
             .when_transition()
             .assert_eq(end.is_terminate, is_terminate);
 
+        builder.when_transition().assert_one(begin.timestamp);
```
zlangley, line 148: We could of course instead just send `1` in the interaction below, but I think this is clearer.
(Otherwise `timestamp` on first row would be unconstrained and potentially confusing.)

jonathanpwang (approved):

PR #1496, opened 2025-03-20
fix: `addr_space_max_bits` in `VolatileBoundaryChip` constructor

jonathanpwang:
We were using `as_height` but it is actually `log2(1 + 2^as_height)`.

### DIFF
--- a/crates/vm/src/system/memory/controller/mod.rs
+++ b/crates/vm/src/system/memory/controller/mod.rs
@@ -232,13 +232,19 @@ impl<F: PrimeField32> MemoryController<F> {
     ) -> Self {
         let range_checker_bus = range_checker.bus();
         let initial_memory = AddressMap::from_mem_config(&mem_config);
+        assert!(mem_config.pointer_max_bits <= F::bits() - 2);
+        assert!(mem_config.as_height <= F::bits() - 2);
+        let addr_space_max_bits = log2_strict_usize(
+            (mem_config.as_offset + 2u32.pow(mem_config.as_height as u32)).next_power_of_two()
+                as usize,
+        );
         Self {
             memory_bus,
             mem_config,
             interface_chip: MemoryInterface::Volatile {
                 boundary_chip: VolatileBoundaryChip::new(
                     memory_bus,
-                    mem_config.as_height,
+                    addr_space_max_bits,
                     mem_config.pointer_max_bits,
                     range_checker.clone(),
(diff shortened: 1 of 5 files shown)


crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -232,13 +232,19 @@ impl<F: PrimeField32> MemoryController<F> {
     ) -> Self {
         let range_checker_bus = range_checker.bus();
         let initial_memory = AddressMap::from_mem_config(&mem_config);
+        assert!(mem_config.pointer_max_bits <= F::bits() - 2);
+        assert!(mem_config.as_height <= F::bits() - 2);
+        let addr_space_max_bits = log2_strict_usize(
+            (mem_config.as_offset + 2u32.pow(mem_config.as_height as u32)).next_power_of_two()
```
zlangley, line 238: nit: You can use `log2_ceil_usize` to skip the `next_power_of_two()`.

crates/vm/tests/integration_test.rs
```diff
@@ -56,7 +57,7 @@ where
 
 fn test_native_config() -> NativeConfig {
     NativeConfig {
-        system: SystemConfig::default(),
+        system: SystemConfig::new(3, MemoryConfig::new(2, 1, 16, 29, 15, 32, 1024), 0),
```
zlangley, line 60: Good to know this config works now, but I'm not sure it's better than the old one for testing since native address space is 4.

zlangley (approved):

crates/vm/src/system/memory/controller/mod.rs
jonathanpwang: thanks, forgot we had this

crates/vm/tests/integration_test.rs
jonathanpwang: 4 is still supported right? it exactly hits the edge case: 4 < 1 + 2^2

zlangley: Ah, right.

crates/vm/src/system/memory/controller/mod.rs
zlangley: Did you mean to change this?

jonathanpwang: hm some weird failed save

PR #1497, opened 2025-03-20
chore: `pub` chip fields necessary for serialization/deserialization

manh9203:
These are public chip fields that were previously used to implement `Stateful`. Exposing them publicly allows an external repo to handle serialization and deserialization of those fields. The list of these fields is tracked here: https://docs.google.com/document/d/1yFUDUlXcObTYrLBWoDV0I-zmBQJ1kIQ9LAv-_fCPb8U


jonathanpwang:
`SharedVariableRangeChecker` has `AsRef<VariableRangeChecker>` so you should just do `chip.as_ref().whatever` so let's not add more methods there. Same with `SharedBitwise...`

jonathanpwang (approved):

PR #1499, opened 2025-03-21
chore: String::new() over "".to_string() 

lfgtwo:
PR: https://github.com/openvm-org/openvm/pull/1487  Introduced `"".to_string()`

other place use `String::new()` keep consistency
https://github.com/openvm-org/openvm/blob/efb782371b0fad77c597b77be64bfab75b20c5a6/crates/prof/src/types.rs#L128


jonathanpwang:
Not merging small cosmetic changes

PR #1506, opened 2025-03-24
chore: clean up deps using `cargo shear`

shuklaayush:
- clean up dependencies by running `cargo shear --fix`
- add exceptions for false positives


jonathanpwang (approved):
LGTM, thanks

PR #1510, opened 2025-03-24
chore: update rustc channel version, make changes suggested by the compiler

Golovanov399:


jonathanpwang:
I'll try this again with stable 1.85

PR #1514, opened 2025-03-24
chore: exclude non-executable targets when building

luffykai:


crates/toolchain/build/src/lib.rs
```diff
@@ -383,6 +383,10 @@ pub fn find_unique_executable<P: AsRef<Path>, Q: AsRef<Path>>(
             {
                 return false;
             }
+            // exclude non-bin targets
+            if !target.kind.contains(&"bin".to_string()) {
```
jonathanpwang, line 387: what about example?

jonathanpwang: this is what `target_filter` is for

jonathanpwang (changes requested):
Not sure this is the correct way to handle it; we introduced `target_filter` for this purpose

jonathanpwang (approved):

PR #1515, opened 2025-03-24
chore: make fields in `MemoryController` public

jonathanpwang:


crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -461,6 +456,10 @@ impl<F: PrimeField32> MemoryController<F> {
 
     fn replay_access_log(&mut self) {
         let log = mem::take(&mut self.memory.log);
+        if log.is_empty() {
```
luffykai, line 459: maybe worse noting in which scenario this would be empty

luffykai (approved):

crates/vm/src/system/memory/controller/mod.rs
jonathanpwang: https://github.com/openvm-org/openvm/pull/1515/commits/e6b750c6421e99acb8d96f95aedf501e26df5700

PR #1519, opened 2025-03-25
feat: `GenericSdk<E: StarkFriEngine<BabyBearPoseidon2Config>>`

Golovanov399:
This change introduces a generic type for Sdk (with a typedef for `Sdk` that uses `BabyBearPoseidon2Engine`). This is to allow Sdk usage with different engines that still use the same `StarkGenericConfig = BabyBearPoseidon2Config`.

This will require changes in other repos, but it is inevitable since `let sdk = Sdk;` is now an invalid assignment.

### DIFF
--- a/benchmarks/src/bin/fib_e2e.rs
+++ b/benchmarks/src/bin/fib_e2e.rs
@@ -30,7 +30,7 @@ async fn main() -> Result<()> {
     ));
     let agg_config = args.agg_config();
 
-    let sdk = Sdk;
+    let sdk = Sdk::default();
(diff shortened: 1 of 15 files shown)


jonathanpwang (changes requested):
Can you revert the change of doing `Sdk::app_keygen`? I prefer `sdk.app_keygen` both because it's more natural UX and also in case `Sdk` struct does store configuration information in the future.

I prefer if we just change to a `Sdk::new()` constructor, but keep the functions the same.

jonathanpwang (approved):
lgtm

PR #1520, opened 2025-03-25
test(ci): run all workflows on Cargo.toml changes

jonathanpwang:


shuklaayush:
should it be `"**/Cargo.toml`?

jonathanpwang:
> should it be `"**/Cargo.toml"`?

I only wanted the root Cargo.toml since that's where package versions change. If you change a sub-Cargo.toml, that should fit into the subdirectory rules

shuklaayush (approved):

PR #1521, opened 2025-03-25
feat: check height constraints after finalization and retry if needed

zlangley:


crates/vm/src/arch/config.rs
```diff
@@ -89,11 +89,11 @@ pub struct SystemConfig {
     /// This field is skipped in serde as it's only used in execution and
     /// not needed after any serialize/deserialize.
     #[serde(skip, default = "get_default_segmentation_strategy")]
-    pub segmentation_strategy: Arc<dyn SegmentationStrategy>,
+    pub segmentation_strategy: SegmentationStrategy,
```
jonathanpwang, line 92: We still want to allow custom user-defined `SegmentationStrategy`s here, which requires the `dyn`. If there is a part that must always be run, we should have it as a separate field?

crates/vm/src/arch/extensions.rs
```diff
@@ -966,15 +977,33 @@ impl<F: PrimeField32, E, P> VmChipComplex<F, E, P> {
...
+        let trace_heights = self.current_trace_heights();
+        if any(&trace_heights, |h| *h > self.max_trace_height) {
+            return Err(GenerationError::TraceHeightsLimitExceeded);
+        }
+        if trace_height_constraints.is_empty() {
+            tracing::warn!("generating proof input without trace height constraints");
+        }
+        for constraint in trace_height_constraints {
+            let value = zip_eq(&constraint.coefficients, &trace_heights)
+                .map(|(c, h)| *c as u64 * *h as u64)
```
jonathanpwang, line 1000:
```suggestion
                .map(|(&c, &h)| c as u64 * h as u64)
```
(easier to read)

crates/vm/src/arch/extensions.rs
```diff
@@ -966,15 +977,33 @@ impl<F: PrimeField32, E, P> VmChipComplex<F, E, P> {
...
+                .sum::<u64>();
+            if value >= constraint.threshold as u64 {
+                return Err(GenerationError::TraceHeightsLimitExceeded);
```
jonathanpwang, line 1003: for debugging, either error with the constraint or constraint index? or log something

crates/vm/src/arch/segment.rs
```diff
@@ -36,25 +37,14 @@ const DEFAULT_MAX_SEGMENT_LEN: usize = (1 << 22) - 100;
...
-{
-    fn should_segment(
-        &self,
-        air_names: &[String],
-        trace_heights: &[usize],
-        trace_cells: &[usize],
-    ) -> bool;
-}
-
 /// Default segmentation strategy: segment if any chip's height or cells exceed the limits.
```
jonathanpwang, line 40: comments needs updating

crates/vm/src/arch/vm.rs
```diff
@@ -154,8 +170,8 @@ where
         &self,
         exe: impl Into<VmExe<F>>,
         input: impl Into<Streams<F>>,
-        mut f: impl FnMut(usize, ExecutionSegment<F, VC>) -> R,
-    ) -> Result<Vec<R>, ExecutionError> {
+        mut f: impl FnMut(usize, ExecutionSegment<F, VC>) -> Result<R, GenerationError>,
```
jonathanpwang, line 173: nit: could we make this function generic in `R, E` and have `f: impl FnMut(..) -> Result<R, E>` and the overall function `-> Result<Vec<R>, E>` and take another closure `map_err: impl Fn(ExecutionError) -> E`?

It seems a little better and means we can localize `GenerationPipelineError` (or even better remove it) to only the place where we use it

jonathanpwang: with this change, hoping we can just make `GenerationError` contain `ExecutionError` as a `#[from]`

crates/vm/src/arch/extensions.rs
```diff
@@ -523,7 +529,7 @@ pub enum SystemPeriphery<F: PrimeField32> {
     Poseidon2(Poseidon2PeripheryChip<F>),
 }
 
-impl<F: PrimeField32> SystemComplex<F> {
+impl<F: PrimeField32 + TwoAdicField> SystemComplex<F> {
```
jonathanpwang, line 532: let's remove this and try to use a setter. I think it complicates the structs / traits too much to add `TwoAdicField` everywhere

jonathanpwang (commented):
Some comments, mostly on interfaces and removing `TwoAdicField`.

Discussing offline about `SegmentationStrategy` and how to handle it now.

crates/vm/src/arch/segment.rs
```diff
@@ -45,10 +48,12 @@ pub trait SegmentationStrategy:
         trace_heights: &[usize],
         trace_cells: &[usize],
     ) -> bool;
+
+    fn stricter_strategy(&self) -> Arc<dyn SegmentationStrategy>;
```
jonathanpwang, line 52: add some docs on what this is

crates/vm/src/arch/extensions.rs
```diff
@@ -930,15 +949,17 @@ impl<F: PrimeField32, E, P> VmChipComplex<F, E, P> {
         E: ChipUsageGetter,
         P: ChipUsageGetter,
     {
-        once(self.program_chip().current_trace_cells())
+        // program_chip, connector_chip
+        [0, 0]
+            .into_iter()
             .chain([self.connector_chip().current_trace_cells()])
```
jonathanpwang, line 955: you still have `chain(connector_chip)`

jonathanpwang (approved):
LGTM!

PR #1522, opened 2025-03-26
chore: update toolchain to rust stable 1.85.1

jonathanpwang:
The default toolchain is rust 1.85.1 now -- the codebase uses stable Rust and does not use unstable features (previously this was blocked by snark-verifier using unstable features).

We still use "group_imports" in rustfmt.toml, which is unstable. So for full functionality, use `cargo +nightly fmt`. In a separate PR, we can add instructions on how to configure IDEs to use nightly for formatting only.

Unfortunately, `cargo openvm build` must still rely on a nightly install: it needs to rebuild the std library of a tier 3 target, which is an unstable `cargo` feature.
I updated the nightly toolchain to `nightly-2025-02-14`, which is when Rust 1.86.0 branched from master.

The book has been updated accordingly.

closes INT-3534

### DIFF
--- a/.github/workflows/benchmark-call.yml
+++ b/.github/workflows/benchmark-call.yml
@@ -133,7 +133,7 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
           repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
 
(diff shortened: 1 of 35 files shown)


jonathanpwang:
the Lint errors for example should go away after merge

PR #1532, opened 2025-03-28
feat: Add static SDK check for recursive verification feasibility

zlangley:


crates/sdk/src/keygen/mod.rs
```diff
@@ -193,6 +262,11 @@ impl AggStarkProvingKey {
             }
         });
         let internal_vm_vk = internal_vm_pk.vm_pk.get_vk();
+        assert!(check_recursive_verifier_size(
+            &internal_vm_vk,
+            config.internal_fri_params,
+            config.internal_fri_params.log_blowup
+        ));
```
zlangley, line 269: Let me know if these are the correct place to do these checks.

crates/sdk/src/keygen/mod.rs
```diff
@@ -155,6 +156,77 @@ where
     }
 }
 
+pub fn check_recursive_verifier_size<SC: StarkGenericConfig>(
```
zlangley, line 159: Also ideally this is private but I needed it to be public to use in the benchmarks because they don't go through `AggStarkProvingKey::keygen`.

crates/sdk/src/keygen/mod.rs
```diff
@@ -155,6 +156,77 @@ where
...
+    vk: &MultiStarkVerifyingKey<SC>,
+    fri_params: FriParameters,
+    next_log_blowup: usize,
```
jonathanpwang, line 162: what does next refer to here?

crates/sdk/src/keygen/mod.rs
```diff
@@ -155,6 +156,77 @@ where
...
+
+    if !prep_widths.is_empty() {
+        rounds.push((prep_widths.iter().sum::<usize>(), prep_widths.len(), 2));
+    }
+
+    let mut main_rounds = vec![];
+    for vk in vk.per_air.iter() {
+        let widths = vk.params.width.main_widths();
+        if widths.len() > main_rounds.len() {
+            main_rounds.resize(widths.len(), (0, 0, 2));
```
jonathanpwang, line 187: I think it's weirder than this: the common main is the last main width in each vk, so you do need to check if there's a common main and if there is, pop it. (It's annoyingly complicated.) Try to just match up with the logic in the rust verifier

crates/sdk/src/keygen/mod.rs
```diff
@@ -155,6 +156,77 @@ where
...
+    rounds.extend(main_rounds);
+
+    let mut after_challenge_rounds = vec![];
+    for vk in vk.per_air.iter() {
+        let widths = &vk.params.width.after_challenge;
+        if widths.len() > after_challenge_rounds.len() {
+            after_challenge_rounds.resize(widths.len(), (0, 0, 2));
+        }
+        for (i, width) in widths.iter().enumerate() {
+            after_challenge_rounds[i].0 += *width;
```
jonathanpwang, line 204: this width is in ext field, multiply by ext degree

crates/sdk/src/keygen/mod.rs
```diff
@@ -155,6 +156,77 @@ where
...
+            after_challenge_rounds[i].1 += 1;
+        }
+    }
+    rounds.extend(after_challenge_rounds);
+
+    let quotient_round = (
+        vk.per_air
+            .iter()
+            .map(|vk| vk.quotient_degree as usize)
```
jonathanpwang, line 213: multiply by ext degree

crates/sdk/src/keygen/mod.rs
zlangley: This is the `log_blowup` of the verifier of the output. It determines the maximum trace height that, say, `FriOpeningChip` can have.

jonathanpwang: makes sense, leave a comment

crates/sdk/src/keygen/mod.rs
jonathanpwang: prefer to make them all warnings (maybe with a helper function) instead of asserts so the sdk doesn't panic

crates/sdk/src/keygen/mod.rs
```diff
@@ -352,6 +452,7 @@ pub fn leaf_keygen(
     let leaf_engine = BabyBearPoseidon2Engine::new(fri_params);
     let leaf_vm_pk = info_span!("keygen", group = "leaf")
         .in_scope(|| VirtualMachine::new(leaf_engine, leaf_vm_config.clone()).keygen());
+    check_recursive_verifier_size(&leaf_vm_pk.get_vk(), fri_params, fri_params.log_blowup);
```
jonathanpwang, line 455: the log_blowup here is not next, but also this function is only used in a weird benchmarking codepath, so we could just remove this line

jonathanpwang (approved):
lgtm

PR #1530, opened 2025-03-28
book: update EVM proof format

jonathanpwang:
closes INT-3663 INT-3577

PR #1538, opened 2025-03-28
docs: add more details into the continuations spec

Golovanov399:
This resolves INT-3712.

This change adds details about different steps of the recursive proving from https://hackmd.io/RidB9Q7rRMSw9bjf4YHSLA.

### DIFF
--- a/docs/specs/continuations.md
+++ b/docs/specs/continuations.md
@@ -194,7 +194,7 @@ Input:
 
 Output: `ContinuationVmProof<SC>` containing:
   - `per_segment`: Collection of STARK proofs for each execution segment
-  - `user_public_values`: Proof of public values from memory
+  - `user_public_values`: Merkle proof of public values from memory Merkle root
 
 Cached Trace Commit:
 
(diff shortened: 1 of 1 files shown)


docs/specs/continuations.md
```diff
@@ -170,9 +188,13 @@ App VM executes an executable with inputs and returns a list of segment proofs.
...
 
-Logical Input:
+Input:
+
+- App VM input stream (`StdIn`)
 
-- App VM input stream
+Output: `ContinuationVmProof<SC>` containing:
+  - `per_segment`: Collection of STARK proofs for each execution segment
+  - `user_public_values`: Proof of public values from memory
```
jonathanpwang, line 197:
```suggestion
  - `user_public_values`: Merkle proof of public values from memory Merkle root
```

jonathanpwang (approved):
LGTM,

can you add somewhere as appropriate, that the Root verifier program hardcodes the `internal_exe_commit` as a constant? This explains why it is not a public value of the final root verifier.

for further completeness, we should add the public values information for the halo2 wrapper from https://github.com/openvm-org/openvm/pull/1530 into this spec as well

jonathanpwang (approved):

PR #1539, opened 2025-03-28
chore: Implement `Send` for `PhantomChip`

manh9203:
Enable `PhantomChip` to implement `Send` so that it can be serialized and deserialized in parallel

### DIFF
--- a/crates/vm/src/arch/execution.rs
+++ b/crates/vm/src/arch/execution.rs
@@ -320,7 +320,7 @@ impl<T: FieldAlgebra> From<(u32, Option<T>)> for PcIncOrSet<T> {
 ///
 /// Phantom sub-instructions are only allowed to use operands
 /// `a,b` and `c_upper = c.as_canonical_u32() >> 16`.
-pub trait PhantomSubExecutor<F> {
+pub trait PhantomSubExecutor<F>: Send {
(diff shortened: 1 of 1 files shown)


crates/vm/src/system/phantom/mod.rs
```diff
@@ -91,7 +91,7 @@ pub struct PhantomChip<F> {
     pub air: PhantomAir,
     pub rows: Vec<PhantomCols<F>>,
     streams: OnceLock<Arc<Mutex<Streams<F>>>>,
-    phantom_executors: FxHashMap<PhantomDiscriminant, Box<dyn PhantomSubExecutor<F>>>,
+    phantom_executors: FxHashMap<PhantomDiscriminant, Box<dyn PhantomSubExecutor<F> + Send>>,
```
jonathanpwang, line 94: do you want to just add `Send` to `PhantomSubExecutor` trait?

jonathanpwang (approved):
lgtm, but since we basically enforce it throughout, we can add `: Send` to the trait itself

crates/vm/src/system/phantom/mod.rs
manh9203: That sounds good. I'll change to do it.

PR #1547, opened 2025-03-31
chore: fix broken links

bodhi-crypo:


jonathanpwang (approved):
mdbook resolves the original and works fine (locally and in deployed version), but your version also works and is a little more correct

PR #1548, opened 2025-03-31
fix: don't segment due to fixed-size traces

zlangley:
This fix only affects logging and segmentation logic during execution under certain misconfigured parameter settings. It does not affect any circuit soundness.

Fixes two separate issues around segmentation with respect to fixed-size traces: the first being that `constant_trace_height` was not implemented for some wrapper types, and the second being that
`VmChipComplex::current_trace_cells` did not check if a chip had constant trace height.


jonathanpwang (approved):
Going to merge this for now, but we will do a refactor soon. 
The current `current_trace_cells` is used for a few different purposes, and it's not clear when we want to know truly the data size of all trace matrices, versus when we only care about the dynamic parts for some segmentation purposes.

PR #1549, opened 2025-03-31
feat: `OpenVmHalo2Verifier`

HrikB:
This PR adds a new verifier contract generation that wraps the original `snark-verifier` output (via inheritance). The goal of this wrapper `OpenVmHalo2Verifier` is to expose a more friendly interface to users that cleanly separates out the guest program public values. `OpenVmHalo2Verifier` exposes the following interface:

```solidity
interface IOpenVmHalo2Verifier {
    function verify(bytes calldata publicValues, bytes calldata proofData, bytes32 appExeCommit, bytes32 appVmCommit)
        external
        view;
}
```

- `publicValues`: The bytes revealed in the OpenVM guest program packed together.
- `proofData`: Defined as `abi.encodePacked(KZG accumulators, publicValuesSuffix)`
- `appExeCommit`: The commitment to the OpenVM application executable whose execution is being verified.
- `appVmCommit`: The commitment to the VM configuration (aka `leaf_exe_commit`)

Once received, the proof is constructed into the format expected by `snark-verifier`. The expected format is `abi.encodePacked(proofData[0:0x180], appExeCommit, appVmExeCommit, publicValuesPayload, proofData[0x180:])` where `publicValuesPayload` is a memory payload with each byte in `publicValues` separated into its own `bytes32` word.

Since `OpenVmHalo2Verifier` inherits the `snark-verifier` output, the proof is forwarded via self-call.

## Verifier Generation

The smart contract is written as a template that does not compile in isolation. During generation, the OpenVM SDK will fill out the maximum amount of public values and the OpenVM version with which the generation happened.

Given an output folder, the relevant contracts are written into the following folder structure:

```
halo2/
├── interfaces/
│   └── IOpenVmHalo2Verifier.sol
├── OpenVmHalo2Verifier.sol
├── Halo2Verifier.sol
```

`cargo openvm setup` will now generate this output directly into the `~/.openvm/` dir.

Closes INT-3710

### DIFF
--- a/crates/cli/tests/app_e2e.rs
+++ b/crates/cli/tests/app_e2e.rs
@@ -24,8 +24,7 @@ fn test_cli_app_e2e() -> Result<()> {
             "--exe-output",
             temp_exe.to_str().unwrap(),
         ],
-    )
-    .unwrap_or_else(|e| panic!("{:?}", e));
+    )?;
 
     run_cmd(
         "cargo",
@@ -39,8 +38,7 @@ fn test_cli_app_e2e() -> Result<()> {
             "--vk-output",
             temp_vk.to_str().unwrap(),
         ],
-    )
-    .unwrap_or_else(|e| panic!("{:?}", e));
+    )?;
 
     run_cmd(
         "cargo",
@@ -52,8 +50,7 @@ fn test_cli_app_e2e() -> Result<()> {
             "--config",
             "example/openvm.toml",
         ],
-    )
-    .unwrap_or_else(|e| panic!("{:?}", e));
+    )?;
 
     run_cmd(
         "cargo",
@@ -68,8 +65,7 @@ fn test_cli_app_e2e() -> Result<()> {
             "--output",
             temp_proof.to_str().unwrap(),
         ],
-    )
-    .unwrap_or_else(|e| panic!("{:?}", e));
+    )?;
 
     run_cmd(
         "cargo",
@@ -82,8 +78,7 @@ fn test_cli_app_e2e() -> Result<()> {
             "--proof",
             temp_proof.to_str().unwrap(),
         ],
-    )
-    .unwrap_or_else(|e| panic!("{:?}", e));
+    )?;
 
     Ok(())
 }
--- a/crates/cli/tests/app_e2e.rs
+++ b/crates/cli/tests/app_e2e.rs
@@ -24,7 +24,8 @@ fn test_cli_app_e2e() -> Result<()> {
             "--exe-output",
             temp_exe.to_str().unwrap(),
         ],
-    )?;
+    )
+    .unwrap_or_else(|e| panic!("{:?}", e));
 
     run_cmd(
         "cargo",
@@ -38,7 +39,8 @@ fn test_cli_app_e2e() -> Result<()> {
             "--vk-output",
             temp_vk.to_str().unwrap(),
         ],
-    )?;
+    )
+    .unwrap_or_else(|e| panic!("{:?}", e));
 
     run_cmd(
         "cargo",
@@ -50,7 +52,8 @@ fn test_cli_app_e2e() -> Result<()> {
             "--config",
             "example/openvm.toml",
         ],
-    )?;
+    )
+    .unwrap_or_else(|e| panic!("{:?}", e));
 
     run_cmd(
         "cargo",
@@ -65,7 +68,8 @@ fn test_cli_app_e2e() -> Result<()> {
             "--output",
             temp_proof.to_str().unwrap(),
         ],
-    )?;
+    )
+    .unwrap_or_else(|e| panic!("{:?}", e));
 
     run_cmd(
         "cargo",
@@ -78,7 +82,8 @@ fn test_cli_app_e2e() -> Result<()> {
             "--proof",
             temp_proof.to_str().unwrap(),
         ],
-    )?;
+    )
+    .unwrap_or_else(|e| panic!("{:?}", e));
 
     Ok(())
 }
--- a/crates/sdk/Cargo.toml
+++ b/crates/sdk/Cargo.toml
@@ -8,6 +8,7 @@ repository.workspace = true
 license.workspace = true
 
 [dependencies]
+
 p3-fri = { workspace = true }
 openvm-algebra-circuit = { workspace = true }
 openvm-algebra-transpiler = { workspace = true }
@@ -34,6 +35,8 @@ openvm-circuit = { workspace = true }
 openvm-continuations = { workspace = true }
 openvm = { workspace = true }
 
+alloy-primitives = { workspace = true }
+alloy-sol-types = { workspace = true, features = ["json"] }
 bitcode = { workspace = true }
 bon = { workspace = true }
 derivative = { workspace = true }
@@ -49,6 +52,10 @@ clap = { workspace = true, features = ["derive"] }
 serde_with = { workspace = true, features = ["hex"] }
 serde_json.workspace = true
 thiserror.workspace = true
+snark-verifier = { workspace = true }
+snark-verifier-sdk.workspace = true
+tempfile.workspace = true
+hex.workspace = true
 
 [features]
 default = ["parallel"]
--- a/crates/sdk/contracts/foundry.toml
+++ b/crates/sdk/contracts/foundry.toml
@@ -1,0 +1,31 @@
+[profile.default]
+src = "src"
+out = "out"
+libs = ["lib"]
+verbosity = 2
+solc = "0.8.19"
+optimizer = true
+optimizer_runs = 100000
+evm_version = "shanghai"
+show_progress = true
+fs_permissions = [{ access = "read", path = "./template"}, { access = "read", path = "./test/helpers/MockDeps.sol"}]
+ffi = true
+
+[profile.default.optimizer_details]
+  constantOptimizer = false
+  yul = false
+
+[fuzz]
+  runs = 256
+
+[fmt]
+  bracket_spacing = true
+  int_types = "long"
+  line_length = 120
+  multiline_func_header = "attributes_first"
+  number_underscore = "thousands"
+  quote_style = "double"
+  single_line_statement_blocks = "single"
+  tab_width = 4
+  wrap_comments = false
+  
--- a/crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
+++ b/crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
@@ -39,12 +39,12 @@ contract OpenVmHalo2Verifier is Halo2Verifier, IOpenVmHalo2Verifier {
     /// proof[12 * 32..13 * 32]: app exe commit
     /// proof[13 * 32..14 * 32]: app vm commit
     /// proof[14 * 32..(14 + PUBLIC_VALUES_LENGTH) * 32]: publicValues[0..PUBLIC_VALUES_LENGTH]
-    /// proof[(14 + PUBLIC_VALUES_LENGTH) * 32..]: Public Values Suffix
+    /// proof[(14 + PUBLIC_VALUES_LENGTH) * 32..]: Proof Suffix
     ///
     /// @param publicValues The PVs revealed by the OpenVM guest program.
     /// @param proofData All components of the proof except the public values and
     /// app exe and vm commits. The expected format is:
-    /// `abi.encodePacked(kzgAccumulators, pvsSuffix)`
+    /// `abi.encodePacked(kzgAccumulators, proofSuffix)`
     /// @param appExeCommit The commitment to the OpenVM application executable whose execution
     /// is being verified.
     /// @param appVmCommit The commitment to the VM configuration.
@@ -99,7 +99,7 @@ contract OpenVmHalo2Verifier is Halo2Verifier, IOpenVmHalo2Verifier {
         // proof[0x180..0x1a0]: app exe commit
         // proof[0x1a0..0x1c0]: app vm commit
         // proof[0x1c0..(0x1c0 + PUBLIC_VALUES_LENGTH * 32)]: publicValues[0..PUBLIC_VALUES_LENGTH]
-        // proof[(0x1c0 + PUBLIC_VALUES_LENGTH * 32)..]: Public Values Suffix
+        // proof[(0x1c0 + PUBLIC_VALUES_LENGTH * 32)..]: Proof Suffix
 
         /// @solidity memory-safe-assembly
         assembly {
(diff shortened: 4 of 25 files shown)


crates/sdk/contracts/foundry.toml
```diff
@@ -0,0 +1,31 @@
...
+src = "src"
+out = "out"
+libs = ["lib"]
+verbosity = 2
+solc = "0.8.19"
+optimizer = true
+optimizer_runs = 100000
+evm_version = "shanghai"
+show_progress = true
+fs_permissions = [{ access = "read", path = "./template"}, { access = "read", path = "./test/helpers/ExampleDeps.sol"}]
```
yi-sun, line 11: Should this be something like DummyDeps?  It was initially confusing to me because it's not really an example of something we need to emulate.

crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    /// @dev Proof verification failed
+    error ProofVerificationFailed();
+
+    /// @dev The length of the proof data, in bytes.
+    uint256 private constant PROOF_DATA_LENGTH = (12 + 43) * 32;
+
+    /// @dev The length of the public values, in bytes. This value is set by
+    /// OpenVM. For security analysis, the contract assumes that OpenVM will
+    /// never set this to a value larger than 10000.
```
yi-sun, line 26: We should make sure this is enforced via an assert in the OpenVM setup.

crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+    }
+
+    /// @dev The assembly code should perform the same function as the following
+    /// solidity code:
+    //
+    /// ```solidity
+    /// bytes memory proof =
+    ///     abi.encodePacked(proofData[0:0x180], appExeCommit, appVmExeCommit, publicValuesPayload, proofData[0x180:]);
```
yi-sun, line 78:
```suggestion
    ///     abi.encodePacked(proofData[0:0x180], appExeCommit, appVmCommit, publicValuesPayload, proofData[0x180:]);
```

yi-sun (changes requested):
Some nits -- on the `appVmCommit` renaming, we should make sure it is consistent with the rest of the codebase or somewhere explain that `appVmCommit` is a renaming of `leafExeCommit` which is easier to understand for devs.

@nyunyunyunyu -- can you also take a look?

crates/sdk/contracts/template/IOpenVmHalo2Verifier.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
```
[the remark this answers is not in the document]
HrikB: I used to have a custom error in here which is only `0.8.4` and beyond but we can drop down to `0.8.0`

crates/sdk/contracts/foundry.toml
HrikB: Yes we can do this

crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
HrikB: Are we ok with 10k as a hard limit?

crates/sdk/src/lib.rs
```diff
@@ -56,13 +73,22 @@ pub mod prover;
...
 
 pub type NonRootCommittedExe = VmCommittedExe<SC>;
 
+const OPENVM_HALO2_VERIFIER_INTERFACE: &str =
+    include_str!("../contracts/template/IOpenVmHalo2Verifier.sol");
+const OPENVM_HALO2_VERIFIER_TEMPLATE: &str =
+    include_str!("../contracts/template/OpenVmHalo2Verifier.sol");
+
+sol! {
+    interface IOpenVmHalo2Verifier {
```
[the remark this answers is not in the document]
HrikB: Yeah I was thinking about this. It can't import the file unfortunately (alloy only supports json for file imports). One potential way to do it would be to make us build the foundry project but even that wouldn't work since those template files don't get compiled during a `forge build`

crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
HrikB: Setting to 8192, as discussed

crates/sdk/src/lib.rs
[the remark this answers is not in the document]
yi-sun: Can you write a `.json` from the interface by file manipulation?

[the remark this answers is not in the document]
HrikB: Lemme try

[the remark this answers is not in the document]
HrikB: I guess this wouldn't really be compile time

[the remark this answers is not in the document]
HrikB: You could do some macro stuff but I don't feel it's worth the complexity

crates/cli/tests/app_e2e.rs
```diff
@@ -24,7 +24,8 @@ fn test_cli_app_e2e() -> Result<()> {
             "--exe-output",
             temp_exe.to_str().unwrap(),
         ],
-    )?;
+    )
+    .unwrap_or_else(|e| println!("{:?}", e));
```
jonathanpwang, line 28: is this intentional? it means it won't error on failure

crates/sdk/Cargo.toml
```diff
@@ -8,6 +8,10 @@ repository.workspace = true
 license.workspace = true
 
 [dependencies]
+alloy-primitives = "0.8.25"
+alloy-sol-types = "0.8.25"
+snark-verifier = { version = "0.2.0", default-features = false }
```
jonathanpwang, line 13: these should all use workspace versions

crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    /// @notice A wrapper that constructs the proof into the right format for
+    /// use with the `snark-verifier` verification.
+    ///
+    /// @dev The verifier expected proof format is:
+    /// proof[..12 * 32]: KZG accumulators
+    /// proof[12 * 32..13 * 32]: app exe commit
+    /// proof[13 * 32..14 * 32]: app vm commit
+    /// proof[14 * 32..(14 + PUBLIC_VALUES_LENGTH) * 32]: publicValues[0..PUBLIC_VALUES_LENGTH]
+    /// proof[(14 + PUBLIC_VALUES_LENGTH) * 32..]: Public Values Suffix
```
jonathanpwang, line 43: why do you call this public values suffix? it is not related to public values at all

crates/sdk/src/lib.rs
```diff
@@ -283,4 +447,81 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
             .map_err(|reason| eyre::eyre!("Sdk::verify_evm_proof: {reason:?}"))?;
         Ok(gas_cost)
     }
+
+    /// `OpenVmHalo2Verifier` wraps the `snark-verifer` contract, meaning that
+    /// the default `fallback` interface can still be used. This function uses
+    /// the fallback interface as opposed to the `verify(..)` interface.
+    pub fn verify_openvm_halo2_proof(
```
jonathanpwang, line 454: why do we even need the old `verify_evm_proof` given this? why do we need the old `EvmVerifier`?

crates/cli/tests/app_e2e.rs
HrikB: Ah yes, this should `panic!` instead of `println!`

crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
HrikB: We had discussed calling it this to denote its relation to the public values, since this basically refers to the portion of the proof that comes after the public values. Do you have a better suggestion for what to call this part of the proof?

crates/sdk/src/lib.rs
HrikB: We don't need `EvmVerifier` anymore. I kept it in for optionality but am open to removing it altogether.

crates/sdk/src/lib.rs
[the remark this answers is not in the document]
HrikB: As discussed, will manually include the abi in `contracts/` and verify in CI that it corresponds to the `IOpenVmHalo2Verifier.sol` file.

crates/sdk/src/lib.rs
```diff
@@ -56,13 +73,22 @@ pub mod prover;
...
 
-use crate::types::EvmProof;
-
 pub mod fs;
 pub mod types;
 
 pub type NonRootCommittedExe = VmCommittedExe<SC>;
 
+const OPENVM_HALO2_VERIFIER_INTERFACE: &str =
+    include_str!("../contracts/template/IOpenVmHalo2Verifier.sol");
```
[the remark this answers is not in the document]
HrikB: Is a named constants worth it for this path? The path is only used here

crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
yi-sun: How about `PROOF_SUFFIX`?

yi-sun (approved):

crates/sdk/contracts/template/OpenVmHalo2Verifier.sol
jonathanpwang: I'm ok with `PROOF_SUFFIX`

HrikB: Ok will go with `PROOF_SUFFIX`

crates/sdk/src/fs.rs
```diff
@@ -98,6 +102,45 @@ pub fn read_evm_verifier_from_folder<P: AsRef<Path>>(folder: P) -> Result<EvmVer
...
+/// ```text
+/// halo2/
+/// ├── interfaces/
+/// │   └── IOpenVmHalo2Verifier.sol
+/// ├── OpenVmHalo2Verifier.sol
+/// └── Halo2Verifier.sol
+/// ```
+///
+/// If the relevant directories do not exist, they will be created.
+pub fn write_openvm_halo2_verifier_to_folder<P: AsRef<Path>>(
```
jonathanpwang, line 116: `write_evm_halo2_verifier_to_folder`

crates/sdk/src/lib.rs
```diff
@@ -273,6 +302,152 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
...
+                start = Some(idx);
+            }
+        }
+        if let Some(last) = start {
+            split.push(&bytes[last..]);
+        }
+        split
+    }
+
     pub fn verify_evm_proof(
```
jonathanpwang, line 451: we discussed offline: will delete this since it's a fallback

crates/sdk/src/lib.rs
```diff
@@ -283,4 +458,81 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
...
+            openvm_verifier.artifact.bytecode.clone(),
+            vec![evm_proof.instances.clone()],
+            evm_proof.proof.clone(),
+        )
+        .map_err(|reason| eyre::eyre!("Sdk::verify_openvm_evm_proof: {reason:?}"))?;
+        Ok(gas_cost)
+    }
+
+    /// Uses the `verify(..)` interface of the `OpenVmHalo2Verifier` contract.
+    pub fn verify_openvm_halo2_proof_with_wrapped_interface(
```
jonathanpwang, line 481: rename this `verify_evm_halo2_proof` since this will be the preferred function

crates/sdk/src/lib.rs
```diff
@@ -273,6 +302,152 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
...
+    ///     "OpenVmHalo2Verifier.sol:OpenVmHalo2Verifier",
+    ///     "=======",
+    ///     "Binary:"
+    ///     "[compiled bytecode]"
+    ///     ...
+    /// ]
+    ///
+    /// Once we find "OpenVmHalo2Verifier.sol:OpenVmHalo2Verifier," we can skip
+    /// to the appropriate offset to get the compiled bytecode.
+    fn extract_binary(output: &[u8], contract_name: &str) -> Vec<u8> {
```
jonathanpwang, line 420: these are internal functions just for EVM verify; make a separate module `evm` for all the evm types and methods not directly callable from `Sdk`

crates/sdk/tests/integration_test.rs
```diff
@@ -373,7 +376,7 @@ fn test_e2e_proof_generation_and_verification() {
         )
         .unwrap();
     let evm_verifier = sdk
-        .generate_snark_verifier_contract(&params_reader, &agg_pk)
+        .generate_openvm_halo2_verifier_contract(&params_reader, &agg_pk)
```
jonathanpwang, line 379: I think this test is actually duplicate / a subset of the new test you added; we don't need both to run in CI, and you can just delete this one

jonathanpwang (commented):
LGTM after deleting the fallback types/APIs from SDK as discussed offline.

To avoid confusion, I think we can also rename the old `EvmVerifier` struct to `FallbackEvmVerifier` (and I think it is now not really used)

crates/sdk/src/fs.rs
```diff
@@ -90,22 +92,68 @@ pub fn write_evm_proof_to_file<P: AsRef<Path>>(proof: EvmProof, path: P) -> Resu
...
-    let sol_code_path = folder.as_ref().join(EVM_VERIFIER_SOL_FILENAME);
-    let sol_code = std::fs::read_to_string(sol_code_path)?;
+pub fn read_evm_halo2_verifier_from_folder<P: AsRef<Path>>(folder: P) -> Result<EvmHalo2Verifier> {
+    let halo2_verifier_code_path = folder.as_ref().join(EVM_HALO2_VERIFIER_PARENT_NAME);
+    let openvm_verifier_code_path = folder.as_ref().join(EVM_HALO2_VERIFIER_BASE_NAME);
+    let interface_path = folder
+        .as_ref()
+        .join("interfaces")
+        .join(EVM_HALO2_VERIFIER_INTERFACE_NAME);
+    let halo2_verifier_code = std::fs::read_to_string(halo2_verifier_code_path)?;
```
jonathanpwang, line 102: when you use it 3 times, better to just import the function

crates/sdk/src/fs.rs
```diff
@@ -90,22 +92,68 @@ pub fn write_evm_proof_to_file<P: AsRef<Path>>(proof: EvmProof, path: P) -> Resu
...
+    }
+
+    let halo2_verifier_code_path = folder.join(EVM_HALO2_VERIFIER_PARENT_NAME);
+    let openvm_verifier_code_path = folder.join(EVM_HALO2_VERIFIER_BASE_NAME);
+    let interface_path = folder
+        .join("interfaces")
+        .join(EVM_HALO2_VERIFIER_INTERFACE_NAME);
+
+    if let Some(parent) = interface_path.parent() {
+        std::fs::create_dir_all(parent)?;
```
jonathanpwang, line 144: ditto

jonathanpwang (approved):
LGTM

I think at a later time, we should feature-gate all the EVM parts since the dependencies on alloy, halo2, etc may be undesirable if the usage of SDK is outside the EVM context, but let's leave it for now.

extensions/native/recursion/src/halo2/wrapper.rs
```diff
@@ -28,7 +28,7 @@ use crate::halo2::{
 };
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct EvmVerifier {
+pub struct FallbackEvmVerifier {
```
nyunyunyunyu, line 31: Add a comment to explain why it's called "FallbackEvmVerifier"

nyunyunyunyu (approved):

PR #1550, opened 2025-04-01
fix: use absolute paths in `VmConfig` derive macro

shuklaayush:
- use absolute paths for structs used only in the derive macro

### DIFF
--- a/crates/vm/derive/src/lib.rs
+++ b/crates/vm/derive/src/lib.rs
@@ -347,14 +347,14 @@ pub fn vm_generic_config_derive(input: proc_macro::TokenStream) -> proc_macro::T
             let periphery_type = Ident::new(&format!("{}Periphery", name), name.span());
 
             TokenStream::from(quote! {
-                #[derive(::openvm_circuit::circuit_derive::ChipUsageGetter, ::openvm_circuit::circuit_derive::Chip, ::openvm_circuit::derive::InstructionExecutor, ::derive_more::From, ::openvm_circuit::derive::AnyEnum)]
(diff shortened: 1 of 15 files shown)


crates/vm/derive/src/lib.rs
```diff
@@ -344,34 +347,34 @@ pub fn vm_generic_config_derive(input: proc_macro::TokenStream) -> proc_macro::T
...
-                pub enum #periphery_type<F: PrimeField32> {
+                #[derive(::openvm_circuit::circuit_derive::ChipUsageGetter, ::openvm_circuit::circuit_derive::Chip, ::derive_more::From, ::openvm_circuit::derive::AnyEnum)]
+                pub enum #periphery_type<F: ::openvm_stark_backend::p3_field::PrimeField32> {
                     #[any_enum]
                     #source_name_upper(#source_periphery_type<F>),
                     #(#periphery_enum_fields)*
                 }
 
-                impl<F: PrimeField32> VmConfig<F> for #name {
+                impl<F: ::openvm_stark_backend::p3_field::PrimeField32> ::openvm_circuit::arch::VmConfig<F> for #name {
```
jonathanpwang, line 364: it seems a good quality of life change to assume you must have `openvm-circuit` imported, but not everything needs `openvm-stark-backend` perhaps (it might just import `openvm-stark-sdk`). we could either just leave `PrimeField32` and require people to figure out how to import it themselves, or we do have a re-export `openvm-stark-backend` from `openvm-circuit`

shuklaayush: makes sense. reused `stark-backend` from `openvm-circuit`. the only other dependency is `derive_more` but i guess that's fine

shuklaayush: makes sense. left it to be imported by the user

jonathanpwang (approved):

jonathanpwang (approved):

PR #1551, opened 2025-04-01
fix: Update Mod Builder symbolic expression multiplication test

Avaneesh-axiom:
The symbolic expression multiplication unit test was being run with an incorrect expected result. The test became out-of-date after #1473 changed the `max_abs` method.

This PR updates the test with the correct expected result and adds the Mod Builder crate's unit tests to the CI workflow.

This change has no impact on the functionality of the Mod Builder crate as it only updates the tests.

### DIFF
--- a/crates/circuits/mod-builder/src/tests.rs
+++ b/crates/circuits/mod-builder/src/tests.rs
@@ -387,9 +387,9 @@ fn test_symbolic_limbs_mul() {
         Box::new(SymbolicExpr::Var(0)),
         Box::new(SymbolicExpr::Var(1)),
     );
-    // x * y = pq, q can be up to p so can limbs as p.
-    // x * y and p * q  both have 63 limbs.
-    let expected_q = 32;
-    let expected_carry = 63;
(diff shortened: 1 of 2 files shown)


crates/circuits/mod-builder/src/tests.rs
```diff
@@ -389,7 +390,7 @@ fn test_symbolic_limbs_mul() {
     );
     // x * y = pq, q can be up to p so can limbs as p.
```
jonathanpwang, line 391: update the comment too

crates/circuits/mod-builder/src/tests.rs
```diff
@@ -345,6 +345,7 @@ fn test_select2() {
 
 fn test_symbolic_limbs(expr: SymbolicExpr, expected_q: usize, expected_carry: usize) {
     let prime = secp256k1_coord_prime();
+    println!("prime: {}", prime);
```
jonathanpwang, line 348: remember to delete this when done debugging

jonathanpwang (approved):

PR #1552, opened 2025-04-01
fix(ci): run openvm-cli tests

shuklaayush:


jonathanpwang (approved):
CLI tests are pretty slow, but SDK tests are already slow, so we can enable and try to optimize them together later

PR #1553, opened 2025-04-01
feat: make provers generic over engine

Golovanov399:
This adds the ability to use different engines for proving.

### DIFF
--- a/benchmarks/src/utils.rs
+++ b/benchmarks/src/utils.rs
@@ -20,10 +20,8 @@ use openvm_sdk::{
 };
 use openvm_stark_backend::utils::metrics_span;
 use openvm_stark_sdk::{
-    config::{
-        baby_bear_poseidon2::{BabyBearPoseidon2Config, BabyBearPoseidon2Engine},
-        FriParameters,
-    },
+    config::{baby_bear_poseidon2::BabyBearPoseidon2Config, FriParameters},
+    engine::StarkFriEngine,
     openvm_stark_backend::Chip,
     p3_baby_bear::BabyBear,
 };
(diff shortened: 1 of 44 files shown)


benchmarks/src/utils.rs
```diff
@@ -144,7 +142,7 @@ impl BenchmarkCli {
         build_bench(manifest_dir, profile)
     }
 
-    pub fn bench_from_exe<VC>(
+    pub fn bench_from_exe<VC, E: StarkFriEngine<SC>>(
```
jonathanpwang, line 145: should/can we keep this function for fixed engine just for convenience? so we don't need to change all the benchmarks?

crates/sdk/src/prover/stark.rs
```diff
@@ -3,18 +3,19 @@ use std::sync::Arc;
...
     keygen::{AggStarkProvingKey, AppProvingKey},
     prover::{agg::AggStarkProver, app::AppProver},
     NonRootCommittedExe, RootSC, StdIn, F, SC,
 };
 
-pub struct StarkProver<VC> {
-    app_prover: AppProver<VC>,
+pub struct StarkProver<VC, E: StarkFriEngine<SC>> {
+    app_prover: AppProver<VC, E>,
     agg_prover: AggStarkProver,
```
jonathanpwang, line 16: It looks like `AggStarkProver` also needs generic `E`

jonathanpwang (commented):
`AggStarkProver` needs similar `<E>`

jonathanpwang (approved):
LGTM

PR #1554, opened 2025-04-02
fix: RootVerifierLocalProver should respect AIR heights

nyunyunyunyu:
- Not a soundness issue. Just an issue of `RootVerifierLocalProver` implementation in SDK.
- In some edge cases, the trace heights of the root verifier circuit are less than the required heights. In this case, `RootVerifierLocalProver` should pad its traces to the required heights.
- [Breaking change] Add a new field `internal_heights` into `RootVerifierProvingKey`.

### DIFF
--- a/crates/sdk/src/keygen/mod.rs
+++ b/crates/sdk/src/keygen/mod.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use derivative::Derivative;
 use dummy::{compute_root_proof_heights, dummy_internal_proof_riscv_app_vm};
 use openvm_circuit::{
-    arch::{VirtualMachine, VmConfig},
+    arch::{VirtualMachine, VmComplexTraceHeights, VmConfig},
     system::{memory::dimensions::MemoryDimensions, program::trace::VmCommittedExe},
 };
 use openvm_continuations::{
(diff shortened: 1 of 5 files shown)


crates/sdk/src/keygen/mod.rs
```diff
@@ -392,9 +393,8 @@ pub struct RootVerifierProvingKey {
     pub root_committed_exe: Arc<VmCommittedExe<RootSC>>,
     /// The constant trace heights, ordered by AIR ID.
     pub air_heights: Vec<usize>,
-    // The following is currently not used:
-    // The constant trace heights, ordered according to an internal ordering determined by the `NativeConfig`.
-    // pub internal_heights: VmComplexTraceHeights,
+    /// The constant trace heights in a semaantic way.
+    pub internal_heights: VmComplexTraceHeights,
```
jonathanpwang, line 397: can we rename `internal_heights` to something else, because I got confused about this versus internal verifier?

jonathanpwang: maybe just `override_heights` or `raw_heights`? It's just a permutation of `air_heights` in the chip complex ordering right?

jonathanpwang (approved):

crates/sdk/src/keygen/mod.rs
jonathanpwang: or just `heights` ...

PR #1556, opened 2025-04-04
feat(test): add execution benchmarks

shuklaayush:
Closes INT-3706

- Moved existing benchmarks to `benchmarks/prove` crate
- Added a new `benchmarks/execute` binary crate that runs execution for a list of guest programs
- Added programs for benchmarking in `benchmarks/guest`
- Added a ci workflow to run execution benchmarks
- Note: Timing metrics for summary are currently being parsed from logs. I'll change it to use proper metrics in a follow up


shuklaayush:
not sure why the cli tests failed on the [first run](https://github.com/openvm-org/openvm/actions/runs/14269489157/job/39999194001#step:13:370). happened once before too

benchmarks/guest/revm_blake2f/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[workspace]
+[package]
+name = "openvm-revm-blake2f-program"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+openvm = { path = "../../../crates/toolchain/openvm", features = ["std"] }
+
+revm-precompile = { git = "https://github.com/axiom-crypto/revm", branch = "v43-openvm", default-features = false, features = [
```
jonathanpwang, line 10: this dependency may lead to some annoyances with version updates in the future

jonathanpwang (commented):
Let us exclude the guest programs that rely on a patched version of revm, especially if it requires an axiom dependency. The benchmarks in this repo should be as self-contained and standalone as possible.

In addition, please update the readme around how to: (1) run the execute benchmarks, (2) if we need to update the saved elfs, how do we do that.

jonathanpwang (approved):
lgtm

PR #1558, opened 2025-04-05
fix: Replace Invalid Curve Points with Safe Defaults

mdqst:
Replaced `Secp256k1Point::from_xy_nonidentity(x, y).unwrap()` with safe, known-good points:

- `p1 = Secp256k1Point::generator()` — base point G  
- `p2 = &p1 + &p1` — doubled point (2G)

This avoids potential `panic!` caused by invalid coordinates that don't lie on the curve.

### Why it matters
`from_xy_nonidentity` returns `None` if the point is not on the curve. Using trusted points eliminates the need for manual validation and ensures the example runs safely and predictably.


jonathanpwang:
Not a fix. The example points are fine, and the unwrap is testing the logic of checking whether the point is on curve

PR #1559, opened 2025-04-06
feat: typeless `AddressMap` with typed APIs

jonathanpwang:
Note: this PR is not targeting `main`.
I've used `TODO` and `TEMP` to mark places in code that will need to be cleaned up before merging to `main`.

Beginning the refactor of online memory to allow different host types in different address spaces.
Going to touch a lot of APIs.
Focusing on stabilizing APIs - currently this PR will not improve performance.

Tests will not all pass because I have intentionally disabled some logging required for trace generation.
Only execution tests will pass (or run the execute benchmark).

In future PR(s):

- [ ] make `Memory` trait for execution read/write API
- [ ] better handling of type conversions for memory image
- [ ] replace the underlying memory implementation with other implementations like mmap

Towards INT-3743

Even with wasteful conversions, execution is faster:
Before: https://github.com/openvm-org/openvm/actions/runs/14318675080
After: https://github.com/openvm-org/openvm/actions/runs/14371335248?pr=1559

### DIFF
--- a/crates/vm/src/system/memory/paged_vec.rs
+++ b/crates/vm/src/system/memory/paged_vec.rs
@@ -123,13 +123,13 @@ impl<const PAGE_SIZE: usize> PagedVec<PAGE_SIZE> {
     /// # Panics
     /// If `from..from + size_of<BLOCK>()` is out of bounds.
     #[inline(always)]
-    pub fn set<BLOCK: Copy>(&mut self, from: usize, values: BLOCK) -> BLOCK {
+    pub fn set<BLOCK: Copy>(&mut self, from: usize, values: &BLOCK) -> BLOCK {
         // Create an uninitialized array for old values.
         let mut result: MaybeUninit<BLOCK> = MaybeUninit::uninit();
         self.set_range_generic(
             from,
             size_of::<BLOCK>(),
-            &values as *const _ as *const u8,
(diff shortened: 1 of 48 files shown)


Golovanov399 (approved):

crates/vm/src/system/memory/paged_vec.rs
```diff
@@ -253,49 +226,109 @@ impl<T: Clone + Default, const PAGE_SIZE: usize> AddressMap<T, PAGE_SIZE> {
...
+                        .map(move |(ptr_idx, x)| {
+                            (
+                                (as_idx as u32 + self.as_offset, ptr_idx as u32),
+                                F::from_canonical_u8(x),
+                            )
+                        })
+                        .collect_vec()
+                } else {
+                    // TEMP
+                    assert_eq!(cell_size, 4);
```
shuklaayush, line 245: Maybe a `CellKind` enum would be better instead of `cell_size`

shuklaayush (approved):

crates/vm/src/system/memory/paged_vec.rs
jonathanpwang: yea I also thought that; I kind of intentionally chose a hacky way to force us to think about it more carefully and change it later

PR #1570, opened 2025-04-11
feat(sdk)!: configurable aggregation tree shape in SDK

jonathanpwang:
Introduce `AggregationTreeConfig` struct for simple configuration of aggregation tree.

These were already in the provers, but not exposed to be changeable. 

Breaking API changes in SDK:
- `AggStarkProver::new`
- `ContinuationProver::new`
- `Sdk` now internally contains `AggregationTreeConfig`, although `Sdk::new()` is unchanged and just sets a default

Towards INT-3302

### DIFF
--- a/crates/sdk/src/prover/mod.rs
+++ b/crates/sdk/src/prover/mod.rs
@@ -3,7 +3,9 @@ use std::sync::Arc;
 use openvm_circuit::arch::VmConfig;
 use openvm_stark_sdk::{engine::StarkFriEngine, openvm_stark_backend::Chip};
 
-use crate::{keygen::AppProvingKey, stdin::StdIn, NonRootCommittedExe, F, SC};
+use crate::{
+    config::AggregationTreeConfig, keygen::AppProvingKey, stdin::StdIn, NonRootCommittedExe, F, SC,
+};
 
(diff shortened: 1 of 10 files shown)


crates/sdk/src/prover/mod.rs
```diff
@@ -25,8 +27,8 @@ pub use stark::*;
 use crate::{keygen::AggProvingKey, prover::halo2::Halo2Prover, types::EvmProof};
 
 pub struct ContinuationProver<VC, E: StarkFriEngine<SC>> {
-    stark_prover: StarkProver<VC, E>,
-    halo2_prover: Halo2Prover,
+    pub stark_prover: StarkProver<VC, E>,
```
nyunyunyunyu, line 30: Why do we want to expose this?

nyunyunyunyu (approved):

crates/sdk/src/prover/mod.rs
jonathanpwang: you gave no way to modify any parameters of `StarkProver`, so if you create `ContinuationProver::new(..)`, there is no way to modify the internal `stark_prover` config in any way

PR #1571, opened 2025-04-11
fix: circuit flamegraphs with segmentation

Avaneesh-axiom:
The circuit flamegraphs being generated were incorrect due to a bug in the multi-segment execution logic. In particular, each segment takes the metric data from the previous segment, but the first segment was using an empty metrics struct instead of being initialized with the correct `fn_bounds` data.


jonathanpwang (approved):
Not sure why lint is failing

jonathanpwang:
@Avaneesh-axiom this was supposed to merge into `main`

PR #1572, opened 2025-04-12
chore: feature gate `revm` and `alloy` imports

jonathanpwang:
Previously we were very sloppy with enabling of features: `openvm-native-recursion` was always enabling the `snark-verifier` feature "revm" which imports a fixed version of revm only used for testing purposes. Since revm version changes rather frequently, this can cause Cargo dependency conflicts if someone wants to use the SDK with another version of revm.

I've added new features "evm-prove" and "evm-verify" to `openvm-native-recursion`, `openvm-sdk`, `cargo-openvm` so that 
- "evm-prove" allows for generation of EVM halo2 proofs, which just enables some additional features in `snark-verifier` but no additional other dependencies.
- "evm-verify" allows for revm simulation of EVM contract calls to verify halo2 proofs. This includes imports of `revm` and some `alloy` crates, with fixed versions. If you enable this feature (on by default), the revm and alloy versions will be fixed in the cargo tree.

Also added some re-exported features to SDK and CLI, and switched default memory allocator to `jemalloc` since that's what we mostly use.

### DIFF
--- a/crates/sdk/src/prover/mod.rs
+++ b/crates/sdk/src/prover/mod.rs
@@ -7,16 +7,16 @@ mod stark;
 pub mod vm;
 
 pub use agg::*;
-#[cfg(feature = "evm-prove")]
-pub use api::*;
 pub use app::*;
+#[cfg(feature = "evm-prove")]
+pub use evm::*;
 #[cfg(feature = "evm-prove")]
 pub use halo2::*;
 pub use root::*;
 pub use stark::*;
 
 #[cfg(feature = "evm-prove")]
-mod api {
+mod evm {
     use std::sync::Arc;
 
     use openvm_circuit::arch::VmConfig;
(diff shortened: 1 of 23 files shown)


jonathanpwang:
I renamed `ContinuationProver` to `EvmHalo2Prover` because the struct is more specialized than just for continuations: you can argue `StarkProver` itself is already a continuation prover. And this struct is only used if you want both EVM-verifiable proof and also using Halo2.

crates/sdk/src/prover/mod.rs
```diff
@@ -1,71 +1,78 @@
...
 mod halo2;
-#[allow(unused_imports)]
-pub use halo2::*;
 mod root;
-pub use root::*;
 mod stark;
 pub mod vm;
 
-#[allow(unused_imports)]
+pub use agg::*;
```
HrikB, line 9: Does `agg::*` need to be behind the `evm-prove` feature flag?

HrikB (approved):

crates/sdk/src/prover/mod.rs
jonathanpwang: `agg` includes internal verifiers so I don't think so

PR #1574, opened 2025-04-13
feat: `GuestMemory` trait

jonathanpwang:
Not merging to main

Add `GuestMemory` trait and implement for `AddressMap`. We are moving more towards a trait based style to re-use code when different types of memory might be swapped out.


jonathanpwang:
@shuklaayush fyi I went ahead and did this since it was easy and the `AddressMap` api will be more stable afterwards

PR #1577, opened 2025-04-14
feat: metered execution

shuklaayush:


crates/vm/src/arch/segment_new.rs
```diff
@@ -0,0 +1,442 @@
...
+/// Prepared instruction type for efficient execution
+pub type PInstruction = Instruction<u32>;
+
+/// Represents the state of the VM during execution
+pub struct VmExecutionState<Mem, Ctx>
+where
+    Mem: GuestMemory,
+{
+    /// Current timestamp representing clock cycles
+    pub timestamp: u32,
```
jonathanpwang, line 43: if this refers to clock cycles, let's name it something other than `timestamp` since timestamp refers to a specific thing in most of the system

jonathanpwang: should this be in `Ctx`? not sure

PR #1582, opened 2025-04-15
feat(new-execution): make vm segment executor generic

shuklaayush:
- make `VmSegmentExecutor` generic on `<Mem, Ctx, Ctrl>` where:
  - `Mem`: struct that implements `GuestMemory`
  - `Ctx`: struct that stores host context during execution
  - `Ctrl`: struct that implements pre/post segment execution hooks, termination condition and instruction execution logic
- add `TracegenVmSegmentExecutor` that implements the current execution flow
- move segmentation strategies to new file


crates/vm/src/arch/execution_control.rs
```diff
@@ -0,0 +1,169 @@
...
+const SEGMENT_CHECK_INTERVAL: usize = 100;
+
+/// Trait for execution control, determining segmentation and stopping conditions
+pub trait ExecutionControl<F, VC>
+where
+    F: PrimeField32,
+    VC: VmConfig<F>,
+{
+    /// Guest memory type
+    type Mem: GuestMemory;
```
jonathanpwang, line 20: what's the reasoning to have these as associated types?

crates/vm/src/arch/execution_control.rs
```diff
@@ -0,0 +1,169 @@
...
+    where
+        F: PrimeField32,
+    {
+        let timestamp = chip_complex.memory_controller().timestamp();
+        let (instruction, _) = chip_complex
+            .base
+            .program_chip
+            .get_instruction(vm_state.pc)?;
+
+        let &Instruction { opcode, .. } = instruction;
```
jonathanpwang, line 150: I did not know you could do `&Instruction`!

crates/vm/src/arch/segment.rs
```diff
@@ -157,12 +79,42 @@ where
...
+
+impl TracegenVmExecutionState {
+    pub fn from_pc_and_memory_controller<F>(
+        pc: u32,
+        memory_controller: &MemoryController<F>,
+    ) -> Self
+    where
+        F: PrimeField32,
+    {
+        let memory = memory_controller.memory_image().clone();
```
jonathanpwang, line 102: this is pretty bad to clone? is this temporary?

crates/vm/src/arch/segment.rs
```diff
@@ -211,133 +163,121 @@ impl<F: PrimeField32, VC: VmConfig<F>> ExecutionSegment<F, VC> {
             .set_override_inventory_trace_heights(overridden_heights.inventory);
     }
 
-    /// Stopping is triggered by should_segment()
-    pub fn execute_from_pc(
+    /// Stopping is triggered by should_stop() or if VM is terminated
+    pub fn execute_from_state(
```
jonathanpwang, line 167: this function is very clean

crates/vm/src/arch/vm.rs
```diff
@@ -473,7 +486,13 @@ where
         if let Some(overridden_heights) = self.overridden_heights.as_ref() {
             segment.set_override_trace_heights(overridden_heights.clone());
         }
-        metrics_span("execute_time_ms", || segment.execute_from_pc(pc_start))?;
+        let mut vm_state = TracegenVmExecutionState::from_pc_and_memory_controller(
```
jonathanpwang, line 489: this flow seems a bit weird - ok if it's just temporary until we get rid of or refactor memory controller

crates/vm/src/arch/segment.rs
```diff
@@ -157,12 +79,42 @@ where
     pub metrics: VmMetrics,
 }
 
-pub struct ExecutionSegmentState {
-    pub pc: u32,
-    pub is_terminated: bool,
+pub struct TracegenCtx {
+    pub timestamp: u32,
```
jonathanpwang, line 83: fyi: timestamp seems like it will need to be inside `TracingMemory` later since that is what really manages it most of time, but I expect we move `Streams` inside `Ctx` eventually.

jonathanpwang (approved):
LGTM, except somehow the current flow of using/constructing `VmExecutionState` seems not quite right. But that seems like it will change anyways with future refactors.

crates/vm/src/arch/execution_control.rs
shuklaayush: we can make them generic too. the reasoning to make these associated types is because i figured that the struct implementing the trait will usually implement it with a specific `Mem` and `Ctx` so the trait would never have to be implemented more than once for a struct and doesn't need to generic

crates/vm/src/arch/segment.rs
shuklaayush: yep, temporary. i need to figure out how to get memory here since it is owned by the memory controller atm

crates/vm/src/arch/vm.rs
shuklaayush: yeah, it's temporary until i figure out how to get rid of memory controller owning the memory and timestamp

crates/vm/src/arch/segment.rs
shuklaayush: okay. i was wondering if it was possible to keep all execution state here and make all the controllers etc. stateless, and take a reference to this vm state as argument

PR #1588, opened 2025-04-16
feat!: rewrite Merkle and boundary chips using a new data structure

Golovanov399:
Created a data structure for merkle and boundary chips (currently in `merkle/tree.rs`), make it so their `finalize` accepts only this structure and touched memory, not the final memory. This is towards INT-3801.

- [ ] Missing tests
- [ ] Missing parallelization
- [ ] Need to look if we can get rid of anything in the controller -- probably keep some redundant information now

The main logic is in the `update` function in the Merkle tree. It serves both for filling this structure initially with memory and for finalization. It has an `UPDATE_LEAVES` const generic (to save runtime) to set if we want to update the values in the leaves -- we always want to do this in the very beginning, and in the scenario where we maintain one copy of this structure throughout all execution (or a very finite number of copies) and **do not** reset it every segment, we also want to update the leaves always, otherwise there is no need to do this at the end of the segment right before dropping this data structure.

So to clarify, these are the usages:
- If we do all segments simultaneously
  1. At the beginning of the segment we call `tree.from_memory(current_memory)`, which calls `update::<true, _>` and fills the leaves.
  2. When we call `controller.finalize`, the boundary chip uses these initial leaf values in its routine. However, there is no need to insert into this leaf map again, so it calls `update::<false, _>`.
- If we do them serially
  1. We create this data structure at the beginning of execution and call `tree.from_memory(initial_memory)`, which calls `update::<true, _>`.
  2. At the end of the segment, we call `controller.finalize`, which does `update::<true, _>`.
  3. Then we use the same instance for the next segment, and then for the next, etc.

Currently the first thing is actually implemented, but all the functionality for the second one is there as well.

No matter the case, on creating this structure we don't need to record anything in the hasher or fill any merkle trace, but we do when `finalize`. Therefore, there is a new auxiliary `enum` that is either `&impl HasherChip`, or `(&mut impl HasherChip, &mut Vec<MerkleColumns>)`.


crates/vm/src/system/memory/merkle/tree.rs
```diff
@@ -0,0 +1,156 @@
...
+        hasher: &mut impl HasherChip<CHUNK, F>,
+        touched: &Equipartition<F, CHUNK>,
+        md: &MemoryDimensions,
+    ) -> FinalState<CHUNK, F> {
+        let mut layer: Vec<_> = touched
+            .iter()
+            .map(|(k, v)| ((1 << self.height) + md.label_to_index(*k), *v))
+            .collect();
+        for (k, v) in touched.iter() {
+            self.leaf_values.insert(*k, *v);
```
jonathanpwang, line 63: is there some way we could literally just set `leaf_values = touched`? inserting needs like wasted time (shouldn't we just use the same data structure for both)

crates/vm/src/system/memory/merkle/tree.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+        let init_root = self.get_node(1);
+        let mut rows = Vec::new();
```
jonathanpwang, line 66: avoid `vec.push`! how can we do a version of this where you allocate a buffer only once and never memory allocate again?

jonathanpwang (commented):
as we discussed offline, I think you need to support starting from an initial state

also is this algorithm parallelizable?

and at least one test to demonstrate it works please

crates/vm/src/system/memory/merkle/tree.rs
Golovanov399: No! `leaf_values` may not be empty in the beginning.
But there is no point in changing the leaf values if we intend to rebuild the tree every segment

crates/vm/src/system/memory/merkle/tree.rs
Golovanov399: called `Vec::with_capacity` for now

crates/vm/src/system/memory/merkle/tree.rs
jonathanpwang: oh I think I misunderstood what was going on

crates/vm/src/system/memory/controller/dimensions.rs
```diff
@@ -17,6 +17,14 @@ pub struct MemoryDimensions {
 }
 
 impl MemoryDimensions {
+    pub fn from_config(config: &MemoryConfig) -> Self {
+        Self {
+            as_height: config.as_height,
+            address_height: config.pointer_max_bits - log2_strict_usize(CHUNK),
```
jonathanpwang, line 23: is this correct? is `MemoryDimensions` just a property of the tree? if so, maybe we should move it to `tree.rs`

PR #1590, opened 2025-04-17
feat(new-execution): add execute functions for rv32im instructions

shuklaayush:
- introduce a new generic `InsExecutorE1` trait 
- add `InsExecutor::execute_e1` for rv32im instructions

### DIFF
--- a/crates/vm/src/arch/integration_api.rs
+++ b/crates/vm/src/arch/integration_api.rs
@@ -22,7 +22,7 @@ use openvm_stark_backend::{
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use super::{
-    ExecutionError, ExecutionState, InsExecutor, InstructionExecutor, Result, VmExecutionState,
+    ExecutionError, ExecutionState, InsExecutorE1, InstructionExecutor, Result, VmExecutionState,
     VmStateMut,
 };
 use crate::system::memory::{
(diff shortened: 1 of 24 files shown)


crates/vm/src/arch/integration_api.rs
```diff
@@ -133,6 +136,14 @@ pub trait VmCoreChip<F, I: VmAdapterInterface<F>> {
         reads: I::Reads,
     ) -> Result<(AdapterRuntimeContext<F, I>, Self::Record)>;
 
+    fn execute_instruction2<Mem, Ctx>(
```
jonathanpwang, line 139: depends how we want to handle this wrapper pattern, indeed you need something like this similar to `SingleTraceStep`. it might be possible to just directly implement `InsExecutor` on the wrapper struct? anyways we can keep this for now if it make life easier

crates/vm/src/arch/integration_api.rs
```diff
@@ -133,6 +136,14 @@ pub trait VmCoreChip<F, I: VmAdapterInterface<F>> {
...
+        &mut self,
+        state: &mut VmExecutionState<Mem, Ctx>,
+        instruction: &Instruction<F>,
+    ) -> Result<()>
+    where
+        Mem: GuestMemory;
```
jonathanpwang, line 145: just to save code, you could make the `todo!` the default impl

shuklaayush: yep, but having them separate makes it easy to search and resolve them one by one

jonathanpwang (approved):
we can merge after moving `execute_instruction2` to separate trait

extensions/rv32im/circuit/src/divrem/core.rs
```diff
@@ -517,6 +522,57 @@ where
...
+        } = *instruction;
+
+        // Determine opcode and operation type
+        let divrem_opcode = DivRemOpcode::from_usize(opcode.local_opcode_idx(self.air.offset));
+
+        // Read input registers
+        let rs1_addr = b.as_canonical_u32();
+        let rs2_addr = c.as_canonical_u32();
+
+        let rs1_bytes: [u8; NUM_LIMBS] = unsafe { state.memory.read(RV32_REGISTER_AS, rs1_addr) };
```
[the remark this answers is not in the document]
shuklaayush: are you saying that we need to check alignment here before each read?

crates/vm/src/arch/integration_api.rs
shuklaayush: i moved the execute functions into a new trait for now since we're deleting `VmCoreChip`

PR #1592, opened 2025-04-18
fix: Reduce size of `FnBounds`

manh9203:
- Extract guest symbol data into a string table and store it in `guest.syms` to reduce the size of metrics.json
- Update `flamegraph.py` to load guest symbols from `guest.syms`
- Fix stack trace unwinding in `update_current_fn()`
- Update `openvm-prof` to use `mmap` for faster reading of large JSON files

### DIFF
--- a/crates/prof/src/lib.rs
+++ b/crates/prof/src/lib.rs
@@ -4,6 +4,7 @@ use aggregate::{
     EXECUTE_TIME_LABEL, PROOF_TIME_LABEL, PROVE_EXCL_TRACE_TIME_LABEL, TRACE_GEN_TIME_LABEL,
 };
 use eyre::Result;
+use memmap2::Mmap;
 
 use crate::types::{Labels, Metric, MetricDb, MetricsFile};
 
@@ -14,7 +15,8 @@ pub mod types;
 impl MetricDb {
     pub fn new(metrics_file: impl AsRef<Path>) -> Result<Self> {
         let file = File::open(metrics_file)?;
-        let metrics: MetricsFile = serde_json::from_reader(file)?;
+        let mmap = unsafe { Mmap::map(&file)? };
+        let metrics: MetricsFile = serde_json::from_slice(&mmap)?;
 
         let mut db = MetricDb::default();
 
--- a/crates/toolchain/instructions/src/exe.rs
+++ b/crates/toolchain/instructions/src/exe.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, sync::Arc};
 
 use openvm_stark_backend::p3_field::Field;
 use serde::{Deserialize, Serialize};
@@ -24,7 +24,7 @@ pub struct VmExe<F> {
     /// Initial memory image.
     pub init_memory: MemoryImage<F>,
     /// Starting + ending bounds for each function.
-    pub fn_bounds: FnBounds,
+    pub fn_bounds: Arc<FnBounds>,
 }
 
 impl<F> VmExe<F> {
@@ -33,7 +33,7 @@ impl<F> VmExe<F> {
             program,
             pc_start: 0,
             init_memory: BTreeMap::new(),
-            fn_bounds: Default::default(),
+            fn_bounds: Arc::new(Default::default()),
         }
     }
     pub fn with_pc_start(mut self, pc_start: u32) -> Self {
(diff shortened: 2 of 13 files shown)


crates/prof/src/lib.rs
```diff
@@ -14,7 +15,8 @@ pub mod types;
 impl MetricDb {
     pub fn new(metrics_file: impl AsRef<Path>) -> Result<Self> {
         let file = File::open(metrics_file)?;
-        let metrics: MetricsFile = serde_json::from_reader(file)?;
+        let mmap = unsafe { Mmap::map(&file)? };
+        let metrics: MetricsFile = serde_json::from_slice(&mmap)?;
```
jonathanpwang, line 19: amazing.
aside: I wonder if just using a buffered reader would have gotten most of the gains, or if `mmap` is really that much better (you don't need to find out)

crates/toolchain/transpiler/Cargo.toml
```diff
@@ -15,6 +15,7 @@ eyre.workspace = true
 thiserror.workspace = true
 elf = "0.7.4"
 rrs-lib.workspace = true
+rustc-demangle = "0.1.24"
```
jonathanpwang, line 18: if this is only used by the "function-span" feature, can you make it optional and enable it only when "function-span" is enabled? just to prevent dependency issues for a non-core functionality

crates/toolchain/transpiler/src/elf.rs
```diff
@@ -93,18 +98,45 @@ impl Elf {
...
                                 end: (symbol.st_value + symbol.st_size - (WORD_SIZE as u64)) as u32,
-                                name: stringtab.get(symbol.st_name as usize).unwrap().to_string(),
+                                name: offsets[&symbol.st_name].to_string(),
                             },
                         );
                     }
                 }
+
+                let guest_symbols_file = std::env::var("GUEST_SYMBOLS_PATH")
+                    .map(|path| std::fs::File::create(path).unwrap());
```
jonathanpwang, line 136: let's not unwrap and propagate the error; also make sure the error displays the path if case it fails to create (this can happen eg if a directory doesn't exist)

crates/toolchain/transpiler/src/elf.rs
```diff
@@ -93,18 +98,45 @@ impl Elf {
...
+                if let Ok(mut guest_symbols_file) = guest_symbols_file {
+                    guest_symbols_file.write_all(buf.as_slice()).unwrap();
```
jonathanpwang, line 138: use `?`

crates/vm/src/metrics/mod.rs
```diff
@@ -116,10 +119,16 @@ impl VmMetrics {
             if pc == self.current_fn.start {
                 self.cycle_tracker.start(self.current_fn.name.clone());
             } else {
-                self.cycle_tracker.force_end();
+                while let Some(name) = self.cycle_tracker.top() {
```
jonathanpwang, line 122: was this the bug you fixed?

crates/toolchain/instructions/src/exe.rs
```diff
@@ -24,7 +24,7 @@ pub struct VmExe<F> {
     /// Initial memory image.
     pub init_memory: MemoryImage<F>,
     /// Starting + ending bounds for each function.
-    pub fn_bounds: FnBounds,
+    pub fn_bounds: Arc<FnBounds>,
```
jonathanpwang, line 27: does Serialize and Deserialize work with Arc? I guess it must otherwise the compiler would complain.

I'm also curious: does this actually make a difference in memory usage? I assume you put the `Arc` because we do `fn_bound.clone()` some places, but I thought it would mostly get compiled away

jonathanpwang (approved):
LGTM after improving the error handling

crates/prof/src/lib.rs
manh9203: I think `mmap` is still better, since JSON parsing tends to involve random data access

crates/vm/src/metrics/mod.rs
manh9203: yes we didn’t properly pop the stack trace

crates/toolchain/instructions/src/exe.rs
manh9203: The compiler doesn’t complain because `stark-backed` already imports `serde` with the "rc" feature, which `serde` needs to work with `Arc`.
I re‑ran the benchmark without `Arc`, and performance was identical, so I’ll leave it as before since `fn_bounds` is cloned at most twice—and `Arc` itself has non‑trivial overhead.

PR #1597, opened 2025-04-23
feat: Add e2e stark proof support

nyunyunyunyu:
- Add new type `E2eStarkProof` as the format of the final internal proof. `E2eStarkProof` supports `Encode` so it can be deserialized in other languages.
- Add function `wrap_e2e_stark_proof` to wrap the final internal proof to `RootVerifierInput`.
- Add `cargo openvm prove stark` to generate e2e stark proof.

closes INT-3784

### DIFF
--- a/crates/continuations/src/verifier/internal/types.rs
+++ b/crates/continuations/src/verifier/internal/types.rs
@@ -30,6 +30,36 @@ pub struct InternalVmVerifierInput<SC: StarkGenericConfig> {
 }
 assert_impl_all!(InternalVmVerifierInput<BabyBearPoseidon2Config>: Serialize, DeserializeOwned);
 
+/// The final output of the internal VM verifier.
+#[derive(Deserialize, Serialize, Derivative)]
+#[serde(bound = "")]
+#[derivative(Clone(bound = "Com<SC>: Clone"))]
+pub struct E2eStarkProof<SC: StarkGenericConfig> {
+    pub proof: Proof<SC>,
+    pub user_public_values: Vec<Val<SC>>,
+}
+assert_impl_all!(E2eStarkProof<BabyBearPoseidon2Config>: Serialize, DeserializeOwned);
+
+/// Aggregated state of all segments
+#[derive(Debug, Clone, Copy, AlignedBorrow)]
+#[repr(C)]
+pub struct InternalVmVerifierPvs<T> {
+    pub vm_verifier_pvs: VmVerifierPvs<T>,
+    pub extra_pvs: InternalVmVerifierExtraPvs<T>,
+}
+
+/// Extra PVs for internal VM verifier except VmVerifierPvs.
+#[derive(Debug, Clone, Copy, AlignedBorrow)]
+#[repr(C)]
+pub struct InternalVmVerifierExtraPvs<T> {
+    /// The commitment of the leaf verifier program.
+    pub leaf_verifier_commit: [T; DIGEST_SIZE],
+    /// For recursion verification, a program need its own commitment, but its own commitment
+    /// cannot be hardcoded inside the program itself. So the commitment has to be read from
+    /// external and be committed.
+    pub internal_program_commit: [T; DIGEST_SIZE],
+}
+
 impl InternalVmVerifierInput<SC> {
     pub fn chunk_leaf_or_internal_proofs(
         self_program_commit: [Val<SC>; DIGEST_SIZE],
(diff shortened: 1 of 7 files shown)


jonathanpwang:
@nyunyunyunyu I made a new feature branch `feat/verify_e2e_stark` we can merge your upcoming PRs into

crates/continuations/src/verifier/internal/types.rs
```diff
@@ -30,6 +30,36 @@ pub struct InternalVmVerifierInput<SC: StarkGenericConfig> {
 }
 assert_impl_all!(InternalVmVerifierInput<BabyBearPoseidon2Config>: Serialize, DeserializeOwned);
 
+/// The final output of the internal VM verifier.
+#[derive(Deserialize, Serialize, Derivative)]
+#[serde(bound = "")]
+#[derivative(Clone(bound = "Com<SC>: Clone"))]
+pub struct E2eStarkProof<SC: StarkGenericConfig> {
```
jonathanpwang, line 37: We should have a better name for this without `E2e` in it

crates/continuations/src/verifier/internal/types.rs
```diff
@@ -30,6 +30,36 @@ pub struct InternalVmVerifierInput<SC: StarkGenericConfig> {
...
+    pub proof: Proof<SC>,
+    pub user_public_values: Vec<Val<SC>>,
+}
+assert_impl_all!(E2eStarkProof<BabyBearPoseidon2Config>: Serialize, DeserializeOwned);
+
+/// Aggregated state of all segments
+#[derive(Debug, Clone, Copy, AlignedBorrow)]
+#[repr(C)]
+pub struct InternalVmVerifierPvs<T> {
```
jonathanpwang, line 46: nit: could you not move the locations of these, just to minimize the code diff?

crates/continuations/src/verifier/internal/types.rs
jonathanpwang: How about `VmStarkProof`? 
we can rename `ContinuationVmProof` to something else if we want to further avoid confusion

nyunyunyunyu: Sounds good. I can do it in the next PR. Do we want `e2e` in SDK?

crates/continuations/src/verifier/internal/types.rs
nyunyunyunyu: It has been merged. I guess we can keep it?

jonathanpwang: if you can undo it in a later PR, I think it may still make the tracking PR for feature branch a bit cleaner?

crates/continuations/src/verifier/internal/types.rs
jonathanpwang: I'm more ok with `e2e` in SDK, although just `generate_stark_proof` would sound a bit nicer.. maybe we go with the latter if we document so it's clear that `app = multiple proofs without agg, stark = e2e single proof`?

crates/sdk/src/lib.rs
```diff
@@ -268,6 +270,23 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
         Ok(proof)
     }
 
+    pub fn generate_e2e_stark_proof<VC: VmConfig<F>>(
+        &self,
+        app_pk: Arc<AppProvingKey<VC>>,
+        app_exe: Arc<NonRootCommittedExe>,
+        agg_stark_pk: AggStarkProvingKey,
```
jonathanpwang, line 277: we should update `agg_keygen`: there should be one version that only generates `AggStarkProvingKey` and a separate one that also does `halo2`. we can discuss naming offline

jonathanpwang: similarly we should polish the CLI flow for e2e stark without halo2

crates/sdk/src/lib.rs
```diff
@@ -268,6 +270,23 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
         Ok(proof)
```
jonathanpwang, line 270: I think we can just delete `generate_root_verifier_input`?

crates/sdk/src/prover/agg.rs
```diff
@@ -83,50 +85,30 @@ impl<E: StarkFriEngine<SC>> AggStarkProver<E> {
         self.generate_root_proof_impl(root_verifier_input)
     }
 
+    pub fn generate_leaf_proofs(&self, app_proofs: &ContinuationVmProof<SC>) -> Vec<Proof<SC>> {
+        self.leaf_controller
+            .generate_proof(&self.leaf_prover, app_proofs)
+    }
+
     pub fn generate_root_verifier_input(
```
jonathanpwang, line 93: ah I see the distinction is this still needs to wrap

crates/sdk/src/lib.rs
jonathanpwang: ok I see it's technically different from `generate_e2e` due to the wrapping. But you would only wrap if you want to do the root proof, so I still say we delete it from SDK

jonathanpwang (commented):
Approved but with comments to be addressed in follow-up PR

PR #1598, opened 2025-04-24
fix: Micro benchmarks

manh9203:
Fix microbenchmarks after circuit flamegraph changes

### DIFF
--- a/ci/scripts/utils.sh
+++ b/ci/scripts/utils.sh
@@ -18,12 +18,6 @@ add_metadata_and_flamegraphs() {
       }')
     echo "inputs: $inputs"
 
-    if [[ "$FLAMEGRAPHS" == 'true' ]]; then
-      repo_root=$(git rev-parse --show-toplevel)
-      python3 ${repo_root}/ci/scripts/metric_unify/flamegraph.py $metric_path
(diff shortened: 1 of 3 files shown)


jonathanpwang (approved):

ci/scripts/utils.sh
```diff
@@ -57,7 +51,7 @@ add_metadata() {
         repo_root=$(git rev-parse --show-toplevel)
         for file in $repo_root/.bench_metrics/flamegraphs/*.svg; do
         filename=$(basename "$file")
-            flamegraph_url=https://openvm-public-data-sandbox-us-east-1.s3.us-east-1.amazonaws.com/benchmark/github/flamegraphs/${CURRENT_SHA}/${filename}
+            flamegraph_url=${S3_FLAMEGRAPHS_PATH}/${CURRENT_SHA}/${filename}
```
jonathanpwang, line 54: not sure it's old but this is giving 's3://openvm-public-data-sandbox-us-east-1/benchmark/github/flamegraphs/1b4b05afb460e753565c358ee8136706415fc4e3/*.svg'

manh9203: you're right—the new url isn't working. I'll change it back.

PR #1601, opened 2025-04-24
feat: access adapters

Golovanov399:


crates/vm/src/system/memory/online/access_adapter.rs
```diff
@@ -0,0 +1,453 @@
...
+
+use crate::system::memory::{
+    offline_checker::{MemoryBus, AUX_LEN},
+    MemoryAddress,
+};
+
+#[repr(C)]
+#[derive(Debug, Clone, AlignedBorrow)]
+/// The columns of the AccessAdapterAir, indicating one split or merge operation.
+pub(crate) struct AccessAdapterCols<T, const LEN: usize, const NUM_BLOCKS: usize> {
```
jonathanpwang, line 27: where was this struct before? if possible keep it where it was since it's part of the AIR

PR #1596, opened 2025-04-23
feat: Build script for calling init macros

Avaneesh-axiom:
Added a build script that automatically adds calls to `moduli_init, complex_init, sw_init` appropriately by reading the vm config.

PR #1603, opened 2025-04-25
feat: Lazily call setup function for moduli and curves on first use

Avaneesh-axiom:
We no longer need to manually call `setup_*` at the start of main for the moduli and curves. Instead, setup is automatically done on first use of each modulus or curve.

Depends on #1596


benchmarks/guest/kitchen-sink/src/main.rs
```diff
@@ -48,10 +48,8 @@ openvm_algebra_guest::complex_macros::complex_init! {
 }
 
 pub fn main() {
-    // Setup will materialize every chip
-    setup_all_moduli();
-    setup_all_complex_extensions();
-    setup_all_curves();
+    // TODO: Since we don't explicitly call setup functions anymore, we should rewrite this test
```
jonathanpwang, line 51: don't forget this TODO before we merge feature branch

docs/specs/RISCV.md
```diff
@@ -132,7 +132,9 @@ generates classes `Bls12381` and `Bn254` that represent the elements of the corr
 
 ### Field Arithmetic
 
-For each created modular class, one must call a corresponding `setup_*` function once at the beginning of the program. For example, for the structs above this would be `setup_0()` and `setup_1()`. This function generates the `setup` intrinsics which are distinguished by the `rs2` operand that specifies the chip this instruction is passed to..
+For each created modular class, one must call a corresponding `setup_*` function before using the intrinsics.
+For example, for the structs above this would be `setup_0()` and `setup_1()`. This function generates the `setup` intrinsics which are distinguished by the `rs2` operand that specifies the chip this instruction is passed to..
```
jonathanpwang, line 136:
```suggestion
For example, for the structs above this would be `setup_0()` and `setup_1()`. This function generates the `setup` intrinsics which are distinguished by the `rs2` operand that specifies the chip this instruction is passed to.
```

docs/specs/RISCV.md
```diff
@@ -132,7 +132,9 @@ generates classes `Bls12381` and `Bn254` that represent the elements of the corr
...
+For convenience, each modulus's `setup_*` function is automatically called on the first use of any of its intrinsics.
```
jonathanpwang, line 137:
```suggestion
For developer convenience, in the Rust function bindings for these intrinsics, each modulus's `setup_*` function is automatically called on the first use of any of its intrinsics.
```

extensions/algebra/complex-macros/README.md
```diff
@@ -79,13 +77,9 @@ mod openvm_intrinsics_ffi_complex {
...
+3. Obviously, `mod_idx` in the `complex_init!` must match the position of the corresponding modulus in the `moduli_init!` macro. The order of the items in `complex_init!` affects what `setup_complex_*` function will correspond to what complex class. Also, it **must match** the order of the moduli in the chip configuration -- more specifically, in the modular extension parameters (the order of numbers in `Fp2Extension::supported_modulus`, which is usually defined with the whole `app_vm_config` in the `openvm.toml` file). However, it again imposes the restriction that we only can invoke `complex_init!` once. Again analogous to the moduli setups, `setup_complex_*` is automatically called on each complex extension on first use of its intrinsics.
```
jonathanpwang, line 82:
```suggestion
3. Obviously, `mod_idx` in the `complex_init!` must match the position of the corresponding modulus in the `moduli_init!` macro. The order of the items in `complex_init!` affects what `setup_complex_*` function will correspond to what complex class. Also, it **must match** the order of the moduli in the chip configuration -- more specifically, in the modular extension parameters (the order of numbers in `Fp2Extension::supported_modulus`, which is usually defined with the whole `app_vm_config` in the `openvm.toml` file). However, it again imposes the restriction that we only can invoke `complex_init!` once. Again analogous to the moduli setups, the rust bindings will automatically call `setup_complex_*` on each complex extension on first use of its intrinsics.
```

extensions/algebra/complex-macros/src/lib.rs
```diff
@@ -110,6 +113,7 @@ pub fn complex_declare(input: TokenStream) -> TokenStream {
                     }
                     #[cfg(target_os = "zkvm")]
                     {
+                        Self::assert_is_setup();
```
jonathanpwang, line 116: side comment, technically shouldn't it be `assert_is_set_up`? not worth changing though

extensions/pairing/tests/programs/openvm_init_fp12_mul_bls12_381.rs
```diff
@@ -1,4 +1,4 @@
 // This file is automatically generated by cargo openvm. Do not rename or edit.
-openvm_algebra_guest::moduli_macros::moduli_init! { "4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559787" }
-openvm_algebra_guest::complex_macros::complex_init! { Bls12_381Fp2 { mod_idx = 0 } }
+openvm_algebra_guest::moduli_macros::moduli_init! { "21888242871839275222246405745257275088696311157297823662689037894645226208583" }
```
jonathanpwang, line 2: this is weird, why did this file change?

extensions/ecc/sw-macros/src/lib.rs
```diff
@@ -527,9 +541,12 @@ pub fn sw_init(input: TokenStream) -> TokenStream {
         let setup_function = syn::Ident::new(&format!("setup_sw_{}", str_path), span.into());
         setups.push(quote::quote_spanned! { span.into() =>
             #[allow(non_snake_case)]
-            pub fn #setup_function() {
+            #[no_mangle]
+            extern "C" fn #setup_function() {
                 #[cfg(target_os = "zkvm")]
                 {
+                    openvm::io::println("setup function called");
```
jonathanpwang, line 548:
```suggestion
```

Avoid noisy print statements in the released version

extensions/ecc/sw-macros/src/lib.rs
```diff
@@ -90,6 +90,8 @@ pub fn sw_declare(input: TokenStream) -> TokenStream {
         create_extern_func!(hint_decompress_extern_func);
         create_extern_func!(hint_non_qr_extern_func);
 
+        let setup_function = syn::Ident::new(&format!("setup_sw_{}", struct_name), span.into());
```
jonathanpwang, line 93: I am ever so slightly concerned `setup_sw_*` isn't unique enough for a no_mangle. I will rename it `setup_sw_extern_func` similar to the other extern functions

jonathanpwang: done in d109b3d0a75d7ce98c8eea1955d184d94eb594c5

extensions/ecc/sw-macros/src/lib.rs
```diff
@@ -98,6 +99,7 @@ pub fn sw_declare(input: TokenStream) -> TokenStream {
                 fn #sw_double_extern_func(rd: usize, rs1: usize);
                 fn #hint_decompress_extern_func(rs1: usize, rs2: usize);
                 fn #hint_non_qr_extern_func();
+                fn #setup_sw_extern_func();
```
jonathanpwang, line 102: noting this does setup for both add_ne and double_ne (this is the way it was originally), which makes sense because you would basically never use just one over the other

extensions/pairing/tests/programs/openvm_init_fp12_mul_bls12_381.rs
jonathanpwang: reverted it in 50bc7569716f6647550d386b11ad14af923845b0

jonathanpwang (approved):
LGTM, I cleaned up the macros a bit so they are safer for `no_mangle` and also conform to the other extern functions better.

extensions/pairing/tests/programs/openvm_init_fp12_mul_bls12_381.rs
Avaneesh-axiom: hm... not sure. looks like a mistake

PR #1604, opened 2025-04-25
refactor: Guest library re-organization (Part 1)

Avaneesh-axiom:
- Deleted the guest library methods for Sha2 and Keccak256 extensions (only bindings remain)
- Added a hint-based square root instruction to the algebra extension

The changes in this repo related to the guest library re-organization have been split into two PRs. This is the first one.

Note: some of the CI jobs are expected to fail since the re-org is not yet complete. For example, the ecdsa tests use keccak, but keccak has been deleted.

Depends on #1603


book/src/custom-extensions/algebra.md
```diff
@@ -26,11 +29,15 @@ To [leverage](./overview.md) compile-time known moduli for performance, you decl
 ```rust
 moduli_declare! {
     Bls12_381Fp { modulus = "0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab" },
-    Bn254Fp { modulus = "21888242871839275222246405745257275088696311157297823662689037894645226208583" },
+    Bn254Fp {
+        modulus = "21888242871839275222246405745257275088696311157297823662689037894645226208583", impl_field = true
```
jonathanpwang, line 33: I am not quite a fan of the `impl_field` field, though I could live with it if we needed it. But I think it might actually be easy inside the proc-macro to just check if the modulus as a big-int is prime, so `impl_field` could be inferred?

extensions/keccak256/guest/src/lib.rs
```diff
@@ -40,7 +20,7 @@ pub fn keccak256(input: &[u8]) -> [u8; 32] {
 #[cfg(target_os = "zkvm")]
 #[inline(always)]
 #[no_mangle]
-extern "C" fn native_keccak256(bytes: *const u8, len: usize, output: *mut u8) {
+pub extern "C" fn native_keccak256(bytes: *const u8, len: usize, output: *mut u8) {
```
jonathanpwang, line 23: is the `pub` needed? I just want to make sure it works with a non-`pub` extern: https://docs.rs/alloy-primitives/latest/src/alloy_primitives/utils/mod.rs.html#166

Avaneesh-axiom:
Replaced by #1613

jonathanpwang:
Assuming this was re-opened by accident

PR #1606, opened 2025-04-26
feat: Add Rv32HintLoadByKey

nyunyunyunyu:
- Add `kv_store` into `Streams`. More details at `docs/specs/ISA.md`.
- Add a new phantom instruction `Rv32HintLoadByKey` which can hint data based on a key at runtime. More details at `docs/specs/ISA.md`.
- SDK support will be added in the future PRs.

close INT-3893

### DIFF
--- a/docs/specs/ISA.md
+++ b/docs/specs/ISA.md
@@ -171,6 +171,9 @@ structures during runtime execution:
 - `hint_space`: a vector of vectors of field elements used to store hints during runtime execution
   via [phantom sub-instructions](#phantom-sub-instructions) such as `NativeHintLoad`. The outer `hint_space` vector is append-only, but
   each internal `hint_space[hint_id]` vector may be mutated, including deletions, by the host.
+- `kv_store`: a read-only key-value store for hints. Executors(e.g. `Rv32HintLoadByKey`) can read data from `kv_store` 
+  at runtime. `kv_store` is designed for general purposes so both key and value are byte arrays. Encoding of key/value
+  are decided by each executor. Users need to use the corresponding encoding when adding data to `kv_store`.
 
 These data structures are **not** part of the guest state, and their state depends on host behavior that cannot be determined by the guest.
 
@@ -204,7 +207,7 @@ which must satisfy the following conditions:
 - The execution has full read/write access to the data memory, except address space `0` must be read-only.
 - User public outputs can be set at any index in `[0, num_public_values)`. If continuations are disabled, a public
   value cannot be overwritten with a different value once it is set.
-- Input stream can only be popped from the front as a queue. Appends are not allowed.
+- Input stream can only be popped from the front as a queue.
 - Full read/write access to the hint stream.
 - Hint spaces can be read from at any index. Hint spaces may be mutated only by append.
 - The program counter is set to a new `to_pc` at the end of the instruction execution.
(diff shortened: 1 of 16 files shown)


docs/specs/ISA.md
```diff
@@ -204,7 +207,7 @@ which must satisfy the following conditions:
 - The execution has full read/write access to the data memory, except address space `0` must be read-only.
 - User public outputs can be set at any index in `[0, num_public_values)`. If continuations are disabled, a public
   value cannot be overwritten with a different value once it is set.
-- Input stream can only be popped from the front as a queue. Appends are not allowed.
```
jonathanpwang, line 207: why was the second line deleted? you are now appending?

docs/specs/ISA.md
```diff
@@ -426,12 +429,12 @@ with user input-output.
...
+| Rv32HintInput     | 0x20         | `_`      | Pops a vector `hint` of field elements from the input stream and resets the hint stream to equal the vector `[(hint.len() as u32).to_le_bytes()), hint].concat()`.                                           |
+| Rv32PrintStr      | 0x21         | `a,b,_`  | Peeks at `[r32{0}(a)..r32{0}(a) + r32{0}(b)]_2`, tries to convert to byte array and then UTF-8 string and prints to host stdout. Prints error message if conversion fails. Does not change any VM state.     |
+| Rv32HintRandom    | 0x22         | `a,_,_`  | Resets the hint stream to `4 * r32{0}(a)` random bytes. The source of randomness is the host operating system (`rand::rngs::OsRng`). Its result is not constrained in any way.                               |
+| Rv32HintLoadByKey | 0x23         | `a,b,_`  | Look up the value by key `[r32{0}{a}:r32{0}{b}]_2` and prepend the value into `input_stream`. Users should use `openvm-rv32im-guest::hint_load_by_key_encode` to encode the value when constructing inputs. |
```
jonathanpwang, line 437: `openvm_rv32im_guest::hint_load_by_key_encode`

jonathanpwang: Is there a way to describe the encoding in words without referring to the code? 
Here it's hard to tell if it is a single list of `F` or multiple lists

docs/specs/ISA.md
nyunyunyunyu: Right, `Rv32HintLoadByKey ` prepends `Vec<F>` in the front of `Input Stream`

extensions/rv32im/guest/src/lib.rs
```diff
@@ -28,4 +30,19 @@ pub enum PhantomImm {
     HintInput = 0,
     PrintStr,
     HintRandom,
+    HintLoadByKey,
+}
+
+/// Encode a 2d-array of field elements into bytes for `hint_load_by_key`
+#[cfg(not(target_os = "zkvm"))]
+pub fn hint_load_by_key_encode<F: p3_field::PrimeField32>(
```
jonathanpwang, line 38: could this just go in `rv32im-circuit`?

extensions/rv32im/tests/programs/examples/hint_load_by_key.rs
```diff
@@ -0,0 +1,32 @@
...
+use openvm::io::{hint_load_by_key, read_vec};
+
+openvm::entry!(main);
+
+pub fn main() {
+    const KEY: &str = "key";
+    hint_load_by_key(KEY.as_bytes());
+
+    let vec = read_vec();
+    // assert_eq!(vec.len(), 4);
```
jonathanpwang, line 12: did assert_eq not work?

docs/specs/ISA.md
nyunyunyunyu: ok

jonathanpwang (commented):
LGTM except some typos.

Noting that it would be nicer if we didn't need to use InputStream because you could go directly from KvStore -> HintStream, but because of how we implemented `Hintable` on all the structs, InputStream is "hardcoded" into the recursion programs so this is needed.

extensions/rv32im/guest/src/lib.rs
nyunyunyunyu: It's irrelevant to `rv32im-circuit`, which has complicated dependencies. How about a new crate `rv32im-host`?

extensions/rv32im/tests/programs/examples/hint_load_by_key.rs
nyunyunyunyu: This is literally copied from `extensions/rv32im/tests/programs/examples/hint.rs`. I can remove it

jonathanpwang: ah I see, yea `assert_eq` works now

extensions/rv32im/guest/src/lib.rs
jonathanpwang: mmm seems too much to make a new crate. technically `rv32im-transpiler` fits this purpose I think. if you feel transpiler isn't right, let's just leave in `guest`

PR #1611, opened 2025-04-29
fix(new-execution): fix some rv32im loadstore tests

shuklaayush:
- fix some loadstore tests
- remove records
- wrap unsafe memory read/writes into safe wrappers


crates/vm/src/system/memory/online.rs
```diff
@@ -145,7 +145,7 @@ impl TracingMemory {
             let num_cells = paged_vec.bytes_capacity() / cell_size;
 
             // TMP: hardcoding for now
-            if i <= 3 {
+            if i < 3 {
```
jonathanpwang, line 148: back to <= ?

shuklaayush: i think we only need to push three times for spaces 1, 2, 3 since the index 0 is already initialized

jonathanpwang: ah... we should really sort out this `as_offset` shift business later

PR #1613, opened 2025-04-30
refactor: guest bindings

Avaneesh-axiom:
This PR removes parts of the guest libraries for each extension. These libraries are to be moved into a new repo as part of the guest library re-org.

Summary of changes for each extension:
- algebra: minor changes
- ecc: the re-export of halo2curves was removed, also minor changes
- pairing: the traits, the `halo2curves_shims` module, along with a minimal set of constants remain in the openvm repo, while the bn254 and bls12_381 implementations were moved out
- bigint: the U256 and I256 structs were deleted. Only Rust bindings remain.
- sha256: the `sha256` and `set_sha256` functions were deleted, leaving behind only the Rust binding
- keccak256: the `keccak256` and `set_keccak256` functions were deleted, leaving behind only the Rust binding

Notes:
- the k256 and p256 modules of the ecc extension were not deleted as they are useful for testing the ecc macros and traits. For this reason, the ecc tests were also not deleted
- the algebra tests were also not deleted for the same reason
- the `mod_sqrt` implementation was moved from the ecc extension to the algebra extension

Benchmarks and examples are now outdated since they depend on guest libraries that no longer exist. These need to be redesigned and rewritten to use the new guest libraries. I wasn't able to easily port the benchmarks and examples because depending on the new guest libraries creates weird linker issues (I'm guessing due to double-importing openvm).

TODO:
- switch all deps to main after merge
- uncomment ecdsa test

Closes INT-3788


book/src/custom-extensions/algebra.md
```diff
@@ -26,11 +29,15 @@ To [leverage](./overview.md) compile-time known moduli for performance, you decl
 ```rust
 moduli_declare! {
     Bls12_381Fp { modulus = "0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab" },
-    Bn254Fp { modulus = "21888242871839275222246405745257275088696311157297823662689037894645226208583" },
+    Bn254Fp {
+        modulus = "21888242871839275222246405745257275088696311157297823662689037894645226208583", impl_field = true
```
jonathanpwang, line 33: see my comment in the old PR on `impl_field` auto detection

jonathanpwang (commented):
LGTM, but don't merge because we will merge the other PR into this one

benchmarks/guest/kitchen-sink/src/main.rs
```diff
@@ -23,33 +29,64 @@ openvm_algebra_guest::moduli_macros::moduli_declare! {
...
-    "0x30644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd47", // Bn254Fp Coordinate field
-    "0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001", // Bn254 Scalar
-    "0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab", // BLS12-381 Coordinate field
-    "0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001", // BLS12-381 Scalar field
-    "0x1fffffffffffffff",
-    "7",
+openvm::init!();
+
+fn materialize_modular_chip<T: IntMod>() {
+    // hopefully the compiler doesn't optimize out the operations
```
jonathanpwang, line 35:
```suggestion
    // ensure the compiler doesn't optimize out the operations
```

benchmarks/guest/kitchen-sink/src/main.rs
```diff
@@ -23,33 +29,64 @@ openvm_algebra_guest::moduli_macros::moduli_declare! {
...
+    black_box(T::ZERO.assert_reduced());
 }
 
-openvm_ecc_guest::sw_macros::sw_init! {
-    Secp256k1Point, P256Point,
-    Bn254G1Affine, Bls12_381G1Affine
+// making this a macro since there's no complex extension trait
+macro_rules! materialize_complex_chip {
+    ($complex_type:ty, $modular_type:ty) => {
+        // hopefully the compiler doesn't optimize out the operations
```
jonathanpwang, line 47:
```suggestion
        // ensure the compiler doesn't optimize out the operations
```

extensions/algebra/guest/src/halo2curves.rs
```diff
@@ -1,55 +1,76 @@
-use core::ops::{Add, Mul, Sub};
-
 use halo2curves_axiom::ff;
 
 use crate::{field::Field, DivAssignUnsafe, DivUnsafe};
 
-impl<'a, F: ff::Field> DivUnsafe<&'a F> for F {
```
jonathanpwang, line 7: it doesn't really matter, but what stopped you from doing a blanket implementation for all `ff::Field`?

extensions/bigint/guest/src/externs.rs
```diff
@@ -5,7 +5,7 @@ use openvm_platform::custom_insn_r;
 use super::{Int256Funct7, BEQ256_FUNCT3, INT256_FUNCT3, OPCODE};
 
 #[no_mangle]
-unsafe extern "C" fn zkvm_u256_wrapping_add_impl(result: *mut u8, a: *const u8, b: *const u8) {
+pub unsafe extern "C" fn zkvm_u256_wrapping_add_impl(result: *mut u8, a: *const u8, b: *const u8) {
```
jonathanpwang, line 8: I don't think the pub matters for extern

jonathanpwang:
It is weird to keep `k256.rs` in `ecc/guest` since there is a lot of code duplication.

guest-libs/ff_derive/src/lib.rs
```diff
@@ -0,0 +1,1686 @@
...
+            ReprEndianness::Little => quote! {self.0.iter().rev()},
+        }
+    }
+}
+
+/// Derive the `PrimeField` trait.
+/// This macro removes the struct definition and inserts our own zkvm-compatible struct into the
+/// token stream.
+// Required attributes: PrimeFieldModulus, PrimeFieldGenerator, PrimeFieldReprEndianness
+// Note: In our fork, we changed the macro from a derive macro to an attribute-style macro because
```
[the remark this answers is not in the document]
jonathanpwang: removed the internal field in https://github.com/openvm-org/openvm/pull/1613/commits/0201ee18ad38b242548b286be139ef826b181700

extensions/bigint/guest/src/externs.rs
jonathanpwang: nvm, in guest-lib you import directly

jonathanpwang (approved):
Two follow-ups, but let's merge this first:

- [ ] constant time algorithms (constant time doesn't really make sense in zkvm context I think, but let's revisit)
- [ ] redundant code for k256.rs

guest-libs/ff_derive/src/lib.rs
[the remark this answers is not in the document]
jonathanpwang: nevermind: this runs into issues in the non-zkvm case because we still use the `ff_derive` macro with monty reduction when target is not zkvm. It would be best if we could just always go through `openvm-algebra-guest` directly, but this would require more work in moving the Tonelli-Shanks sqrt implementation into a place where `Sqrt::sqrt` is properly implemented when `not(target_os = "zkvm"))`.

for now, I just reverted my attempt

PR #1614, opened 2025-04-30
feat: access adapters

Golovanov399:
closes INT-3839

### DIFF
--- a/crates/vm/src/system/memory/online.rs
+++ b/crates/vm/src/system/memory/online.rs
@@ -260,7 +260,6 @@ impl<F: PrimeField32> TracingMemory<F> {
         &mut self,
         address_space: usize,
         pointer: usize,
-        values: &[F],
     ) -> u32 {
         let size = size_of::<T>();
         let seg_size = ALIGN * size;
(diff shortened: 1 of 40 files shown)


crates/vm/src/system/memory/online.rs
```diff
@@ -179,6 +188,150 @@ impl TracingMemory {
...
+            }
+        }
+    }
+
+    /// Returns the timestamp of the previous access to `[pointer:BLOCK_SIZE]_{address_space}`.
+    /// If we need to split/merge/initialize something for this, we first do all the necessary
+    /// actions. In the end of this process, we have this segment intact in our `meta`.
+    ///
+    /// Caller must ensure alignment (e.g. via `assert_alignment`) prior to calling this function.
+    fn prev_access_time<T: Copy + Debug, const BLOCK_SIZE: usize, const ALIGN: usize>(
```
shuklaayush, line 259: looks like this isn't used anywhere. is that still a todo?

shuklaayush (commented):
a couple of minor comments. i'm not sure how to test these changes. if that's still wip and merging would unblock stuff, then i'm happy to approve and merge

crates/vm/src/system/memory/online.rs
Golovanov399: It will be used once I deal with the boundary stuff -- this function is actually what is calling all the necessary splits and merges

crates/vm/src/system/memory/online.rs
```diff
@@ -211,24 +364,32 @@ impl TracingMemory {
...
-            timestamp: t_prev,
-            mut block_size,
-        } = meta.replace(access_idx, &AccessMetadata::new(t_curr, BLOCK_SIZE as u32));
-        // TODO: do we need to handle uninitialized memory?
-        if block_size == 0 {
-            block_size = BLOCK_SIZE as u32;
-        }
-        if (block_size as usize) != BLOCK_SIZE {
-            todo!("split and merge stuff")
+        // TODO: this is wrong and must be replaced with normal logic
```
jonathanpwang, line 374: what does normal logic here mean?

crates/vm/src/system/memory/online.rs
```diff
@@ -211,24 +364,32 @@ impl TracingMemory {
...
-        let AccessMetadata {
-            timestamp: t_prev,
-            mut block_size,
-        } = meta.replace(access_idx, &AccessMetadata::new(t_curr, BLOCK_SIZE as u32));
-        // TODO: do we need to handle uninitialized memory?
-        if block_size == 0 {
-            block_size = BLOCK_SIZE as u32;
-        }
-        if (block_size as usize) != BLOCK_SIZE {
-            todo!("split and merge stuff")
```
jonathanpwang, line 230: what I was expecting is that this todo is resolved in this PR?

crates/vm/src/system/memory/online.rs
```diff
@@ -211,24 +364,32 @@ impl TracingMemory {
     {
         self.assert_alignment(BLOCK_SIZE, ALIGN, address_space, pointer);
         let values = self.data.read(address_space, pointer);
+        let values_f = (0..BLOCK_SIZE)
```
jonathanpwang, line 367: this is bad (for perf) -- why are you doing it?

PR #1615, opened 2025-04-30
feat: Root Verifier ASM

nyunyunyunyu:
- Add a function `build_kernel_asm` into `RootVmVerifierConfig`. The generated ASM can be used in kernel functions for guest program to verify starks. More details could be found in the doc comments of the function.
- Fix a confusing naming. `num_public_values` is actually the number of user public values instead of the number of public values of the proof. Use `num_user_public_values` instead.

closes INT-3894

### DIFF
--- a/crates/cli/src/commands/setup.rs
+++ b/crates/cli/src/commands/setup.rs
@@ -17,7 +17,9 @@ use openvm_sdk::{
     DefaultStaticVerifierPvHandler, Sdk,
 };
 
-use crate::default::{DEFAULT_AGG_PK_PATH, DEFAULT_EVM_HALO2_VERIFIER_PATH, DEFAULT_PARAMS_DIR};
+use crate::default::{
+    DEFAULT_AGG_PK_PATH, DEFAULT_ASM_PATH, DEFAULT_EVM_HALO2_VERIFIER_PATH, DEFAULT_PARAMS_DIR,
+};
 
 #[derive(Parser)]
 #[command(
@@ -56,11 +58,17 @@ impl EvmProvingSetupCmd {
         println!("Generating proving key...");
         let agg_pk = sdk.agg_keygen(agg_config, &params_reader, &DefaultStaticVerifierPvHandler)?;
 
+        println!("Generating root verifier ASM...");
+        let root_verifier_asm = sdk.generate_root_verifier_asm(&agg_pk.agg_stark_pk);
+
         println!("Generating verifier contract...");
         let verifier = sdk.generate_halo2_verifier_solidity(&params_reader, &agg_pk)?;
 
         println!("Writing proving key to file...");
         write_agg_pk_to_file(agg_pk, DEFAULT_AGG_PK_PATH)?;
+
+        println!("Writing root verifier ASM to file...");
+        write(DEFAULT_ASM_PATH, root_verifier_asm)?;
 
         println!("Writing verifier contract to file...");
         write_evm_halo2_verifier_to_folder(verifier, DEFAULT_EVM_HALO2_VERIFIER_PATH)?;
--- a/crates/continuations/src/verifier/internal/vars.rs
+++ b/crates/continuations/src/verifier/internal/vars.rs
@@ -5,8 +5,11 @@ use openvm_native_recursion::{hints::Hintable, vars::StarkProofVariable};
 use openvm_stark_sdk::openvm_stark_backend::proof::Proof;
 
 use crate::{
-    verifier::{internal::types::InternalVmVerifierInput, utils::write_field_slice},
-    C, SC,
+    verifier::{
+        internal::types::{E2eStarkProof, InternalVmVerifierInput},
+        utils::write_field_slice,
+    },
+    C, F, SC,
 };
 
 #[derive(DslVariable, Clone)]
@@ -14,6 +17,12 @@ pub struct InternalVmVerifierInputVariable<C: Config> {
     pub self_program_commit: [Felt<C::F>; DIGEST_SIZE],
     /// The proofs of the execution segments in the execution order.
     pub proofs: Array<C, StarkProofVariable<C>>,
+}
+
+#[derive(DslVariable, Clone)]
+pub struct E2eStarkProofVariable<C: Config> {
+    pub proof: StarkProofVariable<C>,
+    pub user_public_values: Array<C, Felt<C::F>>,
 }
 
 impl Hintable<C> for InternalVmVerifierInput<SC> {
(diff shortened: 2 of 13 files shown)


crates/cli/src/commands/setup.rs
```diff
@@ -56,12 +58,18 @@ impl EvmProvingSetupCmd {
         println!("Generating proving key...");
         let agg_pk = sdk.agg_keygen(agg_config, &params_reader, &DefaultStaticVerifierPvHandler)?;
 
+        println!("Generating root verifier ASM...");
+        let root_verifier_asm = sdk.generate_root_verifier_asm(&agg_pk.agg_stark_pk);
```
jonathanpwang, line 62: is this something we want to always generate? I guess as a one-time `setup` function, it makes sense

crates/continuations/src/verifier/internal/vars.rs
```diff
@@ -16,6 +19,12 @@ pub struct InternalVmVerifierInputVariable<C: Config> {
     pub proofs: Array<C, StarkProofVariable<C>>,
 }
 
+#[derive(DslVariable, Clone)]
+pub struct E2eStarkProofVariable<C: Config> {
```
jonathanpwang, line 23: match renaming `VmStarkProofVariable`

crates/continuations/src/verifier/root/mod.rs
```diff
@@ -40,73 +41,163 @@ impl RootVmVerifierConfig {
...
+        builder.assert_eq::<[Felt<_>; DIGEST_SIZE]>(pvs.exe_commit, expected_pvs.exe_commit);
+        builder.assert_eq::<[Felt<_>; DIGEST_SIZE]>(
+            pvs.leaf_verifier_commit,
+            expected_pvs.leaf_verifier_commit,
+        );
+
+        builder.compile_isa_with_options(self.compiler_options)
+    }
+
+    fn verifier_impl(
```
jonathanpwang, line 139: to trick the git diff, can you move this above `build_kernel_asm`?

crates/continuations/src/verifier/root/mod.rs
```diff
@@ -40,73 +41,163 @@ impl RootVmVerifierConfig {
...
+        pvs.flatten()
+            .into_iter()
+            .for_each(|v| builder.commit_public_value(v));
+        builder.halt();
+        builder.compile_isa_with_options(self.compiler_options)
+    }
+
+    /// Build instructions which can be called as a kernel function in RISC-V guest programs.
+    /// Inputs:
+    /// - expected public values are stored from `HEAP_START_ADDRESS` in the native address space in
```
jonathanpwang, line 64: the guest program is expected to put the public values here?

crates/continuations/src/verifier/root/mod.rs
```diff
@@ -40,73 +41,163 @@ impl RootVmVerifierConfig {
...
+                        + fs[1] * F::from_canonical_u32(1 << 8)
+                        + fs[2] * F::from_canonical_u32(1 << 16)
+                        + fs[3] * F::from_canonical_u32(1 << 24),
+                )
+            })
+            .collect();
+        let expected_pvs = RootVmVerifierPvs::<Felt<F>>::from_flatten(expected_pvs);
+        let user_pvs = builder.array(self.num_user_public_values);
+        for (i, &pv) in expected_pvs.public_values.iter().enumerate() {
+            builder.set(&user_pvs, i, pv);
```
jonathanpwang, line 113: is this just some `Vec` to `Array` conversion?

crates/sdk/src/keygen/asm.rs
```diff
@@ -0,0 +1,107 @@
...
+                asm_output.push_str(" // ");
+                asm_output.push_str(comment);
+            }
+            asm_output.push('\n');
+        }
+    }
+    asm_output
+}
+
+fn u32_to_directive(x: u32) -> String {
```
jonathanpwang, line 28: I feel like I saw a similar function somewhere else (maybe transpiler), could you check?

crates/sdk/src/keygen/asm.rs
```diff
@@ -0,0 +1,107 @@
...
+        simm12 -= 1 << 12;
+    }
+    format!(
+        ".insn i {}, {}, x{}, x{}, {}",
+        opcode, funct3, rd, rs1, simm12
+    )
+}
+
+/// In order to use native instructions in kernel functions, native instructions need to be
+/// converted to assembles(long form instructions) first. Then Rust compiler compiles the whole
```
jonathanpwang, line 44:
```suggestion
/// converted to RISC-V machine code (long form instructions) first. Then Rust compiler compiles the whole
```

crates/sdk/src/keygen/asm.rs
```diff
@@ -0,0 +1,107 @@
...
+    pc_diff: usize,
+) -> Vec<(Vec<u32>, String)> {
+    program
+        .defined_instructions()
+        .iter()
+        .map(|ins| {
+            (
+                vec![
+                    LONG_FORM_INSTRUCTION_INDICATOR,
+                    7,
```
jonathanpwang, line 89: i think we have a NUM_OPERANDS constant

crates/cli/src/commands/setup.rs
nyunyunyunyu: It always sticks with a specific `AggConfig` so it's ok to generate it together

crates/continuations/src/verifier/root/mod.rs
nyunyunyunyu: Maybe use https://platform.text.com/tools/diff-checker to compare? `verifier_impl` is private so I want to put it behind all private functions.

Previous impl: https://github.com/openvm-org/openvm/blob/6d019463b50dc315553aa81007dfc85a0c0736e4/crates/continuations/src/verifier/root/mod.rs#L43
Current impl: https://github.com/openvm-org/openvm/blob/c1829fc5d1a21d3111656ae128bc99c008ed70f2/crates/continuations/src/verifier/root/mod.rs#L146C9-L200C10

crates/continuations/src/verifier/root/mod.rs
nyunyunyunyu: Yes, slightly rephrased here to clarify

crates/continuations/src/verifier/root/mod.rs
nyunyunyunyu: Yes, just `Vec` to `Array`. `builder.vec` could generate a `Array:Fixed`. Not sure if the verification function must need `Array::Dyn` so I implement it manually.

The shape of `expected_pvs` is determined at compile time(see `num_public_values`). So no need to check.

crates/sdk/src/keygen/asm.rs
nyunyunyunyu: Use `IType` from `rrs_lib` instead.

crates/continuations/src/verifier/internal/vars.rs
nyunyunyunyu: Will rename it in a later PR in order to avoid conflicts of open PRs.

crates/continuations/src/verifier/root/mod.rs
jonathanpwang: Ok - let's leave this comment open as a note of the diff

jonathanpwang (approved):
LGTM

After the rust macro is done, let's do another pass on the docs / usage explanation

PR #1616, opened 2025-05-01
fix(new-execution): make rv32im hintstore work

shuklaayush:
- make `Rv32HintStoreChip` use the `NewVmChipWrapper`
- rename `SingleTraceStep` to `TraceStep` and update it to work for chips whose execution creates multiple trace rows
- comment out criterion execute benchmarks for now

### DIFF
--- a/crates/vm/src/arch/integration_api.rs
+++ b/crates/vm/src/arch/integration_api.rs
@@ -216,25 +216,29 @@ pub struct AdapterAirContext<T, I: VmAdapterInterface<T>> {
     pub instruction: I::ProcessedInstruction,
 }
 
-/// Interface for trace generation when the state transition step of a single instruction
-/// uses only one trace row. The trace row is provided as a mutable buffer during both
-/// instruction execution and trace generation.
+/// Interface for trace generation of a single instruction.The trace is provided as a mutable
+/// buffer during both instruction execution and trace generation.
 /// It is expected that no additional memory allocation is necessary and the trace buffer
 /// is sufficient, with possible overwriting.
-pub trait SingleTraceStep<F, CTX> {
+pub trait TraceStep<F, CTX> {
     fn execute(
         &mut self,
         state: VmStateMut<TracingMemory<F>, CTX>,
         instruction: &Instruction<F>,
-        row_slice: &mut [F],
+        // TODO(ayush): combine to a single struct
+        trace: &mut [F],
+        trace_offset: &mut usize,
+        // TODO(ayush): move air inside step and remove width
+        width: usize,
     ) -> Result<()>;
 
     /// Populates `row_slice`. This function will always be called after
-    /// [`SingleTraceStep::execute`], so the `row_slice` should already contain context necessary to
+    /// [`TraceStep::execute`], so the `row_slice` should already contain context necessary to
     /// fill in the rest of the row. This function will be called for each row in the trace which is
     /// being used, and all other rows in the trace will be filled with zeroes.
     ///
     /// The provided `row_slice` will have length equal to the width of the AIR.
+    // TODO(ayush): should this still operate on a single row or be like execute and operate on a set of rows?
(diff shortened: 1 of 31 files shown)


crates/vm/src/system/public_values/core.rs
```diff
@@ -164,7 +164,9 @@ where
         &mut self,
         state: VmStateMut<TracingMemory, CTX>,
         instruction: &Instruction<F>,
-        row_slice: &mut [F],
+        trace: &mut [F],
```
jonathanpwang, line 167: i want to switch this to abstract buffer later

extensions/rv32im/circuit/src/hintstore/mod.rs
```diff
@@ -334,211 +342,135 @@ impl<F: PrimeField32> InstructionExecutor<F> for Rv32HintStoreChip<F> {
...
-                memory.increment_timestamp();
+                row.is_buffer = F::ONE;
+                row.from_state.timestamp = F::from_canonical_u32(state.memory.timestamp);
+
+                state.memory.increment_timestamp();
+                state.memory.increment_timestamp();
             }
 
-            let data: [F; RV32_REGISTER_NUM_LIMBS] =
+            // TODO(ayush): why is hint_stream F if composed of u8 limbs?
```
jonathanpwang, line 411: `hint_stream` is shared with other instructions (native) so it's type `F`

extensions/rv32im/circuit/src/hintstore/mod.rs
```diff
@@ -334,211 +342,135 @@ impl<F: PrimeField32> InstructionExecutor<F> for Rv32HintStoreChip<F> {
...
-        self.height
-    }
-
-    fn trace_width(&self) -> usize {
-        Rv32HintStoreCols::<F>::width()
-    }
-}
+            for half in 0..(RV32_REGISTER_NUM_LIMBS / 2) {
+                self.bitwise_lookup_chip
+                    .request_range(data[2 * half] as u32, data[2 * half + 1] as u32);
```
jonathanpwang, line 419: if easy, move this to `generate_proof_input`

crates/vm/src/arch/integration_api.rs
```diff
@@ -216,25 +216,29 @@ pub struct AdapterAirContext<T, I: VmAdapterInterface<T>> {
...
     ) -> Result<()>;
 
     /// Populates `row_slice`. This function will always be called after
-    /// [`SingleTraceStep::execute`], so the `row_slice` should already contain context necessary to
+    /// [`TraceStep::execute`], so the `row_slice` should already contain context necessary to
     /// fill in the rest of the row. This function will be called for each row in the trace which is
     /// being used, and all other rows in the trace will be filled with zeroes.
     ///
     /// The provided `row_slice` will have length equal to the width of the AIR.
+    // TODO(ayush): should this still operate on a single row or be like execute and operate on a set of rows?
```
jonathanpwang, line 241: Yea it doesn't make sense if `execute` is on whole trace; rename it to `fill_trace` and provide the whole trace, parallel to `execute`?

jonathanpwang (approved):
LGTM, interesting you were able to get it to work with `fill_trace_row` rather than `fill_trace`. It is a bit annoying to have to re-type the `par_iter` each time if we switch to `fill_trace`, but I'm not sure if for example keccak would work without getting the whole trace.

We could add `fill_trace` with a default implementation that calls `fill_trace_row`; not sure how I feel about it?

crates/vm/src/system/public_values/core.rs
shuklaayush: cool, we can modify later

crates/vm/src/arch/integration_api.rs
shuklaayush: done

extensions/rv32im/circuit/src/hintstore/mod.rs
shuklaayush: done

PR #1620, opened 2025-05-02
feat: Macro define_verify_openvm_stark 

nyunyunyunyu:
- Add a RISC-V custom instruction `nativestorew` into `Rv32IoTranspilerExtension`.
- Add macro `define_verify_openvm_stark`, which can define a function for guest program to verify a stark proof.

closes INT-3896


crates/sdk/tests/integration_test.rs
```diff
@@ -580,11 +585,109 @@ fn test_segmentation_retry() {
...
+        .iter()
+        .map(|x| x.as_canonical_u32())
+        .collect();
+    let pvs_u32: Vec<_> = pvs
+        .iter()
+        .flat_map(|x| x.to_le_bytes())
+        .map(|x| x as u32)
+        .collect();
+
+    let key = ASM_FILENAME
```
jonathanpwang, line 670: there should be a function that does this?

nyunyunyunyu: Yes, https://github.com/openvm-org/openvm/pull/1631

docs/specs/RISCV.md
```diff
@@ -72,6 +72,10 @@ the guest must take care to validate all data and account for behavior in cases
 | printstr    | I   | 0001011     | 011    | 0x1       | Tries to convert `[rd..rd + rs1]_2` to UTF-8 string and print to host stdout. Will print error message if conversion fails.                                                |
 | hintrandom  | I   | 0001011     | 011    | 0x2       | Resets the hint stream to `4 * rd` random bytes from `rand::rngs::OsRng` on the host.                                                                                      |
 
+| RISC-V Inst  | FMT | opcode[6:0] | funct3  | funct7  | RISC-V description and notes                                                                                                    |
```
jonathanpwang, line 75: remind me why you put this under IO instead of under native kernel again?

docs/specs/RISCV.md
```diff
@@ -72,6 +72,10 @@ the guest must take care to validate all data and account for behavior in cases
...
+|--------------|-----|-------------|---------|---------|---------------------------------------------------------------------------------------------------------------------------------|
+| nativestorew | R   | 0001011     | 111     | 2       | Stores the 4-byte word `rs1` at address `rd` in native address space. The address `rd` must be aligned to a 4-byte boundary. |
```
jonathanpwang, line 77:
```suggestion
| nativestorew | R   | 0001011     | 111     | 0x2       | Stores the 4-byte word `rs1` at address `rd` in native address space. The address `rd` must be aligned to a 4-byte boundary. |
```
nit: for consistency

docs/specs/RISCV.md
jonathanpwang: oh because it's a pseudo instruction. hm I feel like for the opcode numbering, it might still be better to move under native kernel

nyunyunyunyu: It still needs to be handled by rv32 transplier so it's hard to out it into native

extensions/rv32im/transpiler/src/lib.rs
```diff
@@ -198,6 +196,23 @@ impl<F: PrimeField32> TranspilerExtension<F> for Rv32IoTranspilerExtension {
...
+                let dec_insn = RType::new(instruction_u32);
+                if dec_insn.funct7 != NATIVE_STOREW_FUNCT7 {
+                    return None;
+                }
+                Some(Instruction::large_from_isize(
+                    Rv32LoadStoreOpcode::STOREW.global_opcode(),
+                    (RV32_REGISTER_NUM_LIMBS * dec_insn.rs1) as isize,
+                    (RV32_REGISTER_NUM_LIMBS * dec_insn.rd) as isize,
+                    0,
+                    1,
```
jonathanpwang, line 210: do we have constant for this addr space?

crates/toolchain/openvm/src/verify_stark.rs
```diff
@@ -0,0 +1,56 @@
...
+#[macro_export]
+macro_rules! define_verify_openvm_stark {
+    ($fn_name: ident, $asm_folder: expr, $asm_filename: literal) => {
+        pub fn $fn_name(app_exe_commit: &[u32; 8], app_vm_commit: &[u32; 8], user_pvs: &[u32]) {
+            // The memory location for the start of the heap.
+            const HEAP_START_ADDRESS: u32 = 1 << 24;
+            const FIELDS_PER_U32: u32 = 4;
+            const FILENAME: &str = $asm_filename;
+            // Construct the hint key
+            let hint_key: alloc::vec::Vec<u8> = FILENAME
```
jonathanpwang, line 25: let's make a util function

crates/sdk/tests/integration_test.rs
jonathanpwang, line 670: why is the filename part of the key?
and if it is part of the key, why the base filename and not the full path?

crates/toolchain/openvm/src/verify_stark.rs
```diff
@@ -0,0 +1,56 @@
...
+/// as native fields. `user_pvs` are the exact public values of that proof. Which means, if a guest
+/// program calls `reveal_u32` to publish an u32(e.g. `258`) in the proof, that u32 is decomposed
+/// into 4 field elements in byte. In `user_pvs`, it should be an u32 array, `[2, 1, 0, 0]`.
+/// 2. Provide the corresponding stark proof in the key-value store in OpenVM streams. The key
+///    should
+/// be the concatenation of the filename, `app_exe_commit`, `app_vm_commit`, and `user_pvs` in
+/// little-endian bytes.
+#[macro_export]
+macro_rules! define_verify_openvm_stark {
+    ($fn_name: ident, $asm_folder: expr, $asm_filename: literal) => {
```
jonathanpwang, line 18: why the separation between folder vs filename?

nyunyunyunyu: Well I didn't figure out how to split a constant string in macro at compile time

jonathanpwang (approved):
LGTM except that the key calculation should be moved into a function.

Either in this PR or another, let's move the macro into a dedicated crate as we discussed.

docs/specs/RISCV.md
jonathanpwang: oh? why does it have to be handled by rv32 transpiler?

crates/toolchain/openvm/src/verify_stark.rs
jonathanpwang: I meant why did you need the key to just be the filename

nyunyunyunyu: Ah it's more friendly to add inputs into key-value store. "/usr/abc/xxx/verifier.asm" is much more weird than "verifier.asm"...

docs/specs/RISCV.md
nyunyunyunyu: Well this requires some contexts about RV32. e.g. `Rv32LoadStoreOpcode::STOREW`/`RV32_REGISTER_NUM_LIMBS`. Also I prefer the transpiler should be put in a place close to the opcode implementation.

PR #1630, opened 2025-05-06
feat: make rv32im testing uniform accross all the chips

arayikhalatyan:
Made the rv32im tests pass and made all the testing files to have the same testing interface.
Deleted the `test_adapter`.  Kept all the test cases unchanged. The only commented test case remaining is the `store` test to the address space 4, which is failing because currently memory accesses with block size 4 are not supported with the address space 4. 

All the test files have 3 types of tests: Positive, Negative, and Sanity tests.
All the test files have 2 helper functions: `create_test_chip`, `set_and_execute`.

An important thing to notice about negative tests when expecting an interaction fail (aka ChallangePhase error) is that ther might be an imbalance created for the wrong reasons. For example, there might be an imbalance on the range checker bus created by the interactions:
[send 1] (sent from the chip_air)
[receive 2] (the execution did `add_count(2)` at some point)
This is not a "valid" fail since 1 is still in the range of the range checker. Because of this a manual check is needed for all the negative checks. To see all the imbalances occurred during a test remove the 'disable_debug_builder();' line from the `run_negative_test` function and run the test. I am 95% sure that I wen through all the negative tests and checked that the imbalances occurred are correct. 

The `test_adapter` tried to address this issue by getting rid of interaction imbalances on the memory bus. But even with the `test _adapter` a manual check was necessary.
To solve this I suggest that we somehow keep all the interactions that occur during the test and automatically check that actually an invalid interaction has happened on a specified bus.

Resolves INT-3975


jonathanpwang:
We discussed we can update `stark-backend` so the `debug` function returns the actual hashmaps of messages in the bus to avoid the "manual check". We can do this later, in separate PR

extensions/rv32im/circuit/src/base_alu/tests.rs
```diff
@@ -79,227 +65,258 @@ fn create_test_chip(
...
+
 //////////////////////////////////////////////////////////////////////////////////////
 // POSITIVE TESTS
 //
 // Randomly generate computations and execute, ensuring that the generated trace
 // passes all constraints.
 //////////////////////////////////////////////////////////////////////////////////////
 
-fn run_rv32_alu_rand_test(opcode: BaseAluOpcode, num_ops: usize) {
+#[test_case(ADD, 100)]
```
jonathanpwang, line 115: nice!

extensions/rv32im/circuit/src/base_alu/tests.rs
```diff
@@ -79,227 +65,258 @@ fn create_test_chip(
...
+#[test_case(SUB, 100)]
+#[test_case(XOR, 100)]
+#[test_case(OR, 100)]
+#[test_case(AND, 100)]
+fn rand_rv32_alu_test(opcode: BaseAluOpcode, num_ops: usize) {
+    setup_tracing();
```
jonathanpwang, line 121:
```suggestion
```

I thought this was supposed to automatically be enabled by the tester

jonathanpwang (approved):
This is much much nicer, great improvement!

I briefly looked through all the files, assuming that you kept the test cases themselves the same and mostly focused on getting things into the framework.

Will let @shuklaayush take a look for context and leave him to merge

extensions/rv32im/circuit/src/loadstore/tests.rs
```diff
@@ -410,40 +358,29 @@ fn negative_wrong_address_space_tests() {
...
-    //     None,
-    //     None,
-    //     None,
-    //     None,
-    //     None,
     //     Some(4),
-    //     VerificationError::OodEvaluationMismatch,
+    //     LoadStorePrankValues::default(),
+    //     false,
     // );
```
shuklaayush, line 374: i wonder if we should just remove this test since rv32im loadstore shouldn't operate on the native address space

shuklaayush (approved):
lgtm. this is much better

extensions/rv32im/circuit/src/base_alu/tests.rs
arayikhalatyan: it was enabled when `tester::new` is called but not with `tester::default`, I added it to the `default` one as well

PR #1631, opened 2025-05-07
feat: helper functions for verify_openvm_stark

nyunyunyunyu:
- `AppExecutionCommit` stores `u32` instead of `BabyBear`.
- Rename `E2eStarkProof` to `VmStarkProof`.
- Add `compute_hint_key_for_verify_openvm_stark`/`encode_rv32_public_values` for users to prepare inputs for `verify_openvm_stark`. But it still seems hard to understand what happens.
- Improve docs about `Rv32HintLoadByKey`.


Cargo.toml
```diff
@@ -66,6 +66,8 @@ members = [
     "extensions/pairing/transpiler",
     "extensions/pairing/guest",
     "extensions/pairing/tests",
+    "guest-libs/verify_stark/guest",
```
jonathanpwang, line 69: I feel this folder structure is weird for just a library: can you do just `guest-libs/verify_stark/src` and `guest-libs/verify_stark/tests` (https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory)

I tried it here and it seems to work well: https://github.com/axiom-crypto/openvm-kzg/pull/18

jonathanpwang (approved):
lgtm except folder structure

Cargo.toml
jonathanpwang: I'll tell @Avaneesh-axiom to change as well, the `/guests` only made sense for VM extensions (and also nowadays perhaps `/guests` should become `/bindings`?). For true libs, we should make it more like normal rust lib

Avaneesh-axiom: My reasoning for the folder structure in #1642 is that the `guest-libs/XXX/tests` are separate crates because they have integration tests (full guest programs). This is similar to the structure in e.g. `extensions/algebra/tests`. Really, I just copied the structure of `extensions/XXX/guest` and `extensions/XXX/tests` into `guest-libs/XXX/guest` and `guest-libs/XXX/tests`

Avaneesh-axiom: (oops, I didn't see your previous message because I didn't refresh the page)

PR #1635, opened 2025-05-08
chore: update `getrandom` to v0.3 in `openvm` lib

jonathanpwang:
delete getrandom dependency from platform since getrandom now makes
enabling of the custom backend opt-in via extern function and --cfg flag

closes INT-3934


.github/workflows/lints.yml
```diff
@@ -43,9 +43,10 @@ jobs:
 
       - name: Run clippy
         run: |
-          # list of features generated using:
-          # echo -e "\033[1;32mAll unique features across workspace:\033[0m" && cargo metadata --format-version=1 --no-deps | jq -r '.packages[].features | to_entries[] | .key' | sort -u | sed 's/^/• /'
-          cargo clippy --all-targets --all --tests --features "aggregation bench-metrics bls12_381 bn254 default entrypoint evm-prove evm-verify export-getrandom export-libm function-span getrandom halo2-compiler halo2curves heap-embedded-alloc k256 jemalloc jemalloc-prof nightly-features panic-handler parallel rust-runtime static-verifier std test-utils unstable" -- -D warnings
+          # list of all unique features across workspace generated using:
+          # cargo metadata --format-version=1 --no-deps | jq -r '.packages[].features | to_entries[] | .key' | sort -u | tr '\n' ' ' && echo ""
```
luffykai, line 47: maybe we should automate this step

crates/toolchain/openvm/src/getrandom.rs
```diff
@@ -0,0 +1,14 @@
+//! [getrandom] custom backend implementations. The implementations are feature-gated. The default
+//! feature enables "getrandom-unsupported", which is a backend that always errors. This should be
+//! used when `getrandom` is never called by pulled in as a dependency unavoidably. If no feature is
```
luffykai, line 3:
```suggestion
//! used when `getrandom` is never called but pulled in as a dependency unavoidably. If no feature is
```

extensions/rv32im/tests/programs/examples/getrandom.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    #[cfg(not(feature = "getrandom-unsupported"))]
+    {
+        // not a good random function!
+        assert_eq!(get_random_u128(), Ok(0));
+    }
+    #[cfg(feature = "getrandom-unsupported")]
+    {
+        // not a good random function!
```
luffykai, line 27: rm?

luffykai (approved):

extensions/rv32im/tests/programs/examples/getrandom.rs
jonathanpwang: oh that was a comment because the custom implementation is all 0s so i'm saying it's not a good random function

.github/workflows/lints.yml
jonathanpwang: I still had to remove snark-verifier somewhere manually, not sure what the right fix is

extensions/rv32im/tests/programs/examples/getrandom.rs
luffykai: that's the one above? this one is just non supported right

jonathanpwang: oh, correct

jonathanpwang, line 27:
```suggestion
```

PR #1639, opened 2025-05-08
feat: bigint circuit to compile

arayikhalatyan:
Implemented e1 and e3 for HeapBranch, Heap, and VecHeap adapters. Updated the Bigint circuit correspondingly. Had to make some changes in the interfaces of rv32im Steps. In particular

- Changed Reads type `([u8; N], [u8; N])` into `Into<[[u8;N];2]>` and Writes type `[u8; N]` into `From<[[u8;N];1]>`. This change corresponds to what we used to do with the previous integration API in order to make the interfaces to match.

- Got rid of TraceAdapterContext in a lot of places. This is because the same Step can be using different AdapterSteps that require different TraceContexts. Or even the AdapterStep might require a `TraceContext` that the Step doesn't have. The easy solution was to implement AdapterSteps in a similar way as in the previous integration API. That is, added the necessary fields to the AdapterStep structs. I am thinking maybe deleting the `TraceContext` from the interface makes sense. I am not sure if there is a better way to do this

Important Note: the tests don't run right now because a lot of the read/write operations are done in address space 2 with block size 32 but currently only block size 4 is supported by the memory.

Resolves INT-3980

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4143,6 +4143,7 @@ dependencies = [
  "openvm-stark-sdk",
  "rand",
  "serde",
+ "test-case",
 ]
 
 [[package]]
--- a/crates/vm/src/arch/integration_api.rs
+++ b/crates/vm/src/arch/integration_api.rs
@@ -636,6 +636,18 @@ pub struct SignedImmInstruction<T> {
 mod conversions {
     use super::*;
 
+    // impl<T> From<[T; 1]> for T {
+    //     fn from(v: [T; 1]) -> Self {
+    //         v[0]
+    //     }
+    // }
+
+    // impl<T> From<[T; 2]> for (T, T) {
+    //     fn from(v: [T; 2]) -> Self {
+    //         (v[0], v[1])
+    //     }
+    // }
+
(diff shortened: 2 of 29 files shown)


jonathanpwang (commented):
Hmm we need to discuss this: it is not ideal for a Step to contain a Chip because ideally Step should be more stateless. Maybe @shuklaayush has a view here

jonathanpwang:
seems like we do need to store some type of reference inside the AdapterStep struct. For now we'll use smart pointer, but Arayi left a comment that it's not ideal and we should use directly reference if possible in the future

shuklaayush (approved):
lgtm. hopefully the remaining tests would pass after the memory pr

> Got rid of TraceAdapterContext in a lot of places. This is because the same Step can be using different AdapterSteps that require different TraceContexts. Or even the AdapterStep might require a TraceContext that the Step doesn't have. The easy solution was to implement AdapterSteps in a similar way as in the previous integration API. That is, added the necessary fields to the AdapterStep structs

this is a bit unfortunate but i guess we can't avoid it for now

PR #1640, opened 2025-05-09
feat: memory access adapters, boundary chips, merkle chip E3

Golovanov399:
Resolves INT-3801.

- Added memory access adapters. To improve:
  * Allocate the trace buffer once before filling it as opposed to pushing to `Vec` how it's done now,
  * Maybe not call `get_f` too often (although I don't know how to avoid it normally).
- Added volatile and persistent boundary chips tracegen,
- Added merkle chip tracegen as described [here](https://docs.google.com/document/d/12cH7ZYRFWHgflpPzOILb7bg5XExdyWOL4vwrQ9HFGkQ/edit?tab=t.0#heading=h.hrg0oexxgu9). To improve:
  * Parallelize at least something,
  * Maybe support passing this struct between segments.
- `VmChipTestBuilder` now has `::default_persistent`, so all tests in `extensions/rv32im/circuit` pass both with volatile and persistent memory interface.

### DIFF
--- a/crates/vm/src/arch/segment.rs
+++ b/crates/vm/src/arch/segment.rs
@@ -215,8 +215,9 @@ where
         *prev_backtrace = trace.cloned();
 
         // Execute the instruction using the control implementation
+        // TODO(AG): maybe avoid cloning the instruction?
         self.control
-            .execute_instruction(state, &mut self.chip_complex)?;
+            .execute_instruction(state, &instruction.clone(), &mut self.chip_complex)?;
 
         // Update metrics if enabled
         #[cfg(feature = "bench-metrics")]
--- a/crates/vm/src/arch/testing/memory/mod.rs
+++ b/crates/vm/src/arch/testing/memory/mod.rs
@@ -105,19 +105,7 @@ impl<F: PrimeField32> MemoryTester<F> {
     /// messages. Taking a volatile memory approach: any touched address will be initialized with
     /// zero.
     pub fn finalize(&mut self) {
-        let memory = &self.controller.memory;
-        // TODO[jpw]: assuming the last block size matches initial block size, after adding
-        // adapters, fix this
-        for ((addr_space, ptr), metadata) in memory.touched_blocks() {
-            let block_size = metadata.block_size as usize;
-            let chip = self.chip_for_block.get_mut(&block_size).unwrap();
-            let mut data = F::zero_vec(block_size);
-            chip.send(addr_space, ptr, &data, INITIAL_TIMESTAMP);
-            for (i, v) in data.iter_mut().enumerate() {
-                *v = memory.data().get_f(addr_space, ptr + i as u32);
-            }
-            chip.receive(addr_space, ptr, &data, metadata.timestamp);
-        }
+        panic!("should not be called");
     }
 }
 
--- a/crates/vm/src/arch/testing/memory/mod.rs
+++ b/crates/vm/src/arch/testing/memory/mod.rs
@@ -5,8 +5,6 @@ use openvm_circuit::system::memory::MemoryController;
 use openvm_stark_backend::p3_field::PrimeField32;
 use rand::Rng;
 
-use crate::system::memory::INITIAL_TIMESTAMP;
-
 pub mod air;
 
 /// A dummy testing chip that will add unconstrained messages into the [MemoryBus].
@@ -99,13 +97,6 @@ impl<F: PrimeField32> MemoryTester<F> {
             .get_mut(&N)
             .unwrap()
             .send(addr_space as u32, ptr as u32, &data, t);
-    }
-
-    /// Fills in dummy memory chips to balance the memory bus with the initial and final boundary
-    /// messages. Taking a volatile memory approach: any touched address will be initialized with
-    /// zero.
-    pub fn finalize(&mut self) {
-        panic!("should not be called");
     }
 }
 
--- a/crates/vm/src/arch/testing/mod.rs
+++ b/crates/vm/src/arch/testing/mod.rs
@@ -338,6 +338,14 @@ where
                     zip(airs, air_proof_inputs).filter(|(_, input)| input.main_trace_height() > 0),
                 );
             }
+            eprintln!(
+                "poseidon_chip: {:?}",
+                match &poseidon_chip {
+                    Poseidon2PeripheryChip::Register0(chip) => &chip.records,
+                    Poseidon2PeripheryChip::Register1(chip) => &chip.records,
+                }
+            );
+            self = self.load(poseidon_chip);
             self = self.load(range_checker); // this must be last because other trace generation
                                              // mutates its state
         }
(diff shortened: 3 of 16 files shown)


crates/vm/src/arch/segment.rs
```diff
@@ -215,8 +215,9 @@ where
         *prev_backtrace = trace.cloned();
 
         // Execute the instruction using the control implementation
+        // TODO(AG): maybe avoid cloning the instruction?
         self.control
-            .execute_instruction(state, &mut self.chip_complex)?;
+            .execute_instruction(state, &instruction.clone(), &mut self.chip_complex)?;
```
jonathanpwang, line 220: how come it's both reference and clone?

crates/vm/src/arch/testing/memory/mod.rs
```diff
@@ -105,19 +105,7 @@ impl<F: PrimeField32> MemoryTester<F> {
...
-            let block_size = metadata.block_size as usize;
-            let chip = self.chip_for_block.get_mut(&block_size).unwrap();
-            let mut data = F::zero_vec(block_size);
-            chip.send(addr_space, ptr, &data, INITIAL_TIMESTAMP);
-            for (i, v) in data.iter_mut().enumerate() {
-                *v = memory.data().get_f(addr_space, ptr + i as u32);
-            }
-            chip.receive(addr_space, ptr, &data, metadata.timestamp);
-        }
+        panic!("should not be called");
```
jonathanpwang, line 108: should we just delete this??

crates/vm/src/arch/testing/mod.rs
```diff
@@ -300,27 +335,49 @@ where
     }
 
     pub fn finalize(mut self) -> Self {
-        if let Some(mut memory_tester) = self.memory.take() {
+        if let Some(memory_tester) = self.memory.take() {
             // Balance memory boundaries
-            memory_tester.finalize();
-            let memory_controller = memory_tester.controller;
+            // memory_tester.finalize();
```
jonathanpwang, line 340:
```suggestion
```

crates/vm/src/system/memory/adapter/mod.rs
```diff
@@ -304,48 +318,56 @@ impl<F, const N: usize> GenericAccessAdapterChipTrait<F> for AccessAdapterChip<F
...
-                    (&mut row.lt_aux, &mut row.is_right_larger),
-                );
-            });
-        RowMajorMatrix::new(values, width)
+        let mut trace = self.trace;
+        let height = trace.height();
+        trace.pad_to_height(height.next_power_of_two(), F::ZERO);
+        trace
+        // let width = BaseAir::<F>::width(&self.air);
+        // let height = if let Some(oh) = self.overridden_height {
```
jonathanpwang, line 326: will we handle overridden height elsewhere? leave a TODO where we need to handle it

crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -463,46 +440,151 @@ impl<F: PrimeField32> MemoryController<F> {
...
+                    continue;
+                }
+                assert!(
+                    metadata.block_size >= RV32_REGISTER_NUM_LIMBS as u32,
+                    "How come some block size is less than RV32_REGISTER_NUM_LIMBS, isn't this the alignment?"
+                );
+                let values = (0..metadata.block_size)
+                    .map(|i| self.memory.data.get_f::<F>(addr_space, ptr + i))
+                    .collect::<Vec<_>>();
+                for start in (0..metadata.block_size).step_by(RV32_REGISTER_NUM_LIMBS) {
```
jonathanpwang, line 463: I think the step_by size needs to depend on address space? i.e., it is 4 for address space <=3, but 1 for address space 4

crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -463,46 +440,151 @@ impl<F: PrimeField32> MemoryController<F> {
...
+                    for (i, &v) in values.iter().enumerate() {
+                        final_memory.insert(
+                            (addr_space, ptr + i as u32),
+                            TimestampedValues {
+                                timestamp,
+                                values: [v],
+                            },
+                        );
+                    }
+                    self.memory.execute_splits::<1, false>(
```
jonathanpwang, line 500: does the size in `execute_splits` need to be const generic? life may be easier for you if it isn't in the future

crates/vm/src/system/memory/merkle/trace.rs
```diff
@@ -122,136 +95,6 @@ impl<const CHUNK: usize, F: PrimeField32> ChipUsageGetter for MemoryMerkleChip<C
     }
 }
 
-struct TreeHelper<'a, const CHUNK: usize, F: PrimeField32> {
```
jonathanpwang, line 125: yayyyyy

crates/vm/src/system/memory/online.rs
```diff
@@ -186,12 +190,22 @@ impl<F: PrimeField32> TracingMemory<F> {
         );
     }
 
-    fn execute_splits<const ALIGN: usize>(
+    pub(crate) fn execute_splits<const ALIGN: usize, const UPDATE_META: bool>(
```
jonathanpwang, line 193: as mentioned, your life may be easier if `ALIGN` doesn't need to be const generic anymore

jonathanpwang (approved):
LGTM except comments around (1) avoiding const generic `ALIGN` when possible because (2) the boundary memories need to account for per-address space min block size.

crates/vm/src/arch/segment.rs
Golovanov399: It's because
- design-wise I prefer passing a reference
- currently it's hard to do without tricks, because this reference is inside `self.something`, and `self.execute_instruction` requires a mutable reference to `self`. That is why I did this hack and added a `// TODO` above

crates/vm/src/system/memory/online.rs
```diff
@@ -249,6 +271,34 @@ impl<F: PrimeField32> TracingMemory<F> {
...
+    fn set_meta_block<const ALIGN: usize>(
+        &mut self,
+        address_space: usize,
+        pointer: usize,
+        block_size: usize,
+        timestamp: u32,
+    ) {
+        let ptr = pointer / ALIGN;
+        let meta = unsafe { self.meta.get_unchecked_mut(address_space) };
+        meta.set(
```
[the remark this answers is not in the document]
Golovanov399: What would we do in this case? We still want to set the meta block. Or do you think that iterating over `1..1` is slower than branch prediction in practice?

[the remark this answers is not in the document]
jonathanpwang: oh hm I somehow thought we didn't need to set the meta block in this case, but I guess you need to update the timestamp regardless?

crates/vm/src/arch/segment.rs
jonathanpwang: I meant does `&instruction` not work?

PR #1642, opened 2025-05-09
feat: Add primitive guest libraries

Avaneesh-axiom:


jonathanpwang (commented):
Before anything else, as mentioned in Xinding's PR, I feel for `guest-libs` the crate organization is unnecessarily complicated because it is a single rust library, not some collection like in VM extensions. Please follow the pattern in this PR: https://github.com/axiom-crypto/openvm-kzg/pull/18

benchmarks/guest/ecrecover/src/main.rs
jonathanpwang, line 1: the best test now would be if you could patch `k256` with `openvm-k256` and see if `ecrecover` from unpatched `revm` still works

benchmarks/guest/ecrecover/src/main.rs
```diff
@@ -1,7 +1,10 @@
 use alloy_primitives::{Bytes, B256, B512};
-use k256::ecdsa::{Error, RecoveryId, Signature, VerifyingKey};
+// Be careful not to import k256::ecdsa::{Signature, VerifyingKey}
+// because those are type aliases that their (non-zkvm) implementations
+use ecdsa::{Signature, VerifyingKey};
+use k256::ecdsa::{Error, RecoveryId};
 use openvm::io::read_vec;
-use openvm_k256::Secp256k1Point;
+use openvm_k256::{Secp256k1, Secp256k1Point};
```
jonathanpwang, line 7: I was wondering if you could just not have the `openvm-k256` import at all here, and instead in Cargo.toml patch `k256` with `openvm-k256`? that's how we would likely use it in practice

PR #1647, opened 2025-05-13
feat: openvm build can build multiple targets + additional cargo options

stephenh-axiom-xyz:
Resolves INT-4000, INT-3883, and INT-3662. In general, we would like `cargo openvm build` to align as closely with `cargo build` as possible. OpenVM users should be able to almost completely replace the latter with the former in their development workflows.

Implements the following:

- Parity between `cargo openvm build` and `cargo build` options
- Ability to build multiple targets when building at the workspace level
- `cargo openvm build` works outside of project root
- Default paths in OpenVM CLI evaluate environment variable `HOME` at runtime

OpenVM-generated artifacts being stored in `${CARGO_TARGET_DIR}/.openvm` is still to be implemented, but in order to maintain coherency between `cargo openvm build` and `cargo openvm run/prove` this change will be added later.


nyunyunyunyu (approved):
I don't fully understand all the cargo flags but overall LGTM

crates/cli/src/commands/build.rs
```diff
@@ -21,225 +22,432 @@ use openvm_transpiler::{elf::Elf, openvm_platform::memory::MEM_SIZE};
...
+    #[arg(
+        long,
+        value_name = "DIR",
+        help = "Directory for all generated artifacts and intermediate files",
+        help_heading = "Output Options"
+    )]
+    pub target_dir: Option<PathBuf>,
 
-    #[arg(long, default_value = OPENVM_DEFAULT_INIT_FILE_NAME, help = "Name of the init file")]
-    pub init_file_name: String,
```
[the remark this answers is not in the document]
jonathanpwang: ah nvm just hard to tell the diff

crates/cli/src/commands/build.rs
```diff
@@ -21,225 +22,432 @@ use openvm_transpiler::{elf::Elf, openvm_platform::memory::MEM_SIZE};
...
+        help = "Skips transpilation into exe when set",
+        help_heading = "OpenVM Options"
     )]
     pub no_transpile: bool,
 
     #[arg(
         long,
         default_value = DEFAULT_APP_CONFIG_PATH,
-        help = "Path to the SDK config .toml file that specifies the transpiler extensions"
+        help = "Path to the SDK config .toml file that specifies the transpiler extensions",
```
jonathanpwang, line 60:
```suggestion
        help = "Path to the OpenVM config .toml file that specifies the VM extensions",
```
I think we use it more generally than just transpiler

crates/cli/src/commands/build.rs
```diff
@@ -21,225 +22,432 @@ use openvm_transpiler::{elf::Elf, openvm_platform::memory::MEM_SIZE};
...
     }
 
+    // Build (allowing passed options to decide what gets built)
+    let target_dir = build_generic(&guest_options)?;
+    if cargo_args.lib {
+        return Ok(None);
+    }
+
+    // Write to init file
     let app_config = read_config_toml_or_default(&build_args.config)?;
```
jonathanpwang, line 359: noting that this means we will try to use the same openvm.toml for all build targets within this command. I thought about it and it does seem the most reasonable option: if you wanted to compile just a select target with a different config, you should run the command with the right target and `--config` custom

crates/cli/src/commands/build.rs
```diff
@@ -21,225 +22,432 @@ use openvm_transpiler::{elf::Elf, openvm_platform::memory::MEM_SIZE};
...
+                (cargo_args.package.is_empty() || cargo_args.package.contains(&pkg.name))
+                    && !cargo_args.exclude.contains(&pkg.name)
+            })
+            .collect()
+    } else {
+        vec![get_package(manifest_dir)]
     };
 
+    // Find elf paths of all targets for all built packages
+    let elf_paths: Vec<PathBuf> = packages
```
jonathanpwang, line 379: does this still make sense or does it need to be generalized?

crates/cli/src/commands/build.rs
```diff
@@ -21,225 +22,432 @@ use openvm_transpiler::{elf::Elf, openvm_platform::memory::MEM_SIZE};
...
-        }
-        write(
-            &build_args.exe_commit_output,
-            committed_exe_as_bn254(&committed_exe).value.to_bytes(),
-        )?;
-        if let Some(parent) = build_args.committed_exe_output.parent() {
-            create_dir_all(parent)?;
+        for elf_path in &elf_paths {
+            println!("[openvm] Transpiling the package...");
+            let output_path = &build_args.exe_output;
```
jonathanpwang, line 415: this doesn't seem to be doing what the comment above says?

jonathanpwang (commented):
I'd like to get the file paths sorted out and also resolve why the new `build_generic` function was introduced

crates/toolchain/build/src/lib.rs
```diff
@@ -370,6 +403,72 @@ pub fn build_guest_package(
     }
 }
 
+/// Generic wrapper call to cargo build
+pub fn build_generic(guest_opts: &GuestOptions) -> eyre::Result<PathBuf> {
```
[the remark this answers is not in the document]
stephenh-axiom-xyz: We do, right above it is another function called `build_guest_package`. The primary difference is that `build_guest_package` is meant to build a single package only, while `build_generic` more so passes the options to `cargo build` and lets it do whatever it's natural behavior is.

One issue encountered while `cargo openvm build` allow multiple targets is that much of our code expects single target builds only - for example, `build_guest_package` is used in the SDK, where it does more or less make sense to do so. Further discussion on multi-target builds throughout OpenVM would be appreciated!

crates/cli/src/commands/build.rs
stephenh-axiom-xyz: Assuming you mean getting the ELF paths, I think it still makes sense? We still need to find the targets compiled by `cargo build` to transpile them. We currently only allow `bin` and `example` targets (some stuff breaks with certain other targets I think), so I believe some additional processing is still needed.

crates/toolchain/build/src/lib.rs
```diff
@@ -349,24 +387,17 @@ pub fn build_guest_package(
...
     } else {
-        Ok(get_dir_with_profile(
-            &target_dir,
-            profile,
-            target_filter
-                .as_ref()
-                .map(|t| t.kind == "example")
-                .unwrap_or(false),
-        ))
+        Ok(get_dir_with_profile(target_dir, profile, false))
```
jonathanpwang, line 400: maybe we don't need the last arg of this function anymore, but ok to leave as well

jonathanpwang (approved):
lgtm

PR #1651, opened 2025-05-14
fix: new block initialization in memory access adapters

Golovanov399:
I used to handle creating new blocks in a wrong way when `align > initial_block_size`, now I hopefully do it right.
Also added persistent base alu tests, although nothing changed for the persistent case,
and added a dummy access in all of them that used to fail.


arayikhalatyan (approved):
lgtm, the mod_eq test that was failing now works as well

Golovanov399:
Yay
Okay pushing then to unblock

PR #1652, opened 2025-05-15
feat(new-execution): metered execution

shuklaayush:
- Added a `execute_metered` function that keeps a track of `trace_heights` during execution and suspends execution if the trace height, number of cells or number of interactions goes above a fixed threshold

Resolves INT-3752


extensions/rv32-adapters/src/eq_mod.rs
```diff
@@ -394,9 +395,14 @@ impl<
     type ReadData = [[u8; TOTAL_READ_SIZE]; NUM_READS];
     type WriteData = [u8; RV32_REGISTER_NUM_LIMBS];
 
-    fn read<Mem>(&self, memory: &mut Mem, instruction: &Instruction<F>) -> Self::ReadData
+    fn read<Mem, Ctx>(
+        &self,
+        state: &mut VmStateMut<Mem, Ctx>,
```
jonathanpwang, line 400:
```suggestion
        state: VmStateMut<Mem, Ctx>,
```

shuklaayush: i'll take a look at this again after rebasing but this was causing the compiler to complain so i had to add the `&mut`

crates/vm/src/arch/extensions.rs
```diff
@@ -337,6 +337,24 @@ impl<E, P> VmInventory<E, P> {
         self.executors.get_mut(*id)
     }
 
+    pub fn get_mut_executor_with_index(&mut self, opcode: &VmOpcode) -> Option<(&mut E, usize)> {
+        let id = *self.instruction_lookup.get(opcode)?;
+
+        self.executors.get_mut(id).map(|executor| {
+            let insertion_id = self
+                .insertion_order
```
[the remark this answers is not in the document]
shuklaayush: yes, but i'm not sure how to avoid it atm. ideally this should be a constant vector and O(1). i added a todo and i'll fix this in a follow up

extensions/rv32-adapters/src/eq_mod.rs
shuklaayush: this doesn't work because i need to pass the state as a reference now and call a callback to update `ctx.trace_heights` on every memory operation. ideally this should just be a `&mut VmState<Mem, Ctx>` but for that we'll need to take memory out of the controller

https://github.com/openvm-org/openvm/blob/7508e8278ad6fd48fe2ca45d73e5056363549068/extensions/rv32im/circuit/src/adapters/alu.rs#L304-L305
https://github.com/openvm-org/openvm/blob/7508e8278ad6fd48fe2ca45d73e5056363549068/extensions/rv32im/circuit/src/adapters/mod.rs#L93-L105

crates/vm/src/arch/execution_mode/metered/exact.rs
```diff
@@ -0,0 +1,391 @@
...
+            } else {
+                // Initial memory access only happens at CHUNK boundary
+                let chunk_ratio = 1 << (CHUNK_BITS - align_bits);
+                let chunk_offset = curr_block & (chunk_ratio - 1);
+                curr_block -= chunk_offset;
+                CHUNK
+            };
+
+            // TODO(ayush): why are we splitting in the case when we only
+            //              read at mutually exclusive CHUNK boundaries?
```
Golovanov399, line 129: What do you mean

crates/vm/src/arch/execution.rs
```diff
@@ -115,8 +118,18 @@ pub trait InstructionExecutor<F> {
 pub trait InsExecutorE1<F> {
     fn execute_e1<Ctx>(
         &mut self,
-        state: VmStateMut<GuestMemory, Ctx>,
+        state: &mut VmStateMut<GuestMemory, Ctx>,
```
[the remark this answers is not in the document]
shuklaayush: i can't get around this. it's annoying. if you create a struct of mutable references, then you can't pass the struct as an immutable reference and edit the members. you need to make the outer reference mutable too. hence the double mut

crates/vm/src/arch/execution_mode/metered/exact.rs
shuklaayush: i think this is outdated. i'll remove it. i was trying to figure out why we need to split on the first read

crates/vm/src/arch/execution_mode/tracegen.rs
```diff
@@ -0,0 +1,117 @@
...
+        _state: &mut VmSegmentState<Self::Ctx>,
+        chip_complex: &VmChipComplex<F, VC::Executor, VC::Periphery>,
+    ) -> bool {
+        // Avoid checking segment too often.
+        if self.since_last_segment_check != SEGMENT_CHECK_INTERVAL {
+            self.since_last_segment_check += 1;
+            return false;
+        }
+        self.since_last_segment_check = 0;
+        chip_complex.config().segmentation_strategy.should_segment(
```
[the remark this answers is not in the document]
shuklaayush: it should be execution control. i kept it like this to make e3 compatible with how it was working earlier but yeah, this needs to be cleaned up. 
i don't think it's being duplicated on the segmentation strategy side. this should be the only place where the check is happening

crates/vm/src/arch/execution.rs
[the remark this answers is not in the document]
jonathanpwang: I also ran into this, basically when we do stuff and pass `&mut ..` around, actually the compiler is doing something extra. But when we're in a trait, you can't trick the compiler anymore

crates/vm/src/arch/execution_mode/metered/bounded.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        let num_blocks = (size + CHUNK - 1) >> CHUNK_BITS;
+        for i in 0..num_blocks {
+            let addr = ptr.wrapping_add((i * CHUNK) as u32);
+            let block_id = addr >> CHUNK_BITS;
+            let leaf_id = self
+                .memory_dimensions
+                .label_to_index((address_space, block_id));
+
+            if let Err(insert_idx) = self.leaf_indices.binary_search(&leaf_id) {
```
jonathanpwang, line 60: ok I didn't read this, but it seems too complicated for something that must be done during execution

crates/vm/src/arch/execution_mode/metered/exact.rs
jonathanpwang, line 1: why is there both bounded and exact? this is to benchmark/compare how bad the estimate is?

jonathanpwang (commented):
Took a brief look, deferring to @Golovanov399 's review

I think we all have the sense that the structure / traits should be revisited, but I don't think we should block merge.

crates/vm/src/arch/execution_mode/metered/exact.rs
shuklaayush: Yes. And to sanity check that the calculations are correct

crates/vm/src/arch/execution_mode/metered/bounded.rs
shuklaayush: I can change it to add the height of the Merkle tree every time but that might be a huge overestimate
I think the better way is to not maintain a set here and make GuestMemory return a Boolean indicating a page fault. We can then add a fixed height on every page fault
Fixed height would be height of the merkle tree up to page size + size of the full page subtree

shuklaayush: @Golovanov399 wdyt?

Golovanov399: I thought for myself that it'll do for now, because if it's too slow, we'll surely notice

Golovanov399:
Yeah I'm all for merging

PR #1657, opened 2025-05-15
feat: e1, e3 implementation for sha2 + keccak extensions

arayikhalatyan:
Implemented e1, e3 for sha2 and keccak extensions. I feel like the trace generation of sha2 is more readable now. 
Also, implemented an arbitrary length read function (used in both e1 executions)

I think eventually we should reimplement Plonky3's trace generation for Keccak so we don't have to allocate a separate trace matrix for the perm cols. 

TODOs:

- port the tests of keccak to the new framework
- spend some time (not much) on optimizing the keccak trace gen 

Resolves INT-3966
Resolves INT-3921


jonathanpwang:
@arayikhalatyan we shouldn't have any block size restrictions on E1 (but likely need it for E2 in some light form). My imagination is indeed that E1 should just look like `input = read_bytes(); return sha256(input)`

arayikhalatyan:
> @arayikhalatyan we shouldn't have any block size restrictions on E1 (but likely need it for E2 in some light form). My imagination is indeed that E1 should just look like `input = read_bytes(); return sha256(input)`

Yeah I agree. Then we have to make some changes to the GuestMemory interface because it expects the block size as a const generic.

jonathanpwang:
> > @arayikhalatyan we shouldn't have any block size restrictions on E1 (but likely need it for E2 in some light form). My imagination is indeed that E1 should just look like `input = read_bytes(); return sha256(input)`
> 
> Yeah I agree. Then we have to make some changes to the GuestMemory interface because it expects the block size as a const generic.

ah yea

crates/circuits/sha256-air/src/trace.rs
jonathanpwang, line 1: this was a lot cleaner than I thought!

crates/circuits/sha256-air/src/trace.rs
```diff
@@ -1,41 +1,47 @@
...
+
 /// The trace generation of SHA256 should be done in two passes.
 /// The first pass should do `get_block_trace` for every block and generate the invalid rows through
 /// `get_default_row` The second pass should go through all the blocks and call
 /// `generate_missing_cells`
-impl Sha256Air {
+impl Sha256StepHelper {
+    pub fn new() -> Self {
+        Self {
+            row_idx_encoder: Encoder::new(18, 2, false),
```
jonathanpwang, line 37: did you have to hardcode 18 because `Padding::COUNT` is not accessible in this crate?

jonathanpwang: maybe this function should be renamed and moved to `tests.rs` if it's only used there

extensions/keccak256/circuit/src/trace.rs
```diff
@@ -1,150 +1,216 @@
...
+            )
+        };
+
+        // Due to the AIR constraints, the final memory timestamp should be the following:
+        let final_timestamp = state.memory.timestamp()
+            + (remaining_len + KECCAK_ABSORB_READS + KECCAK_DIGEST_WRITES) as u32;
+        let num_blocks = num_keccak_f(remaining_len);
+        let mut hasher = Keccak::v256();
+
+        trace
```
jonathanpwang, line 78: note: I think the way this is done makes sense ATM, but we'll have to change it / re-think it pretty carefully for the e4 rewrite

jonathanpwang (approved):
lgtm

tbh I didn't review the keccak execute/step too closely because (1) tests pass, (2) we'll need to re-review after the next (and last) e4 rewrite

crates/circuits/sha256-air/src/trace.rs
arayikhalatyan: `18` comes from `NUM_ROWS+1`, I think it makes sense to keep it inside the constructor but with constants. eventually should be in `colsref` after Avaneesh's PR

PR #1658, opened 2025-05-15
fix: verifier read dir

HrikB:
Fixes the path from which verifier contract is read.

I tested the full CI flow on an aws instance to double check everything is working.


jonathanpwang (approved):
lgtm

PR #1659, opened 2025-05-16
feat: rv32im working execution + tracegen + prove (aka integration tests)

Golovanov399:
This resolves INT-3913.

As a _side effect_, this removes `GuestMemory` trait -- it is a struct now with underlying `AddressMap<PAGE_SIZE>` (I didn't make `type GuestMemory = ...` because the vaguely called `read` and `write` methods would be too vaguely called for `AddressMap`). `VmStateMut` is generic over `MEM` though.

I didn't fully implement `TraceStep` and `StepExecutorE1` for the phantom chip because it's relatively easy and I'm not sure it would be better expressible in terms of `NewVmChipWrapper`.

`PhantomSubExecutor` also changed a little (now accepts `u32` instead of `F`, for example, and also `GuestMemory` instead of what it needed before).


jonathanpwang (approved):
lgtm

PR #1662, opened 2025-05-16
feat: cargo CLI update target directory

stephenh-axiom-xyz:
Resolves INT-3882, and records changes in the book as per INT-4038. Most importantly, reworks command outputs to primarily use the `target/` directory that Cargo uses and assumes full control over.

- User can specify `target_dir`; if unspecified, `cargo openvm` will use the Cargo default
- Keys are stored in `${target_dir}/openvm/`, while all other artifacts are stored in `${target_dir}/openvm/${profile}`
- User can copy outputs to `${output_dir}`
- **NOTE** - generated proofs are stored at `.`, as this file should generally be the final output

Other changes include:
- `build` no longer generates app commits
- `run` will call `build` before attempting to execute
- `prove` will call `build` before attempting to generate a proof
- Replaced all references to `evm-proving-setup` with references to `setup`
- `openvm.toml`'s default location is in the working directory instead of `.`

The basic workflow should now be `cargo openvm prove ...` into `cargo openvm verify ...`.


crates/cli/src/default.rs
```diff
@@ -4,14 +4,14 @@ use openvm_sdk::config::{AppConfig, SdkVmConfig, DEFAULT_APP_LOG_BLOWUP, DEFAULT
 use openvm_stark_sdk::config::FriParameters;
 
 pub const DEFAULT_MANIFEST_DIR: &str = ".";
-
 pub const DEFAULT_APP_CONFIG_PATH: &str = "./openvm.toml";
```
stephenh-axiom-xyz, line 7: I feel like the default behavior should be to look for `openvm.toml` in the same directory as `Cargo.toml` (which can be set using `--manifest-path`, instead of being relative to the working directory. Thoughts?

crates/cli/src/commands/keygen.rs
```diff
@@ -8,48 +8,84 @@ use openvm_sdk::{
...
     #[arg(
         long,
-        action,
-        help = "Path to output app proving key file",
-        default_value = DEFAULT_APP_PK_PATH
+        help = "Output path for the app proving key, by default will be ${target_dir}/openvm/app.pk",
+        help_heading = "OpenVM Options"
     )]
-    output: PathBuf,
+    app_pk: Option<PathBuf>,
```
stephenh-axiom-xyz, line 32: I think perhaps we should only let the user set the verifying key output, and then store `app_pk` according to the `target` directory? Then `prove` would (to the user) generate a proof and verifying key.

crates/cli/src/commands/build.rs
```diff
@@ -57,19 +52,10 @@ pub struct BuildArgs {
...
-    pub exe_output: PathBuf,
-
-    #[arg(
-        long,
-        default_value = DEFAULT_COMMITTED_APP_EXE_PATH,
-        help = "Output path for the committed program",
-        help_heading = "OpenVM Options"
-    )]
-    pub committed_exe_output: PathBuf,
+    pub output_dir: Option<PathBuf>,
```
stephenh-axiom-xyz, line 58: This should probably be a copy directory, i.e. we store OpenVM artifacts both at `${target_dir}/openvm/${profile}` instead of one or the other. IMO this semantic both preserves the control we assume over the `${target_dir}/openvm` directory and allows users more customizability.

crates/cli/src/commands/keygen.rs
```diff
@@ -8,48 +8,84 @@ use openvm_sdk::{
...
+    )]
+    pub(crate) manifest_path: Option<PathBuf>,
 }
 
 impl KeygenCmd {
     pub fn run(&self) -> Result<()> {
-        keygen(&self.config, &self.output, &self.vk_output)?;
+        let (manifest_path, _) = manifest_path_and_dir(&self.cargo_args.manifest_path)?;
+        let target_dir = target_dir(&self.cargo_args.target_dir, &manifest_path);
+        let app_pk_path = app_pk_path(&self.app_pk, &target_dir);
```
luffykai, line 68: i think function name should be `get_app_pk_path`

crates/cli/src/global.rs
```diff
@@ -0,0 +1,138 @@
...
+    } else {
+        find_manifest_dir(PathBuf::from("."))?
+    };
+    let manifest_path = MANIFEST_PATH.get_or_init(|| manifest_dir.join("Cargo.toml"));
+    Ok((manifest_path.clone(), manifest_dir))
+}
+
+pub(crate) fn target_dir(target_dir: &Option<PathBuf>, manifest_path: &PathBuf) -> PathBuf {
+    TARGET_DIR
+        .get_or_init(|| {
```
luffykai, line 46: this means, if this function is called second time with a different target_dir, it won't have effect right? i think this can be confusing? 
what's the desired behavior here?

crates/cli/src/commands/prove.rs
```diff
@@ -146,29 +189,57 @@ impl ProveCmd {
...
+                )?;
+
+                write_evm_proof_to_file(evm_proof, proof)?;
             }
         }
         Ok(())
     }
 
-    fn prepare_execution(
+    fn get_app_pk(
```
luffykai, line 207: the function name should change, as this potentially run keygen.

crates/cli/src/commands/prove.rs
```diff
@@ -146,29 +189,57 @@ impl ProveCmd {
...
+        app_pk: &Option<PathBuf>,
+        run_args: &RunArgs,
+        cargo_args: &RunCargoArgs,
+    ) -> Result<Arc<AppProvingKey<SdkVmConfig>>> {
+        let (manifest_path, _) = manifest_path_and_dir(&cargo_args.manifest_path)?;
+        let target_dir = target_dir(&cargo_args.target_dir, &manifest_path);
+
+        let app_pk_path = app_pk_path(app_pk, &target_dir);
+        if app_pk.is_none() {
```
luffykai, line 216: even if app pk exists (say i ran keygen before) this will rerun keygen right? can we not do that?

crates/cli/src/commands/prove.rs
```diff
@@ -146,29 +189,57 @@ impl ProveCmd {
...
+            let app_vk_path = app_vk_path(&None, &target_dir);
+            keygen(&run_args.config, &app_pk_path, &app_vk_path)?;
+        }
+
+        Ok(Arc::new(read_app_pk_from_file(app_pk_path)?))
+    }
+
+    fn get_commit_exe(
```
luffykai, line 224: same here, should probably change name

luffykai (commented):
I think the major comment is we should discuss and finalize the behavior first (or if we've done this before, we should document somewhere). e.g. do we always do keygen in prove/run even though key is already there? 

Also the functions in `global.rs` should have clear docs, as their names don't match what they do. (implicit build/keygen happens in `get_xx` function

crates/cli/src/commands/build.rs
stephenh-axiom-xyz: Sorry, both to `${target_dir}/openvm/${profile}` and `output_dir`.

crates/cli/src/commands/build.rs
```diff
@@ -366,73 +385,60 @@ pub fn build(build_args: &BuildArgs, cargo_args: &BuildCargoArgs) -> Result<Vec<
     };
 
     // Find elf paths of all targets for all built packages
-    let elf_paths: Vec<PathBuf> = packages
+    let elf_targets = packages
+        .iter()
+        .flat_map(|pkg| pkg.targets.iter())
+        .filter(|target| {
+            // We only build bin and example targets (note they are mutually exclusive
```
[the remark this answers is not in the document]
stephenh-axiom-xyz: Yes I believe so, as you have to define them as `[[bin]]` and `[[example]]` in `Cargo.toml`

crates/cli/src/commands/prove.rs
stephenh-axiom-xyz: There's no guarantee the key is up to date

crates/cli/src/global.rs
stephenh-axiom-xyz: Some commands call other commands, so instead of deriving the manifest path, target directory, etc. each time we use these global `OnceLock` variables to save a bit of compute.

luffykai: i am not sure if we need to save compute like this. my concern is that it's not obvious that this function will ignore the input parameters when target_dir is init before, and if we later ever use it on different target_dir input, it's going to be very confusing. I think we can just compute it everytime

crates/cli/src/global.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub(crate) fn get_app_pk_path(target_dir: &Path) -> PathBuf {
+    target_dir.join("openvm").join(DEFAULT_APP_PK_NAME)
+}
+
+pub(crate) fn get_app_vk_path(target_dir: &Path) -> PathBuf {
+    target_dir.join("openvm").join(DEFAULT_APP_VK_NAME)
+}
+
+pub(crate) fn get_single_target_name(cargo_args: &RunCargoArgs) -> Result<String> {
```
luffykai, line 63: add a comment on the logic of this fn

crates/cli/src/commands/verify.rs
```diff
@@ -36,8 +60,20 @@ impl VerifyCmd {
...
         let sdk = Sdk::new();
         match &self.command {
-            VerifySubCommand::App { app_vk, proof } => {
-                let app_vk = read_app_vk_from_file(app_vk)?;
+            VerifySubCommand::App {
+                app_vk,
+                proof,
+                cargo_args,
+            } => {
+                let (manifest_path, _) = get_manifest_path_and_dir(&cargo_args.manifest_path)?;
```
luffykai, line 68: this is needed only in else branch below?

book/src/writing-apps/build.md
```diff
@@ -9,135 +9,140 @@ The command `cargo openvm build` compiles the program on host to an executable f
...
+- `--output_dir <OUTPUT_DIR>`
 
-  **Description**: Specifies the directory where the guest binary will be built. If not specified, the default target directory is used.
+  **Description**: Output directory for OpenVM artifacts to be copied to. Keys will be placed in `${output-dir}/`, while all other artifacts will be in `${output-dir}/${profile}`.
 
-  **Default**: The `target` directory in the package root directory.
+- `--init-file-name <INIT_FILE_NAME>`
 
-  **Usage Example**: To build the guest binary in the `my_target` directory:
+  **Description**: Name of the generated initialization file, which will be written into the manifest directory.
```
luffykai, line 34: what's init file?

book/src/writing-apps/build.md
```diff
@@ -9,135 +9,140 @@ The command `cargo openvm build` compiles the program on host to an executable f
...
-  ```bash
-  cargo openvm build --bin my_bin
-  ```
+  **Description**: Excludes the specified packages. Must be used in conjunction with `--workspace`. This flag may be specified multiple times or as a comma-separated list.
 
-- `--example <NAME>`
+### Target Selection
 
-  **Description**: Restricts the build to the example target with the given name, similar to `cargo build --example <NAME>`. Projects often include code samples or demos under the examples directory, and this flag focuses on compiling a specific example.
+By default all package libraries and binaries will be built. To build samples or demos under the `examples` directory, use either the `--example` or `--examples` option.
```
luffykai, line 56: how do we decide which one to transpile

book/src/writing-apps/overview.md
```diff
@@ -36,20 +36,23 @@ If you are providing input for a struct of type `T` that will be deserialized by
 
 ## Generating a Proof
 
-To generate a proof, you first need to generate a proving and verifying key:
+To generate a proof, you can run: first need to generate a proving and verifying key:
```
luffykai, line 39: remove things after run: ?

crates/cli/src/global.rs
stephenh-axiom-xyz: I see what you mean, but I still think it's important to ensure commands are using the directories. Let's maybe keep the `OnceLock`s, but move their initialization elsewhere?

book/src/writing-apps/build.md
stephenh-axiom-xyz: All targets that are built are transpiled

book/src/writing-apps/build.md
```diff
@@ -9,135 +9,140 @@ The command `cargo openvm build` compiles the program on host to an executable f
...
 Right now we use `riscv32im-risc0-zkvm-elf` target which is available in the [Rust toolchain](https://doc.rust-lang.org/rustc/platform-support/riscv32im-risc0-zkvm-elf.html), but we will contribute an OpenVM target to Rust in the future.
 
-## Build flags
+## Build Flags
 
-The following flags are available for the `cargo openvm build` command:
+The following flags are available for the `cargo openvm build` command. You can run `cargo openvm build --help` for this list within the command line.
 
-- `--manifest-dir <MANIFEST_DIR>`
+Generally, outputs will always be built to the **target directory**, which will either be determined by the manifest path or explicitly set using the `--target-dir` option. OpenVM-specific artifacts will be placed in `${target_dir}/openvm/`, but if `--output-dir` is specified they will be copied to `${output-dir}/` as well.
```
luffykai, line 16: might be worth to mention the default target directory when not explicitly set

book/src/writing-apps/build.md
```diff
@@ -9,135 +9,140 @@ The command `cargo openvm build` compiles the program on host to an executable f
...
-  **Description**: Restricts the build to the example target with the given name, similar to `cargo build --example <NAME>`. Projects often include code samples or demos under the examples directory, and this flag focuses on compiling a specific example.
+By default all package libraries and binaries will be built. To build samples or demos under the `examples` directory, use either the `--example` or `--examples` option.
 
-  **Usage Example**:
+- `--lib`
 
-  ```bash
-  cargo openvm build --example my_example
-  ```
+  **Description**: Builds the package's library.
```
luffykai, line 60: when is this useful? i thought the use case always involves some kind of executable?

book/src/writing-apps/build.md
```diff
@@ -9,135 +9,140 @@ The command `cargo openvm build` compiles the program on host to an executable f
...
+
+  **Description**: Builds the specified example. This flag may be specified multiple times or as a comma-separated list.
+
+- `--examples`
+
+  **Description**: Builds all example targets.
+
+- `--all-targets`
+
+  **Description**: Builds all package targets. Equivalent to specifying `--lib` `--bins` `--examples`.
```
luffykai, line 80: i thought previous we way bin and example are mutually exclusive?

book/src/writing-apps/build.md
```diff
@@ -9,135 +9,140 @@ The command `cargo openvm build` compiles the program on host to an executable f
...
-```bash
-cargo openvm run
-    --exe <path_to_transpiled_program>
-    --config <path_to_app_config>
-    --input <path_to_input>
-```
+  **Description**: Prevents Cargo from accessing the network for any reason.
 
-If `--exe` and/or `--config` are not provided, the command will search for these files in `./openvm/app.vmexe` and `./openvm.toml` respectively. If `./openvm.toml` is not present, a default configuration will be used.
+- `--ignore-rust-version`
```
luffykai, line 146: dupe?

book/src/writing-apps/run.md
```diff
@@ -0,0 +1,141 @@
...
+  **Description**: Runs with the given profile. Common profiles are `dev` (faster builds, less optimization) and `release` (slower builds, more optimization). For more information on profiles, see [Cargo's reference page](https://doc.rust-lang.org/cargo/reference/profiles.html).
+
+  **Default**: `release`
+
+### Output Options
+
+- `--target_dir <TARGET_DIR>`
+
+  **Description**: Directory for all generated artifacts and intermediate files. Defaults to directory `target/` at the root of the workspace.
+
```
luffykai, line 89: maybe just point to build page for the rest

crates/cli/src/commands/run.rs
```diff
@@ -2,33 +2,259 @@ use std::path::PathBuf;
...
+            config: args.config,
+            output_dir: args.output_dir,
+            init_file_name: args.init_file_name,
+            ..Default::default()
+        }
+    }
+}
+
+#[derive(Clone, Parser)]
+pub struct RunCargoArgs {
```
luffykai, line 77: is there no way to avoid this large dupe code?

crates/cli/src/util.rs
```diff
@@ -50,3 +57,89 @@ pub fn find_manifest_dir(mut current_dir: PathBuf) -> Result<PathBuf> {
     }
     Ok(current_dir)
 }
+
+pub(crate) fn get_manifest_path_and_dir(
```
luffykai, line 61: can you just make the functions here pub instead of `pub(crate)`, I think some might be useful for the cloud CLI

luffykai (approved):

book/src/writing-apps/build.md
stephenh-axiom-xyz: Doesn't really do anything special, mainly here for alignment with `cargo build`. I guess it will build the library to the risc0 target?

book/src/writing-apps/build.md
stephenh-axiom-xyz: Each individual target can either be a bin or example, but `build` can compile everything at once.

crates/cli/src/commands/run.rs
stephenh-axiom-xyz: There are probably ways of using marginally less dupe code, but the sets of instructions (as well as their descriptions) are different

PR #1667, opened 2025-05-20
feat: commit command for cargo openvm

stephenh-axiom-xyz:
Resolves INT-4066

### DIFF
--- a/book/src/writing-apps/prove.md
+++ b/book/src/writing-apps/prove.md
@@ -49,3 +49,15 @@ The `app` subcommand generates an application-level proof, the `stark` command g
 > In order to run the `evm` subcommand, you must have previously called the costly `cargo openvm setup`, which requires very large amounts of computation and memory (~200 GB).
 
 See [EVM Proof Format](./verify.md#evm-proof-json-format) for details on the output format for `cargo openvm prove evm`.
+
+## Commit Hashes
+
+To see the commit hash for an executable, you may run:
+
+```bash
+cargo openvm commit
+    --app-pk <path_to_app_pk>
+    --exe <path_to_transpiled_program>
+```
+
+The `commit` command has all the auxiliary options that `prove` does, and outputs both u32 and Bn254 commits for both your executable and VM. Commits are written to  `${target_dir}/openvm/` (and `--output-dir` if specified).
--- a/book/src/writing-apps/prove.md
+++ b/book/src/writing-apps/prove.md
@@ -60,4 +60,4 @@ cargo openvm commit
     --exe <path_to_transpiled_program>
 ```
 
-The `commit` command has all the auxiliary options that `prove` does, and outputs both u32 and Bn254 commits for both your executable and VM. Commits are written to  `${target_dir}/openvm/` (and `--output-dir` if specified).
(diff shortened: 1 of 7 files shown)


crates/sdk/src/fs.rs
```diff
@@ -21,85 +22,124 @@ pub const EVM_HALO2_VERIFIER_PARENT_NAME: &str = "Halo2Verifier.sol";
...
-    serde_json::to_writer(File::create(path)?, &proof)?;
-    Ok(())
+    write_to_file_json(&path, proof).map_err(|e| write_error("EVM proof", path, e))
+}
+
+pub fn read_app_exe_commit_from_file<P: AsRef<Path>>(path: P) -> Result<AppExecutionCommit> {
+    read_from_file_json(&path).map_err(|e| read_error("app execution commit", path, e))
+}
+
+pub fn write_app_exe_commit_to_file<P: AsRef<Path>>(
```
jonathanpwang, line 124: can we stop adding more of these single line functions? they aren't really helpful and @luffykai pointed out a while back we should just have a single `write_to_file_json` function for all object types

stephenh-axiom-xyz: We can, but we'll have to keep in mind that the read/write errors that are naturally returned won't be as descriptive to the user.

jonathanpwang: I think the error context could probably be provided at the point of call. and also the most descriptive error message is often just printing the file not found path

stephenh-axiom-xyz: I agree with you, but right now the error (when you try to read an `app_pk` that doesn't exist`) simply says "File does not exist" without any other context. I'll use `write_to_file_json` for the commits, but everything else let's change at a later time?

jonathanpwang: that should be an error message that `write_to_file_json` always prints out (the file path)

book/src/writing-apps/prove.md
```diff
@@ -49,3 +49,15 @@ The `app` subcommand generates an application-level proof, the `stark` command g
...
+
+To see the commit hash for an executable, you may run:
+
+```bash
+cargo openvm commit
+    --app-pk <path_to_app_pk>
+    --exe <path_to_transpiled_program>
+```
+
+The `commit` command has all the auxiliary options that `prove` does, and outputs both u32 and Bn254 commits for both your executable and VM. Commits are written to  `${target_dir}/openvm/` (and `--output-dir` if specified).
```
jonathanpwang, line 63: update to remove u32 commit

crates/sdk/src/fs.rs
```diff
@@ -198,6 +212,19 @@ pub(crate) fn write_to_file_bitcode<T: Serialize, P: AsRef<Path>>(path: P, data:
     Ok(())
 }
 
+pub fn read_from_file_json<T: DeserializeOwned, P: AsRef<Path>>(path: P) -> Result<T> {
+    let ret: T = serde_json::from_reader(File::open(path)?)?;
```
jonathanpwang, line 216: here you should add an error context of the `path`

jonathanpwang (approved):
LGTM after updating book comment

On the errors, I think the shared function can always log the path in the error, which addresses the main issue. If you still prefer not to address it, we can leave it

PR #1669, opened 2025-05-20
test unaligned keccak calls 

arayikhalatyan:


extensions/keccak256/guest/src/lib.rs
```diff
@@ -41,14 +43,42 @@ pub fn keccak256(input: &[u8]) -> [u8; 32] {
...
-        rs1 = In bytes,
-        rs2 = In len
-    );
+    let input = if bytes as usize % 4 != 0 {
+        let mut aligned_buff: Vec<u8> = alloc_zkvm::vec::Vec::with_capacity(len);
+        // let mut aligned_buff = [0u8; 1000];
+        unsafe {
+            core::ptr::copy_nonoverlapping::<u8>(bytes, aligned_buff.as_mut_ptr() as *mut u8, len);
+        }
+        black_box(aligned_buff.as_ptr() as *const u8)
```
arayikhalatyan, line 52: probably spent an hour debugging this line. Without the `black_box` this doesn't work for some reason

Golovanov399: Does `aligned_buff` live long enough?

Golovanov399, line 52:
```suggestion
    let mut aligned_buff = Vec<u8>::new();
    let input = if bytes as usize % 4 != 0 {
        aligned_buff = alloc_zkvm::vec::Vec::with_capacity(len);
        unsafe {
            core::ptr::copy_nonoverlapping::<u8>(bytes, aligned_buff.as_mut_ptr() as *mut u8, len);
        }
        aligned_buff.as_ptr() as *const u8
```

extensions/keccak256/guest/src/lib.rs
```diff
@@ -1,7 +1,9 @@
 #![no_std]
 
 #[cfg(target_os = "zkvm")]
-use core::mem::MaybeUninit;
+extern crate alloc as alloc_zkvm;
```
jonathanpwang, line 4: huh why was this rename needed

extensions/keccak256/guest/src/lib.rs
```diff
@@ -41,14 +43,42 @@ pub fn keccak256(input: &[u8]) -> [u8; 32] {
...
-    openvm_platform::custom_insn_r!(
-        opcode = OPCODE,
-        funct3 = KECCAK256_FUNCT3,
-        funct7 = KECCAK256_FUNCT7,
-        rd = In output,
-        rs1 = In bytes,
-        rs2 = In len
-    );
+    let input = if bytes as usize % 4 != 0 {
+        let mut aligned_buff: Vec<u8> = alloc_zkvm::vec::Vec::with_capacity(len);
```
jonathanpwang, line 47: I don't understand why this would align it; this is not the right way to align things properly

jonathanpwang: oh because we assume our allocators always align to 4

arayikhalatyan: I thought in guest programs the allocator always allocates 4 byte aligned memory. I tried allocating using some memory `Layout` to provide the alignment as well but couldn't find any no-std solutions

jonathanpwang: put a note of this form: https://github.com/openvm-org/openvm/blob/c771a213f5e7f0732e0ddbafb273e15d99c5049d/crates/toolchain/openvm/src/io/mod.rs#L72

extensions/keccak256/guest/src/lib.rs
jonathanpwang: yes you are returning the ptr but it will drop the Vec

you can do `ManuallyDrop`, but I think it's better if you can just directly pass in the vector's pointer to avoid weirdness (if you manually drop and then don't drop at the end of function, you've caused a memory leak)

PR #1670, opened 2025-05-20
feat: init command for cargo openvm

stephenh-axiom-xyz:
Resolves INT-4001

### DIFF
--- a/crates/cli/templates/main.rs
+++ b/crates/cli/templates/main.rs
@@ -1,0 +1,10 @@
+#![cfg_attr(target_os = "zkvm", no_main)]
+#![cfg_attr(target_os = "zkvm", no_std)]
+
+use openvm::process::exit;
+
+openvm::entry!(main);
+
+pub fn main() {
+    exit();
+}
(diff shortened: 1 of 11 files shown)


crates/cli/templates/main.rs
```diff
@@ -0,0 +1,10 @@
+#![cfg_attr(target_os = "zkvm", no_main)]
```
jonathanpwang, line 1: look at the book's quickstart: for starter templates we use std and enable the "std" feature now

crates/cli/templates/main.rs
```diff
@@ -0,0 +1,10 @@
...
+#![cfg_attr(target_os = "zkvm", no_std)]
+
+use openvm::process::exit;
+
+openvm::entry!(main);
+
+pub fn main() {
+    exit();
```
jonathanpwang, line 9: maybe fibonacci is a better example?

crates/cli/templates/openvm.toml
```diff
@@ -0,0 +1,2 @@
+[app_vm_config.rv32i]
+[app_vm_config.rv32m]
```
jonathanpwang, line 2: add io

jonathanpwang (approved):
Re-read the book's quickstart and writing programs section: we recommend `std` for bins and `no_std` for libs. Please update templates to match if possible. (If there is any `bin`, you can enable the `std` feature)

PR #1671, opened 2025-05-21
fix: handle keccak256 input/output not 4-byte aligned

jonathanpwang:
We removed support for memory read/writes that aren't 4-byte aligned (except risc-v loadh, loadb), so now the keccak guest binding must handle the input misalignment cases


jonathanpwang:
@arayikhalatyan you can do the same for sha

Golovanov399 (approved):

jonathanpwang:
@arayikhalatyan @Golovanov399 maybe we should move `AlignedBuffer` to `openvm_platform` since SHA will need it?

PR #1674, opened 2025-05-21
fix: misaligned accesses to sha256

arayikhalatyan:
Fixes misaligned accesses to sha256. Moved Aligned_buf to openvm platform so both `sha2` and `keccak` can access it

### DIFF
--- a/crates/toolchain/platform/src/aligned_buf.rs
+++ b/crates/toolchain/platform/src/aligned_buf.rs
@@ -1,0 +1,65 @@
+extern crate alloc;
+use {core::mem::MaybeUninit};
+
+use alloc::alloc::{alloc, dealloc, handle_alloc_error, Layout};
+use core::ptr::NonNull;
+
+// Register, Memory, and IO address spaces require 4-byte alignment
+pub const MIN_ALIGN: usize = 4;
+
+/// Bytes aligned to 4 bytes.
+pub struct AlignedBuf {
+    pub ptr: *mut u8,
+    pub layout: Layout,
+}
+
+impl AlignedBuf {
+    /// Allocate a new buffer whose start address is aligned to 4 bytes.
+    pub fn uninit(len: usize) -> Self {
+        let layout = Layout::from_size_align(len, MIN_ALIGN).unwrap();
+        if layout.size() == 0 {
+            return Self {
+                ptr: NonNull::<u32>::dangling().as_ptr() as *mut u8,
+                layout,
+            };
+        }
+        // SAFETY: `len` is nonzero
+        let ptr = unsafe { alloc(layout) };
+        if ptr.is_null() {
+            handle_alloc_error(layout);
+        }
+        AlignedBuf { ptr, layout }
+    }
+
+    /// Allocate a new buffer whose start address is aligned to 4 bytes
+    /// and copy the given data into it.
+    ///
+    /// # Safety
+    /// - `bytes` must not be null
+    /// - `len` should not be zero
+    ///
+    /// See [alloc]. In particular `data` should not be empty.
+    pub unsafe fn new(bytes: *const u8, len: usize) -> Self {
+        let buf = Self::uninit(len);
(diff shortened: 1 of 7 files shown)


crates/toolchain/platform/src/aligned_buf.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Bytes aligned to 4 bytes.
+pub struct AlignedBuf {
+    pub ptr: *mut u8,
+    pub layout: Layout,
+}
+
+impl AlignedBuf {
+    /// Allocate a new buffer whose start address is aligned to 4 bytes.
+    pub fn uninit(len: usize) -> Self {
```
jonathanpwang, line 18: let's make this take `align: usize` as an argument to generalize

crates/toolchain/platform/src/aligned_buf.rs
```diff
@@ -0,0 +1,65 @@
+extern crate alloc;
+use {core::mem::MaybeUninit};
+
+use alloc::alloc::{alloc, dealloc, handle_alloc_error, Layout};
+use core::ptr::NonNull;
+
+// Register, Memory, and IO address spaces require 4-byte alignment
+pub const MIN_ALIGN: usize = 4;
```
jonathanpwang, line 8: I think this doesn't need to be here and should be in keccak / sha respectively.

For example, doesn't sha actually need MIN_ALIGN = 8? or you will let the adapter handle 8 <> 4?

crates/toolchain/platform/src/lib.rs
```diff
@@ -12,6 +12,9 @@ mod getrandom;
 pub mod heap;
 #[cfg(all(feature = "export-libm", target_os = "zkvm"))]
 mod libm_extern;
+#[cfg(target_os = "zkvm")]
+pub mod aligned_buf;
```
jonathanpwang, line 16: nit: maybe we call it `alloc`?

extensions/sha256/guest/src/lib.rs
```diff
@@ -8,25 +13,71 @@ pub const SHA256_FUNCT7: u8 = 0x1;
...
+/// # Safety
+///
+/// The VM accepts the preimage by pointer and length, and writes the
+/// 32-byte hash.
+/// - `bytes` must point to an input buffer at least `len` long.
+/// - `output` must point to a buffer that is at least 32-bytes long.
+/// - `bytes` and `output` must be 4-byte aligned.
+#[cfg(target_os = "zkvm")]
+#[inline(always)]
+fn __native_sha256(bytes: *const u8, len: usize, output: *mut u8) {
```
jonathanpwang, line 41: nit move this private function under the extern function

extensions/sha256/guest/src/lib.rs
```diff
@@ -8,25 +13,71 @@ pub const SHA256_FUNCT7: u8 = 0x1;
 /// The sha256 cryptographic hash function.
 #[inline(always)]
 pub fn sha256(input: &[u8]) -> [u8; 32] {
-    let mut output = [0u8; 32];
-    set_sha256(input, &mut output);
-    output
+    #[cfg(not(target_os = "zkvm"))]
+    {
+        let mut output = [0u8; 32];
+        set_sha256(input, &mut output);
```
jonathanpwang, line 19: to reduce the diff (for security), can we keep `sha256` the same? I think it will amount to the same thing in the end anyways

jonathanpwang (approved):
lgtm except nits

I think `AlignedBuffer` constructors `uninit, new` should be generalized to not hardcode the min alignment.

In terms of the actual min alignment used, just wanted to check for sha what your thinking was

crates/toolchain/platform/src/aligned_buf.rs
arayikhalatyan: That's a good point, changed so when we need to allocate a new memory we allocate with 'preferred' alignment.

PR #1675, opened 2025-05-21
feat: CLI prove use bin name for output path

stephenh-axiom-xyz:
Resolves INT-4074. Primarily accomplishes the following:

- `cargo openvm prove` outputs proofs to `${bin_name}.app.proof` instead of `app.proof`, where `bin_name` is the file stem of the executable (same for `stark` and `evm`)
- `cargo openvm verify` by default searches the working directory for files with extension `.app.proof`

Additionally, the following are also in this PR:

- `cargo openvm init ${dir}` creates directory `dir` if it doesn't already exist
- Updates to the book to reflect recent CLI changes

### DIFF
--- a/book/src/writing-apps/verify.md
+++ b/book/src/writing-apps/verify.md
@@ -10,7 +10,9 @@ cargo openvm verify app
     --proof <path_to_proof>
 ```
 
-Options `--manifest-path`, `--target-dir` are also available to `verify`. If you omit `--app_vk` and/or `--proof`, the command will search for those files at `${target_dir}/openvm/app.vk` and `./app.proof` respectively.
+Options `--manifest-path`, `--target-dir` are also available to `verify`. If you omit `--app_vk` the command will search for the verifying key at `${target_dir}/openvm/app.vk`.
+
+If you omit `--proof`, the command will search the working directory for files with the `.app.proof` extension. Note that for this default case a single proof is expected to be found, and `verify` will fail otherwise.
 
 ## EVM Level
 
@@ -93,11 +95,11 @@ cargo openvm prove evm --input <path_to_input>
 cargo openvm verify evm --proof <path_to_proof>
 ```
 
-If `proof` is omitted, the `verify` command will search for the proof at `./openvm/evm.proof`.
+If `proof` is omitted, the `verify` command will search for a file with extension `.evm.proof` in the working directory.
 
 ### EVM Proof: JSON Format
 
-The EVM proof is written to `evm.proof` as a JSON of the following format:
+The EVM proof is written as a JSON of the following format:
 
 ```json
 {
--- a/crates/cli/src/commands/prove.rs
+++ b/crates/cli/src/commands/prove.rs
@@ -38,11 +38,10 @@ enum ProveSubCommand {
         #[arg(
             long,
             action,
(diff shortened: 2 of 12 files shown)


book/src/writing-apps/verify.md
```diff
@@ -10,7 +10,9 @@ cargo openvm verify app
     --proof <path_to_proof>
 ```
 
-Options `--manifest-path`, `--target-dir` are also available to `verify`. If you omit `--app_vk` and/or `--proof`, the command will search for those files at `${target_dir}/openvm/app.vk` and `./app.proof` respectively.
+Options `--manifest-path`, `--target-dir` are also available to `verify`. If you omit `--app_vk` the command will search for the verifying key at `${target_dir}/openvm/app.vk`.
+
+If you omit `--proof`, the command will search the working directory for files with the `.app.proof` extension. Note that for this default case a single proof is expected to be found, and `verify` will fail otherwise.
```
luffykai, line 15: I think `verify` should always pass in `--proof` to be explicit

crates/cli/src/commands/prove.rs
```diff
@@ -243,5 +257,8 @@ pub(crate) fn load_or_build_and_commit_exe(
 
     let app_exe = read_exe_from_file(exe_path)?;
     let committed_exe = sdk.commit_app_exe(app_pk.app_fri_params(), app_exe)?;
-    Ok(committed_exe)
+    Ok((
+        committed_exe,
+        exe_path.file_stem().unwrap().to_str().unwrap().to_string(),
```
luffykai, line 262: maybe `exe_path.file_stem().unwrap().to_string_lossy().into_owned()`?

book/src/writing-apps/verify.md
stephenh-axiom-xyz: Not necessarily opposed to this, but our CLI toolchain generally has the semantic where we allow file inputs to be omitted if there's exactly one option. Given that we error out if there isn't exactly one option, I think it might be good to maintain this behavior in `verify`. Lmk what you think!

luffykai: I think when you do `prove`, because the input (exe) is a generated, and is never really being "used" directly outside of the openvm context, we can use some reasonable default.
But I think the "proof" is a bit different in that it's the final object that when will used elsewhere (onchain?) so would make sense to be explicit. Want to avoid the case that user think a proof is verified, but actually they verify something else (on the default path)

wdyt?

stephenh-axiom-xyz: From earlier discussions with Jonathan, I think we should assume that all of the generated artifacts (i.e. `.pk`, `.vk`, `.vmexe`, etc.) can and likely will be used somewhere. Default naming is done using the project's structure, which the user should in theory know about and set.

I don't think it's a stretch to assume that users will be able to deduce that a `verify` call with no `--proof` specified will look in the working directory, especially since that's the output behavior of `prove`? In general I feel like if a user does `prove --proof file.proof` they should know to do `verify --proof file.proof`.

That being said, regardless of what we choose we should probably output a success message with the proof path. As of now there's no messaging, which does indeed make it difficult to confirm what's going on 😅

luffykai: oh i think printing the proof path is a good idea. ok i think this is probably fine

PR #1676, opened 2025-05-21
fix: integration tests

arayikhalatyan:


jonathanpwang:
fixed by https://github.com/openvm-org/openvm/pull/1678

PR #1677, opened 2025-05-21
feat(new-execution): add interfaces for running different execute functions

shuklaayush:
- this is targeting `feat/new-exec-native-ext` since it was built on top of it


benchmarks/execute/src/main.rs
```diff
@@ -173,30 +182,50 @@ fn main() -> Result<()> {
...
                     .unzip()
             };
 
             let executor = VmExecutor::new(vm_config);
-            executor
-                .execute_e1(exe.clone(), vec![])
-                // .execute_metered(exe.clone(), vec![], widths, interactions)
-                .expect("Failed to execute program");
+
+            // E2 to find segment points
```
jonathanpwang, line 205: this was just for testing? don't we want to keep the benchmark E1?

crates/vm/src/arch/execution_mode/tracegen/normal.rs
```diff
@@ -0,0 +1,104 @@
...
+    ) -> bool {
+        state.clk >= self.clk_end
+    }
+
+    fn on_start(
+        &mut self,
+        state: &mut VmSegmentState<Self::Ctx>,
+        chip_complex: &mut VmChipComplex<F, VC::Executor, VC::Periphery>,
+    ) {
+        let timestamp = chip_complex.memory_controller().timestamp();
```
jonathanpwang, line 51: I think this is always 1, but depends on how we're using memory_controller

crates/vm/src/arch/execution_mode/tracegen/normal.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fn on_suspend_or_terminate(
+        &mut self,
+        state: &mut VmSegmentState<Self::Ctx>,
+        chip_complex: &mut VmChipComplex<F, VC::Executor, VC::Periphery>,
+        exit_code: Option<u32>,
+    ) {
+        // TODO(ayush): this should ideally not be here
+        self.final_memory = Some(chip_complex.base.memory_controller.memory_image().clone());
```
jonathanpwang, line 64: why do we need it? why do we need final_memory at all?

@Golovanov399 wdyt?

jonathanpwang (approved):
LGTM, but I think the execution tests should be reverted?

benchmarks/execute/src/main.rs
shuklaayush: Yes, I'll fix this. Though with codspeed, we can probably remove the old execute tests that were using this file

crates/vm/src/arch/execution_mode/tracegen/normal.rs
shuklaayush: I'll check

PR #1679, opened 2025-05-22
fix: integration test

arayikhalatyan:
All the integration tests should pass now. Merged with Native's PR so some of the tests could compile, I think for simplicity this can be merged after the native's PR.


jonathanpwang:
the tests involving native were infinite looping due to the bug @Golovanov399 is fixing

jonathanpwang (approved):

PR #1681, opened 2025-05-22
feat: port public values chip and native adapter

shuklaayush:


crates/vm/src/system/public_values/core.rs
```diff
@@ -120,6 +118,8 @@ pub struct PublicValuesRecord<F> {
 /// the proof but in the perspective of constraints, it could be any value.
 pub struct PublicValuesCoreStep<A, F> {
     adapter: A,
+    // TODO(ayush): put air here and take from air
```
arayikhalatyan, line 121: I thought we wanted to decouple the AIR and STEP? There is a similar situation for sha256's STEP

crates/vm/src/system/public_values/core.rs
```diff
@@ -169,32 +170,48 @@ where
...
-        // // TODO(ayush): add this check
-        // // debug_assert_eq!(core_row.width(), BaseAir::<F>::width(&self.air));
+        Ok(())
+    }
 
-        // core_row.is_valid = F::ONE;
-        // core_row.value = record.value;
-        // core_row.index = record.index;
+    fn fill_trace_row(&self, mem_helper: &MemoryAuxColsFactory<F>, row_slice: &mut [F]) {
+        let (adapter_row, core_row) = unsafe { row_slice.split_at_mut_unchecked(A::WIDTH) };
```
arayikhalatyan, line 204:
```suggestion
        let (adapter_row, core_row) = row_slice.split_at_mut(A::WIDTH);
```

Is there a benefit to doing an unsafe split?

crates/vm/src/system/native_adapter/mod.rs
```diff
@@ -214,7 +180,34 @@ where
...
+            reads[0][0] = tracing_read_or_imm_native(
+                memory,
+                e.as_canonical_u32(),
+                b,
+                &mut cols.reads_aux[0].address.address_space,
+                &mut cols.reads_aux[0].read_aux,
+            );
+        }
+        if R >= 2 {
+            cols.reads_aux[0].address.pointer = b;
```
arayikhalatyan, line 201:
```suggestion
            cols.reads_aux[1].address.pointer = c;
```

was it supposed to be like this?

arayikhalatyan (approved):
lgtm, just minor nits and comments

crates/vm/src/system/public_values/core.rs
shuklaayush: i think it avoids bounds checking at runtime. not sure if the marginal benefit is worth it though. i just followed the patter of how it was being done in other areas of the code

PR #1687, opened 2025-05-23
fix(new-execution): add missing pc increment

shuklaayush:
- added a missing pc increment step in `ModularIsEqualStep`


arayikhalatyan (approved):
wow, good catch! I am wondering how this didn't come up earlier, ie with integration tests

PR #1689, opened 2025-05-23
feat: ability to verify e2e STARK proof + CLI command

stephenh-axiom-xyz:
Resolves INT-4064

### DIFF
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -340,14 +340,9 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
             return Err(eyre::eyre!("Missing program AIR"));
         } else if proof.proof.per_air[1].air_id != CONNECTOR_AIR_ID {
             return Err(eyre::eyre!("Missing connector AIR"));
+        } else if proof.proof.per_air[2].air_id != PUBLIC_VALUES_AIR_ID {
+            return Err(eyre::eyre!("Missing public values AIR"));
         }
-
-        let public_values_air_proof_data = proof
-            .proof
-            .per_air
-            .iter()
-            .find(|p| p.air_id == PUBLIC_VALUES_AIR_ID)
-            .ok_or_else(|| eyre::eyre!("Missing public values AIR"))?;
 
         let vm_commit = proof.proof.commitments.main_trace[PROGRAM_CACHED_TRACE_INDEX].as_ref();
         let internal_commit: &[_; CHUNK] = &agg_stark_pk
@@ -363,6 +358,7 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
         let e = E::new(vm_pk.fri_params);
         e.verify(&vm_pk.vm_pk.get_vk(), &proof.proof)?;
 
+        let public_values_air_proof_data = &proof.proof.per_air[2];
         let pvs: &VmVerifierPvs<_> =
             public_values_air_proof_data.public_values[..VmVerifierPvs::<u8>::width()].borrow();
 


crates/sdk/src/lib.rs
```diff
@@ -322,6 +326,51 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
         Ok(proof)
     }
 
+    pub fn verify_e2e_stark_proof(
```
nyunyunyunyu, line 329: Taking https://github.com/openvm-org/openvm/blob/4da0af980e5baebb92e61839e83f8ddb17943adf/crates/vm/src/arch/vm.rs#L735 as an example and see what's missing.

crates/sdk/src/lib.rs
```diff
@@ -322,6 +326,51 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
...
+        &self,
+        agg_stark_pk: AggStarkProvingKey,
+        proof: &VmStarkProof<SC>,
+    ) -> Result<[F; CHUNK]> {
+        let program_commit =
```
nyunyunyunyu, line 334: Check if all required AIRs are present. Otherwise `.main_trace[PROGRAM_CACHED_TRACE_INDEX]` could be about another AIR.

crates/sdk/src/lib.rs
```diff
@@ -322,6 +326,51 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
         Ok(proof)
     }
 
+    pub fn verify_e2e_stark_proof(
+        &self,
+        agg_stark_pk: AggStarkProvingKey,
```
nyunyunyunyu, line 331: Pass a reference instead of a value.

crates/sdk/src/lib.rs
```diff
@@ -322,6 +326,51 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
...
+        proof: &VmStarkProof<SC>,
+    ) -> Result<[F; CHUNK]> {
```
nyunyunyunyu, line 333: Return `AppExecutionCommit` instead of just exe commit.

crates/sdk/src/lib.rs
```diff
@@ -322,6 +326,51 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
...
+        };
+        let e = E::new(vm_pk.fri_params);
+        e.verify(&vm_pk.vm_pk.get_vk(), &proof.proof)?;
+
+        let public_values_air_proof_data = proof
+            .proof
+            .per_air
+            .iter()
+            .find(|p| p.air_id == PUBLIC_VALUES_AIR_ID)
+            .unwrap();
```
nyunyunyunyu, line 354: Since this function returns a `Result`, throw an error instead of panic when `PublicValueAir` is absent.

crates/sdk/src/lib.rs
```diff
@@ -322,6 +326,51 @@ impl<E: StarkFriEngine<SC>> GenericSdk<E> {
...
+        let pvs: &VmVerifierPvs<_> =
+            public_values_air_proof_data.public_values[..VmVerifierPvs::<u8>::width()].borrow();
+
```
nyunyunyunyu, line 357: Check if program terminates with exit code = 0.

nyunyunyunyu (changes requested):

crates/sdk/src/lib.rs
stephenh-axiom-xyz: `vm_commit` should be `internal_committed_exe.get_program_commit()`? Or should I get the program commit from the root verifier program?

crates/sdk/src/lib.rs
nyunyunyunyu: The above checks cannot guarantee `proof.proof.commitments.main_trace[PROGRAM_CACHED_TRACE_INDEX]` is the program trace?

crates/sdk/src/lib.rs
nyunyunyunyu: It's the program commit of leaf verifier or a public value of internal verifier. See here: https://github.com/openvm-org/openvm/blob/0888016a053b04148f4395d7d39c408815a0cde6/crates/continuations/src/verifier/root/mod.rs#L73

nyunyunyunyu (approved):

PR #1691, opened 2025-05-24
feat: Patch ecdsa `VerifyingKey::recover_from_prehash` to skip message verification

Avaneesh-axiom:
The k256 and p256 guest libraries use the `ecdsa::VerifyingKey::recover_from_prehash` method which recovers the key and then verifies the message with that key. Our ecdsa implementation in the ECC extension does not do the final verification step. Furthermore, the latest pre-release version of the `ecdsa` crate also skips this step.

This PR patches the `VerifyingKey` struct to skip the final message verification in the `recover_from_prehash` method. This is achieved by wrapping `ecdsa::VerifyingKey`.

Changes:
- ECC extension
    - The `VerifyingKey` and `PublicKey` structs have been renamed to `OpenVMVerifyingKey` and `OpenVMPublicKey`
    - A new struct `VerifyingKey` which patches the function was created
- Guest libraries
    - the k256 and p256 guest libraries now re-export the new `openvm_ecc_guest::ecdsa::VerifyingKey` struct.

Guest programs should now import `openvm_k256::ecdsa::VerifyingKey` instead of `ecdsa::VerifyingKey` to access the new changes.


jonathanpwang (approved):
LGTM

I tried several things besides `OnceBox` for `set_up_once` but somehow it gives the best compiler optimizations by a long shot.

PR #1692, opened 2025-05-27
feat: e4 implementation for rv32im and bigint

arayikhalatyan:
After a discussion with @jonathanpwang deleted all the `zerocopy` traits.

[Open for suggestions and discussion] Instead used something very similar to `AlignedBorrow`. Made a derive macro called `AlignedBytesBorrow` which similar to `AlignedBorrow` can be derived on `Sized` structs. This will ensure that the struct is aligned and `repr(c)` makes sure the layout of the fields is correct so we don't need to add padding fields manually. 
For `non-Sized` record structs (e.g. those that use vector) a specialized implementation will be needed for `Borrow` and `BorrowMut` traits. Seems like the `zerovec` crate might be useful for this.

Update: All the chips `Bigint` and `RV32im` are now ported except `hintstore` 
Note: some things are commented out to make it compile 

My philosophy of the record structs is to make them as small as theoretically possible and make the `execute` as light as possible. Because of that `fill_trace_row` suffers. There is a call to `run_..` in both `execute` and `fill_trace_row`. Whenever trivial made a `fast_run_...` method to be called in `execute`. Take a look at `branch_eq` for an example. 

TODO:

- [ ] Remove unnecessary opcode conversions
- [x] Make `AdapterTraceFiller : AdapterTraceStep` (implementing `TraceStep : TraceFiller` introduced unnecessary complications)
- [x] Move out of unsafe when possible
- [ ] Jalr keeps duplicate `from_pc`, perhaps we can avoid that

Resolves INT-4048
Resolves INT-4050

### DIFF
--- a/crates/circuits/primitives/derive/src/lib.rs
+++ b/crates/circuits/primitives/derive/src/lib.rs
@@ -74,26 +74,22 @@ pub fn aligned_borrow_derive(input: TokenStream) -> TokenStream {
 }
 
 /// `S` is the type the derive macro is being called on
-/// Implements Borrow<S> and BorrowMut<S> for [AT]
-/// [AT] has to have (checked via `debug_assert!`s)
+/// Implements Borrow<S> and BorrowMut<S> for [u8]
+/// [u8] has to have (checked via `debug_assert!`s)
 /// - at least size_of(S) length
-/// - at least align_of(S) alignemnt
+/// - at least align_of(S) alignment
 #[proc_macro_derive(AlignedBytesBorrow)]
 pub fn aligned_bytes_borrow_derive(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as DeriveInput);
-    let mut ast_clone = ast.clone();
     let name = &ast.ident;
 
     // Get impl generics, type generics, where clause
     // Note, need to add the new type generic to the `impl_generics`
-    let (_, type_generics, where_clause) = ast.generics.split_for_impl();
-    ast_clone.generics.params.push(syn::parse_quote!(T));
-    let impl_generics = ast_clone.generics.split_for_impl().0;
+    let (impl_generics, type_generics, where_clause) = ast.generics.split_for_impl();
 
     let methods = quote! {
-        impl #impl_generics core::borrow::Borrow<#name #type_generics> for [T]
+        impl #impl_generics core::borrow::Borrow<#name #type_generics> for [u8]
         where
-            T: Sized,
             #where_clause
         {
             fn borrow(&self) -> &#name #type_generics {
@@ -104,9 +100,8 @@ pub fn aligned_bytes_borrow_derive(input: TokenStream) -> TokenStream {
             }
         }
 
-        impl #impl_generics core::borrow::BorrowMut<#name #type_generics> for [T]
+        impl #impl_generics core::borrow::BorrowMut<#name #type_generics> for [u8]
         where
-            T: Sized,
             #where_clause
         {
             fn borrow_mut(&mut self) -> &mut #name #type_generics {
--- a/crates/vm/src/arch/integration_api.rs
+++ b/crates/vm/src/arch/integration_api.rs
@@ -203,10 +203,17 @@ pub struct MatrixRecordArena<F> {
 }
 
 impl<F: Field> MatrixRecordArena<F> {
-    pub fn alloc_single_row(&mut self) -> &mut [F] {
+    pub fn alloc_single_row(&mut self) -> &mut [u8] {
         let start = self.trace_offset;
         self.trace_offset += self.width;
-        &mut self.trace_buffer[start..self.trace_offset]
+        let row_slice = &mut self.trace_buffer[start..self.trace_offset];
+        let size = size_of_val(row_slice);
+        let ptr = row_slice as *mut [F] as *mut u8;
+        // SAFETY:
+        // - `ptr` is non-null
+        // - `size` is correct
+        // - alignment of `u8` is always satisfied
+        unsafe { &mut *std::ptr::slice_from_raw_parts_mut(ptr, size) }
     }
 }
 
@@ -236,7 +243,6 @@ impl<F: Field> RowMajorMatrixArena<F> for MatrixRecordArena<F> {
 /// A record arena struct that can be used by chips that
 /// - have a single row per instruction
 /// - have trace row = [adapter_row, core_row]
-/// TODO(arayi): come up with a better name
 pub struct AdapterCoreRecordArena<F> {
     inner: MatrixRecordArena<F>,
 }
@@ -254,11 +260,12 @@ impl<'a, F: Field, A, C> RecordArena<'a, AdapterCoreLayout, (&'a mut A, &'a mut 
 where
     A: Sized,
     C: Sized,
-    [F]: BorrowMut<A> + BorrowMut<C>,
+    [u8]: BorrowMut<A> + BorrowMut<C>,
 {
     fn alloc(&'a mut self, layout: AdapterCoreLayout) -> (&'a mut A, &'a mut C) {
         let buffer = self.inner.alloc_single_row();
-        let (adapter_buffer, core_buffer) = buffer.split_at_mut(layout.adapter_width);
+        let (adapter_buffer, core_buffer) =
+            buffer.split_at_mut(layout.adapter_width * size_of::<F>());
 
         let adapter_record: &mut A = adapter_buffer.borrow_mut();
         let core_record: &mut C = core_buffer.borrow_mut();
@@ -267,7 +274,6 @@ where
     }
 }
 
-// TODO: make a macro
 impl<F: Field> RowMajorMatrixArena<F> for AdapterCoreRecordArena<F> {
     fn with_capacity(height: usize, width: usize) -> Self {
         Self {
--- a/extensions/rv32-adapters/src/vec_heap.rs
+++ b/extensions/rv32-adapters/src/vec_heap.rs
@@ -439,7 +439,18 @@ impl<
 {
     fn fill_trace_row(&self, mem_helper: &MemoryAuxColsFactory<F>, adapter_row: &mut [F]) {
         let record = unsafe {
-            let record_buffer = &*slice_from_raw_parts(adapter_row.as_ptr(), adapter_row.len());
+            let record_buffer = &*slice_from_raw_parts(
+                adapter_row.as_ptr() as *const u8,
+                size_of::<
+                    Rv32VecHeapAdapterRecord<
+                        NUM_READS,
+                        BLOCKS_PER_READ,
+                        BLOCKS_PER_WRITE,
+                        READ_SIZE,
+                        WRITE_SIZE,
+                    >,
+                >(),
+            );
             let record: &Rv32VecHeapAdapterRecord<
                 NUM_READS,
                 BLOCKS_PER_READ,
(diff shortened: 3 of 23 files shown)


crates/vm/src/system/memory/offline_checker/mod.rs
```diff
@@ -5,10 +5,59 @@ mod columns;
...
+        (*self).into()
+    }
+}
+
+impl AsMut<u32> for Ru32 {
+    // **SAFETY** Self must be 4-byte aligned
+    #[inline(always)]
+    fn as_mut(&mut self) -> &mut u32 {
+        unsafe { &mut *(self.0.as_ptr() as *mut u32) }
+    }
```
arayikhalatyan, line 40: Didn't want to add a runtime check here. However, we need to make sure all the `Ru32` fields in `record` structs are 4-byte aligned by adding padding fields as necessary.

extensions/rv32im/circuit/src/adapters/alu.rs
```diff
@@ -176,17 +175,16 @@ pub struct Rv32BaseAluAdapterStep<const LIMB_BITS: usize> {
...
+#[derive(FromBytes, IntoBytes, KnownLayout, Immutable, Debug)]
 pub struct Rv32BaseAluAdapterRecord {
     pub from_pc: u32,
     pub from_timestamp: u32,
 
     // Pack u8 together for alignment
-    pub rd_ptr: u8,
-    pub rs1_ptr: u8,
+    pub rd_ptr: u32,
+    pub rs1_ptr: u32,
```
arayikhalatyan, line 185: register pointers can be in the range `[0, 2^pointer_max_bits)`, however if we change that to be small enough to keep in a `u8` it would save a lot of space.

extensions/rv32im/circuit/src/adapters/alu.rs
```diff
@@ -176,17 +175,16 @@ pub struct Rv32BaseAluAdapterStep<const LIMB_BITS: usize> {
...
     /// 1 if rs2 was a read, 0 if an immediate
-    pub rs2_as: u8,
-    pub _padding: u8,
+    pub rs2_as: u32,
```
arayikhalatyan, line 187: Can make `rs2_as` have `u8` type but we need to pad after it anyways so it doesn't make any difference in this case

jonathanpwang:
@arayikhalatyan I hadn't wanted to use `Unaligned` because it loses info and so you're manually checking the alignment stuff in the end anyways. I specialized to `u8` to get `zerocopy` to work, but I wasn't thinking we needed to use zerocopy everywhere. In places where we would have to not use const generic, I think we should just directly implement the `unsafe {}` version where we manually check the safety assertions

extensions/rv32im/circuit/src/adapters/alu.rs
```diff
@@ -223,26 +222,26 @@ impl<F: PrimeField32, CTX, const LIMB_BITS: usize> AdapterTraceStep<F, CTX>
             e.as_canonical_u32() == RV32_REGISTER_AS || e.as_canonical_u32() == RV32_IMM_AS
         );
 
-        record.rs1_ptr = b.as_canonical_u32() as u8;
+        record.rs1_ptr = b.as_canonical_u32();
```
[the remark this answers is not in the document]
jonathanpwang: see INT-3755

crates/circuits/primitives/derive/src/lib.rs
```diff
@@ -73,6 +73,40 @@ pub fn aligned_borrow_derive(input: TokenStream) -> TokenStream {
     TokenStream::from(methods)
 }
 
+/// Implements Borrow<T> and BorrowMut<T> for [u8]
+#[proc_macro_derive(AlignedBytesBorrow)]
```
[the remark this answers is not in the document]
jonathanpwang: ok their `NoUninit` trait doesn't allow padding bytes

crates/vm/src/system/memory/offline_checker/mod.rs
```diff
@@ -5,10 +5,16 @@ mod columns;
...
+#[derive(Debug)]
 pub struct MemoryReadAuxRecord {
     pub prev_timestamp: u32,
 }
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct MemoryWriteAuxRecord<const NUM_LIMBS: usize> {
+    pub prev_timestamp: u32,
+    pub prev_data: [u8; NUM_LIMBS],
```
jonathanpwang, line 19: I hadn't put this struct here (in this crate) because `u8` is specialized to RV32 stuff (in principle the fully general version should use `F`)

extensions/rv32-adapters/src/vec_heap.rs
```diff
@@ -408,44 +458,118 @@ impl<
             WRITE_SIZE,
         > = adapter_row.borrow_mut();
 
-        let mut timestamp = cols.from_state.timestamp.as_canonical_u32();
-        let mut timestamp_pp = || {
-            timestamp += 1;
-            timestamp - 1
-        };
+        unsafe {
```
jonathanpwang, line 461: @shuklaayush had a comment that we only need to wrap the truly unsafe part in `unsafe`
technically the entire block is indeed unsafe since you continually need to worry about this overlapping issue, so I'm not sure which way to fall here

extensions/rv32im/circuit/src/adapters/jalr.rs
```diff
@@ -139,6 +144,19 @@ impl<AB: InteractionBuilder> VmAdapterAir<AB> for Rv32JalrAdapterAir {
...
 }
 
+#[repr(C)]
+#[derive(AlignedBytesBorrow, Debug)]
+pub struct Rv32JalrAdapterRecord {
+    pub from_pc: u32,
+    pub from_timestamp: u32,
+
+    pub rs1_ptr: u32,
+    pub rd_ptr: Option<u32>,
```
jonathanpwang, line 154: I think we should avoid `Option` if possible? It has bad alignment. I think we can use `u32::MAX` as an indicator

extensions/rv32im/circuit/src/adapters/mod.rs
```diff
@@ -38,13 +35,6 @@ pub const RV_B_TYPE_IMM_BITS: usize = 13;
 
 pub const RV_J_TYPE_IMM_BITS: usize = 21;
 
-#[repr(C)]
-#[derive(FromBytes, IntoBytes, KnownLayout, Immutable)]
-pub struct Rv32WordWriteAuxRecord {
```
jonathanpwang, line 43: see my note above

jonathanpwang (approved):
changes make sense: after looking into alternatives like `bytemuck`, seems like `AlignedBytesBorrow` is still the simplest for what we want

overall lgtm

extensions/rv32im/circuit/src/jalr/core.rs
```diff
@@ -337,43 +339,25 @@ where
...
 
-        // TODO(ayush): should this be [u8; 4]?
-        let (to_pc, rd) = run_jalr(local_opcode, *state.pc, imm_extended, rs1);
+        let (to_pc, rd) = run_jalr(*state.pc, rs1, c.as_canonical_u32() as u16, g.is_one());
         let rd = rd.map(|x| x as u8);
 
         self.adapter.write(state.memory, instruction, &rd);
 
-        *state.pc = to_pc;
+        *state.pc = to_pc & !1;
```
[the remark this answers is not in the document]
arayikhalatyan: I kinda needed two different `run` functions but the only difference would have been the `& !1` with `to_pc`. So moved that out of the function to avoid having 2 almost exactly the same functions😅

extensions/rv32-adapters/src/vec_heap.rs
arayikhalatyan: The only argument I have here is that in more complicated chips we might want to have other `unsafe` parts and we won't want it all to be wrapped in one big unsafe block

crates/vm/src/arch/integration_api.rs
```diff
@@ -235,6 +240,64 @@ impl<F: Field> RowMajorMatrixArena<F> for MatrixRecordArena<F> {
     }
 }
 
+/// A record arena struct that can be used by chips that
+/// - have a single row per instruction
+/// - have trace row = [adapter_row, core_row]
+/// TODO(arayi): come up with a better name
+pub struct AdapterCoreRecordArena<F> {
```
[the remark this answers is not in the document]
jonathanpwang: but it's just trait implementations on it, so doesn't seem like there's any harm in using same struct?

crates/vm/src/system/memory/offline_checker/mod.rs
jonathanpwang: I think it's hard, which is why I was thinking we just have 2 separate structs, one for `u8` and one for `F`

extensions/rv32-adapters/src/vec_heap.rs
jonathanpwang: yea so maybe we just localize the unsafe?

extensions/rv32im/circuit/src/jalr/core.rs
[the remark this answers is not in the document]
jonathanpwang: what is this `& !1` doing? and why wasn't it there in the old version?

crates/vm/src/system/memory/offline_checker/mod.rs
arayikhalatyan: uuh what's the benefit of 2 different structs vs `WriteAux<u8>+WriteAux<F>` 
The issue with `loadstore` is that sometimes we need a record of type `F` and sometimes of type `u8`. Probably will have to keep a `u32` array for `prev_data` in `laodstore` since can't have two different types of records

crates/vm/src/arch/integration_api.rs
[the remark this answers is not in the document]
arayikhalatyan: Do you mean having different implementations of `RecordArena` trait for `MatrixRecordArena`? I thought that's not allowed in Rust

crates/vm/src/system/memory/offline_checker/mod.rs
jonathanpwang: I wasn't sure you would be able to get `AlignedByteBorrow` to work?

crates/circuits/primitives/derive/src/lib.rs
```diff
@@ -73,6 +73,54 @@ pub fn aligned_borrow_derive(input: TokenStream) -> TokenStream {
     TokenStream::from(methods)
 }
 
+/// `S` is the type the derive macro is being called on
+/// Implements Borrow<S> and BorrowMut<S> for [AT]
+/// [AT] has to have (checked via `debug_assert!`s)
+/// - at least size_of(S) length
+/// - at least align_of(S) alignemnt
```
jonathanpwang, line 80: typo

crates/vm/src/arch/integration_api.rs
```diff
@@ -235,6 +233,61 @@ impl<F: Field> RowMajorMatrixArena<F> for MatrixRecordArena<F> {
...
+        let (adapter_buffer, core_buffer) = buffer.split_at_mut(layout.adapter_width);
+
+        let adapter_record: &mut A = adapter_buffer.borrow_mut();
+        let core_record: &mut C = core_buffer.borrow_mut();
+
+        (adapter_record, core_record)
+    }
+}
+
+// TODO: make a macro
```
jonathanpwang, line 270: since you made the trait, can remove this comment

jonathanpwang (approved):
lgtm after reverting `AlignedBytesBorrow` to only borrow on `&[u8]`

also fix a small typo

PR #1695, opened 2025-05-28
feat: make base_alu tests pass and make register address u8

arayikhalatyan:
This PR makes the `base_alu` unit tests run so we have a working example with the new `e4` traits. As discussed with @jonathanpwang we can set the register addresses to be in `0..255` so we can keep them in `u8` to save space

The main PR for porting the chips is [here](https://github.com/openvm-org/openvm/pull/1692)

### DIFF
--- a/extensions/rv32im/circuit/src/adapters/alu.rs
+++ b/extensions/rv32im/circuit/src/adapters/alu.rs
@@ -19,13 +19,12 @@ use openvm_circuit::{
 use openvm_circuit_primitives::{
     bitwise_op_lookup::{BitwiseOperationLookupBus, SharedBitwiseOperationLookupChip},
     utils::not,
-    TraceSubRowGenerator,
 };
 use openvm_circuit_primitives_derive::AlignedBorrow;
 use openvm_instructions::{
     instruction::Instruction,
     program::DEFAULT_PC_STEP,
-    riscv::{RV32_IMM_AS, RV32_REGISTER_AS},
+    riscv::{RV32_IMM_AS, RV32_MAX_REGISTER_ADDRESS, RV32_REGISTER_AS},
 };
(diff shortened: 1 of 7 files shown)


extensions/rv32im/circuit/src/adapters/alu.rs
```diff
@@ -218,10 +217,14 @@ impl<F: PrimeField32, CTX, const LIMB_BITS: usize> AdapterTraceStep<F, CTX>
     ) -> Self::ReadData {
         let &Instruction { b, c, d, e, .. } = instruction;
 
-        debug_assert_eq!(d.as_canonical_u32(), RV32_REGISTER_AS);
-        debug_assert!(
-            e.as_canonical_u32() == RV32_REGISTER_AS || e.as_canonical_u32() == RV32_IMM_AS
-        );
+        #[cfg(debug_assertions)]
+        {
+            assert!(b.as_canonical_u32() <= RV32_MAX_REGISTER_ADDRESS);
```
jonathanpwang, line 222: should it be `<`?

jonathanpwang (commented):
I realized it might be difficult to restrict address space 1 to only <256:
> The address space `e` can be `0`, `1`, or `2` for load instructions, and `2`, `3`, or `4` for store instructions.

extensions/rv32im/circuit/src/adapters/alu.rs
jonathanpwang: ah max=255

jonathanpwang:
My bad on suggesting this: it would require changes to OpenVM ISA (ban `e=1` in LOAD*_RV32 instructions) to ensure all OpenVM assembly code has address space 1 pointers <256, so we should just go with `u32` for now.

PR #1697, opened 2025-05-29
feat: ensure all CLI commit outputs are in hex + fix verify STARK

stephenh-axiom-xyz:
Resolves INT-3880 and INT-4073. Implements the following:

- All CLI commit outputs are in hex
- `cargo openvm prove stark` outputs a JSON similar to `cargo openvm prove evm`
- Updates STARK verification to check app executable and VM commits

`cargo openvm prove stark` output looks like this:
```
{
  "app_exe_commit": "004807b0289c7eeaa279e610af936bec5fa6f677d4f2da64bb53f39d37debe3c",
  "app_vm_commit": "00249a27204ce75ed124a9cc45a722cab52744d57400c6967901260e5a510cd2",
  "user_public_values": "200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  "proof": "0100000002000..."
}
```

### DIFF
--- a/crates/sdk/src/commit.rs
+++ b/crates/sdk/src/commit.rs
@@ -19,8 +19,9 @@ use serde_with::serde_as;
 
 use crate::{types::BN254_BYTES, NonRootCommittedExe, F, SC};
 
-/// Wrapper for an array of big-endian bytes, representing a Bn254. Each commit can be
-/// converted to a Bn254 and/or base-F::MODULUS number as a u32 digest.
+/// Wrapper for an array of big-endian bytes, representing an unsigned big integer. Each commit can
+/// be converted to a Bn254Fr using the trivial identification as natural numbers or into a `u32`
+/// digest by decomposing the big integer base-`F::MODULUS`.
 #[serde_as]
 #[derive(Copy, Clone, Debug, Serialize, Deserialize)]
 pub struct CommitBytes(#[serde_as(as = "serde_with::hex::Hex")] [u8; BN254_BYTES]);
--- a/crates/sdk/src/commit.rs
+++ b/crates/sdk/src/commit.rs
@@ -14,15 +14,15 @@ use openvm_stark_sdk::{
     p3_bn254_fr::Bn254Fr,
 };
 use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
 
-use crate::{NonRootCommittedExe, F, SC};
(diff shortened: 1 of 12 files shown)


crates/sdk/src/commit.rs
```diff
@@ -108,3 +111,39 @@ pub fn commit_app_exe(
     let app_engine = BabyBearPoseidon2Engine::new(app_fri_params);
     Arc::new(VmCommittedExe::<SC>::commit(exe, app_engine.config.pcs()))
 }
+
+/// Byte representation of AppExecutionCommit. Because the commits in AppExecutionCommit
+/// are stored as an array representing a base-F number, the 256 bits that make up each
+/// commit are different than actual the 256-bit representation of the Bn254. This class
+/// stores the latter to enable consistent serialization between different proof outputs.
```
stephenh-axiom-xyz, line 118: In general we currently do need this struct (in addition to `AppExecutionCommit`) for hex serialization. Because `u32`s in `AppExecutionCommit` represent field elements (which don't have typical division/modulo operations), I believe we can't efficiently convert the raw bytes of a `Bn254` back into its `[F; DIGEST_SIZE]` representation.

crates/sdk/src/commit.rs
```diff
@@ -108,3 +111,39 @@ pub fn commit_app_exe(
...
+#[serde_as]
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub struct AppExecutionCommitBytes {
+    #[serde_as(as = "serde_with::hex::Hex")]
+    /// 1 Bn254Fr public value for app exe commit in big-endian bytes.
+    pub app_exe_commit: [u8; BN254_BYTES],
+    #[serde_as(as = "serde_with::hex::Hex")]
+    /// 1 Bn254Fr public value for app vm commit in big-endian bytes.
+    pub app_vm_commit: [u8; BN254_BYTES],
```
jonathanpwang, line 127: I'd prefer is `[u8; BN254_BYTES]` is some protected newtype that's opaque (so user just seems it's some Commit type) and then we convert, as you've done, to Bn254Fr, [F;8], etc etc

yi-sun (approved):
I only checked the interface formats and they look good.

jonathanpwang (approved):
LGTM

PR #1698, opened 2025-05-29
fix(new-execution): use unpadded height for segmentation

shuklaayush:
- use unpadded height when calculating max height and total cells during segmentation
- use default segmentation thresholds from reth benchmark


crates/vm/src/arch/execution_mode/metered/mod.rs
```diff
@@ -71,31 +70,29 @@ impl<'a> MeteredExecutionControl<'a> {
 
     fn should_segment(&mut self, state: &mut VmSegmentState<MeteredCtx>) -> bool {
         let trace_heights = state.ctx.trace_heights_if_finalized();
-        let max_trace_cells = MAX_TRACE_CELLS_PER_CHIP * trace_heights.len();
         for (i, &height) in trace_heights.iter().enumerate() {
-            let padded_height = height.next_power_of_two();
-            if padded_height > MAX_TRACE_HEIGHT {
+            if height > MAX_TRACE_HEIGHT {
```
jonathanpwang, line 74: `MAX_TRACE_HEIGHT` isn't configurable?

jonathanpwang (approved):

PR #1699, opened 2025-05-30
refactor: VM Executor for E1/E2

nyunyunyunyu:
- Add `InterpretedInstance` for E1/E2.
- Enforce `ExecutionControl` to be stateless.
- `InterpretedInstance` E2 is not tested because context initialization is tricky. Will address in the future PR.

close INT-4043

### DIFF
--- a/crates/vm/src/arch/execution_control.rs
+++ b/crates/vm/src/arch/execution_control.rs
@@ -4,6 +4,10 @@ use openvm_stark_backend::p3_field::PrimeField32;
 use super::{ExecutionError, VmChipComplex, VmConfig, VmSegmentState};
 
 /// Trait for execution control, determining segmentation and stopping conditions
+/// Invariants:
+/// - `ExecutionControl` should be stateless.
+/// - For E3/E4, `ExecutionControl` is for a specific execution and cannot be used for another
+///   execution with different inputs or segmentation criteria.
 pub trait ExecutionControl<F, VC>
 where
     F: PrimeField32,
--- a/crates/vm/src/arch/execution_control.rs
+++ b/crates/vm/src/arch/execution_control.rs
@@ -12,30 +12,32 @@ where
     /// Host context
     type Ctx;
 
+    fn initialize_context(&self) -> Self::Ctx;
+
     /// Determines if execution should suspend
     fn should_suspend(
-        &mut self,
+        &self,
         state: &mut VmSegmentState<Self::Ctx>,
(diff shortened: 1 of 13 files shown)


crates/vm/src/arch/vm.rs
```diff
@@ -415,100 +427,100 @@ where
...
-                if code != ExitCode::Success as u32 {
-                    return Err(ExecutionError::FailedWithExitCode(code));
-                }
-            }
-            None => return Err(ExecutionError::DidNotTerminate),
-        };
-
-        Ok(executor.ctrl.segments)
-    }
+    // pub fn execute_metered(
```
jonathanpwang, line 430: probably need to uncomment this before merging

crates/vm/src/arch/execution_control.rs
```diff
@@ -4,6 +4,10 @@ use openvm_stark_backend::p3_field::PrimeField32;
 use super::{ExecutionError, VmChipComplex, VmConfig, VmSegmentState};
 
 /// Trait for execution control, determining segmentation and stopping conditions
+/// Invariants:
```
nyunyunyunyu, line 7: Clarify some invariants about `ExecutionControl` here. @shuklaayush @jonathanpwang

crates/vm/src/arch/vm.rs
nyunyunyunyu: I commented this out because it's not covered by any tests. Added this back but not 100% sure it's working.

jonathanpwang (approved):
lgtm

crates/vm/src/arch/execution_control.rs
shuklaayush: i feel like stateless might be too strict. i'd say more like immutable, in the sense that it should be initialized with some configuration parameters that should not change during execution. all the things that do change will be in `VmState/VmSegmentState/Ctx`

crates/vm/src/arch/execution_control.rs
```diff
@@ -12,38 +16,40 @@ where
     /// Host context
     type Ctx;
 
+    fn initialize_context(&self) -> Self::Ctx;
```
shuklaayush, line 19: i feel like we might need a different constructor for each `ctx` with different arguments. for e3, the context will have the trace buffers and would need the trace heights and widths as input during initialization
for your approach to work, we need to keep everything that is needed to initialize `ctx` in execution control. maybe that's fine but it's's not clear to me atm

PR #1705, opened 2025-06-03
feat: e4 rewrite of mod-builder

teokitan:
Implemented E4 for mod-builder. Created new `FieldExpressionCoreRecord` that represents minimal execution data needed for deferred trace generation, saved by CPU during the call to `execute()` and that GPU could use to reconstruct and do tracegen in a future implementation (currently this still falls back to CPU). Unit tests testing it in isolation pass but further testing should be done with adapters once they are ready for E4.

Currently `FieldExpressionCoreRecord` stores input limbs and flags in a fixed-size array, which should be changed to vectors once variable sized containers are supported (as discussed with @arayikhalatyan).

This resolves INT-4122.

### DIFF
--- a/crates/circuits/mod-builder/src/core_chip.rs
+++ b/crates/circuits/mod-builder/src/core_chip.rs
@@ -1,5 +1,3 @@
-use std::ptr::slice_from_raw_parts;
-
 use itertools::Itertools;
 use num_bigint::BigUint;
 use num_traits::Zero;
@@ -178,27 +176,26 @@ pub struct FieldExpressionMetadata {
     pub total_input_limbs: usize, // num_inputs * limbs_per_input
 }
 
-pub struct FieldExpressionCoreRecord<'a> {
+pub struct FieldExpressionCoreRecordMut<'a> {
     pub opcode: &'a mut u8,
-    pub input_limbs: &'a mut [u32],
+    pub input_limbs: &'a mut [u8],
 }
 
-impl<'a> CustomBorrow<'a, FieldExpressionCoreRecord<'a>, FieldExpressionMetadata> for [u8] {
+impl<'a> CustomBorrow<'a, FieldExpressionCoreRecordMut<'a>, FieldExpressionMetadata> for [u8] {
     fn custom_borrow(
         &'a mut self,
         metadata: FieldExpressionMetadata,
-    ) -> FieldExpressionCoreRecord<'a> {
-        let (opcode_buf, rest) = unsafe { self.split_at_mut_unchecked(1) };
+    ) -> FieldExpressionCoreRecordMut<'a> {
+        let (opcode_buf, input_limbs_buff) = unsafe { self.split_at_mut_unchecked(1) };
 
-        let (_, input_limbs_buff, _) = unsafe { rest.align_to_mut::<u32>() };
-        FieldExpressionCoreRecord {
+        FieldExpressionCoreRecordMut {
             opcode: &mut opcode_buf[0],
             input_limbs: &mut input_limbs_buff[..metadata.total_input_limbs],
         }
     }
 }
 
-impl<'a> FieldExpressionCoreRecord<'a> {
+impl<'a> FieldExpressionCoreRecordMut<'a> {
     pub fn new_from_execution_data(
         buffer: &'a mut [u8],
         inputs: &[BigUint],
@@ -217,12 +214,7 @@ impl<'a> FieldExpressionCoreRecord<'a> {
         // Rust will assert that length of `data` and `self.input_limbs` are the same
         // That is `data.len() == num_inputs * limbs_per_input`
         *self.opcode = opcode;
-
-        let input = self.input_limbs as *mut _ as *mut u8;
-        unsafe {
-            // Copying bytes to u32s, assuming everything is in little endian
-            std::ptr::copy_nonoverlapping(data.as_ptr(), input, data.len());
-        }
+        self.input_limbs.copy_from_slice(data);
     }
 }
 
@@ -301,7 +293,7 @@ where
         >,
 {
     type RecordLayout = AdapterCoreLayout<FieldExpressionMetadata>;
-    type RecordMut<'a> = (A::RecordMut<'a>, FieldExpressionCoreRecord<'a>);
+    type RecordMut<'a> = (A::RecordMut<'a>, FieldExpressionCoreRecordMut<'a>);
 
     fn execute<'buf, RA>(
         &mut self,
@@ -363,7 +355,7 @@ where
 
         self.adapter.fill_trace_row(mem_helper, adapter_row);
 
-        let record: FieldExpressionCoreRecord = unsafe {
+        let record: FieldExpressionCoreRecordMut = unsafe {
             get_record_from_slice(
                 &mut core_row,
                 FieldExpressionMetadata {
@@ -412,30 +404,11 @@ where
         Ctx: E1E2ExecutionCtx,
     {
         let data: &[u8] = &self.adapter.read(state, instruction).into().0;
-        let data_ptr = data.as_ptr();
-        let writes = if data_ptr as usize % align_of::<u32>() != 0 {
-            // Can't safely transmute bytes to u32s, because `data` is not guaranteed to be aligned
-            let data_u32s = data
-                .chunks(4)
-                .map(|chunk| u32::from_le_bytes(chunk.try_into().unwrap()))
-                .collect_vec();
-            run_field_expression(
-                self,
-                &data_u32s,
-                instruction.opcode.local_opcode_idx(self.offset) as usize,
-            )
-            .0
-        } else {
-            let data_u32s = unsafe {
-                &*slice_from_raw_parts(data_ptr as *const u32, self.expr.canonical_num_limbs())
-            };
-            run_field_expression(
-                self,
-                data_u32s,
-                instruction.opcode.local_opcode_idx(self.offset) as usize,
-            )
-            .0
-        };
+        let (writes, _, _) = run_field_expression(
+            self,
+            data,
+            instruction.opcode.local_opcode_idx(self.offset) as usize,
+        );
 
         self.adapter.write(state, instruction, &writes.into());
         *state.pc = state.pc.wrapping_add(DEFAULT_PC_STEP);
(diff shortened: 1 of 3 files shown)


arayikhalatyan (commented):
Update: Added implementation for `mod_is_equal` and `vec_heap_two_reads` so @teokitan  can run/test the `mod-builder` on `ecc` and `algebra` modular tests

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -24,11 +29,129 @@ use openvm_stark_backend::{
...
+#[repr(C)]
+#[derive(AlignedBytesBorrow, Debug)]
+pub struct FieldExpressionCoreRecord {
+    pub from_pc: u32,
+    pub from_timestamp: u32,
+    pub opcode: u8,
+    pub num_inputs: u8,
+    pub limbs_per_input: u8,
+    pub num_flags: u8,
+}
```
arayikhalatyan, line 44: I think we don't need these in the record for `FieldExression`. `from_pc`, `from_timestamp` will be taken care of by the adapter. `num_inputs`, `limbs_pre_input`,  and `num_flags` can be derived from the `Step`'s fields (`expr`) so no need to keep them in a record. Perhaps we only need the `opcode` and we can use that opcode to reconstruct the `flag_states`. So we can remove the flag states from the variable length stuff. Btw because of how we keep the records this can be a reason why the tests fail

arayikhalatyan (commented):
Merged the `hintstore-e4` into this branch. Edited the `AdapterCoreRecordArena` so that it also supports `CustomBorrow` with metadata. I think it is better if we merge this PR after the `hintstore-e4` is merged

arayikhalatyan (approved):
All the unit tests pass and about half the integration tests pass. The failing integration tests give `segment proofs should verify: StarkError(OodEvaluationMismatch)` error which is unrelated to the `mod-builder` and is happening for yet to be known reasons. 

I made some changes that broke the `core_chip`'s unit tests. LGTM after those are fixed.

arayikhalatyan (approved):
Ok after the `basealu` adapter fix on `hintstore`'s PR all the integration tests pass for `algebra` and `pairing`. Unable to compile the `ecc` int tests because it needs e4 for native and keccak

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -170,6 +170,63 @@ where
     }
 }
 
+#[derive(Clone)]
```
jonathanpwang, line 173: you can derive Copy

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -170,6 +170,63 @@ where
...
+        inputs: &[BigUint],
+        _limb_bits: usize,
+        limbs_per_input: usize,
+    ) -> Self {
+        let record_info = FieldExpressionMetadata {
+            total_input_limbs: inputs.len() * limbs_per_input,
+        };
+
+        let record: Self = buffer.custom_borrow(record_info);
+        // record.fill_from_execution_data(opcode, );
```
jonathanpwang, line 217: ? delete if old comment

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -170,6 +170,63 @@ where
...
+            inner: record_buf.borrow_mut(),
+            input_limbs: &mut input_limbs_buf[..metadata.total_input_limbs],
+        }
+    }
+}
+
+impl<'a> FieldExpressionCoreRecordMut<'a> {
+    pub fn new_from_execution_data(
+        buffer: &'a mut [u8],
+        _opcode: u8,
```
jonathanpwang, line 207: why are there these unused fields?

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -170,6 +170,63 @@ where
...
 }
 
+#[derive(Clone)]
+pub struct FieldExpressionMetadata {
+    pub total_input_limbs: usize, // num_inputs * limbs_per_input
+}
+
+#[repr(C)]
+#[derive(AlignedBytesBorrow, Debug)]
+pub struct FieldExpressionCoreRecord {
```
jonathanpwang, line 180: I got very confused by this: you should only call `FieldExpressionCoreRecordMut` something that is equivalent to a mutable reference of `FieldExpressionCoreRecord`. Currently it's confusing that `Record` is inside `RecordMut`. If only for implementation of `custom_buffer` you could just split via `u8` directly.

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -242,59 +299,101 @@ where
...
+        A::start(*state.pc, state.memory, &mut adapter_record);
 
         let data: DynArray<_> = self
             .adapter
-            .read(state.memory, instruction, adapter_row)
+            .read(state.memory, instruction, &mut adapter_record)
             .into();
 
-        let (writes, inputs, flags) = run_field_expression(self, &data, instruction);
+        core_record.fill_from_execution_data(
```
jonathanpwang, line 332: maybe it's ok for now, but it seems what we really want is that the `adapter` doesn't return `data.0` but directly fills into `core_record`?

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -242,59 +299,101 @@ where
...
+    }
+}
+
+impl<F, CTX, A> TraceFiller<F, CTX> for FieldExpressionStep<A>
+where
+    F: PrimeField32 + Send + Sync + Clone,
+    A: 'static + AdapterTraceFiller<F, CTX>,
+{
+    fn fill_trace_row(&self, mem_helper: &MemoryAuxColsFactory<F>, row_slice: &mut [F]) {
+        // NOTE[teokitan]: This is where GPU acceleration should happen in the future
```
jonathanpwang, line 361: remove: `TraceFiller` is for cpu only, this entire function and more will be replaced by a separate kernel

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -242,59 +299,101 @@ where
...
+
+        // Get the core record from the row slice
+        let (adapter_row, core_row) = unsafe { row_slice.split_at_mut_unchecked(A::WIDTH) };
+
+        self.adapter.fill_trace_row(mem_helper, adapter_row);
+
+        let core_bytes = unsafe {
+            std::slice::from_raw_parts_mut(
```
jonathanpwang, line 369: hm maybe this is my bad, but it does seem pretty verbose if we keep writing this stuff to convert `&[F]` to `&[u8]`. @arayikhalatyan did you have any better way? Maybe we should at least make a helper function.

crates/circuits/mod-builder/src/tests.rs
jonathanpwang, line 1: I'm not sure, but I wonder if we could add some utility functions to decrease the verbosity of the tests? it could get harder to maintain later

jonathanpwang (commented):
I reviewed the mod-builder part and it LGTM aside from some code quality comments.

But @arayikhalatyan will need to provide some more context on all the other stuff that ended up happening in this PR.

crates/circuits/mod-builder/src/core_chip.rs
teokitan: using it as a raw `u8` would mean we lose the safety provided by `#[repr(C)]`, no?

arayikhalatyan: Ah yes, at first there were other fields besides the opcode in this struct but then they turned out to be unnecessary. But now this struct doesn't really make sense.
@teokitan if you have only one field in a `repr(C)` struct the memory layout of the struct is the same as of the inner field

crates/circuits/mod-builder/src/core_chip.rs
arayikhalatyan: I tried doing this some time ago but was getting a lot of errors from rust. But tried it again now and it worked😅

crates/circuits/mod-builder/src/core_chip.rs
arayikhalatyan: Hmm, do you mean making the `adapter::read` function to directly write into some buffer instead of returning some reads?

jonathanpwang: yes

crates/circuits/mod-builder/src/core_chip.rs
```diff
@@ -170,6 +171,53 @@ where
     }
 }
 
+#[derive(Copy, Clone)]
+pub struct FieldExpressionMetadata {
+    pub total_input_limbs: usize, // num_inputs * limbs_per_input
+}
+
+pub struct FieldExpressionCoreRecord<'a> {
+    pub opcode: &'a mut u8,
```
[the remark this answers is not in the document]
arayikhalatyan: The size of `opcode` doesn't affect the alignment of `input_limbs` whatsoever, since we use `align_of`. Also, the alignment of `input_limbs` doesn't matter because when converting it to `BigUints` it does a lot of non-trivial conversions - so probably copies stuff anyway. 
But one thing I realized we can do, is that since we have fixed the limbs to be bytes, we can remove the `limbs_to_biguint` utility function and just use `BigUint::from_le_bytes` or `from_be_bytes` cc @teokitan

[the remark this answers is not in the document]
arayikhalatyan: On second thought, after looking into `from..` stuff in `BigUint` I think it makes sense to make the `input_limbs` as `u32`'s and use `BigUint::from_slice`. Then we can just unsafely copy the `u8`'s we get from adapter into that so not even `u32::from_le_bytes` is needed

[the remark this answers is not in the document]
arayikhalatyan: Pushed the changes for this

jonathanpwang (approved):
LGTM but please either revert the u32 buffer change, or don't user call u32 a limb. Just call it u32 or u32_digit

Also I don't think width should go inside the layout, but let's merge this first and then address that in a separate PR.

crates/circuits/mod-builder/src/core_chip.rs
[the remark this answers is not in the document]
arayikhalatyan: Reverted it back, now it uses `BigUint::from_le_bytes` for conversions (stiil better than it was before). 
Probably `BiguInt::from_le_bytes` vs `BigUint::from_slice` don't have a big perf diff.

PR #1707, opened 2025-06-03
refactor: Move `Streams` into `VmStateMut`

nyunyunyunyu:
Previously some chips hold `Arc<Mutex<Streams>>` to read from `Streams`. Move `Streams` into `VmStateMut` so it has all data about VM state.

### DIFF
--- a/crates/vm/src/arch/vm.rs
+++ b/crates/vm/src/arch/vm.rs
@@ -252,7 +252,6 @@ where
         let chip_complex = create_and_initialize_chip_complex(
             &self.config,
             exe.program.clone(),
-            from_state.input,
             Some(from_state.memory),
         )
         .unwrap();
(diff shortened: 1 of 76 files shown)


nyunyunyunyu:
> To check, the logic to move into `VmStateMut` directly instead of `Ctx` is because it is shared regardless of `Ctx`, so might as well make it global?

Yes, it's not specific to any execution mode

crates/vm/src/arch/vm.rs
```diff
@@ -390,15 +383,10 @@ where
         );
         #[cfg(feature = "bench-metrics")]
         {
-            segment.metrics = state.metrics;
+            segment.metrics = Default::default();
```
jonathanpwang, line 386: why is it changed?

jonathanpwang (approved):
LGTM except I don't understand why the metrics thing changed

crates/vm/src/arch/vm.rs
nyunyunyunyu: Because `state` is unnecessary anymore

jonathanpwang: oh there's just no state at all

PR #1716, opened 2025-06-05
docs: Add comment explaining the `is_identity` optimization

Avaneesh-axiom:
A recent PR (#1709) introduced the optimization of overriding the default implementation of `Group::is_identity` in the `impl_sw_group_ops` macro. The overridden implementation was identical to the default one. This PR adds a comment to explain the code duplication.


jonathanpwang (approved):
LGTM, I rebased onto `main` because the feature branch has been merged

PR #1722, opened 2025-06-09
feat: e{3+4} implementation of sha2 extension

arayikhalatyan:
Implemented the `e{3+4}` for `sha2` extension. All the unit tests and the integration tests pass for it.
It is unfortunate but we have to zero out the trace in the `fill_trace` in order to avoid having accidental garbage values in the trace that overflow the prime.

Resolves INT-4051


crates/vm/src/arch/integration_api.rs
```diff
@@ -368,26 +332,31 @@ where
...
-    }
-
-    fn width(&self) -> usize {
-        self.width
-    }
-
-    fn trace_offset(&self) -> usize {
-        self.trace_offset
+impl<F: Field> MatrixRecordArena<F> {
+    pub fn alloc_single_row(&mut self) -> &mut [u8] {
```
jonathanpwang, line 336: shouldn't this now just call `alloc_buffer(1)`

extensions/sha256/circuit/src/sha256_chip/trace.rs
```diff
@@ -1,44 +1,118 @@
...
+    pub dst_ptr: u32,
+    pub src_ptr: u32,
+    pub len: u32,
+
+    pub register_reads_aux: [MemoryReadAuxRecord; SHA256_REGISTER_READS],
+    pub write_aux: MemoryWriteAuxRecord<SHA256_WRITE_SIZE>,
+}
+
+pub struct Sha256VmRecordMut<'a> {
+    pub inner: &'a mut Sha256VmRecord,
```
jonathanpwang, line 70: you should not call this `Record`, it's confusing when `RecordMut` contains more info than `Record`

extensions/sha256/circuit/src/sha256_chip/trace.rs
```diff
@@ -1,44 +1,118 @@
...
+/// provided at runtime, followed by a slice of `MemoryReadAuxRecord`'s of length
+/// `SHA256_NUM_READ_ROWS * num_blocks`. Uses `align_to_mut()` to make sure the slice is properly
+/// aligned to `MemoryReadAuxRecord`. Has debug assertions that check the size and alignment of the
+/// slices.
+impl<'a> CustomBorrow<'a, Sha256VmRecordMut<'a>, Sha256VmMetadata> for [u8] {
+    fn custom_borrow(&'a mut self, metadata: Sha256VmMetadata) -> Sha256VmRecordMut<'a> {
+        let (record_buf, rest) =
+            unsafe { self.split_at_mut_unchecked(size_of::<Sha256VmRecord>()) };
+
+        let (input, rest) = unsafe {
```
jonathanpwang, line 87: do you not need any alignment on `input` because it's truly raw bytes?

extensions/sha256/circuit/src/sha256_chip/trace.rs
```diff
@@ -1,44 +1,118 @@
...
+            rest.split_at_mut_unchecked((metadata.num_blocks as usize) * SHA256_BLOCK_CELLS)
+        };
+        let (_, read_aux_buf, _) = unsafe { rest.align_to_mut::<MemoryReadAuxRecord>() };
```
jonathanpwang, line 90: I feel like custom borrow maybe should have another function that just returns the number of bytes this thing will actually take, to be used for the dense RecordArena implementation; otherwise it's hard to keep track

extensions/sha256/circuit/src/sha256_chip/trace.rs
```diff
@@ -1,44 +1,118 @@
...
+    pub len: u32,
+
+    pub register_reads_aux: [MemoryReadAuxRecord; SHA256_REGISTER_READS],
+    pub write_aux: MemoryWriteAuxRecord<SHA256_WRITE_SIZE>,
+}
+
+pub struct Sha256VmRecordMut<'a> {
+    pub inner: &'a mut Sha256VmRecord,
+    // Having a continuous slice of the input is useful for fast hashing in `execute`
+    pub input: &'a mut [u8],
```
jonathanpwang, line 72: shouldn't you be able to get this slice from memory? technically read_aux will have it all already so it's wasteful

jonathanpwang: maybe a technicality because memory could be paged

jonathanpwang: putting it here still feels weird - maybe switch back to making a `Vec`, or keep a scratch `Vec` inside the Step itself, and after @pjabbarzade 's PR is merged, we can try to switch to using pure slice

extensions/sha256/circuit/src/sha256_chip/trace.rs
```diff
@@ -48,371 +122,443 @@ impl<F: PrimeField32, CTX> TraceStep<F, CTX> for Sha256VmStep {
...
+                sizes.push((0, num_blocks_so_far));
+                break;
+            } else {
+                let record = unsafe {
+                    let row = trace.as_ptr() as *const u8;
+                    let row = &*slice_from_raw_parts(row, size_of::<Sha256VmRecord>());
+                    let record: &Sha256VmRecord = row.borrow();
+                    record
+                };
+                let num_blocks = ((record.len << 3) as usize + 1 + 64).div_ceil(SHA256_BLOCK_BITS);
```
jonathanpwang, line 241: there should be a helper function for this

extensions/sha256/circuit/src/sha256_chip/mod.rs
```diff
@@ -127,7 +132,7 @@ impl<F: PrimeField32> StepExecutorE1<F> for Sha256VmStep {
         instruction: &Instruction<F>,
         chip_index: usize,
     ) -> Result<()> {
-        let &Instruction {
+        let Instruction {
```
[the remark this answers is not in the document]
arayikhalatyan: Initially, I thought that e2 should be just an e1 wrapper. So changed it to just call `execute_e1` in my commits, then changed it back to the correct form and made it similar to `execute`. But there is no logical change from the Ayush's version. I just added some debug asserts and made address spaces constant.

extensions/sha256/circuit/src/sha256_chip/trace.rs
arayikhalatyan: Yeah I agree, that was my idea as well after `DenceRecordArena` is merged

extensions/sha256/circuit/src/sha256_chip/trace.rs
```diff
@@ -48,371 +122,443 @@ impl<F: PrimeField32, CTX> TraceStep<F, CTX> for Sha256VmStep {
...
+        // First pass over the trace to get the number of blocks for each instruction
+        // and divide the matrix into chunks of needed sizes
+        loop {
+            if num_blocks_so_far * SHA256_ROWS_PER_BLOCK >= rows_used {
+                // Push all the padding rows as a single chunk and break
+                chunks.push(trace);
+                sizes.push((0, num_blocks_so_far));
+                break;
+            } else {
+                let record = unsafe {
```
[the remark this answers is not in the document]
arayikhalatyan: I did it in `keccak-e4` branch accidentally. The result of having too many branches lol

extensions/sha256/circuit/src/sha256_chip/trace.rs
arayikhalatyan: Changed the name to `Sha256VmRecordHeader`

arayikhalatyan:
> LGTM, although we should test this on many sha2 test cases to make sure there's no gaps.
> 
> Also, it seems like perhaps for SHA (and possibly Keccak), we should just always use a dense RecordArena...?
Yeah, dense RecordArena probably makes sense here for sha and keccak🤔

PR #1724, opened 2025-06-09
refactor: make adapter width constant + single struct for record arena

arayikhalatyan:
First, deleted the `AdapterCoreRecordArena` and `MultiRowRecordArena` struct and have a single `MatrixRecordArena` struct. Previously thought Rust doesn't allow multiple implementations of the same trait on a single struct. Turns out it allows as long as they are not conflicting with each other (ie the generics are strictly non-overlapping). In this PR I made 2 implementations of `RecordArena` trait one with `AdapterCoreLayout` and one with `MultiRowLayout`. 
Second, removed the `adapter_width` field from `AdapterCoreLayout`, instead added a generic `A` which is the `AdapterStep` struct and has the `A::WIDTH` constant. Initially, wanted to make it work with a const generic `ADAPTER_WIDTH` but couldn't because rust doesn't allow generics in const expressions.

### DIFF
--- a/crates/vm/src/arch/integration_api.rs
+++ b/crates/vm/src/arch/integration_api.rs
@@ -3,7 +3,7 @@ use std::{
     array::from_fn,
     borrow::{Borrow, BorrowMut},
     marker::PhantomData,
-    ptr::{slice_from_raw_parts, slice_from_raw_parts_mut},
+    ptr::slice_from_raw_parts_mut,
     sync::Arc,
 };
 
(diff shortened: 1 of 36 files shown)


crates/vm/src/arch/integration_api.rs
```diff
@@ -367,26 +332,31 @@ where
...
+impl<F: Field> MatrixRecordArena<F> {
+    pub fn alloc_single_row(&mut self) -> &mut [u8] {
+        let start = self.trace_offset;
+        self.trace_offset += self.width;
+        let row_slice = &mut self.trace_buffer[start..self.trace_offset];
+        let size = size_of_val(row_slice);
+        let ptr = row_slice as *mut [F] as *mut u8;
+        // SAFETY:
+        // - `ptr` is non-null
+        // - `size` is correct
```
Golovanov399, line 344: Well technically you didn't check the bounds of the slice, but they are checked in debug so we probably _do_ want to omit this check in release

Golovanov399 (approved):
I was wondering whether we need the `width` field of the `MatrixRecordArena`, but we probably want, at least for now (for reasons unrelated to allocation and related to `RowMajorMatrixArena` trait)

crates/vm/src/arch/integration_api.rs
Golovanov399: I checked, you are right: bounds are checked even in release

arayikhalatyan: Right, good point! will merge this branch then

PR #1727, opened 2025-06-10
feat: e{3+4} implementation for keccak256 extention

arayikhalatyan:
Implemented the e{3+4} for keccak extension. All the unit tests and the integration tests pass for it.
It is unfortunate but we have to zero out some of the trace in the fill_trace in order to avoid having accidental garbage values in the trace that overflow the prime.

I think it's better if we merge in this order
1. mod-builder-e4
2. sha2-e4
3. refactor record arena
4. keccak

Resolves INT-4052

### DIFF
--- a/extensions/keccak256/circuit/src/lib.rs
+++ b/extensions/keccak256/circuit/src/lib.rs
@@ -4,7 +4,6 @@
 use openvm_circuit_primitives::bitwise_op_lookup::SharedBitwiseOperationLookupChip;
 use openvm_stark_backend::p3_field::PrimeField32;
 use p3_keccak_air::NUM_ROUNDS;
-use tiny_keccak::{Hasher, Keccak};
 
 pub mod air;
 pub mod columns;
@@ -21,7 +20,7 @@ pub use air::KeccakVmAir;
 use openvm_circuit::{
     arch::{
         execution_mode::{metered::MeteredCtx, E1E2ExecutionCtx},
-        ExecutionBridge, NewVmChipWrapper, Result, StepExecutorE1, VmStateMut,
+        ExecutionBridge, MultiRowRecordArena, NewVmChipWrapper, Result, StepExecutorE1, VmStateMut,
     },
     system::memory::online::GuestMemory,
 };
@@ -33,9 +32,11 @@ use openvm_instructions::{
 };
 use openvm_keccak256_transpiler::Rv32KeccakOpcode;
 use openvm_rv32im_circuit::adapters::{
-    memory_read_from_state, memory_write_from_state, read_rv32_register_from_state,
+    memory_write_from_state, read_rv32_register, read_rv32_register_from_state,
 };
 use utils::num_keccak_f;
+
+use crate::utils::keccak256;
 
 // ==== Constants for register/memory adapter ====
 /// Register reads to get dst, src, len
@@ -68,7 +69,7 @@ pub const KECCAK_DIGEST_BYTES: usize = 32;
 /// Number of 64-bit digest limbs.
 pub const KECCAK_DIGEST_U64S: usize = KECCAK_DIGEST_BYTES / 8;
 
-pub type KeccakVmChip<F> = NewVmChipWrapper<F, KeccakVmAir, KeccakVmStep>;
+pub type KeccakVmChip<F> = NewVmChipWrapper<F, KeccakVmAir, KeccakVmStep, MultiRowRecordArena<F>>;
 
 //#[derive(derive_new::new)]
 pub struct KeccakVmStep {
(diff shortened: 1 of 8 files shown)


extensions/keccak256/circuit/src/lib.rs
```diff
@@ -109,31 +110,27 @@ impl<F: PrimeField32> StepExecutorE1<F> for KeccakVmStep {
...
+            .on_memory_operation(RV32_MEMORY_AS, src, len as u32);
+        let message = unsafe {
+            state
+                .memory
+                .memory
+                .get_slice((RV32_MEMORY_AS, src), len as usize)
+        };
+
+        let output = keccak256(&message);
+        memory_write_from_state(state, RV32_MEMORY_AS, dst, &output);
```
shuklaayush, line 133: same. wouldn't work for e2

extensions/keccak256/circuit/src/trace.rs
```diff
@@ -44,326 +120,410 @@ impl<F: PrimeField32, CTX> TraceStep<F, CTX> for KeccakVmStep {
...
-                            row[..NUM_KECCAK_PERM_COLS].copy_from_slice(p3_keccak_row);
+                            // Need to get rid of the accidental garbage data that might overflow the
+                            // F's prime field. Unfortunately, there is no good way around this
+                            unsafe {
+                                std::ptr::write_bytes(
+                                    row.as_mut_ptr().add(NUM_KECCAK_PERM_COLS) as *mut u8,
+                                    0,
+                                    (NUM_KECCAK_VM_COLS - NUM_KECCAK_PERM_COLS) * size_of::<F>(),
+                                );
+                            }
```
shuklaayush, line 277: can we do something like `row[NUM_KECCAK_PERM_COLS..NUM_KECCAK_VM_COLS].fill(F::ZERO)` here instead?

extensions/keccak256/circuit/src/trace.rs
```diff
@@ -44,326 +120,410 @@ impl<F: PrimeField32, CTX> TraceStep<F, CTX> for KeccakVmStep {
...
-
-                    cols.instruction = first_row.instruction;
-                    cols.sponge.block_bytes = first_row.sponge.block_bytes;
-                    cols.sponge.is_padding_byte = first_row.sponge.is_padding_byte;
-                    cols.mem_oc.partial_block = first_row.mem_oc.partial_block;
+                // Copy the read aux records and inner record to another place
+                // to safely fill in the trace matrix without overwriting the record
+                let mut read_aux_records = Vec::with_capacity(num_reads);
+                read_aux_records.extend_from_slice(record.read_aux);
+                let vm_record = record.inner.clone();
```
shuklaayush, line 299: this is unfortunate

extensions/keccak256/circuit/src/trace.rs
```diff
@@ -44,326 +120,410 @@ impl<F: PrimeField32, CTX> TraceStep<F, CTX> for KeccakVmStep {
...
+                                // Update bitwise lookup (i.e. xor) chip state: order matters!
+                                if idx != 0 {
+                                    self.bitwise_lookup_chip
+                                        .request_xor(byte as u32, s_byte as u32);
+                                }
+                                *s ^= (byte as u64) << (i * 8);
+                            }
+                        }
+                        states.push(state);
+                        keccakf(&mut state);
```
shuklaayush, line 336: we'll need to figure out how to do this on gpu if it's not in record

shuklaayush (commented):
dropped some comments

extensions/keccak256/circuit/src/lib.rs
```diff
@@ -109,31 +110,27 @@ impl<F: PrimeField32> StepExecutorE1<F> for KeccakVmStep {
...
+
+        state
+            .ctx
+            .on_memory_operation(RV32_MEMORY_AS, src, len as u32);
+        let message = unsafe {
+            state
+                .memory
+                .memory
+                .get_slice((RV32_MEMORY_AS, src), len as usize)
+        };
```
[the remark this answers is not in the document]
arayikhalatyan: Yeah, realized this after speaking to you the other day. Will work on it today. Loadstore and Hintstore are incorrect too

extensions/keccak256/circuit/src/trace.rs
arayikhalatyan: My reasoning here was that this copying is still parallel over instructions, and it's better to copy here so it allows us to do parallel trace gen over blocks/rows later.

extensions/keccak256/circuit/src/trace.rs
shuklaayush: do you mean to avoid bounds checks?

extensions/keccak256/circuit/src/trace.rs
shuklaayush: makes sense

extensions/keccak256/circuit/src/trace.rs
shuklaayush: sounds good

extensions/keccak256/circuit/src/trace.rs
arayikhalatyan: I think the idea is more like the compiler will optimize a lot if it's zeroing out a slice of memory but might not be able to optimize filling with `F::ZERO`. I had a similar thing with `mem_helper.fill_zero` and @jonathanpwang preferred the raw `write_bytes`

extensions/keccak256/circuit/src/lib.rs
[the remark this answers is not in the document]
arayikhalatyan: Pushed the keccak fixes here, and made a PR for `loadstore` and `hintstore` fixes

extensions/keccak256/circuit/src/lib.rs
```diff
@@ -109,31 +110,27 @@ impl<F: PrimeField32> StepExecutorE1<F> for KeccakVmStep {
...
+        debug_assert_eq!(d.as_canonical_u32(), RV32_REGISTER_AS);
+        debug_assert_eq!(e.as_canonical_u32(), RV32_MEMORY_AS);
+
+        let dst = read_rv32_register_from_state(state, a.as_canonical_u32());
+        let src = read_rv32_register_from_state(state, b.as_canonical_u32());
+        let len = read_rv32_register_from_state(state, c.as_canonical_u32());
+
+        state
+            .ctx
+            .on_memory_operation(RV32_MEMORY_AS, src, len as u32);
```
shuklaayush, line 124:
```suggestion
```

extensions/keccak256/circuit/src/lib.rs
```diff
@@ -109,31 +110,24 @@ impl<F: PrimeField32> StepExecutorE1<F> for KeccakVmStep {
...
+                .memory
+                .memory
+                .get_slice((RV32_MEMORY_AS, src), len as usize)
+        };
 
-        let mut output = [0u8; 32];
-        hasher.finalize(&mut output);
-        memory_write_from_state(state, e, dst, &output);
+        let output = keccak256(&message);
+        memory_write_from_state(state, RV32_MEMORY_AS, dst, &output);
```
shuklaayush, line 130: if we're just doing e1, we don't need to register the callback and can directly write to memory

PR #1733, opened 2025-06-12
[wip] feat: new execution e4 for memory adapters

Golovanov399:


crates/vm/src/system/memory/adapter/records.rs
```diff
@@ -0,0 +1,129 @@
...
+            pointer,
+            block_size: block_size_field,
+            data,
+            prev_data,
+            timestamps,
+        }
+    }
+}
+
+impl<'a> RecordArena<'a, AccessLayout, AccessRecordMut<'a>> for DenseRecordArena {
```
arayikhalatyan, line 111: hmm can't this just use the `DenseRecordArena` with `MultiRowLayout`?

crates/vm/src/system/memory/adapter/records.rs
```diff
@@ -0,0 +1,129 @@
...
+pub(crate) struct AccessRecordMut<'a> {
+    /// If we need to merge before, this has the 31st bit set.
+    /// If we need to split after, this has the 30th bit set.
+    pub timestamp_and_mask: &'a mut u32,
+    pub address_space: &'a mut u32,
+    pub pointer: &'a mut u32,
+    /// The block size is not defined by the access adapter's `N`,
+    /// because the record denotes an operation of sort
+    /// "split/merge every `block_size` chunk of this data"
+    pub block_size: &'a mut u32,
```
arayikhalatyan, line 14: nit: maybe grouping these 4 fields into a some `header` struct would simplify? You could also derive `AlignedBytesBorrow` on that struct and it would make the `custom_borrow` simpler as well

crates/vm/src/system/memory/adapter/records.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl SizedRecord<AccessLayout, AccessRecordMut<'_>> for DenseRecordArena {
+    fn size(&self, layout: &AccessLayout) -> usize {
+        size_by_layout(layout)
+    }
+
+    fn alignment(&self, layout: &AccessLayout) -> usize {
+        debug_assert_eq!(layout.cell_size * layout.type_size, 4);
+        4
```
arayikhalatyan, line 49:
```suggestion
        align_of::<32>()
```

crates/vm/src/system/memory/adapter/records.rs
arayikhalatyan: I see, but since it is just used with `DenseRecordArena` it's fine we don't need the `num_rows`

crates/vm/src/arch/integration_api.rs
```diff
@@ -489,48 +489,43 @@ impl DenseRecordArena {
...
-    pub fn alloc_one<'a, T>(&mut self) -> &'a mut T {
-        let begin = self.records_buffer.position();
+    /// Returns the current size of the allocated buffer so far.
+    pub fn current_size(&self) -> usize {
+        self.records_buffer.position() as usize
+    }
+
+    /// Transmutes a single record of the given type from some offset and returns a mutable
+    /// reference to it.
+    // TODO(AG): remove this, replace with the record-based read we have in the near future
```
jonathanpwang, line 499: is this still needed

crates/vm/src/system/memory/adapter/records.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct AccessLayout {
+    /// The size of the block in elements.
+    pub block_size: usize,
+    /// The size of the minimal block (1 or 4 depending on the address space)
+    pub cell_size: usize,
+    /// The size of the type in bytes (1 for u8, 4 for F).
+    // TODO(AG): is it true that cell_size * type_size = 4?
```
jonathanpwang, line 28: there is no hard rule; more of a coincidence

crates/vm/src/system/memory/adapter/records.rs
```diff
@@ -0,0 +1,129 @@
...
+            std::slice::from_raw_parts_mut(
+                self.as_mut_ptr().add(offset),
+                layout.block_size * layout.type_size,
+            )
+        };
+        offset += layout.block_size * layout.type_size;
+
+        // timestamps: [u32] (block_size / cell_size * 4 bytes)
+        let timestamps = unsafe {
+            std::slice::from_raw_parts_mut(
```
jonathanpwang, line 93: alignment?

crates/vm/src/system/memory/online.rs
```diff
@@ -580,24 +620,46 @@ impl<F: PrimeField32> TracingMemory<F> {
...
-                    .filter_map(move |(idx, metadata)| {
-                        (metadata.block_size != 0
-                            && metadata.block_size != AccessMetadata::OCCUPIED)
-                            .then_some(((addr_space as u32, idx as u32 * align), metadata))
-                    })
-            })
+    /// Returns the list of all touched blocks. The list is sorted by address.
+    /// If a block hasn't been explicitly accessed and is created by a split,
+    /// the corresponding metadata has `offset` set to `AccessMetadata::UNSPLITTABLE`.
+    pub fn touched_blocks(&mut self) -> Vec<(Address, AccessMetadata)> {
```
jonathanpwang, line 626: I had changed this to use `par_iter`. The issue is if we use LinearMemory for metadata, it can actually be quite big.

crates/vm/src/system/memory/online.rs
```diff
@@ -143,35 +149,21 @@ impl GuestMemory {
...
-    },
-    IncrementTimestampBy(u32),
-}
-
-// perf[jpw]: since we restrict `timestamp < 2^29`, we could pack `timestamp, log2(block_size)`
-// into a single u32 to save half the memory, since `block_size` is a power of 2 and its log2
-// is less than 2^3.
 #[repr(C)]
 #[derive(Clone, Copy, PartialEq, Eq, Debug, derive_new::new)]
 pub struct AccessMetadata {
```
jonathanpwang, line 154: Is TracingMemory still holding essentially a Vec of these via LinearMemory? If so, ideally we can make the size smaller: right now for touched_blocks iterating through the `meta` can be slow, also it is a good amount of memory. Alternatively, we could switch back to a slower paged vec approach to `meta`

extensions/native/circuit/src/poseidon2/chip.rs
```diff
@@ -859,18 +859,18 @@ impl<F: PrimeField32, const SBOX_REGISTERS: usize> NativePoseidon2Step<F, SBOX_R
             );
             debug_assert_eq!(data_address_space, F::from_canonical_u32(AS::Native as u32));
             let [output_pointer]: [F; 1] =
-                memory_read_native(state.memory, output_register.as_canonical_u32());
```
jonathanpwang, line 862: this seems an unrelated change? maybe just needs merge

crates/vm/src/system/memory/adapter/records.rs
Golovanov399: Huh, it is fine because whenever `type_size < 4` we have `block_size` at least 4, but I'll move `timestamps` up just to be safe

crates/vm/src/system/memory/online.rs
Golovanov399: I think I'll need additional stuff in metadata if I want to be able to `par_iter()`, but with return to paged vec approach it shouldn't be a big deal (both doing `.iter()` and storing additional stuff)

jonathanpwang: recording we discussed offline that we will switch to `PagedVec` for metadata

jonathanpwang: ideally we can combine this with the `touched_blocks_to_equipartition` to avoid the new vec allocation

crates/vm/src/system/memory/online.rs
```diff
@@ -515,133 +443,271 @@ impl<F: PrimeField32> TracingMemory<F> {
...
-                        align,
-                        INITIAL_TIMESTAMP,
-                    );
-                    current_metadata = AccessMetadata::new(INITIAL_TIMESTAMP, align as u32);
+
+        let first_meta = unsafe {
+            self.meta[address_space].read::<AccessMetadata>(begin * size_of::<AccessMetadata>())
+        };
+        let mut need_to_merge = first_meta.block_size != BLOCK_SIZE as u32;
+        let existing_metadatas = (0..num_segs)
```
jonathanpwang, line 599: feels like this does a collect always, I still want use to break very early if `BLOCK_SIZE = align` and there's nothing to do

crates/vm/src/system/memory/online.rs
```diff
@@ -515,133 +443,271 @@ impl<F: PrimeField32> TracingMemory<F> {
...
+                        let block_start = (begin + i) & !(self.initial_block_size / align - 1);
+                        if (address_space as u32) < NATIVE_AS {
+                            let initial_values = unsafe {
+                                self.data
+                                    .memory
+                                    .get_slice::<u8>(
+                                        (address_space as u32, (block_start * align) as u32),
+                                        self.initial_block_size,
+                                    )
+                                    .to_vec() // TODO: small vec?
```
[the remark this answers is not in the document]
Golovanov399: Okay here comes `unsafe { from_raw_parts }`

crates/vm/src/system/memory/controller/mod.rs
```diff
@@ -345,152 +346,151 @@ impl<F: PrimeField32> MemoryController<F> {
...
 
     /// Returns the equipartition of the touched blocks.
-    /// Has side effects (namely setting the traces for the access adapters).
+    /// Modifies records and adds new to account for the initial/final segments.
     fn touched_blocks_to_equipartition<const CHUNK: usize, const INITIAL_MERGES: bool>(
         &mut self,
         touched_blocks: Vec<((u32, u32), AccessMetadata)>,
     ) -> TimestampedEquipartition<F, CHUNK> {
-        let mut current_values = [F::ZERO; CHUNK];
+        let mut final_memory = Vec::new();
```
jonathanpwang, line 354: can we estimate the capacity?

crates/vm/src/system/memory/online.rs
```diff
@@ -739,50 +853,119 @@ impl<F: PrimeField32> TracingMemory<F> {
...
-                meta.par_iter().filter_map(move |(idx, metadata)| {
-                    (metadata.block_size != 0 && metadata.block_size != AccessMetadata::OCCUPIED)
-                        .then_some(((addr_space as u32, idx as u32 * align), metadata))
-                })
-            })
+    /// Returns the list of all touched blocks. The list is sorted by address.
+    /// If a block hasn't been explicitly accessed and is created by a split,
+    /// the corresponding metadata has `offset` set to `AccessMetadata::UNSPLITTABLE`.
+    pub fn touched_blocks(&mut self) -> Vec<(Address, AccessMetadata)> {
+        let mut blocks = Vec::new();
```
jonathanpwang, line 860: with capacity

crates/vm/src/system/memory/online.rs
```diff
@@ -739,50 +853,119 @@ impl<F: PrimeField32> TracingMemory<F> {
...
+        for (addr_space, (page, &align)) in zip_eq(&self.meta, &self.min_block_size).enumerate() {
```
jonathanpwang, line 861: I feel like we need to be able to make this parallelizable or there will be problems

jonathanpwang: maybe the paged vec helps

crates/vm/src/system/memory/online.rs
```diff
@@ -510,124 +438,317 @@ impl<F: PrimeField32> TracingMemory<F> {
...
+                        block_size: BLOCK_SIZE,
+                        lowest_block_size: align,
+                        type_size: size_of::<T>(),
+                    },
+                )
+                .unwrap();
+
+            record_mut.header.timestamp_and_mask = self.timestamp;
+            let data_slice = unsafe {
+                from_raw_parts(values.as_ptr() as *const u8, BLOCK_SIZE * size_of::<T>())
```
jonathanpwang, line 730: feels like you do this a lot; we should make some `cast_to_raw_bytes` function?

jonathanpwang (commented):
The implemented logic looks good.

It doesn't seem to affect perf in reth-benchmark [comparison](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/15963867443) but in micro-benchmarks E3 seems to have noticeably degraded?

Overall I am still a bit worried that this process where you mutate records after the fact via `offset` and to update the split/merge flag might either slow down e3 (it's not immediately evident it does, except it does make the metadata size bigger and possibly makes memory more random access) or affect parallelism.

If you are going to stick with this way, then when you inevitably need to implement the first scan before parallelism, please add that to this repo.

crates/vm/src/system/memory/adapter/mod.rs
```diff
@@ -53,18 +72,65 @@ impl<F: Clone + Send + Sync> AccessAdapterInventory<F> {
...
+            chip.set_override_trace_height(oh);
         }
     }
 
+    pub fn set_arena_from_trace_heights(&mut self, trace_heights: &[u32]) {
+        assert_eq!(trace_heights.len(), self.chips.len());
+        // At the very worst, each row in `Adapter<N>`
+        // corresponds to a unique record of `block_size` being `2 * N`,
+        // and its `lowest_block_size` is at least 1 and `type_size` is at most 4.
+        // However, there may be "blank" records -- the ones that don't generate any rows.
```
jonathanpwang, line 103: do blank records still exist?

crates/vm/src/system/memory/online.rs
```diff
@@ -418,216 +434,195 @@ impl<F: PrimeField32> TracingMemory<F> {
         );
     }
 
-    pub(crate) fn execute_splits<const UPDATE_META: bool>(
+    /// Updates the metadata with the given block.
+    #[inline]
+    fn set_meta_block(
         &mut self,
-        address: MemoryAddress<u32, u32>,
+        address_space: usize,
```
jonathanpwang, line 441: don't need to change it, but using `u32` instead of `usize` might allow the compiler to do some stuff more efficiently(?)

jonathanpwang (approved):
LGTM, nice job!

jonathanpwang:
Latest reth benchmark comparison: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/16016116215

crates/vm/src/system/memory/online.rs
Golovanov399: I believe the compiler should be able to find that we usually essentially do `address_space as u32` so it should treat it like `u32`, at least most of the time, but yes, I keep in mind that we probably need to unify those `usize` <-> `u32` conversions more

PR #1737, opened 2025-06-13
feat: `DenseRecordArena` for dense records

Golovanov399:
This resolves INT-4184.

This **does not** introduce a new chip wrapper. Instead, this introduces a new struct that `impl RecordArena`. More specifically:

- A new trait arrived:
```
pub trait SizedRecord<Layout, RecordMut> {
    fn width_bytes(&self, layout: &Layout) -> usize;
}
```
This trait tells how many bytes to allocate for a pair `(Layout, RecordMut)`. Different record arenas can require different amount of bytes. This can generalize things, probably.

- `pub struct DenseRecordArena` for dense records.
- Some default implementations of `RecordArena` and `RecordWidthFromLayout` for it.
- `impl<F, AIR, STEP> NewVmChipWrapper` is now two different implementations for `MatrixRecordArena<F>` and `DenseRecordArena`.
- It works kinda well with simple cases where we only care about `size_of::<Record>()`. It can work with other layouts and records as well, it's just one needs to implement those things separately for the corresponding layout-record pair.
- As before, there is a test in AUIPC as an illustration of how dense records would work.
- As before, some random changes by `cargo +nightly fmt --all` because I have the corresponding pre-push hook.

### DIFF
--- a/crates/vm/src/arch/integration_api.rs
+++ b/crates/vm/src/arch/integration_api.rs
@@ -105,7 +105,10 @@ pub struct AdapterAirContext<T, I: VmAdapterInterface<T>> {
 }
 
 pub trait SizedRecord<Layout, RecordMut> {
-    fn num_bytes(&self, layout: &Layout) -> usize;
+    /// The size of the record in bytes
+    fn size(&self, layout: &Layout) -> usize;
+    /// The alignment of the record in bytes
+    fn alignment(&self, layout: &Layout) -> usize;
 }
 
 /// Given some minimum metadata of type `Layout` that specifies the record size, the `RecordArena`
@@ -308,13 +311,13 @@ pub struct DenseRecordArena {
     pub records_buffer: Cursor<Vec<u8>>,
 }
 
-const ALIGNMENT: usize = 4;
+const MAX_ALIGNMENT: usize = 32;
 
 impl DenseRecordArena {
     /// Creates a new [DenseRecordArena] with the given capacity in bytes.
     pub fn with_capacity(size_bytes: usize) -> Self {
-        let buffer = vec![0; size_bytes + ALIGNMENT];
-        let offset = (ALIGNMENT - (buffer.as_ptr() as usize % ALIGNMENT)) % ALIGNMENT;
+        let buffer = vec![0; size_bytes + MAX_ALIGNMENT];
+        let offset = (MAX_ALIGNMENT - (buffer.as_ptr() as usize % MAX_ALIGNMENT)) % MAX_ALIGNMENT;
         let mut cursor = Cursor::new(buffer);
         cursor.set_position(offset as u64);
         Self {
@@ -360,9 +363,17 @@ impl DenseRecordArena {
 
     pub fn allocated(&self) -> &[u8] {
         let size = self.records_buffer.position() as usize;
-        let offset =
-            (ALIGNMENT - (self.records_buffer.get_ref().as_ptr() as usize % ALIGNMENT)) % ALIGNMENT;
+        let offset = (MAX_ALIGNMENT
+            - (self.records_buffer.get_ref().as_ptr() as usize % MAX_ALIGNMENT))
+            % MAX_ALIGNMENT;
         &self.records_buffer.get_ref()[offset..offset + size]
+    }
+
+    pub fn align_to(&mut self, alignment: usize) {
+        debug_assert!(MAX_ALIGNMENT % alignment == 0);
+        let offset =
+            (alignment - (self.records_buffer.get_ref().as_ptr() as usize % alignment)) % alignment;
+        self.records_buffer.set_position(offset as u64);
     }
 
     /// Disclaimer: it shouldn't even be called here (only on GPU),
@@ -411,8 +422,12 @@ impl<Record> SizedRecord<(), &mut Record> for DenseRecordArena
 where
     [u8]: BorrowMut<Record>,
 {
-    fn num_bytes(&self, _layout: &()) -> usize {
+    fn size(&self, _layout: &()) -> usize {
         size_of::<Record>()
+    }
+
+    fn alignment(&self, _layout: &()) -> usize {
+        align_of::<Record>()
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/vm/src/arch/integration_api.rs
```diff
@@ -341,6 +458,29 @@ where
...
+impl<'a, AS, A, C, I: Clone> RecordArena<'a, AdapterCoreLayout<AS, I>, (A, C)> for DenseRecordArena
+where
+    [u8]: CustomBorrow<'a, A, I> + CustomBorrow<'a, C, I>,
+    DenseRecordArena: RecordWidthFromLayout<I, A> + RecordWidthFromLayout<I, C>,
+{
+    fn alloc(&'a mut self, layout: AdapterCoreLayout<AS, I>) -> (A, C) {
+        let adapter_width =
+            <Self as RecordWidthFromLayout<I, A>>::width_bytes(self, &layout.metadata);
+        let core_width = <Self as RecordWidthFromLayout<I, C>>::width_bytes(self, &layout.metadata);
+        let buffer = self.alloc_bytes(adapter_width + core_width);
```
arayikhalatyan, line 473: My worry about this is that what if `AdapterRecord` has alignment 4 and `CoreRecord` has alignment 1, like `rv32_basealu`'s core record is 9 `u8`'s . Then after one `alloc` the offset won't be 4 byte aligned and the next `AdapterRecord` won't be allocated properly aligned

crates/vm/src/arch/integration_api.rs
```diff
@@ -341,6 +458,29 @@ where
     }
 }
 
+/// RecordArena implementation for [DenseRecordArena], with [AdapterCoreLayout]
+/// `A` is the adapter record type and `C` is the core record type
```
arayikhalatyan, line 462:
```suggestion
/// `A` is the adapter RecordMut type and `C` is the core RecordMut type
```

This is a mistake from my side, missed it in the above comment.

crates/vm/src/arch/integration_api.rs
```diff
@@ -283,6 +303,76 @@ pub struct MatrixRecordArena<F> {
...
+/// not a trace row to be filled later.
+pub struct DenseRecordArena {
+    pub records_buffer: Cursor<Vec<u8>>,
+}
+
+impl DenseRecordArena {
+    /// Creates a new [DenseRecordArena] with the given capacity in bytes.
+    pub fn with_capacity(size_bytes: usize) -> Self {
+        Self {
+            records_buffer: Cursor::new(vec![0; size_bytes]),
```
arayikhalatyan, line 318: Should we somehow impose an explicit (maybe 4-byte) alignment for the `buffer`?

crates/vm/src/arch/integration_api.rs
```diff
@@ -283,6 +303,76 @@ pub struct MatrixRecordArena<F> {
...
+        }
+    }
+
+    /// Allocates a single record of the given type and returns a mutable reference to it.
+    pub fn alloc_one<'a, T>(&mut self) -> &'a mut T {
```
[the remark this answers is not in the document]
Golovanov399: I will use it for memory adapters at least, but in general when `RecordMut = &mut Record`

crates/vm/src/arch/integration_api.rs
```diff
@@ -283,6 +303,76 @@ pub struct MatrixRecordArena<F> {
...
+                    .get_mut()
+                    .as_mut_ptr()
+                    .add(begin as usize) as *mut T,
+                count,
+            )
+        }
+    }
+
+    pub fn alloc_bytes<'a>(&mut self, count: usize) -> &'a mut [u8] {
+        self.alloc_many::<u8>(count)
```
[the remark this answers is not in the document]
Golovanov399: Idk, why?

crates/vm/src/arch/integration_api.rs
arayikhalatyan: Yeah, it's either that or keeping separate buffers for Adapter and Core records. But as I understand we don't wanna do that?
hmm, how can `AlignedBytesBorrow` do that? Do you mean making sure `width` is always a multiple of `4`?

Golovanov399: I meant that maybe the derive proc-macro for `AlignedBytesBorrow` can add `#[repr(C, align(4))]` to the struct but probably not. But who knows, maybe there is a way to enforce alignment inside the macro somehow

crates/vm/src/arch/integration_api.rs
Golovanov399: I think it shouldn't matter for non-zkvm target, and for zkvm target our allocator does that already

jonathanpwang: this will never be used in a zkvm target

crates/vm/src/arch/integration_api.rs
```diff
@@ -292,6 +300,79 @@ pub struct MatrixRecordArena<F> {
...
+
+    /// Disclaimer: it shouldn't even be called here (only on GPU),
+    /// so this function is for testing purposes.
+    /// Also, it only covers the basic case where everything is placed
+    /// densely and consequently and its size is just sizeof
+    pub fn extract_records<Record: Sized + Clone>(&self) -> Vec<Record> {
+        let bytes = self.allocated();
+        let size = size_of::<Record>();
+        debug_assert!(bytes.len() % size == 0);
+        unsafe { std::slice::from_raw_parts(bytes.as_ptr() as *const Record, bytes.len() / size) }
```
jonathanpwang, line 371: if `bytes` is not aligned to `align_of::<Record>()` this violates safety

jonathanpwang: there is a special function called like `align_to` that exactly does this with debug asserts

crates/vm/src/arch/integration_api.rs
```diff
@@ -321,8 +402,32 @@ impl<F: Field> RowMajorMatrixArena<F> for MatrixRecordArena<F> {
...
+{
+    fn width_bytes(&self, _layout: &AdapterCoreLayout<AS, I>) -> usize {
+        AS::WIDTH * size_of::<F>()
+    }
+}
+
+impl<F, I, R> SizedRecord<MultiRowLayout<I>, R> for MatrixRecordArena<F> {
+    fn width_bytes(&self, layout: &MultiRowLayout<I>) -> usize {
+        (layout.num_rows as usize) * self.width * size_of::<F>()
+    }
```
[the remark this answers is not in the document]
jonathanpwang: I also commented above in the trait def

crates/vm/src/arch/integration_api.rs
jonathanpwang: I'm not sure `align(4)` helps? what is `size_of<Adapter>` is not a multiple of 4?

crates/vm/src/arch/integration_api.rs
jonathanpwang: does it really not matter though? see below, if your starting pointer is not multiple of `Record`'s alignment, all safety is gone

crates/vm/src/arch/integration_api.rs
Golovanov399: This function is not intended to be called on CPU, and it is here for the sake of testing stuff before we made it generated on GPU, as its doc-comment says. So I'm not going to make this function provably correct

extensions/rv32im/circuit/src/auipc/tests.rs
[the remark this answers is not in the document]
jonathanpwang: In principle if you call `fill_trace_row` on a dense record, it should still work? It won't have any overlapping issues

jonathanpwang (approved):
I am now happy

extensions/rv32im/circuit/src/auipc/tests.rs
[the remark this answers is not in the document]
arayikhalatyan: `fill_trace_row` does `record = trace_row.custom_borrow()`. We can write the records into `trace_rows` before calling these function just for testing purposes

PR #1738, opened 2025-06-13
refactor: Rewrite E1

nyunyunyunyu:


crates/vm/src/arch/segment.rs
```diff
@@ -43,12 +48,68 @@ impl<F, Ctx> VmSegmentState<F, Ctx> {
...
+                GuestMemory::new(Default::default())
+            },
             streams,
             ctx,
             exit_code: None,
         }
     }
+    /// Runtime read operation for a block of memory
+    #[inline(always)]
+    pub fn vm_read<T: Copy + Debug, const BLOCK_SIZE: usize>(
```
jonathanpwang, line 63: you can just call it `read`? seems no risk of confusion

PR #1739, opened 2025-06-13
fix: e2 memory accesses for hintstore and loadstore

arayikhalatyan:
I have been using `execute_e1` in `metered_execution` in most places, but after an offline discussion with @ayush realized that memory accesses in e2 should be exactly the same as in e3. This PR fixes it for `loadstore` and `hintstore`. Fixed e2 `keccak` and `sha2` in corresponding branches since they haven't been merged yet. I believe all other places should be correct.

Resolves INT-4204


extensions/rv32im/circuit/src/adapters/loadstore.rs
```diff
@@ -579,24 +581,43 @@ where
...
-                let imm_extended = c.as_canonical_u32() + g.as_canonical_u32() * 0xffff0000;
-                let ptr_val = rs1.wrapping_add(imm_extended);
-                assert!(
-                    ptr_val < (1 << self.pointer_max_bits),
-                    "ptr_val: {ptr_val} = rs1_val: {rs1} + imm_extended: {imm_extended} >= 2 ** {}",
-                    self.pointer_max_bits
-                );
-                memory_read_from_state(state, e.as_canonical_u32(), ptr_val)
+                let read_data = memory_read_from_state(state, e.as_canonical_u32(), ptr_val);
+                read_data
```
shuklaayush, line 601: clippy might complain
```suggestion
                memory_read_from_state(state, e.as_canonical_u32(), ptr_val)
```

shuklaayush (approved):
small nit. otherwise lgtm

PR #1744, opened 2025-06-16
fix(k256): ecdsa verify needs custom hook

jonathanpwang:
For some reason, unlike other libraries, the k256 crate has a special requirement that the signature is normalized ahead of time: https://github.com/RustCrypto/elliptic-curves/blob/5ac8f5d77f11399ff48d87b0554935f6eddda342/k256/src/ecdsa.rs#L203 To handle this, we introduce a new trait `VerifyCustomHook` that is called both in `recover_prehashed` and `verify_prehash`.

- [x]  Provide a proof that valid pubkey recovery implies the ECDSA signature verification algorithm will automatically pass, so there is no need for `recover_prehashed` to call `verify_prehash` again.


extensions/ecc/tests/src/lib.rs
```diff
@@ -188,9 +189,9 @@ mod tests {
                 .app_vm_config;
         let elf = build_example_program_at_path_with_features(
             get_programs_dir!(),
-            "ecdsa_recover",
+            "ecdsa_recover_p256",
             ["p256"],
-            &config,
+            &NoInitFile, // using already created file
```
Avaneesh-axiom, line 194: Why aren't we using the build script here?

extensions/ecc/tests/src/lib.rs
```diff
@@ -199,6 +200,24 @@ mod tests {
...
+    #[test]
+    fn test_k256_ecdsa_recover() -> Result<()> {
+        let config =
+            toml::from_str::<AppConfig<SdkVmConfig>>(include_str!("../programs/openvm_k256.toml"))?
+                .app_vm_config;
+        let elf = build_example_program_at_path_with_features(
+            get_programs_dir!(),
+            "ecdsa_recover_k256",
+            ["k256"],
+            &NoInitFile, // using already created file
```
Avaneesh-axiom, line 212: same here

Avaneesh-axiom (approved):

extensions/ecc/guest/ECDSA.md
```diff
@@ -0,0 +1,68 @@
...
+### Proof of signature verification
+
+Above, we skip the verification of the original signature again with the recovered $Q$. We give a proof below that the recovered $Q$ will always pass the signature verification.
+
+We refer to the steps in verification as V1-7 and the steps in recovery as R1-7.
+- V1 is automatic from R1.
+- V2 is the same as R5.
+- V3: We compute $u = e \cdot s^{-1} \mod n$ and $v = r \cdot s^{-1} \mod n$.
+- V4: compute $$R_1 = uG + vQ = e \cdot s^{-1} G + (r \cdot s^{-1})\cdot r^{-1} \cdot (s R - e G) = R.$$
+R7 guarantees $R_1$ is not identity.
```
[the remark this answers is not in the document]
jonathanpwang: ah good call

PR #1749, opened 2025-06-17
chore: add serde to segment

luffykai:


jonathanpwang:
note: we may want to just write a custom serializer to bytes

PR #1751, opened 2025-06-17
fix(new-execution): cleanup `vm.rs`

shuklaayush:
- remove unused functions from `VmExecutor`
- add `execute_from_state` and `execute_from_state_and_generate` for executing from an initial state
- remove `TracegenExecutionControlWithSegmentation`
- bumped `stark-backend` to `0879de1` to get rid of `get_widths_and_interactions_from_vkey`


benchmarks/execute/benches/fibonacci_execute.rs
```diff
@@ -1,44 +1,44 @@
-use criterion::{criterion_group, criterion_main, Criterion};
```
shuklaayush, line 1: commented out this file now. this was doing the old execute (e3) which now requires segments as input. we should probably delete it or benchmark e3 in some other way

crates/vm/src/arch/vm.rs
```diff
@@ -519,139 +329,49 @@ where
...
+        self.execute_metered_from_state(exe, state, widths, interactions)
     }
 
-    fn execute_and_generate_impl<SC: StarkGenericConfig>(
+    /// Base execution function that operates from a given state
+    /// After each segment is executed, call the provided closure on the execution result.
+    /// Returns the results from each closure, one per segment.
+    ///
+    /// The closure takes `f(segment_idx, segment) -> R`.
+    pub fn execute_and_then_from_state<R, E>(
```
jonathanpwang, line 365: later: we should call this `execute_preflight` or `execute_e3`

jonathanpwang (approved):
LGTM for now. There is still more cleanup of `execute_*` functions to do but I think it should wait for more general `InterpretedInstance` structuring

PR #1755, opened 2025-06-18
chore(new-execution): clk/cycles -> instret/insns

shuklaayush:
renamed `clk` to `instret` and `cycles` to `insns` in segments


jonathanpwang (approved):
lgtm but rename `insts` to `insns`

PR #1763, opened 2025-06-18
fix(new-execution): prevent segmentation in single segment executor

shuklaayush:
- added a `segment_check_insns` paramter in `MeteredCtx`. this is set to `u64::MAX` to prevent segmentation in `SingleSegmentExecutor`
- removed unused functions from `SingleSegmentExecutor`

### DIFF
--- a/crates/vm/src/arch/vm.rs
+++ b/crates/vm/src/arch/vm.rs
@@ -645,82 +645,6 @@ where
         self.trace_height_constraints = constraints;
     }
 
-    /// Executes a program, compute the trace heights, and returns the public values.
-    pub fn execute_and_compute_heights(
-        &self,
-        exe: impl Into<VmExe<F>>,
-        input: impl Into<Streams<F>>,
-    ) -> Result<SingleSegmentVmExecutionResult<F>, ExecutionError> {
-        let segment = {
-            let mut segment = self.execute_impl(exe.into(), input.into(), None)?;
-            segment.chip_complex.finalize_memory();
-            segment
-        };
-        let air_heights = segment.chip_complex.current_trace_heights();
-        let vm_heights = segment.chip_complex.get_internal_trace_heights();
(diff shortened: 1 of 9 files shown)


crates/vm/src/arch/vm.rs
```diff
@@ -752,7 +686,14 @@ where
             executor.execute_from_state(&mut exec_state)
         })?;
 
-        check_exit_code(exec_state.exit_code)?;
+        match exec_state.exit_code {
```
jonathanpwang, line 689: why did you have to get rid of `check_exit_code`?

crates/vm/src/arch/vm.rs
```diff
@@ -851,7 +835,7 @@ where
     }
 
     /// Executes a program, compute the trace heights, and returns the public values.
-    pub fn execute_with_max_heights_and_compute_heights(
+    pub fn execute_and_compute_heights(
```
jonathanpwang, line 838: why is it computing heights but also taking in `max_trace_heights`

crates/vm/src/arch/vm.rs
shuklaayush: i inlined it because in cases when you're running for a fixed number of cycles, you don't want to throw a `DidNotTerminate` error when `exit_code` is `None`. so the name `check_exit_code` seemed a bit confusing

crates/vm/src/arch/vm.rs
shuklaayush: this is probably an artifact of the past. let me check if just using `max_trace_heights` works

shuklaayush: ah no. `compute_height` returns this `VmComplexTraceHeights` struct. we can probably replace it with a `Vec` of trace heights, but it'll require more changes

crates/vm/src/arch/vm.rs
shuklaayush: added it back as `check_termination`

PR #1765, opened 2025-06-19
feat: Poseidon2 E4

nyunyunyunyu:
Rewrite Native Poseidon2 chip for E4. The implementation follows the existing logic(directly writing into trace buffer) instead of creating a new minimal record struct. Rationale:
1. Simpler.
2. I also think poseidon2 columns computation is the bottleneck so a new minimal record struct won't help too much,
3. Each row can run in parallel.

close INT-4101

### DIFF
--- a/crates/circuits/mod-builder/src/core_chip.rs
+++ b/crates/circuits/mod-builder/src/core_chip.rs
@@ -1,5 +1,3 @@
-use std::{borrow::BorrowMut, mem::size_of};
-
 use itertools::Itertools;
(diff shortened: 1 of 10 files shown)


shuklaayush:
all workflows except lint should have been passing after this 🤔 

this test seems to fail on `NativeLoadStore` so i'm not sure if it's related to this pr
https://github.com/openvm-org/openvm/actions/runs/15750688796/job/44395087829?pr=1765

shuklaayush (approved):

PR #1768, opened 2025-06-19
fix: Elliptic curve point decompression doesn't check if `y = 0`

Avaneesh-axiom:
The point decompression function takes the x-coordinate and the parity of the y-coordinate. Then it solves $y^2 = x^3 + ax + b$ for $y$ up to parity. If the recovered $y$ doesn't have the correct parity, it returns the point $(x, -y)$, otherwise it returns $(x, y)$.

The problem happens if the recovered $y$ is $0$ and `rec_id = 1`, in which case $-y$ has the same parity as $y$ (namely, even), so although the resulting point $(x, y)$ is a valid point on the curve, it does not correspond to `rec_id = 1`.

This is not an issue for Secp256k1 and Secp256r1 since these curves do not have points with $y = 0$ (such a point would have order 2 since $P = -P$ and the order of these curves is not even) so any such decompression will fail.

However, a curve could be constructed for which this is an issue.

This PR adds a failing test with such a curve and then fixes the bug by checking the special case.
Note that the curve equation for `CurvePoint5mod8` in the `decompress.rs` test was changed so that a point with y-coordinate zero exists in that curve.

### DIFF
--- a/extensions/ecc/tests/src/lib.rs
+++ b/extensions/ecc/tests/src/lib.rs
@@ -116,7 +116,6 @@ mod tests {
                     scalar: BigUint::from_str("1000000007")
                         .unwrap(),
                     a: BigUint::ZERO,
-                    // b: BigUint::from_str("3").unwrap(),
                     b: BigUint::from_str("6").unwrap(),
                 },
                 CurveConfig {


extensions/ecc/tests/src/lib.rs
```diff
@@ -116,7 +116,8 @@ mod tests {
                     scalar: BigUint::from_str("1000000007")
                         .unwrap(),
                     a: BigUint::ZERO,
-                    b: BigUint::from_str("3").unwrap(),
+                    // b: BigUint::from_str("3").unwrap(),
```
jonathanpwang, line 119:
```suggestion
```

jonathanpwang (approved):
LGTM, good find

PR #1769, opened 2025-06-20
fix(new-execution): move hint rng to vmstate

shuklaayush:
Resolves INT-4218


crates/vm/src/arch/execution.rs
```diff
@@ -107,6 +109,7 @@ pub trait InstructionExecutor<F> {
         &mut self,
         memory: &mut MemoryController<F>,
         streams: &mut Streams<F>,
+        rng: &mut StdRng,
```
jonathanpwang, line 112: do we really need this in the interface of the trait? it's not great because the `rng` is only used so sparingly

jonathanpwang: discussed offline; we'll soon change the traits to take a `VmStateMut` to hide the state away

PR #1770, opened 2025-06-20
perf(new-execution): use page-based approach for merkle tree metering

shuklaayush:
- Replace the vector of leaf indices with a `BitSet` of page indices
- Update merkle tree chip height based on page faults
- Fix merkle tree metering when continuations is disabled
- Fix access adapter metering on finalization
- Delete `test_segmentation_retry` since it's not relevant now

Towards INT-4068


jonathanpwang (approved):
LGTM, although we would still prefer if we could skip the page approach altogether (I'm assuming you tried this one first since it was faster to implement)

PR #1773, opened 2025-06-20
feat(sdk): derive Debug traits in config

Qumeric:
It might be convenient to have while developing.


jonathanpwang:
Can you re-open targeting `feat/new-execution`? that's probably where you'll want it anyways

PR #1780, opened 2025-06-24
feat: better merkle building

Golovanov399:
For each address space, find where the last nonzero value is and only use that prefix in `memory_to_vec_partition`


jonathanpwang (approved):
LGTM, I still feel like `memory_to_vec_partition` overall should be avoided somehow (it still clones an entire vec of the memory size into memory)

PR #1789, opened 2025-06-24
refactor(new-execution): make `ExecutionCtrl`s stateless and move state to `Ctx`

shuklaayush:


crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
jonathanpwang, line 1: ok for now, but it might make sense to move this to `system/memory/online` later

jonathanpwang (approved):
I'm assuming most of the large diffs are just from copy pasting to different files

PR #1794, opened 2025-06-26
feat(new-execution): change default `PAGE_BITS` value in e2 to 6

shuklaayush:
- change default page size to 6 bits which seems better on benchmarks


jonathanpwang:
This makes execute_metered time slower? 

Can you run a comparison reth benchmark workflow?

jonathanpwang (approved):
LGTM after you provide the benchmarks to back it up

PR #1798, opened 2025-06-26
feat(new-execution): ignore register addr space for memory ops

shuklaayush:


crates/vm/src/arch/execution_mode/metered/ctx.rs
```diff
@@ -1,11 +1,13 @@
-use openvm_instructions::riscv::RV32_IMM_AS;
+use openvm_instructions::riscv::{RV32_IMM_AS, RV32_REGISTER_AS, RV32_REGISTER_NUM_LIMBS};
 
 use super::{
     memory_ctx::MemoryCtx,
     segment_ctx::{Segment, SegmentationCtx},
 };
 use crate::{arch::execution_mode::E1E2ExecutionCtx, system::memory::dimensions::MemoryDimensions};
 
+const RV32I_NUM_REGISTERS: usize = 32;
```
jonathanpwang, line 9: we should probably move this somewhere else since we'll want to use it elsewhere too. Also I don't think you need "I" -- this is a general property of RV32 without "e"

jonathanpwang:
Is the tracegen time increase because you didn't merge in the `PagedVec` commit?

shuklaayush:
> Is the tracegen time increase because you didn't merge in the PagedVec commit?

i think so since this shouldn't affect tracegen at all. i reran the benchmark and updated the link

jonathanpwang (approved):

PR #1800, opened 2025-06-27
fix: account for E1 / E2 execution in metrics summary

yi-sun:


crates/prof/src/aggregate.rs
```diff
@@ -206,6 +207,15 @@ impl AggregateMetrics {
                             *total_par_proof_time.diff.as_mut().unwrap() += diff / 1000.0;
                         }
                     }
+
+                    if let Some(execute_e1_stats) = execute_e1_stats {
```
jonathanpwang, line 211: we will want to handle this differently if we start to run e1 per segment, so this is a temporary patch

jonathanpwang (approved):

PR #1803, opened 2025-06-27
fix(new-execution): adapter offset should be boundary_idx + 1

shuklaayush:
@Golovanov399 pointed out that the `adapter_offset` was being calculated incorrectly. i did a couple of things in this pr:
- fixed `adapter_offset` and index finding logic
- added a `debug_assert` to verify that the calculated indices are correct


Golovanov399 (approved):
Seems good, thanks!

PR #1809, opened 2025-06-30
feat: new traits for VM extension

jonathanpwang:
Previously, `VmExtension` had only a single notion of chip, where some chips could be executors, and the notions of execution, circuit, tracegen were all deeply intertwined. With the new design, we have multiple modes of execution which are more disconnected from the circuit design, and we also want tracegen to support different device / prover backends.

The overall realization is that a true VM extension, meaning a modular way to extend the functionality of a working zkVM (including the prover) has three _separate_ components:
- extending the execution methods for custom opcodes (this includes different execution modes)
- extending the circuit itself at the level of AIRs
- extending how tracegen and ultimately proving of the circuit from execution traces is actually performed

In this PR, we aim to separate the notion of `VmExtension` into three separate traits `VmExecutionExtension, VmCircuitExtension, VmProverExtension` to better encapsulate this.
- The `VmExecutionExtension` is in principle separate from any circuit considerations: it is just a way to specify hooks to opcodes. Although practically, there is still some connection, and we still take an enum dispatch approach (for now) for storage purposes. This is intended to be compatible with the fn pointer changes being made in parallel.
- The `VmCircuitExtension` is the most security critical, and it should have **no** dependencies on the other two components. This will be made so that the verifying keys do not change even after this PR, and also this trait will not depend on any prover backend differences.
- `VmProverExtension` is the most customizable, and hence unfortunately will have a bunch of generics. We will allow even the system chips to be customized according to the prover backend.

Also
- Deleted `NewVmChipWrapper` because the wrapper pattern is not good. Now each chip owns `mem_helper` -- it is perhaps more correct if just the adapter chip owns `mem_helper`, but since we aren't so convinced about adapter chips existing in the future, I just did the more immediate refactor.
- Deleted the `Shared**Chip` wrapper structs because we don't need them anymore. I made them typedef to `Arc<..>`.

## Design Questions

I am still somewhat on the fence about `Chip` being generic in `RA` versus doing the old approach where the `**Chip` struct is generic in `RA` and starts off with empty records which get swapped out with the records from the `TracegenCtx` (see before this PR). However it feels weird when the clear use case is now that the "preflight trace" is recorded elsewhere and only fed into the chip at time of tracegen.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5411,8 +5411,7 @@ dependencies = [
 
 [[package]]
 name = "openvm-stark-backend"
-version = "1.1.1"
-source = "git+https://github.com/openvm-org/stark-backend.git?rev=37aa9a7e8e5535908720bf0dedcf9bdc96eda4ad#37aa9a7e8e5535908720bf0dedcf9bdc96eda4ad"
+version = "1.2.0-rc.0"
 dependencies = [
  "bitcode",
  "cfg-if",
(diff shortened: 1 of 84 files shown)


jonathanpwang:
merging into my branch so I can merge in updates from `feat/new-execution`

PR #1813, opened 2025-07-02
feat: extract_layout implementation for native Poseidon2

stephenh-axiom-xyz:
Resolves INT-4288.

Discussed with @arayikhalatyan, for trace generation purposes we can actually treat native Poseidon2 as if it uses fixed-size records. This simplifies the GPU tracegen and allows us to not send any auxiliary information to the GPU device.


arayikhalatyan (approved):
I believe this idea should work, and even makes it much easer to do parallel tracegen over rows on GPU.
Note, this is different from other chips in the sense that we will 'decode' the records from buffer differently from how they were 'allocated'. During execution we allocate 1 record (aka `NativePoseidon2RecordMut`) per _instruction_ but in buffer decoding during tracegen we will treat it as 1 record (aka `&mut NativePoseidon2Cols`) per row

PR #1819, opened 2025-07-03
feat: modify fri records + tests

arayikhalatyan:
Towards INT-4341
This PR is blocking the related PR in axiom-gpu

Made some changes to `NativeFri` 
- In `WorkLoadRecord` instead of `b` now I keep keep the current `result`. `b` can be computed using `a`, `alpha`, `result`, and `previous_result`. This allows for tracegen to skip a serial pass calculating all the intermediate `results` at the start. 
- Separated `a_prev_data` from `WrokLoadRecord` so we can skip allocating memory for it when a setup instruction. This is an optimization that reduces record sizes since we don't always need the `prev_data`
- Ported the test file to the new framework
- divided the WorkloadRows tracegen into two parts. One lighter serial pass over all the `WorkLoadRows` that  writes the records into the corresponding places. And a second parallel pass that fills the rest of the trace.
- Also, there was a small bug related to not zeroing out the `a_prev_data` when no write happened

### DIFF
--- a/extensions/native/circuit/src/fri/mod.rs
+++ b/extensions/native/circuit/src/fri/mod.rs
@@ -1023,7 +1023,8 @@ where
             // - Then, a parallel pass to fill the rest of the records into the trace
             // Note, the first pass is done to avoid overwriting the records
 
-            // Copy of `a_write_prev_data` to avoid overwriting it and to use it in the parallel pass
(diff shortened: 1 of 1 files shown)


shuklaayush (approved):
lgtm. just do `cargo +nightly fmt`
nice work in reducing the record size!

PR #1823, opened 2025-07-07
feat(execute-opt): use interpreter instance in vm executor `execute_e1`

shuklaayush:
- make `execute_e1` use interpreter instance

### DIFF
--- a/crates/vm/src/arch/interpreter.rs
+++ b/crates/vm/src/arch/interpreter.rs
@@ -26,9 +26,9 @@ use crate::{
 
 /// VM pure executor(E1/E2 executor) which doesn't consider trace generation.
 /// Note: This executor doesn't hold any VM state and can be used for multiple execution.
-pub struct InterpretedInstance<'a, F: PrimeField32, VC: VmConfig<F>> {
+pub struct InterpretedInstance<F: PrimeField32, VC: VmConfig<F>> {
     exe: VmExe<F>,
-    vm_config: &'a VC,
+    vm_config: VC,
 }
 
 #[derive(AlignedBytesBorrow, Clone)]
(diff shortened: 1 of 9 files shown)


crates/vm/src/arch/interpreter.rs
```diff
@@ -12,26 +12,23 @@ use openvm_instructions::{
...
     system::memory::{online::GuestMemory, AddressMap},
 };
 
 /// VM pure executor(E1/E2 executor) which doesn't consider trace generation.
 /// Note: This executor doesn't hold any VM state and can be used for multiple execution.
-pub struct InterpretedInstance<F: PrimeField32, VC: VmConfig<F>> {
+pub struct InterpretedInstance<'a, F: PrimeField32, VC: VmConfig<F>> {
     exe: VmExe<F>,
-    vm_config: VC,
+    vm_config: &'a VC,
```
nyunyunyunyu, line 31: Why a reference here? I think `InterpretedInstance` needs to exist for multiple runs so it could own `vm_config` in order to reduce the annoying handle about the lifecycle

shuklaayush: i can revert it. i don't have a strong opinion. i just didn't want to clone from `VmSegmentExecutor`. i'm not sure if we'll ever have multiple `InterpretedInstance` simultaneously but if we do, then we can share a config

jonathanpwang: config is usually small, so cloning to save lifetime misery seems nicer.

but I did envision maybe we have `VmExecutor` which is owned but no loaded program, and when you load a program, you create an instance, which can have a lifetime because it might borrow something (e.g., executor structs) from the main `VmExecutor`. Just a thought (atm I just clone when I need to though)

PR #1824, opened 2025-07-08
fix(execute-opt): misc fixes to rv32/bigint e1 execute

shuklaayush:
- fixed some execute functions to match the spec. details in comments

### DIFF
--- a/extensions/bigint/circuit/src/branch_eq.rs
+++ b/extensions/bigint/circuit/src/branch_eq.rs
@@ -42,7 +42,7 @@ impl Rv32BranchEqual256Step {
 #[derive(AlignedBytesBorrow, Clone)]
 #[repr(C)]
 struct BranchEqPreCompute {
-    c: u32,
+    imm: isize,
     a: u8,
     b: u8,
 }
@@ -71,12 +71,18 @@ impl<F: PrimeField32> StepExecutorE1<F> for Rv32BranchEqual256Step {
             e,
             ..
         } = inst;
+        let c = c.as_canonical_u32();
+        let imm = if F::ORDER_U32 - c < c {
+            -((F::ORDER_U32 - c) as isize)
+        } else {
+            c as isize
+        };
         let e_u32 = e.as_canonical_u32();
         if d.as_canonical_u32() != RV32_REGISTER_AS || e_u32 != RV32_MEMORY_AS {
             return Err(InvalidInstruction(pc));
         }
         *data = BranchEqPreCompute {
-            c: c.as_canonical_u32(),
+            imm,
             a: a.as_canonical_u32() as u8,
             b: b.as_canonical_u32() as u8,
         };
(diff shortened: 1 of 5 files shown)


extensions/rv32im/circuit/src/divrem/core.rs
```diff
@@ -640,20 +640,20 @@ struct RemuOp;
...
-            [(1 << (RV32_CELL_BITS - 1)) - 1; 4]
-        } else {
-            let rs1 = i32::from_le_bytes(rs1);
-            let rs2 = i32::from_le_bytes(rs2);
-            (rs1 / rs2).to_le_bytes()
+        let rs1_i32 = i32::from_le_bytes(rs1);
+        let rs2_i32 = i32::from_le_bytes(rs2);
+        match (rs1_i32, rs2_i32) {
+            (_, 0) => [u8::MAX; 4],
+            (i32::MIN, -1) => rs1,
```
shuklaayush, line 647: > If division overflow (i.e. only rs1 is -2^(XLEN - 1) and rs2 is -1), the quotient is -2^(XLEN - 1).

fixed according to spec

extensions/rv32im/circuit/src/less_than/core.rs
```diff
@@ -405,9 +405,9 @@ unsafe fn execute_e1_impl<
 
     let rs1 = vm_state.vm_read::<u8, 4>(RV32_REGISTER_AS, pre_compute.b as u32);
     let rs2 = if E_IS_IMM {
-        imm_to_bytes(pre_compute.c as u32)
+        imm_to_bytes(pre_compute.c)
```
shuklaayush, line 408: `c` can be immediate so should be `u32`

extensions/rv32im/circuit/src/loadstore/core.rs
```diff
@@ -490,22 +509,33 @@ unsafe fn execute_e1_impl<
...
 }
+
 trait LoadStoreOp {
     const IS_LOAD: bool;
     const HOST_READ: bool;
 
     /// Return if the operation is valid.
     fn compute_write_data(
-        write_data: &mut [u8; RV32_REGISTER_NUM_LIMBS],
+        write_data: &mut [u32; RV32_REGISTER_NUM_LIMBS],
```
shuklaayush, line 538: i had to change this to `u32` to have the same function support stores into the native address space which should be supported according to the spec
https://github.com/openvm-org/openvm/blob/feat/new-execution/docs/specs/ISA.md#loadstore

extensions/bigint/circuit/src/branch_eq.rs
```diff
@@ -71,12 +71,18 @@ impl<F: PrimeField32> StepExecutorE1<F> for Rv32BranchEqual256Step {
             e,
             ..
         } = inst;
+        let c = c.as_canonical_u32();
+        let imm = if F::ORDER_U32 - c < c {
+            -((F::ORDER_U32 - c) as isize)
+        } else {
+            c as isize
+        };
```
shuklaayush, line 79: copied from rv32 to handle jumps with negative offsets

extensions/rv32im/circuit/src/less_than/core.rs
```diff
@@ -340,7 +340,7 @@ where
 struct LessThanPreCompute {
     a: u8,
     b: u8,
-    c: u8,
+    c: u32,
```
nyunyunyunyu: Move `c` to top

extensions/rv32im/circuit/src/loadstore/core.rs
nyunyunyunyu: The current implementation needs to convert 4 `F`s into `u32`s then convert them back to `F`. This is not optimal.
We can let `LoadStoreOp ` trait take a generic `F` or `u8` and have different implementations. We also need a trait for writing back into memory.

PR #1830, opened 2025-07-09
chore(sdk): remove #[serde(flatten)]

Qumeric:
`bincode` does not support `#[serde(flatten)]` (see [this](https://github.com/bincode-org/bincode/issues/245)). 

I think this likely a strong enough reason to remove it from here as the benefit of having it seems minor.

Of course, it can be worked around on the client with wrapper type and/or custom serializer, so it's more like nice to have, not must have.


Qumeric:
Test failure were unrelated, I rebased and it passes now

jonathanpwang:
if it's not causing an issue, perhaps we should close it for now?

PR #1839, opened 2025-07-09
feat: use cached max trace heights in leaf/internal aggregation

stephenh-axiom-xyz:
Resolves INT-4375.
- Hardcodes `NativeConfig` max trace heights (derived from the `kitchen-sink` benchmark) and uses them during leaf/internal verification.
- Adds metering to phantom chip, removes metering from public values chip (since estimation is derived from config)


crates/sdk/src/prover/vm/local.rs
```diff
@@ -36,26 +37,46 @@ impl<SC: StarkGenericConfig, VC, E: StarkFriEngine<SC>> VmLocalProver<SC, VC, E>
...
-            overridden_heights,
+            continuation_heights: overridden_heights,
+            single_segment_heights: None,
             _marker: PhantomData,
         }
     }
 
-    pub fn set_override_trace_heights(&mut self, overridden_heights: VmComplexTraceHeights) {
-        self.overridden_heights = Some(overridden_heights);
+    pub fn new_with_overridden_single_segment_trace_heights(
```
shuklaayush, line 60: maybe we should just have a `default()` and follow the builder pattern which consumes `mut self` and returns `Self` to do `with_overridden_single_segment_trace_heights()` and `with_single_segment_trace_heights()`

benchmarks/prove/src/bin/kitchen_sink.rs
```diff
@@ -88,6 +90,68 @@ fn main() -> Result<()> {
...
+                    full_agg_pk
+                        .agg_stark_pk
+                        .leaf_vm_pk
+                        .vm_pk
+                        .trace_height_constraints
+                        .clone(),
+                );
+                executor
+            };
+            let max_trace_heights = executor
```
jonathanpwang, line 118: try removing this

crates/sdk/src/prover/vm/local.rs
```diff
@@ -27,7 +27,8 @@ use crate::prover::vm::{
 pub struct VmLocalProver<SC: StarkGenericConfig, VC, E: StarkFriEngine<SC>> {
     pub pk: Arc<VmProvingKey<SC, VC>>,
     pub committed_exe: Arc<VmCommittedExe<SC>>,
-    overridden_heights: Option<VmComplexTraceHeights>,
+    continuation_heights: Option<VmComplexTraceHeights>,
```
jonathanpwang, line 30: this is really weird and questionable, but maybe I will just fix it in my PR

PR #1840, opened 2025-07-09
feat: implement new vm extension traits on sha256, keccak256, bigint, algebra, pairing

arayikhalatyan:


extensions/algebra/circuit/src/config.rs
```diff
@@ -1,23 +1,32 @@
...
+    #[extension(generics = false)]
     pub base: Rv32I,
-    #[extension]
+    #[extension(generics = false)]
     pub mul: Rv32M,
-    #[extension]
+    #[extension(generics = false)]
     pub io: Rv32Io,
-    #[extension]
+    #[extension(generics = false)]
```
jonathanpwang, line 29: huh do none of these need `<F>`? maybe we should make `generics = false` the default

extensions/algebra/circuit/src/fp2_extension.rs
```diff
@@ -62,144 +73,293 @@ impl Fp2Extension {
...
+    Val<SC>: PrimeField32,
+{
+    fn extend_prover(
+        &self,
+        inventory: &mut ChipInventory<SC, RA, CpuBackend<SC>>,
+    ) -> Result<(), ChipInventoryError> {
+        let range_checker = inventory.range_checker()?.clone();
+        let timestamp_max_bits = inventory.airs().config().memory_config.clk_max_bits;
+        // TODO: add a getter for pointer_max_bits
+        let pointer_max_bits = 29;
```
jonathanpwang, line 275: wait I see above you could get `pointer_max_bits` from `memory_config`. should we stop propagating these TODOs?

extensions/algebra/circuit/src/fp2_extension.rs
jonathanpwang, line 1: it's sad that the code here did get longer from the slight variations of each trait impl 3 times. but I can't see how to do it any better than what you have

extensions/algebra/circuit/src/config.rs
arayikhalatyan: yeahh, most of the `Step`s don't need `F`

extensions/algebra/circuit/src/fp2_extension.rs
arayikhalatyan: `AirInventory` has it but I think `ExecutorInventoryBuilder` and `ChipInventory` don't keep the `pointer_max_bits` and I didn't know the best way to add it to those structs

jonathanpwang: `ChipInventory` has `AirInventory` inside

`ExecutorInventoryBuilder`: ok we should add `SystemConfig` into `ExecutorInventory`

PR #1841, opened 2025-07-10
test: emit execution time metrics for each instruction

stephenh-axiom-xyz:
Part of INT-4376 resolution, introduces a way to track total E3 time each opcode takes to execute.

TODO: decide how to gate this metric collection


stephenh-axiom-xyz:
Closing since I am no longer working on the ticket, but will leave the branch open

PR #1842, opened 2025-07-10
fix(test): use local `openvm` for testing `cargo openvm init`

jonathanpwang:
Currently if we change the workspace version before tagging, the integration test for `cargo openvm init` will break because it creates a guest crate that references the tagged version. I updated the test so it patches to the local openvm if the environmental variable `USE_LOCAL_OPENVM=1` is set.


stephenh-axiom-xyz (approved):
LGTM

PR #1843, opened 2025-07-10
fix(new-execution): add register contribtions on `reset_segment`

shuklaayush:
for metering, we assume all registers are going to be read during execution and don't add their contributions in the hot loop.
this pr fixes the `reset_segment` function to also add the register contributions


stephenh-axiom-xyz (approved):
LGTM

jonathanpwang (approved):

PR #1847, opened 2025-07-10
feat: new vm extension traits

arayikhalatyan:
all the int tests compile and run now but some still fail: investigating that now
Also, fixed `pointer_max_bits` related TODOs, but there are still 3 TODOs let related to getting the `VariableRangeCheckerBus` from `ExecutorInventory`

update: all int tests pass now except `ruint`

### DIFF
--- a/crates/vm/src/arch/extensions.rs
+++ b/crates/vm/src/arch/extensions.rs
@@ -394,7 +394,7 @@ impl<F, E> ExecutorInventoryBuilder<'_, F, E> {
     }
 
     /// Returns the maximum number of bits used to represent addresses in memory
-    pub fn address_bits(&self) -> usize {
+    pub fn pointer_max_bits(&self) -> usize {
         self.new_inventory.config().memory_config.pointer_max_bits
     }
 }
--- a/crates/vm/src/arch/extensions.rs
+++ b/crates/vm/src/arch/extensions.rs
@@ -108,6 +108,7 @@ where
 // ======================= Different Inventory Struct Definitions =============================
 
 pub struct ExecutorInventory<E> {
+    config: SystemConfig,
     /// Lookup table to executor ID.
     /// This is stored in a hashmap because it is _not_ expected to be used in the hot path.
     /// A direct opcode -> executor mapping should be generated before runtime execution.
@@ -203,8 +204,9 @@ where
 impl<E> ExecutorInventory<E> {
     /// Empty inventory should be created at the start of the declaration of a new extension.
     #[allow(clippy::new_without_default)]
-    pub fn new() -> Self {
+    pub fn new(config: SystemConfig) -> Self {
         Self {
+            config,
             instruction_lookup: Default::default(),
             executors: Default::default(),
             ext_start: vec![0],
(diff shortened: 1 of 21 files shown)


crates/vm/src/arch/extensions.rs
```diff
@@ -384,6 +392,11 @@ impl<F, E> ExecutorInventoryBuilder<'_, F, E> {
             .iter()
             .filter_map(|e| e.as_any_kind().downcast_ref())
     }
+
+    /// Returns the maximum number of bits used to represent addresses in memory
+    pub fn address_bits(&self) -> usize {
```
jonathanpwang, line 397: maybe we should keep it `pointer_max_bits` to avoid confusion?

extensions/algebra/circuit/src/fp2_extension.rs
```diff
@@ -89,69 +90,67 @@ impl<F: PrimeField32> VmExecutionExtension<F> for Fp2Extension {
         &self,
         inventory: &mut ExecutorInventoryBuilder<F, Fp2ExtensionExecutor>,
     ) -> Result<(), ExecutorInventoryError> {
-        // TODO: add getter for pointer_max_bits
-        let pointer_max_bits = 29;
-        let range_checker_bus = inventory.range_checker().bus;
-        for (i, modulus) in self.supported_moduli.iter().enumerate() {
+        let pointer_max_bits = inventory.address_bits();
+        // TODO: somehow get the range checker bus from `ExecutorInventory`
```
jonathanpwang, line 94: wait this is unsafe? or is this a dummy bus?

extensions/algebra/circuit/src/fp2_extension.rs
```diff
@@ -89,69 +90,67 @@ impl<F: PrimeField32> VmExecutionExtension<F> for Fp2Extension {
...
+        let range_checker_bus = VariableRangeCheckerBus::new(1, 22);
```
jonathanpwang, line 95: instead of 1, use `u16::MAX` or something and name this `dummy_range_bus` so it's clear it's not used

crates/vm/src/arch/extensions.rs
```diff
@@ -459,6 +472,11 @@ impl<SC: StarkGenericConfig> AirInventory<SC> {
         }
         builder.generate_pk()
     }
+
+    /// Returns the maximum number of bits used to represent addresses in memory
+    pub fn address_bits(&self) -> usize {
```
jonathanpwang, line 477: pointer_max_bits

crates/vm/src/arch/extensions.rs
```diff
@@ -566,6 +584,10 @@ where
             })
             .collect()
     }
+
+    pub fn timestamp_max_bits(&self) -> usize {
+        self.airs.config().memory_config.clk_max_bits
```
jonathanpwang, line 589: should we just rename the field in memory_config to timestamp_max_bits too...?

jonathanpwang: ok maybe we do that later to avoid merge conflicts

PR #1857, opened 2025-07-11
fix: rv32im unit tests

arayikhalatyan:
all the rv32im unit tests pass + sha256 + keccak256 + native (except VirtualMachine)

should we rename `tester.address_bits -> tester.pointer_max_bits`? (I wanna do that after all the tests are working so it would be just a rename symbol operation)

### DIFF
--- a/extensions/rv32im/circuit/src/base_alu/tests.rs
+++ b/extensions/rv32im/circuit/src/base_alu/tests.rs
@@ -1,11 +1,9 @@
 use std::{array, borrow::BorrowMut};
 
-use openvm_circuit::arch::{
-    testing::{VmChipTestBuilder, BITWISE_OP_LOOKUP_BUS},
-    VmAirWrapper,
-};
+use openvm_circuit::arch::testing::{TestChipHarness, VmChipTestBuilder, BITWISE_OP_LOOKUP_BUS};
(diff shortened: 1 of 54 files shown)


jonathanpwang:
> should we rename tester.address_bits -> tester.pointer_max_bits? (I wanna do that after all the tests are working so it would be just a rename symbol operation)

yep, you can make a note so we won't forget

extensions/rv32im/circuit/src/base_alu/tests.rs
```diff
@@ -25,50 +23,60 @@ use test_case::test_case;
...
-    let mut chip = Rv32BaseAluChip::new(
-        VmAirWrapper::new(
-            Rv32BaseAluAdapterAir::new(
-                tester.execution_bridge(),
-                tester.memory_bridge(),
-                bitwise_bus,
-            ),
-            BaseAluCoreAir::new(bitwise_bus, BaseAluOpcode::CLASS_OFFSET),
+    let bitwise_chip =
+        SharedBitwiseOperationLookupChip::<RV32_CELL_BITS>::new(BitwiseOperationLookupChip::<
```
jonathanpwang, line 51: use `Arc` because `Shared..` is a typedef

PR #1862, opened 2025-07-14
feat: make all unit tests pass with new traits

arayikhalatyan:
Resolves INT-4423

did some renaming:
`Rv32IsEqualModeAdapterFiller->Rv32IsEqualModAdapterFiller`
`Rv32IsEqualModeAdapterStep->Rv32IsEqualModAdapterStep`
 `...multdiv...->...muldiv...` to match the naming of the related stuff

It seems like the only commented out tests remaining are `system/memory/tests` and `system/memory/merkle/tests`, but those were commented out before `new-exec-device` changes

### DIFF
--- a/benchmarks/execute/benches/execute.rs
+++ b/benchmarks/execute/benches/execute.rs
@@ -1,151 +1,171 @@
-// use std::{path::Path, sync::OnceLock};
+use std::{path::Path, sync::OnceLock};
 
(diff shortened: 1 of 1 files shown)


jonathanpwang:
I'll fix the vm integration tests -- making some util function updates.
ecc integration test fails because we need sdk for a config that has k256+keccak -- will address later

PR #1871, opened 2025-07-15
chore: Compact E1/E2 PreCompute Memory 

nyunyunyunyu:
Previously all pre-computes take the same size of memory, which is the maximum pre-compute size of all instructions. Now each pre-compute only takes the minimal necessary memory. There are still paddings but it's much more efficient(save ~75% memory, but it's usually just less than 1MB).


jonathanpwang:
please rebase for `feat/new-exec-device`

PR #1879, opened 2025-07-17
chore: Optimize E2 Height Meter

nyunyunyunyu:
Changes:
- Increase the default segment check frequency from 100 to 1000.
- Postpone adapter height change to segment check instead of every memory access. There is still 1 adapter height update in `on_memory_operation` but postponing it will make benchmark slower.

Benchmarks: 
Codespeed and my local show significant performance gain. However, [reth-benchmark](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/16380276462) doesn't have any change. I don't have a good explanation yet.

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
@@ -9,13 +9,13 @@ pub struct BitSet {
 }
 
 impl BitSet {
-    pub fn new(size_bits: usize) -> Self {
-        let num_words = 1 << size_bits.saturating_sub(6);
+    pub fn new(num_bits: usize) -> Self {
         Self {
-            words: vec![0; num_words].into_boxed_slice(),
+            words: vec![0; num_bits.div_ceil(u64::BITS as usize)].into_boxed_slice(),
         }
     }
 
+    #[inline(always)]
     pub fn insert(&mut self, index: usize) -> bool {
         let word_index = index / 64;
         let bit_index = index % 64;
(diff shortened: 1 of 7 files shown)


crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -26,6 +26,39 @@ impl BitSet {
         !was_set
     }
 
+    /// Set all bits within [start, end) to 1, return the number of flipped bits.
+    #[inline(always)]
+    pub fn insert_range(&mut self, start: usize, end: usize) -> usize {
```
jonathanpwang, line 31: debug assert `end != 0`?

crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -172,4 +188,63 @@ impl<const PAGE_BITS: usize> MemoryCtx<PAGE_BITS> {
             trace_heights[adapter_idx] += num << (size_bits - adapter_bits + 1);
         }
     }
+
+    #[inline(always)]
+    pub(crate) fn lazy_update_heights(&mut self, trace_heights: &mut [u32]) {
```
jonathanpwang, line 193: can you add a comment about what heights are actually updated here

jonathanpwang: from what I can tell this is just for boundary

crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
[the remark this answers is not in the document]
nyunyunyunyu: Well i used to try to use this to handle the case that an access touches multiple pages. Instead of iterating over all blocks, this can batch proceed 64 pages together. Unfortunately it turns out to be a negative optimization :(. I hope this function could be useful at some time so I kept it.

PR #1884, opened 2025-07-20
fix(new-exec-device): sdk and force fixed trace heights

jonathanpwang:
resolves INT-4425


jonathanpwang:
@nyunyunyunyu I will merge to consolidate branches, but please review the Root keygen and prover logic.

PR #1885, opened 2025-07-21
fix(new-exec-device): update `VmProverExtension` and `VmProverBuilder`

jonathanpwang:
Update the traits to allow implementing prover extensions for a VM extension and prover backend in a separate crate that may not own the circuit extension or prover backend.

Unfortunately changed the SDK to take an extra `app_vm_builder` argument because the config is in proving key, which is backend independent, but we might use a different `VmBuilder` based on the backend (cpu vs gpu).


jonathanpwang:
@arayikhalatyan still wip but works for rv32im. I might go back to a wrapper struct just for the ProverConfig to avoid cluttering the SDK functions

PR #1888, opened 2025-07-22
fix(ci): install CLI with --locked

yi-sun:
Passing run: https://github.com/openvm-org/openvm/actions/runs/16460596310/job/46526942880


jonathanpwang (approved):

HrikB:
> @HrikB do you know what the --vk-output error is? did we remove it

I'm not sure which error you're referring to but `--vk-output` was replaced with `--output-dir`

PR #1891, opened 2025-07-23
feat: Parse struct name in algebra + ECC init macros as string

Avaneesh-axiom:
Currently, we use the struct path as a shared id between the `*_declare!` and `*_init!` macros so that they could agree on the extern func names. However, this requires the struct to be in scope (as well as some traits like `WeierstrassPoint` and `IntMod`). This isn't good for UX with the guest library re-org since these structs and traits are internal, and ideally should be invisible to the user.

This PR modifies the `complex_init!`, `sw_init!`, and `te_init!` macros to take in the struct name as a string. This requires some logic from the setup extern funcs to be moved to the ECC struct's `set_up_once` associated function. In particular, the setup extern function is now just a wrapper on the custom asm instructions. It's purpose is simply to select the correct opcode (which is how it should have been originally).


jonathanpwang:
You can merge #1945 and this branch into your twisted edwards branch

PR #1901, opened 2025-07-25
fix(new-execution): replace segmentation strategy with segmentation limits

shuklaayush:
Resolves INT-4316


crates/vm/src/arch/config.rs
```diff
@@ -258,14 +250,12 @@ impl SystemConfig {
...
-        self.segmentation_strategy = Arc::new(
-            DefaultSegmentationStrategy::new_with_max_segment_len(max_segment_len),
-        );
+        self.segmentation_limits.max_trace_height = max_segment_len as u32;
         self
     }
 
-    pub fn set_segmentation_strategy(&mut self, strategy: Arc<dyn SegmentationStrategy>) {
-        self.segmentation_strategy = strategy;
+    pub fn set_segmentation_limits(&mut self, limits: SegmentationLimits) {
```
jonathanpwang, line 257: nit: you could use `WithSetters` from `getset`

crates/vm/src/arch/execution_mode/metered/segment_ctx.rs
```diff
@@ -30,6 +30,23 @@ impl Default for SegmentationLimits {
     }
 }
 
+impl SegmentationLimits {
+    pub fn with_max_trace_height(mut self, max_trace_height: u32) -> Self {
```
jonathanpwang, line 34: ditto

jonathanpwang (approved):
LGTM

PR #1905, opened 2025-07-25
fix: native `FriReducedOpeningRecord` size calculation

teokitan:
The size of `a_prev_buf` (i.e. `a_prev_size`) was not being factored in the size calculation for `FriReducedOpeningRecordMut`, sometimes making the record too small.


arayikhalatyan (approved):
LGTM, good catch

PR #1917, opened 2025-07-30
fix(docs): update isa spec to specify alignment requirements

shuklaayush:
also updated `CHANGELOG.md`

### DIFF
--- a/docs/specs/ISA.md
+++ b/docs/specs/ISA.md
@@ -133,12 +133,15 @@ to. Address space `0` is considered a read-only array with `[a]_0 = a` for any `
 #### Memory Accesses and Block Accesses
 
 VM instructions can access (read or write) a contiguous list of cells (called a **block**) in a single address space.
-The block size must be in the set `{1, 2, 4, 8, 16, 32}`, and the access has alignment requirements that are configurable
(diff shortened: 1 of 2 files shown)


docs/specs/ISA.md
```diff
@@ -133,9 +133,11 @@ to. Address space `0` is considered a read-only array with `[a]_0 = a` for any `
 #### Memory Accesses and Block Accesses
 
 VM instructions can access (read or write) a contiguous list of cells (called a **block**) in a single address space.
-The block size must be in the set `{1, 2, 4, 8, 16, 32}`, and the access does not need to be aligned, meaning that
-it can start from any pointer address, even those not divisible by the block size. An access is called a **block access
-** if it has size greater than 1. Block accesses are not supported for address space `0`.
+The block size must be in the set `{1, 2, 4, 8, 16, 32}`, and the access has alignment requirements that depend on the
+address space. For non-native address spaces (address spaces other than `4`), all memory accesses must be 4-byte aligned,
```
jonathanpwang, line 137: make an extra clarification that the RISC-V `lb,lh,sb,sh` instructions still work with alignment equal to size of the access because the block size differs from the size of the access in those cases.

jonathanpwang (approved):
LGTM after comments

PR #1922, opened 2025-08-01
refactor(new-execution): make `VmSegmentState` a superset of `VmState`

shuklaayush:
Resolves INT-4447


crates/vm/src/arch/segment.rs
```diff
@@ -19,37 +21,39 @@ use crate::{
     },
 };
 
+/// Represents the full execution state of a VM during segment execution.
 pub struct VmSegmentState<F, MEM, CTX> {
```
jonathanpwang, line 25: maybe we should rename it VmExecutionState

PR #1924, opened 2025-08-03
feat: `InterpretedInstance` holds pre-computed handlers

jonathanpwang:
For ultra-low latency, all data that can be pre-computed once the program binary is known should be held in `InterpretedInstance` and not re-computed on the fly. This PR does this and closes INT-4470

- `VmExecutor` is constructed once the VM config is known and holds all data that can be determined before given a program binary. (For execution this is only the executors)
- `InterpretedInstance` should be created once the program binary is known and holds all pre-computed buffers and function pointers, so that `execute` can focus on only runtime specific computations.
- Refactored so that the `InterpretedInstance` now provides the only `execute` and `execute_metered` functions, and `VmExecutor` only provides constructors for the `InterpretedInstance`, which is separated between pure (E1) and metered (E2) via the `Ctx` generic.
- Also added the ability to `execute_from_state` for future E1 purposes.
- (Out of scope here) Note that E2 always executes until completion -- the current API seems not designed for any kind of async execution of segments. I think to do this perhaps the easiest is if a separate thread is created with view access to the `SegmentationCtx` which can poll for new `Segment` creation and start off other processes.
- (Out of scope) I want to do a similar refactor for `VirtualMachine` vs a new `VmInstance` type, but going to hold off and try to switch E3 to use `InterpretedInstance` first

The codspeed benchmarks should be largely ignored since any perf improvements are mostly from decreasing the pre-computation time.

### DIFF
--- a/crates/vm/src/arch/vm.rs
+++ b/crates/vm/src/arch/vm.rs
@@ -52,9 +52,9 @@ use crate::{
         hasher::poseidon2::vm_poseidon2_hasher,
         interpreter::InterpretedInstance,
         AirInventoryError, AnyEnum, ChipInventoryError, ExecutionState, ExecutorInventory,
-        ExecutorInventoryError, InsExecutorE2, InstructionExecutor, SystemConfig, TraceFiller,
-        VmBuilder, VmCircuitConfig, VmExecutionConfig, VmSegmentExecutor, VmSegmentState,
-        PUBLIC_VALUES_AIR_ID,
+        ExecutorInventoryError, InsExecutorE2, InstructionExecutor, StaticProgramError,
+        SystemConfig, TraceFiller, VmBuilder, VmCircuitConfig, VmExecutionConfig,
+        VmSegmentExecutor, VmSegmentState, PUBLIC_VALUES_AIR_ID,
     },
     execute_spanned,
     system::{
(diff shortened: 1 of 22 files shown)


jonathanpwang:
@shuklaayush the pairing bench keeps failing on codspeed but not locally for me https://github.com/openvm-org/openvm/actions/runs/16709485569/job/47292860089?pr=1924

crates/vm/src/arch/vm.rs
```diff
@@ -592,15 +521,26 @@ where
         })
     }
 
-    /// Same as [`VmExecutor::create_initial_state`] but sets more information for performance
-    /// metrics when feature "perf-metrics" is enabled.
+    /// Same as [`InterpretedInstance::create_initial_state`] but sets more information for
+    /// performance metrics when feature "perf-metrics" is enabled.
```
shuklaayush, line 525: not a fan of this duplication. i feel it should be a separate function that takes config (and maybe an optional reference to pk or that can be a separate method) as input and not a struct method.

jonathanpwang: let me move this to a constructor of `VmState` in https://github.com/openvm-org/openvm/pull/1926

PR #1930, opened 2025-08-05
fix: setup e12 should also write to match e3 state

luffykai:


extensions/ecc/circuit/src/weierstrass_chip/double.rs
```diff
@@ -327,21 +397,22 @@ unsafe fn execute_e1_impl<
...
+    const CURVE_TYPE: u8,
 >(
     pre_compute: &[u8],
     vm_state: &mut VmExecState<F, GuestMemory, CTX>,
 ) {
     let pre_compute: &EcDoublePreCompute = pre_compute.borrow();
 
-    execute_e12_setup_impl::<_, _, BLOCKS, BLOCK_SIZE>(pre_compute, vm_state);
+    execute_e12_impl::<_, _, BLOCKS, BLOCK_SIZE, CURVE_TYPE>(pre_compute, vm_state, true);
 }
```
shuklaayush, line 416: i guess we don't need this function anymore

extensions/ecc/circuit/src/weierstrass_chip/double.rs
```diff
@@ -386,6 +458,7 @@ unsafe fn execute_e12_impl<
 >(
     pre_compute: &EcDoublePreCompute,
     vm_state: &mut VmExecState<F, GuestMemory, CTX>,
+    is_setup: bool,
```
shuklaayush, line 461: can you make this a const generic instead?

extensions/ecc/circuit/src/weierstrass_chip/double.rs
```diff
@@ -239,25 +239,60 @@ impl<F: PrimeField32, const BLOCKS: usize, const BLOCK_SIZE: usize> InsExecutorE
...
+        if let Some(curve_type) = {
             let modulus = &pre_compute.expr.builder.prime;
             let a_coeff = &pre_compute.expr.setup_values[0];
             get_curve_type(modulus, a_coeff)
         } {
-            match curve_type {
-                CurveType::K256 => {
-                    Ok(execute_e1_impl::<_, _, BLOCKS, BLOCK_SIZE, { CurveType::K256 as u8 }>)
+            if is_setup {
+                match curve_type {
```
shuklaayush, line 248: maybe `match(is_setup, curve_type)`

shuklaayush (changes requested):

extensions/ecc/circuit/src/weierstrass_chip/add_ne.rs
```diff
@@ -382,27 +424,28 @@ unsafe fn execute_e2_impl<
...
     vm_state: &mut VmExecState<F, GuestMemory, CTX>,
 ) {
-    let pre_compute: &E2PreCompute<EcAddNePreCompute> = pre_compute.borrow();
+    let e2_pre_compute: &E2PreCompute<EcAddNePreCompute> = pre_compute.borrow();
     vm_state
         .ctx
-        .on_height_change(pre_compute.chip_idx as usize, 1);
-    execute_e12_impl::<_, _, BLOCKS, BLOCK_SIZE, FIELD_TYPE>(&pre_compute.data, vm_state);
+        .on_height_change(e2_pre_compute.chip_idx as usize, 1);
+    execute_e12_impl::<_, _, BLOCKS, BLOCK_SIZE, FIELD_TYPE, false>(&pre_compute[4..], vm_state);
```
luffykai, line 431: not sure if there is a better way than `4..`?

extensions/ecc/circuit/src/weierstrass_chip/double.rs
```diff
@@ -239,29 +239,75 @@ impl<F: PrimeField32, const BLOCKS: usize, const BLOCK_SIZE: usize> InsExecutorE
...
-                    Ok(execute_e1_impl::<_, _, BLOCKS, BLOCK_SIZE, { CurveType::BN254 as u8 }>)
-                }
-                CurveType::BLS12_381 => {
-                    Ok(execute_e1_impl::<_, _, BLOCKS, BLOCK_SIZE, { CurveType::BLS12_381 as u8 }>)
+                (true, CurveType::P256) => {
+                    Ok(
+                        execute_e12_impl::<_, _, BLOCKS, BLOCK_SIZE, { CurveType::P256 as u8 }, true>,
+                    )
                 }
+                (true, CurveType::BN254) => Ok(execute_e12_impl::<
```
jonathanpwang, line 258: there was no way to avoid this...?

jonathanpwang: ok I think technically if you made another function that did the routing, you could decrease code, but then you need another function

jonathanpwang (approved):

PR #1934, opened 2025-08-05
feat(new-execution): vectorize native fri chip execution

shuklaayush:


shuklaayush:
closing since perf doesn't improve on x86 with avx2

PR #1937, opened 2025-08-06
fix: sw_init macro failed to compile

lightsing:
`as_ref` is ambiguous when other candidate are available.


lightsing:
the repo is here, if you are interested in details

https://github.com/scroll-tech/ecies-rs

I current use my own fork with this fix, without this patch, you could reproduce this bug.

(possibly caused by [hybrid-array](https://docs.rs/hybrid-array/latest/hybrid_array/), not digging into this too much.)

jonathanpwang:
Thanks - I think it may have been fixed by https://github.com/openvm-org/openvm/pull/1945 which we will merge to `main` soon. Will keep this open until we confirm it's fixed.

lightsing:
thx

PR #1941, opened 2025-08-07
perf(new-execution): try to optimize tracing memory

shuklaayush:


crates/vm/src/system/memory/online.rs
```diff
@@ -373,13 +373,11 @@ impl GuestMemory {
...
 pub struct AccessMetadata {
-    /// The starting pointer of the access
-    pub start_ptr: u32,
-    /// The block size of the memory access
-    pub block_size: u32,
-    /// The timestamp of the last access.
-    /// We don't _have_ to store it, but this is probably faster
-    /// in terms of cache locality
+    /// Offset to block start (in ALIGN units). If 0, this IS the block start.
+    pub offset_to_start: u8,
```
jonathanpwang, line 377: nit: to follow Xinding's principles, smaller sized items should be at the bottom (here it makes no difference, sizeof is still 8)

shuklaayush: makes sense

crates/vm/src/system/memory/online.rs
```diff
@@ -524,128 +600,203 @@ impl TracingMemory {
...
+            // Process merge
+            let mut prev_ts_buf = [0u32; MAX_SEGMENTS];
+
+            let mut max_timestamp = INITIAL_TIMESTAMP;
+
+            let mut ptr = merge_ptr;
+            let end_ptr = merge_ptr + merge_size;
+            let mut seg_idx = 0;
+            while ptr < end_ptr {
+                let (_, block_metadata) = self.get_block_metadata::<ALIGN>(address_space, ptr);
```
jonathanpwang, line 719: does this work properly now that `PagedVec::get` does not create the page? is it because you do `unwrap_or_default()` and then the page gets created in a later `set`?

crates/vm/src/system/memory/online.rs
```diff
@@ -524,128 +600,203 @@ impl TracingMemory {
...
 
+    /// Calculate splits and merges needed for a memory access.
+    /// Returns Some((splits, merge)) or None if no operations needed.
+    #[inline(always)]
+    #[allow(clippy::type_complexity)]
+    fn calculate_splits_and_merges<const BLOCK_SIZE: usize, const ALIGN: usize>(
+        &mut self,
+        address_space: usize,
+        pointer: usize,
+    ) -> Option<(Vec<(usize, usize)>, (usize, usize))> {
```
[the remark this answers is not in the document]
shuklaayush: Good point

crates/vm/src/system/memory/online.rs
shuklaayush: yes, i don't think we need to allocate pages if we're just fetching the values. i just return the default (zero) value. we only allocate pages on set

PR #1948, opened 2025-08-08
chore: remove an old `assert!(address_space < NATIVE_AS)`

Golovanov399:
This resolves INT-4549.

The assertion was a sanity check and it essentially checked that the "type size" should be greater than one. I replaced it with an `debug_assert_eq!(align, min block size from config)`.

While we are there, I also replaced some `cell_size` being `size_of::<T>()` or 1 with another value from config.

### DIFF
--- a/crates/vm/src/system/memory/online.rs
+++ b/crates/vm/src/system/memory/online.rs
@@ -650,12 +650,13 @@ impl TracingMemory {
     }
 
     #[inline(always)]
-    fn split_by_meta<T: Copy, const MIN_BLOCK_SIZE: usize>(
+    fn split_by_meta<const MIN_BLOCK_SIZE: usize>(
         &mut self,
         start_ptr: u32,
         timestamp: u32,
         block_size: u8,
         address_space: usize,
+        cell_size: usize,
     ) {
         if block_size == MIN_BLOCK_SIZE as u8 {
             return;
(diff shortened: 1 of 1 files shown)


crates/vm/src/system/memory/online.rs
```diff
@@ -650,12 +650,13 @@ impl TracingMemory {
...
 
     #[inline(always)]
-    fn split_by_meta<T: Copy, const MIN_BLOCK_SIZE: usize>(
+    fn split_by_meta<const MIN_BLOCK_SIZE: usize>(
         &mut self,
         start_ptr: u32,
         timestamp: u32,
         block_size: u8,
         address_space: usize,
+        cell_size: usize,
```
jonathanpwang, line 659: isn't this worse for performance though? if you can know the size statically, that is better

crates/vm/src/system/memory/online.rs
```diff
@@ -690,6 +691,15 @@ impl TracingMemory {
         pointer: usize,
         prev_values: &[T; BLOCK_SIZE],
     ) -> u32 {
+        debug_assert_eq!(ALIGN, self.data.memory.config[address_space].min_block_size);
+        let cell_size = unsafe {
```
jonathanpwang, line 695: better if you make this a debug assert and then the compiler still uses `size_of::<T>()`

jonathanpwang (commented):
LGTM except for performance let us keep `size_of::<T>()`

PR #1955, opened 2025-08-11
chore(new-execution): make executors stateless

jonathanpwang:
The `PreflightExecutor` already only needed `&self` instead of `&mut self` everywhere.
But I removed the `Mutex` from public values executor because it is confusing and can lead to mistakes when re-using the executor structs across runs.


crates/vm/src/system/public_values/core.rs
```diff
@@ -239,17 +246,20 @@ where
...
     fn generate_public_values(&self) -> Vec<F> {
-        let custom_pvs = self.custom_pvs.lock().unwrap();
-        custom_pvs.iter().map(|&x| x.unwrap_or(F::ZERO)).collect()
+        assert_eq!(
+            self.public_values.len(),
+            self.num_custom_pvs,
+            "Did not set public values"
+        );
+        self.public_values.clone()
     }
```
shuklaayush, line 255: would have been great if we could just get rid of public values altogether from the chip but i guess it's needed by `generate_proving_ctx`

PR #1957, opened 2025-08-11
feat: `VmInstance` struct to hold program-specific setup

jonathanpwang:
I renamed `VmLocalProver` to `VmInstance` and modified it so it holds `PreflightInterpretedInstance` so the execution doesn't includes the setup time to construct the interpreted instance. To avoid some self-reference ouroboros things, I used `Arc` instead of lifetimes.


crates/sdk/src/prover/vm/mod.rs
```diff
@@ -23,9 +23,6 @@ where
     let vm = VirtualMachine::new(engine, vm_builder, vm_pk.vm_config.clone(), d_pk)?;
     let cached_program_trace = vm.transport_committed_exe_to_device(committed_exe);
     // TODO[jpw]: remove this clone
-    Ok(VmLocalProver::new(
-        vm,
-        committed_exe.exe.clone(),
-        cached_program_trace,
-    ))
+    let instance = VmInstance::new(vm, committed_exe.exe.clone(), cached_program_trace)?;
```
shuklaayush, line 26: unrelated, but maybe we can remove this clone and instead clone at call sites

crates/vm/src/arch/interpreter_preflight.rs
```diff
@@ -1,31 +1,119 @@
...
+/// Note: This executor doesn't hold any VM state and can be used for multiple execution.
 pub struct PreflightInterpretedInstance<F, E> {
-    pub handler: ProgramHandler<F, E>,
-    executor_idx_to_air_idx: Vec<usize>,
+    // NOTE[jpw]: we use an Arc so that VmInstance can hold both VirtualMachine and
+    // PreflightInterpretedInstance. All we really need is to borrow `executors: &'a [E]`.
+    inventory: Arc<ExecutorInventory<E>>,
+
+    /// This is a map from (pc - pc_base) / pc_step -> [PcEntry].
+    /// We will map to `u32::MAX` if the program has no instruction at that pc.
```
shuklaayush, line 25: maybe clarify that `u32:::MAX` is for the `executor_idx`

crates/vm/src/arch/interpreter_preflight.rs
```diff
@@ -1,31 +1,119 @@
...
+    // PERF[jpw/ayush]: We could map directly to the raw pointer(u64) for executor, but storing the
+    // u32 may be better for cache efficiency.
```
shuklaayush, line 27: unrelated, but i think before that we should come up with something like precompute for e3 so that we don't have to store the full `Instruction<F>` struct

shuklaayush (approved):
looks fine. i would've preferred making `initial_state` an argument and making it the responsibility of the caller to reset it since it's easy to forget adding a `self.reset_state()` in some of these functions. but we can revisit this later

crates/vm/src/arch/interpreter_preflight.rs
jonathanpwang: yes, I tried this in that other PR; it was just too painful with our `AdapterTraceStep` trait as-is :(

PR #1958, opened 2025-08-11
fix(new-execution): rename trait/structs referencing e1/e2/e3

shuklaayush:
- rename e1, e2 -> metered, e3 -> preflight
- move variables in `MeteredCtx` into appropriate structs


crates/vm/src/arch/execution_mode/mod.rs
```diff
@@ -1,15 +1,15 @@
 use crate::{arch::VmExecState, system::memory::online::GuestMemory};
 
-pub mod e1;
 pub mod metered;
-pub mod tracegen;
+pub mod normal;
```
jonathanpwang, line 4: I had to be referring to it as `pure` -- I'm also ok with normal but thought `pure` might be a bit less ambiguous?

PR #1965, opened 2025-08-13
chore: special case insns metric to avoid double counting

PangZhi:
Currently the insns value is not correct as we double count the value from e1 (which does not have segment label). This PR special case the handling in MetricsDB to avoid double counting and also do validation the sum and e1 value match.

addresses INT-4687


jonathanpwang (changes requested):
Please demonstrate this works with some existing metrics.json and show the markdown output.

PangZhi:
I don't think we need the special case anymore with fix #1970

PR #1971, opened 2025-08-14
chore: migrate CUDA include header files

stephenh-axiom-xyz:


crates/circuits/primitives/cuda/buffer_view.cuh
jonathanpwang, line 1: was it always that here the `.cuh` are in `cuda/` and not `cuda/include/`?

PR #1974, opened 2025-08-15
chore: migrate CUDA primitive chips + test

stephenh-axiom-xyz:
No build yet, need to update histogram tests to not use `GpuChipTestBuilder`


Golovanov399:
> @arayikhalatyan @Golovanov399 how do you want to handle a) the `stark-backend-gpu` dependency and b) building? I currently use a local relative path for the former, and Grigorii is creating a build utility crate so I left out `build.rs` for now. Also, I've set `cuda` to be a default feature for now.

I am currently trying to build by copying `build.rs` and making the changes so that it includes everything it needs. Not sure about the stark-backend-gpu dependency yet

crates/circuits/poseidon2-air/src/lib.rs
```diff
@@ -28,6 +28,9 @@ pub use babybear::*;
 pub use config::*;
 pub use permute::*;
 
+#[cfg(all(feature = "cuda", test))]
```
Golovanov399, line 31: Do we not want this on in reth benchmark, for example? Why only test?

crates/circuits/primitives/cuda/include/utils.cuh
Golovanov399, line 1: ah it was `inlcude`, I was looking at this rename for a minute trying to understand what was going on

Golovanov399 (approved):

stephenh-axiom-xyz:
I've updated the workspace `Cargo.toml`, will leave without a `build.rs` for now. @Golovanov399 please lmk when you have your `build.rs` ready 🙏

crates/circuits/primitives/src/range_tuple/tests.rs
```diff
@@ -1,125 +0,0 @@
-use std::{array, iter, sync::Arc};
```
arayikhalatyan, line 1: Why are we deleting this file?

PR #1976, opened 2025-08-15
feat: migrate test builder

arayikhalatyan:
Towards INT-4700

Migrating test builder gpu


crates/vm/Cargo.toml
```diff
@@ -51,7 +54,13 @@ openvm-native-compiler.workspace = true
 openvm-rv32im-transpiler.workspace = true
 
 [features]
-default = ["parallel", "jemalloc"]
+default = ["parallel", "jemalloc", "cuda"]
+cuda = ["dep:stark-backend-gpu"]
```
stephenh-axiom-xyz, line 58: Add `openvm-circuit-primitives/cuda` and `openvm-poseidon2-air/cuda`

crates/vm/src/arch/testing/cuda.rs
```diff
@@ -64,23 +51,16 @@ use stark_backend_gpu::{
...
-pub mod memory;
-pub mod program;
-mod utils;
-pub use utils::*;
+// use crate::{
+//     primitives::{
+//         bitwise_op_lookup::BitwiseOperationLookupChipGPU, range_tuple::RangeTupleCheckerChipGPU,
+//         var_range::VariableRangeCheckerChipGPU,
+//     },
+//     system::{poseidon2::Poseidon2PeripheryChipGPU, DIGEST_WIDTH},
```
stephenh-axiom-xyz, line 59: How does this compile with only includes commented out?

crates/vm/src/arch/testing/cuda_abi.rs
```diff
@@ -11,8 +11,8 @@ use stark_backend_gpu::{
 pub mod execution_testing {
     use super::*;
 
-    extern "C" {
-        fn _execution_testing_tracegen(
+    unsafe extern "C" {
+        unsafe fn _execution_testing_tracegen(
```
stephenh-axiom-xyz, line 15: Do these need to be `unsafe`?

crates/vm/src/arch/testing/execution/mod.rs
```diff
@@ -12,6 +12,10 @@ use openvm_stark_backend::{
 use crate::arch::{ExecutionBus, ExecutionState};
 
 pub mod air;
+#[cfg(feature = "cuda")]
+mod cuda;
+#[cfg(feature = "cuda")]
+pub use cuda::*;
```
stephenh-axiom-xyz, line 18: Any reason not to just `pub mod cuda`?

stephenh-axiom-xyz (commented):
According to our spec `cuda_abi.rs` should be at a direct child of `src`, and the `cuda` folder containing the kernels should be a sibling. You can gate things in `cuda_abi.rs` using `#[cfg(any(test, feature = "test-utils"))]` I believe

crates/vm/src/arch/testing/cuda.rs
arayikhalatyan: This file isn't included, I have commented out the `mod cuda ` in `mod.rs`

crates/vm/src/arch/testing/cuda_abi.rs
arayikhalatyan: all extern "C" functions can only be unsafe I just added the explicit `unsafe` keyword because rust-analyzer was showing me smth about this

crates/vm/src/arch/testing/cuda.rs
stephenh-axiom-xyz: Hmm, if that's the case should we just wait for `system` to be finished? If we won't be able to test stuff yet it might be nice to do it all in one PR

crates/vm/src/arch/testing/execution/mod.rs
arayikhalatyan: hmm it's usually nicer to do this way so that in test files we do `use testing:BuilderGpu` instead of `use testing::cuda::BuilderGpu`. But now I'm thinking maybe it's cleaner to always have cuda in pathname when it's all feature gated🤔

crates/vm/src/arch/testing/cuda.rs
arayikhalatyan: Sure, I'm fine holding off merging this PR until system is merged

PR #1978, opened 2025-08-15
chore: migrate system

Golovanov399:
Towards INT-4744.

This moves the files from `tracegen/{cuda/,}src/system` to somewhere in openvm. This currently compiles both with and without `--features cuda` (well on a machine without cuda it won't compile with `--features cuda`). The tests don't compile, but it's because the testing utilities are missing.

---
todo list:

- [x] move stuff to `cuda/system/`
- [x] make tests compile (except the "undefined test utility" thing)
- [x] feature gate cuda dependencies
- [x] feature gate build script


crates/vm/Cargo.toml
```diff
@@ -74,3 +80,10 @@ mimalloc = ["openvm-stark-backend/mimalloc"]
 jemalloc = ["openvm-stark-backend/jemalloc"]
 jemalloc-prof = ["openvm-stark-backend/jemalloc-prof"]
 nightly-features = ["openvm-stark-sdk?/nightly-features"]
+cuda = [
+    "openvm-stark-sdk", # for BabyBearPoseidon2Config
```
stephenh-axiom-xyz, line 84: Should this be `dep:openvm-stark-sdk`? Not sure about the semantics

crates/vm/build.rs
```diff
@@ -0,0 +1,22 @@
...
+fn main() {
+    #[cfg(feature = "cuda")]
+    {
+        if !cuda_available() {
+            return; // Skip CUDA compilation
+        }
+
+        let builder = CudaBuilder::new()
+            .include_from_dep("DEP_CUDA_COMMON_INCLUDE") // Import headers
+            .include("../../cuda-includes")
```
stephenh-axiom-xyz, line 13: These header files have been moved to `circuits/mod-builder/cuda/include` and `circuits/poseidon2-air/cuda/include`

stephenh-axiom-xyz (changes requested):
Mostly LGTM, but please remove `cuda-includes`. The headers you're looking for are in `crates/circuits/mod-builder/cuda/include` and `crates/circuits/poseidon2-air/cuda/include`.

crates/vm/Cargo.toml
Golovanov399: honestly just took from `test-utils` from several lines before, and it fixed stuff. Maybe you're right though

crates/vm/build.rs
Golovanov399: I'll rewrite this but we probably want something very common to be outside of all crates

crates/vm/Cargo.toml
stephenh-axiom-xyz: Tbh am OK to leave it, what works works XD

crates/vm/build.rs
```diff
@@ -0,0 +1,23 @@
...
+    #[cfg(feature = "cuda")]
+    {
+        if !cuda_available() {
+            return; // Skip CUDA compilation
+        }
+
+        let builder = CudaBuilder::new()
+            .include_from_dep("DEP_CUDA_COMMON_INCLUDE") // Import headers
+            .include("../circuits/primitives/cuda/include")
+            .include("../circuits/mod-builder/cuda/include")
```
arayikhalatyan, line 14: hmm, does this need `mod-builder`?

stephenh-axiom-xyz (approved):
LGTM, we can move include headers later

PR #1982, opened 2025-08-16
perf: Batch Tracing Memory

nyunyunyunyu:


crates/vm/src/arch/record_arena.rs
```diff
@@ -162,8 +162,43 @@ impl<F: Field> RowMajorMatrixArena<F> for MatrixRecordArena<F> {
...
+    }
+
+    pub fn get_mut(&mut self) -> &mut T {
+        &mut self.inner
+    }
+}
+
 pub struct DenseRecordArena {
-    pub records_buffer: Cursor<Vec<u8>>,
+    pub records_buffer: AtomicCursor<Vec<u8>>,
```
jonathanpwang, line 201: I think we shouldn't do this at the record arena level because most of the time we don't want atomic cursor

I think we should just use the raw `Cursor` type, maybe replicating it ourselves, and then only when needed we use the atomic stuff. If we use it everywhere it will be slower

PR #2003, opened 2025-08-19
feat(ci): add benches for internal verifier execution

shuklaayush:
- add methods to generate fixtures for internal verifier
- update execute script to execute internal verifier
- add codspeed bench for internal verifier execution


crates/sdk/src/prover/agg.rs
```diff
@@ -264,7 +264,7 @@ where
 
     #[cfg(feature = "evm-prove")]
     #[instrument(name = "agg_layer", skip_all, fields(group = "root", idx = 0))]
-    fn generate_root_proof_impl(
+    pub fn generate_root_proof_impl(
```
jonathanpwang, line 267: do we really need to make this pub? I hoped to avoid it

jonathanpwang (approved):
I would get rid of the root benchmarks if it means we can keep the internals of root prover private and not `pub`.

Otherwise LGTM but please give instructions on how to run these benchmarks.

crates/sdk/src/prover/agg.rs
shuklaayush: changed it back to private

PR #2013, opened 2025-08-20
feat(nightly): execution `become`s faster

jonathanpwang:
Uses the nightly `become` keyword to tell LLVM to `musttail`. This did not work at first when the `Handler` type returned `Result` but after I changed it to return nothing it does.

Ideas taken from https://github.com/xacrimon/tcvm/blob/main/src/interp.rs

To remove code duplication, I created local declarative macros `dispatch!` for any "enum dispatch" logic of function pointers we were doing. The benefit is that this is now shared across four functions: {e1,e2} x {pre_compute,handler}. I did not make a single general purpose macro because Rust token parsing rules actually make this hard because it doesn't let you reference ident's you don't declare, and the ident's used actually vary based on the function.

Removed `pc_base` and just leave some empty space to avoid `pc - pc_base` calculation. This is because `self.pc_base` is a runtime variable, so we don't want to have to use a register (or worse load/store) to access it.

To complete this:
- [x] For each crate, add "tco" feature, and then add `#[create_tco_handler]` attribute above `execute_e1_impl` functions. Then for each `Executor` implementation, copy the `pre_compute` function implementation verbatim but switch to `handler` function signature and return the tco handler fn pointer instead.
- [x] Do the same for metered execution.
- [x] ~~Switch to x86 global asm instead of relying on LLVM if we want to be extra safe.~~ this seemed complicated and hard to do fully properly so I prefer `become`.

Closes INT-4309


shuklaayush:
> This is the claim that llvm always eliminates tail calls when functions tail call with the same function signature.

i'm skeptical of this claim but if it works on e1 of 1mn fibonacci then seems like it does work at least for rv32 opcodes. maybe we should also try kitchen-sink or just the whole suite

jonathanpwang:
I tried to make a macro to unify the `pre_compute` and `handler` methods, but the find replace of the function name seems surprisingly difficult.

jonathanpwang:
E1 comparison: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/17107140175

.github/workflows/benchmarks-execute.yml
```diff
@@ -66,12 +66,12 @@ jobs:
 
       - name: Build benchmarks
         working-directory: benchmarks/execute
-        run: cargo codspeed build --profile maxperf
+        run: cargo $TOOLCHAIN codspeed build --profile maxperf --features tco
```
jonathanpwang, line 69: @shuklaayush i'm not sure what the right syntax is here

shuklaayush: i think you only need to add `--features tco` to build, not run

crates/vm/derive/src/tco.rs
```diff
@@ -0,0 +1,128 @@
...
+
+            if exec_state.exit_code.is_err() {
+                // stop execution
+                return;
+            }
+            if #ctx_type::should_suspend(exec_state) {
+                return;
+            }
+            // exec_state.pc should have been updated by execute_impl at this point
+            let next_handler = interpreter.get_handler(exec_state.vm_state.pc);
```
shuklaayush, line 64: both `get_handler` and `get_pre_compute` call `get_pc_index`. maybe we can calculate the index once and reuse it

crates/vm/derive/src/tco.rs
```diff
@@ -0,0 +1,128 @@
...
+            if next_handler.is_none() {
```
shuklaayush, line 65: not required rn but maybe we can just return a null pointer instead of `Option`

jonathanpwang: again I hope the compiler just optimizes this out

crates/vm/derive/src/tco.rs
```diff
@@ -0,0 +1,128 @@
...
+        quote! { #fn_name(pre_compute, exec_state) }
+    } else {
+        quote! { #fn_name::<#(#generic_args),*>(pre_compute, exec_state) }
+    };
+
+    // Generate the TCO handler function
+    let handler_fn = quote! {
+        #[inline(never)]
+        unsafe fn #handler_name #handler_generics (
+            interpreter: &::openvm_circuit::arch::interpreter::InterpretedInstance<#f_type, #ctx_type>,
```
shuklaayush, line 42: not sure if it matters but maybe we just pass like a slimmed down instruction table that contains a function/mapping from pc -> (data, handler) instead of a reference to this whole struct

benchmarks/prove/Cargo.toml
```diff
@@ -33,8 +33,9 @@ metrics.workspace = true
 [dev-dependencies]
 
 [features]
-default = ["parallel", "jemalloc", "metrics", "evm"]
```
shuklaayush, line 36: not sure why this was changed

crates/vm/src/arch/execution.rs
```diff
@@ -72,6 +70,9 @@ pub enum ExecutionError {
     Inventory(#[from] ExecutorInventoryError),
     #[error("static program error: {0}")]
     Static(#[from] StaticProgramError),
+    // Placeholder error type for tco
+    #[error("error in VmExecState")]
+    ExecStateError,
```
shuklaayush, line 75: i don't think this is used

crates/vm/src/arch/interpreter.rs
```diff
@@ -42,15 +45,22 @@ pub struct InterpretedInstance<'a, F, Ctx> {
     #[allow(dead_code)]
     pre_compute_buf: AlignedBuf,
     /// Instruction table of function pointers and pointers to the pre-computed buffer. Indexed by
-    /// `pc_index = (pc - pc_base) / DEFAULT_PC_STEP`.
+    /// `pc_index = pc / DEFAULT_PC_STEP`.
+    /// SAFETY: The first `pc_base / DEFAULT_PC_STEP` entries will be unreachable. We do this to
+    /// avoid needing to subtract `pc_base` during runtime.
```
shuklaayush, line 50: is `pc_base` guaranteed to be bounded?

crates/vm/src/arch/interpreter.rs
```diff
@@ -62,17 +72,46 @@ struct TerminatePreCompute {
...
+            // - it is the responsibility of each Executor to ensure that pre_compute_insts contains
+            //   valid function pointers and pre-computed data
+            #[cfg(not(feature = "tco"))]
+            unsafe {
+                tracing::debug!("execute_trampoline");
+                execute_trampoline(&mut $exec_state, &$interpreter.pre_compute_insns);
+            }
+            #[cfg(feature = "tco")]
+            {
+                tracing::debug!("execute_tco");
```
shuklaayush, line 93: slightly pedantic, but i feel the tail-recursive function is more like a "trampoline"

crates/vm/src/arch/interpreter.rs
```diff
@@ -62,17 +72,46 @@ struct TerminatePreCompute {
...
+                let handler = $interpreter
+                    .get_handler($exec_state.pc)
+                    .ok_or(ExecutionError::PcOutOfBounds($exec_state.pc))?;
```
shuklaayush, line 96: set `exit_code`?

crates/vm/src/arch/interpreter.rs
```diff
@@ -42,15 +45,22 @@ pub struct InterpretedInstance<'a, F, Ctx> {
     #[allow(dead_code)]
     pre_compute_buf: AlignedBuf,
     /// Instruction table of function pointers and pointers to the pre-computed buffer. Indexed by
-    /// `pc_index = (pc - pc_base) / DEFAULT_PC_STEP`.
+    /// `pc_index = pc / DEFAULT_PC_STEP`.
+    /// SAFETY: The first `pc_base / DEFAULT_PC_STEP` entries will be unreachable. We do this to
+    /// avoid needing to subtract `pc_base` during runtime.
     pre_compute_insns: Vec<PreComputeInstruction<'a, F, Ctx>>,
```
shuklaayush, line 51: maybe only keep this when `#[cfg(not(feature = "tco"))]`?

crates/vm/derive/src/tco.rs
shuklaayush: oh actually this is the new pc

shuklaayush: still think there's duplicate work happening here that won't be optimized out

crates/vm/src/arch/state.rs
```diff
@@ -94,6 +97,7 @@ impl<F: Clone> VmState<F, GuestMemory> {
 /// The global state is generic in guest memory `MEM` and additional context `CTX`.
 /// The host state is execution context specific.
 // @dev: Do not confuse with `ExecutionState` struct.
+#[derive(WithSetters)]
```
shuklaayush, line 100: doesn't look like this is used

extensions/native/circuit/src/poseidon2/execution.rs
```diff
@@ -173,6 +173,33 @@ impl<F: PrimeField32, const SBOX_REGISTERS: usize> Executor<F>
...
+            } else {
+                Ok(execute_pos2_e1_tco_handler::<_, _, SBOX_REGISTERS, false>)
+            }
+        } else {
+            let verify_batch_data: &mut VerifyBatchPreCompute<F, SBOX_REGISTERS> =
+                data.borrow_mut();
+            self.pre_compute_verify_batch_impl(pc, inst, verify_batch_data)?;
+            Ok(execute_verify_batch_e1_tco_handler::<_, _, SBOX_REGISTERS>)
+        }
+    }
```
shuklaayush, line 202: maybe you can add a `dispatch!` macro here too to avoid duplication

extensions/native/circuit/src/loadstore/execution.rs
```diff
@@ -114,8 +136,33 @@ where
...
+
+        let local_opcode = self.pre_compute_impl(pc, inst, &mut pre_compute.data)?;
+
+        let fn_ptr = match local_opcode {
+            NativeLoadStoreOpcode::LOADW => execute_e2_loadw_tco_handler::<F, Ctx, NUM_CELLS>,
+            NativeLoadStoreOpcode::STOREW => execute_e2_storew_tco_handler::<F, Ctx, NUM_CELLS>,
+            NativeLoadStoreOpcode::HINT_STOREW => {
+                execute_e2_hint_storew_tco_handler::<F, Ctx, NUM_CELLS>
+            }
+        };
```
shuklaayush, line 159: `dispatch!`?

extensions/native/circuit/src/jal_rangecheck/execution.rs
```diff
@@ -109,6 +109,28 @@ where
...
+
+        if is_jal {
+            let jal_data: &mut JalPreCompute<F> = data.borrow_mut();
+            self.pre_compute_jal_impl(pc, inst, jal_data)?;
+            Ok(execute_jal_e1_tco_handler)
+        } else {
+            let range_check_data: &mut RangeCheckPreCompute = data.borrow_mut();
+            self.pre_compute_range_check_impl(pc, inst, range_check_data)?;
+            Ok(execute_range_check_e1_tco_handler)
+        }
```
shuklaayush, line 132: `dispatch!`?

shuklaayush (commented):
looks good. there's an argument that we can have a single handler function instead of separate `precompute` and `handler` functions but i think the current approach with using macros to avoid duplication is fine too

crates/vm/derive/src/tco.rs
jonathanpwang: oh you mean like because new pc doesn't get passed across function boundary

crates/vm/derive/src/tco.rs
jonathanpwang: yea will do with the register pinning

benchmarks/prove/Cargo.toml
jonathanpwang: I didn't want evm in default to compile stuff faster

crates/vm/src/arch/interpreter.rs
jonathanpwang: well it's bounded by the ELF size

crates/vm/derive/src/tco.rs
shuklaayush: yep

crates/vm/src/arch/interpreter.rs
jonathanpwang: the `execute_from_state` fn only uses `exit_code` internally, so here I just want the entire function to return error

crates/vm/src/arch/interpreter.rs
jonathanpwang: nevermind it was easy

extensions/native/circuit/src/loadstore/execution.rs
jonathanpwang: I feel like it's not worth it because the macro definition and code length are about the same

extensions/native/circuit/src/jal_rangecheck/execution.rs
jonathanpwang: this one is borderline, but again feels like the code length difference is negligible

PR #2020, opened 2025-08-20
feat(test): add test to check metered/metered cost cells are similar

shuklaayush:


jonathanpwang:
@shuklaayush was this something to merge or do you want to close it?

PR #2021, opened 2025-08-20
chore: migrate bigint CUDA code

stephenh-axiom-xyz:


Golovanov399 (approved):

extensions/bigint/circuit/build.rs
```diff
@@ -0,0 +1,28 @@
...
+        if !cuda_available() {
+            return; // Skip CUDA compilation
+        }
+
+        let builder = CudaBuilder::new()
+            .include_from_dep("DEP_CUDA_COMMON_INCLUDE")
+            .include("../../../crates/circuits/primitives/cuda/include")
+            .include("../../rv32im/circuit/cuda/include")
+            .include("../../rv32-adapters/cuda/include")
+            .include("../../../crates/vm/cuda/src")
```
gaxiom, line 16: Why do you need that if you don't include something from there?

PR #2029, opened 2025-08-21
chore: migrate SDK CUDA extensions

stephenh-axiom-xyz:
Currently blocked by remaining extensions

### DIFF
--- a/.github/workflows/extension-tests.yml
+++ b/.github/workflows/extension-tests.yml
@@ -76,7 +76,7 @@ jobs:
       - uses: taiki-e/install-action@nextest
 
       - name: Check CUDA status and set environment variables
-        if: contains(matrix.extension.family, 'g4dn')
+        if: contains(matrix.extension.image, 'gpu')
         run: |
           nvidia-smi
           nvidia-smi -L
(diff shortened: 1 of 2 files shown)


gaxiom (approved):
LGTM!

crates/sdk/src/lib.rs
```diff
@@ -153,6 +158,9 @@ where
 
 pub type Sdk = GenericSdk<BabyBearPoseidon2Engine, SdkVmCpuBuilder, NativeCpuBuilder>;
```
Golovanov399, line 159: Is it the case that we only use `Sdk` in `not(feature = "cuda")` and `GpuSdk` in `feature = "cuda"`? If so, maybe it makes sense to do the same as in integration tests?

Golovanov399 (approved):

crates/sdk/src/lib.rs
stephenh-axiom-xyz: Mmm `Sdk` is used in a lot more places than `air_test` is, I thought it'd be best to leave them separate for now in case it broke anything

.github/workflows/extension-tests.yml
```diff
@@ -27,19 +27,19 @@ jobs:
       matrix:
         extension:
           - { name: "rv32im", path: "rv32im", family: "c7g.16xlarge", image: "ubuntu24-full-arm64" }
-          - { name: "rv32im", path: "rv32im", family: "g6.2xlarge", image: "ubuntu24-gpu-x64" }
+          - { name: "rv32im", path: "rv32im", family: "g4dn.2xlarge", image: "ubuntu24-gpu-x64" }
```
gaxiom, line 30: Do we actually need 2xlarge in unit tests?

stephenh-axiom-xyz: Not sure, my next PR will be a bunch of CI stuff

PR #2030, opened 2025-08-21
chore: migrate algebra, ecc, pairing extensions

arayikhalatyan:
Relates to INT-4744

Migrate algebra, ecc, pairing extension


stephenh-axiom-xyz (approved):
LGTM, bunch of nits we can take care of later (if so, please update the cleanup ticket)

stephenh-axiom-xyz:
@arayikhalatyan I think you can push even with ecc integration test failures, I will push a fix in mine

PR #2032, opened 2025-08-21
chore: migrate riscv test vectors 

arayikhalatyan:
Relates to INT-4744

update workflow and test function to run riscv test vectors on gpu


jonathanpwang (approved):
extension tests have some separate issue, but riscv test vector workflows pass

PR #2033, opened 2025-08-21
chore: CUDA benchmarks and other CI

stephenh-axiom-xyz:
Resolves INT-4699

### DIFF
--- a/.github/workflows/benchmark-call.yml
+++ b/.github/workflows/benchmark-call.yml
@@ -7,11 +7,16 @@ on:
         type: string
         required: true
         description: The name of the benchmark to run
-      instance_type:
+      family:
         type: string
         required: false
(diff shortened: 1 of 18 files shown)


jonathanpwang (commented):
do NOT merge this yet. I do not like the current handling of the yml

.github/runs-on.yml
```diff
@@ -5,3 +5,39 @@ images:
     owner: "405894839083"
     name: "arm64-rust-riscv-dev-32gb"
     ami: "ami-040a96edd406c9478"
+
+runners:
+  test-64cpu-arm64:
```
jonathanpwang, line 10: none of this works since it's not on `main` branch, also I prefer for cpu we just use runson runner types

PR #2039, opened 2025-08-22
fix: test_moduli_setup

arayikhalatyan:


crates/circuits/mod-builder/cuda/include/bigint_ops.cuh
```diff
@@ -3,6 +3,7 @@
 #include <assert.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdio.h>
```
gaxiom, line 6: no need

crates/circuits/mod-builder/cuda/include/bigint_ops.cuh
```diff
@@ -320,9 +321,8 @@ struct BigUintGpu {
             result -= r2;
         }
 
-        while (result >= modulus) {
-            result -= modulus;
-        }
+        BigUintGpu temp_quotient;
+        divrem(temp_quotient, result, modulus);
```
gaxiom, line 325: Seems like we can create divrem that wil work without temp BigUintGpu.
Or at least put it in {} brackets, so temp_quotient will be dropped right after divrem call

gaxiom (approved):
some comments left, but conceptually good

crates/circuits/mod-builder/cuda/include/bigint_ops.cuh
```diff
@@ -199,6 +199,61 @@ struct BigUintGpu {
...
+            uint32_t v = limbs[limb] & mask;
+            if (v != 0) {
+                int leading = __clz(v);
+                int bit_index = 31 - leading;
+                msb_pos = limb * limb_bits + bit_index;
+                break;
+            }
+        }
+
+        if (msb_pos == -1)
```
Golovanov399, line 223: We can assert that it's not the case, can't we? Because it only happens when `*this` is zero, and therefore either `divisor` is also zero or `*this < divisor`, which we handled above

crates/circuits/mod-builder/cuda/include/bigint_ops.cuh
```diff
@@ -199,6 +199,61 @@ struct BigUintGpu {
         return result;
     }
 
+    __device__ BigUintGpu rem(const BigUintGpu &divisor) const {
```
gaxiom, line 202: Seems like you can use `rem` in `divrem` to avoid code duplication

arayikhalatyan: I don't think this is possible without having to do double iteration🤔

PR #2049, opened 2025-08-25
feat: add metered execution commands to sdk/cli

shuklaayush:
- add `execute_metered` and `execute_metered_cost` to sdk
- add `cargo openvm run --mode segment` and `cargo openvm run --mode meter` to call the above sdk functions

Resolves INT-4787

### DIFF
--- a/crates/cli/tests/integration.rs
+++ b/crates/cli/tests/integration.rs
@@ -20,7 +20,7 @@ fn install_cli() {
     });
 }
 
-fn build_fibonacci_once() -> Result<()> {
+fn build_fibonacci_once() -> Result<&'static str> {
     static BUILD_ONCE: OnceLock<Result<()>> = OnceLock::new();
     BUILD_ONCE
         .get_or_init(|| {
@@ -37,7 +37,7 @@ fn build_fibonacci_once() -> Result<()> {
             )
         })
         .as_ref()
-        .map(|_| ())
+        .map(|_| "tests/programs/fibonacci/target/openvm/release/openvm-cli-example-test.vmexe")
         .map_err(|e| eyre::eyre!("Failed to build fibonacci: {}", e))
 }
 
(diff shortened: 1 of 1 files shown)


crates/cli/src/commands/run.rs
```diff
@@ -1,16 +1,30 @@
...
 
+#[derive(Clone, Debug, ValueEnum)]
+pub enum ExecutionMode {
+    /// Pure execution (default)
+    Pure,
+    /// Execute with cost metering (execute_metered_cost)
+    Meter,
+    /// Execute with segmentation (execute_metered)
+    Segment,
+}
```
shuklaayush, line 26: open to suggestions for better names

crates/cli/tests/integration.rs
```diff
@@ -222,6 +222,127 @@ fn test_cli_init_build() -> Result<()> {
...
+            "--mode",
+            "segment",
+        ],
+    )?;
+
+    Ok(())
+}
+
+#[test]
+fn test_cli_run_meter() -> Result<()> {
```
jonathanpwang, line 311: should we consolidate the tests to build just once to make things faster?

crates/sdk/src/lib.rs
```diff
@@ -362,6 +363,68 @@ where
...
+    ) -> Result<(Vec<u8>, (u64, u64)), SdkError> {
+        let exe = self.convert_to_exe(app_exe)?;
+        let app_prover = self.app_prover(exe.clone())?;
+
+        let vm = &app_prover.instance().vm;
+        let ctx = vm.build_metered_cost_ctx();
+        let executor_idx_to_air_idx = vm.executor_idx_to_air_idx();
+
+        let interpreter = vm
+            .executor()
```
jonathanpwang, line 411: use `vm.metered_cost_instance` (or create such a function if it doesn't exist)

jonathanpwang (approved):
Some tidying comments and also please double-check the place keygen inside `run --mode meter` outputs matches the keygen CLI command.

crates/cli/tests/integration.rs
shuklaayush: added a `BUILD_ONCE` static

PR #2053, opened 2025-08-25
chore: add script to generate .clangd

jonathanpwang:


scripts/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+authors.workspace = true
+
+[dependencies]
+openvm-cuda-common = { workspace = true, optional = true }
+serde = { workspace = true }
+serde_yaml = "0.9"
+walkdir = "2.5"
+
+[features]
+cuda = ["dep:openvm-cuda-common"]
```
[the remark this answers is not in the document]
stephenh-axiom-xyz: It's not supposed to be anything without `cuda`, should be a CUDA-specific script

scripts/Cargo.toml
jonathanpwang, line 1: it felt slightly weird to me that the entire scripts folder is a rust crate, but maybe we change this only later if necessary

scripts/generate_clangd/run.sh
jonathanpwang, line 1: move this to just `scripts/generate_clangd.sh` ?

PR #2058, opened 2025-08-25
perf(tco): use more registers instead of references

jonathanpwang:


jonathanpwang:
what would be most impactful is if `pc, instret` could stay in registers, but this is too hard to do right now without changing all `execute_impl`

PR #2064, opened 2025-08-26
ci: fix merge queue concurrency

jonathanpwang:
`github.workflow` refers to the caller's workflow name, so merge queue controller messes up all the concurrency of callee jobs.

context: https://stackoverflow.com/questions/79511940/using-workflow-filename-in-concurrency-group-for-workflows-started-by-workflow-c

if reviewers have better ideas how to handle merge queue with required status checks, I'm all ears

### DIFF
--- a/.github/workflows/benchmarks-execute.yml
+++ b/.github/workflows/benchmarks-execute.yml
@@ -18,7 +18,7 @@ on:
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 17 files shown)


jonathanpwang:
> Can we use [workflow_ref](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#:~:text=in%20the%20repository.-,github.workflow_ref,-string)?

Claude told me this, but then it seemed like it is still the ref of the caller and not the callee?

PR #2085, opened 2025-08-27
docs: add execution record docs

Golovanov399:
This resolves INT-4887

### DIFF
--- a/docs/crates/records.md
+++ b/docs/crates/records.md
@@ -35,7 +35,7 @@ fn execute(
 
 ### Types of record arenas
 
-We have two implementations of record arenas: `MatrixRecordArena` for generating trace on CPU, and `DenseRecordArena` for GPU. The distinction is due to the fact that, for GPU tracegen, we have to first send the records to the GPU, and then generate the trace on its side. This motivates the arena to pack the record densely, hence the name. On the other hand, if we don't need to send the records anywhere, then they may already represent a partially filled trace -- maybe with gaps between their elements, which will be filled during tracegen.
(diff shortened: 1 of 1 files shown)


docs/crates/records.md
```diff
@@ -0,0 +1,282 @@
...
+    // ...
+}
+```
+
+> [!NOTE]
+> One can see from this example that a chip does not own its record arena, which is provided to it with the state instead.
+
+### Types of record arenas
+
+We have two implementations of record arenas: `MatrixRecordArena` for generating trace on CPU, and `DenseRecordArena` for GPU. The distinction is due to the fact that, for GPU tracegen, we have to first send the records to the GPU, then generate the trace on its side, and send the trace back to CPU. This motivates the arena to pack the record densely, hence the name. On the other hand, if we don't need to send the records anywhere, then they may already represent a partially filled trace -- maybe with gaps between their elements, which will be filled during tracegen.
```
shuklaayush, line 38: > and send the trace back to CPU

i don't think this is correct

jonathanpwang: Ayush is right: we **never** send buffers from device back to host except for debugging.

@shuklaayush can you fix this directly?

docs/crates/records.md
```diff
@@ -0,0 +1,282 @@
...
+    // ...
+}
+```
+
+> [!NOTE]
+> One can see from this example that a chip does not own its record arena, which is provided to it with the state instead.
+
+### Types of record arenas
+
+We have two implementations of record arenas: `MatrixRecordArena` for generating trace on CPU, and `DenseRecordArena` for GPU. The distinction is due to the fact that, for GPU tracegen, we have to first send the records to the GPU, and then generate the trace on its side. This motivates the arena to pack the record densely, hence the name. On the other hand, if we don't need to send the records anywhere, then they may already represent a partially filled trace -- maybe with gaps between their elements, which will be filled during tracegen.
```
jonathanpwang, line 38:
```suggestion
We have two implementations of record arenas: `MatrixRecordArena` for generating trace on CPU, and `DenseRecordArena` for GPU. The distinction is due to the fact that, for GPU tracegen, we have to first send the records to the GPU, and then generate the trace on device. This motivates the arena to pack the record densely, hence the name. On the other hand, if we don't need to send the records anywhere, then they may already represent a partially filled trace -- maybe with gaps between their elements, which will be filled during tracegen.
```

PR #2086, opened 2025-08-28
feat: use hybrid CPU tracegen for modular, complex, and ECC extensions

jonathanpwang:
We have found that for these extensions, it is more efficient to use the CPU due to the benefits/convenience of bigint arithmetic with variable number of limbs on CPU, while the use of GPU acceleration is overkill and doesn't lead to much (if any) performance gain.


Golovanov399 (approved):
Ok if works

PR #2091, opened 2025-08-28
feat: support v1.2/v1.3 guests

lispc:


crates/sdk/Cargo.toml
```diff
@@ -122,6 +122,16 @@ cuda = [
     "openvm-rv32im-circuit/cuda",
 ]
 
+# Legacy compatibility features for backwards compatibility with v1.2/v1.3
+legacy = ["legacy-evm-verifier", "legacy-mem-align"]
```
jonathanpwang, line 126: not sure on standard syntax, but legacy is ambiguous, should we say `legacy-v1-3` here and all other uses of `legacy`?

crates/vm/Cargo.toml
```diff
@@ -100,3 +100,4 @@ touchemall = [
     "openvm-cuda-backend/touchemall",
     "openvm-cuda-common/touchemall",
 ]
+legacy-mem-align = []
```
jonathanpwang, line 103: ditto on legacy-v1-3

extensions/rv32im/circuit/src/adapters/mod.rs
```diff
@@ -112,7 +112,16 @@ pub fn timed_read<const N: usize>(
     // SAFETY:
     // - address space `RV32_REGISTER_AS` and `RV32_MEMORY_AS` will always have cell type `u8` and
     //   minimum alignment of `RV32_REGISTER_NUM_LIMBS`
-    unsafe { memory.read::<u8, N, RV32_REGISTER_NUM_LIMBS>(address_space, ptr) }
+    #[cfg(feature = "legacy-mem-align")]
+    if address_space == RV32_MEMORY_AS {
+        unsafe { memory.read::<u8, N, 1>(address_space, ptr) }
+    } else {
+        unsafe { memory.read::<u8, N, 4>(address_space, ptr) }
```
jonathanpwang, line 119: 4 -> RV32_REGISTER_NUM_LIMBS

jonathanpwang:
@shuklaayush does anything in metered execution need to be considered with this change in the legacy feature, or will it be auto-handled?

crates/sdk/Cargo.toml
lispc: solved

extensions/rv32im/circuit/src/adapters/mod.rs
lispc: solved

PR #2092, opened 2025-08-28
ci(gpu): use less `test-threads`

jonathanpwang:


Golovanov399 (commented):
Is `2` test threads always less than it was before?

jonathanpwang:
> Is `2` test threads always less than it was before?

Yes because it's based on vCPU by default.

PR #2096, opened 2025-08-28
fix: explicit panic on oob memory access

shuklaayush:
- add bounds checking for memory access in memmap unless `unprotected` feature is enabled (disabled by default)

[comparison benchmark](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/17308694277)


jonathanpwang (commented):
please run a comparison benchmark after adding the feature to cli and sdk

jonathanpwang (approved):

PR #2102, opened 2025-08-29
chore: update input doc with an example

shayanh:
Noting that the example repo will be updated after merging https://github.com/openvm-org/openvm-example-fibonacci/pull/6 and https://github.com/openvm-org/openvm-example-fibonacci/pull/4.

Closes INT-4938


jonathanpwang (approved):
LGTM

PR #2105, opened 2025-08-29
chore: remove `openvm::entry!(main)` from examples

Golovanov399:
Also I added `use openvm` just in case because we say:

> If you write a program that only imports openvm in Cargo.toml but does not import it anywhere in your crate, the Rust linker may optimize away the dependency, which will cause a compile error. To fix this, you may need to explicitly import the openvm crate in your code.

### DIFF
--- a/examples/algebra/src/main.rs
+++ b/examples/algebra/src/main.rs
@@ -1,5 +1,6 @@
 extern crate alloc;
 
+use openvm;
 use openvm_algebra_guest::{moduli_macros::*, IntMod};
 
 // This macro will create two structs, `Mod1` and `Mod2`,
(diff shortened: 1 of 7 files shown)


examples/algebra/src/main.rs
```diff
@@ -1,5 +1,6 @@
 extern crate alloc;
 
+use openvm;
```
jonathanpwang, line 3: if you import `openvm_*` then you don't need `openvm` itself I believe

stephenh-axiom-xyz (commented):
I think I would lean towards leaving in `openvm::entry!(main);` than `use openvm;` since the former explicitly doesn't do anything if `std` is enabled, but I don't feel too strongly about either

examples/algebra/src/main.rs
shuklaayush: i think if you do `use openvm as _`, clippy wouldn't complain

jonathanpwang: for the sake of uniformity, I put `use openvm as _;` everywhere

PR #2118, opened 2025-09-02
ci: skip debug during CI air test

stephenh-axiom-xyz:
Resolves INT-4854

### DIFF
--- a/.github/workflows/guest-lib-tests.cuda.yml
+++ b/.github/workflows/guest-lib-tests.cuda.yml
@@ -114,7 +114,7 @@ jobs:
       - name: Check CUDA status and set environment variables
         run: |
           nvcc --version
-          echo "NEXTEST_ENV=CUDA_OPT_LEVEL=3" >> $GITHUB_ENV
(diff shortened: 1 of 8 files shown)


stephenh-axiom-xyz:
@jonathanpwang the CUDA failures are because of this:
```
CUDA kernel error at cuda/supra/include/ntt/parameters.cuh:253: named symbol not found
```
They passed before the redistribution of tests, not sure why they're so consistent now...?

.github/workflows/guest-lib-tests.cuda.yml
```diff
@@ -30,8 +30,8 @@ jobs:
       matrix:
         crates: # group crates on the same runner based on test time
           - { names: "sha2 keccak256 verify_stark" }
-          - { names: "ruint" }
-          - { names: "ff_derive k256 p256 pairing" }
+          - { names: "ruint k256 p256" }
+          - { names: "ff_derive pairing" }
```
jonathanpwang, line 34: does this balance the tests more for speed?

jonathanpwang (approved):
LGTM, although if we're just doing it across the board, then you can put is as 
```
env:
  OPENVM_SKIP_DEBUG: 1
```
in the yml

PR #2122, opened 2025-09-03
ci: script + workflow to run clang-tidy during CUDA lints

stephenh-axiom-xyz:
Resolves INT-4946


extensions/native/circuit/cuda/src/poseidon2.cu
```diff
@@ -77,6 +77,7 @@ template <size_t SBOX_REGISTERS> struct Poseidon2Wrapper {
         VariableRangeChecker range_checker,
         uint32_t timestamp_max_bits
     ) {
+        assert(row.is_valid());
```
jonathanpwang, line 80: why was this added

PR #2126, opened 2025-09-06
perf: only do exit check for instructions that can exit

shuklaayush:
- reduce exit checks in tco execute handlers

[benchmark comparison](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/17506995716)

Towards INT-4921


crates/vm/derive/src/lib.rs
```diff
@@ -835,7 +835,14 @@ fn parse_executor_type(
 pub fn create_tco_handler(_attr: TokenStream, item: TokenStream) -> TokenStream {
     #[cfg(feature = "tco")]
     {
-        tco::tco_impl(item)
+        // Parse the attribute to check for can_exit flag
+        let can_exit = if _attr.is_empty() {
+            false
+        } else {
+            let attr_str = _attr.to_string();
+            attr_str.contains("can_exit") && attr_str.contains("true")
```
jonathanpwang, line 843: perhaps not the best attribute parser

crates/vm/derive/src/lib.rs
```diff
@@ -835,7 +835,14 @@ fn parse_executor_type(
 pub fn create_tco_handler(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
jonathanpwang, line 835: update doc comment on `can_exit`.
Mention that `can_exit` actually means whether the handler can error in any way (since right now you cannot terminate without error except via `terminate`)

jonathanpwang (commented):
The macro change LGTM.
But how can we statically / non-manually ensure that the `e1/2_impl` never creates an error? The current way where I assume you manually checked each function for whether it can error or not doesn't seem very good/maintanable

crates/vm/derive/src/lib.rs
shuklaayush: fixed

PR #2137, opened 2025-09-15
fix: take program air height into account in metered execution

shuklaayush:
- add `ProgramAir` height as a constant trace height in metered execution

### DIFF
--- a/extensions/native/circuit/src/utils.rs
+++ b/extensions/native/circuit/src/utils.rs
@@ -97,9 +97,9 @@ pub mod test_utils {
         let input = input_stream.into();
 
         let engine = E::new(FriParameters::new_for_testing(1));
-        let (vm, _) = VirtualMachine::new_with_keygen(engine, builder, config)?;
-        let ctx = vm.build_metered_ctx();
         let exe = VmExe::new(program);
+        let (vm, _) = VirtualMachine::new_with_keygen(engine, builder, config)?;
+        let ctx = vm.build_metered_ctx(&exe);
         let (mut segments, _) = vm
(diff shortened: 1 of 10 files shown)


extensions/native/circuit/src/utils.rs
```diff
@@ -97,9 +97,9 @@ pub mod test_utils {
         let input = input_stream.into();
 
         let engine = E::new(FriParameters::new_for_testing(1));
-        let (vm, _) = VirtualMachine::new_with_keygen(engine, builder, config)?;
-        let ctx = vm.build_metered_ctx();
         let exe = VmExe::new(program);
+        let (vm, _) = VirtualMachine::new_with_keygen(engine, builder, config)?;
+        let ctx = vm.build_metered_ctx(&exe);
         let (mut segments, _) = vm
             .metered_interpreter(&exe)?
```
jonathanpwang, line 104: I had vaguely observed this before, but do we have a reason why the `InterpretedInstance` doesn't store/own `ctx`?

Otherwise we could just have `vm.metered_interpreter(&exe)` also create the context.

shuklaayush: i think we didn't make `InterpretedInstance` own `ctx` (or state) to be able to do multiple runs using the same interpreter instance. i can modify `vm.metered_interpreter` to return `ctx` instead of having a separate function

shuklaayush: on second thought, this would make it inconsistent with the other `vm.*_interpreter` functions so i feel it should be kept as how it is

PR #2138, opened 2025-09-15
fix: ensure segments are below thresholds

shuklaayush:
Replace the optimistic execution/segmentation with a checkpointing approach that checkpoints the last `trace_height`/`instret` value that is below the thresholds and use these values for the segments. This should make the segmentation more predictable for downstream usage since the segments should satisfy the thresholds (with the only caveat being if segmentation happens when there is no checkpoint to fall back to i.e. if we overshoot the threshold before the first segmentation check) 

This requires storing some extra state and results in a higher segment count compared to earlier for the same thresholds. Also makes execution slightly slower since we're doing some extra work now

[benchmark run](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/17750607194) with 0.7B max cells
[benchmark run](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/17771662984) with 1.2B max cells

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/segment_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/segment_ctx.rs
@@ -46,6 +46,10 @@ pub struct SegmentationCtx {
     pub instret_last_segment_check: u64,
     #[getset(set_with = "pub")]
     pub segment_check_insns: u64,
+    /// Checkpoint of trace heights at last known state where all thresholds satisfied
+    pub(crate) checkpoint_trace_heights: Vec<u32>,
+    /// Instruction count at the checkpoint
+    checkpoint_instret: u64,
 }
 
 impl SegmentationCtx {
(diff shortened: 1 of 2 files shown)


nyunyunyunyu (approved):
LGTM. This's more decent than subtracting DEFAULT_MAX_TRACE_HEIGHT with a random number.

crates/vm/src/arch/execution_mode/metered/segment_ctx.rs
```diff
@@ -204,16 +217,84 @@ impl SegmentationCtx {
...
+        segment_heights: &[u32],
+        is_trace_height_constant: &[bool],
+    ) {
+        for ((trace_height, &segment_height), &is_trace_height_constant) in trace_heights
+            .iter_mut()
+            .zip(segment_heights.iter())
+            .zip(is_trace_height_constant.iter())
+        {
+            if !is_trace_height_constant {
+                *trace_height -= segment_height;
```
jonathanpwang, line 288: just for safety, could we do checked_sub and unwrap

jonathanpwang (approved):
Actual implementation is ok, but you should leave comments explaining the rational about when to use padded versus unpadded and the considerations that went into it

PR #2140, opened 2025-09-16
feat: E2 supports suspension & VmExecState supports clone

nyunyunyunyu:
- E2 execution supports suspension every segment. [Reth-benchmark](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/17782856127) doesn't show performance difference.
- `VmExecState`/`VmState` supports clone. Benchmark shows cloning a VM state which only uses address space 2 takes 0.3~0.6ms.

closes INT-5066
closes INT-5067

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/ctx.rs
@@ -23,7 +23,7 @@ pub struct MeteredCtx<const PAGE_BITS: usize = DEFAULT_PAGE_BITS> {
     pub is_trace_height_constant: Vec<bool>,
     pub memory_ctx: MemoryCtx<PAGE_BITS>,
     pub segmentation_ctx: SegmentationCtx,
-    pub segment_suspend: bool,
+    pub suspend_on_segment: bool,
 }
 
 impl<const PAGE_BITS: usize> MeteredCtx<PAGE_BITS> {
@@ -76,7 +76,7 @@ impl<const PAGE_BITS: usize> MeteredCtx<PAGE_BITS> {
             is_trace_height_constant,
             memory_ctx,
             segmentation_ctx,
-            segment_suspend: false,
+            suspend_on_segment: false,
         };
         if !config.continuation_enabled {
             // force single segment
@@ -109,7 +109,7 @@ impl<const PAGE_BITS: usize> MeteredCtx<PAGE_BITS> {
     }
 
     pub fn with_segment_suspend(mut self, segment_suspend: bool) -> Self {
-        self.segment_suspend = segment_suspend;
+        self.suspend_on_segment = segment_suspend;
         self
     }
 
@@ -212,7 +212,7 @@ impl<const PAGE_BITS: usize> ExecutionCtxTrait for MeteredCtx<PAGE_BITS> {
         exec_state
             .ctx
             .check_and_segment(instret, segment_check_insns)
-            && exec_state.ctx.segment_suspend
+            && exec_state.ctx.suspend_on_segment
     }
 
     #[inline(always)]
--- a/crates/vm/src/arch/execution_mode/metered/ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/ctx.rs
@@ -208,7 +208,9 @@ impl<const PAGE_BITS: usize> ExecutionCtxTrait for MeteredCtx<PAGE_BITS> {
         segment_check_insns: u64,
         exec_state: &mut VmExecState<F, GuestMemory, Self>,
     ) -> bool {
-        // If `segment_suspend` is set, suspend every segment. Otherwise, execute until termination.
+        // If `segment_suspend` is set, suspend when a segment is determined (but the VM state might
+        // be after the segment boundary because the segment happens in the previous checkpoint).
+        // Otherwise, execute until termination.
         exec_state
             .ctx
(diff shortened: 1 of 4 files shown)


crates/vm/src/arch/execution_mode/metered/ctx.rs
```diff
@@ -23,6 +23,7 @@ pub struct MeteredCtx<const PAGE_BITS: usize = DEFAULT_PAGE_BITS> {
     pub is_trace_height_constant: Vec<bool>,
     pub memory_ctx: MemoryCtx<PAGE_BITS>,
     pub segmentation_ctx: SegmentationCtx,
+    pub segment_suspend: bool,
```
shuklaayush, line 26: nit
```suggestion
    pub suspend_on_segment: bool,
```

crates/vm/src/arch/interpreter.rs
```diff
@@ -405,22 +409,38 @@ where
...
-        let segmentation_check_insns = exec_state.ctx.segmentation_ctx.segment_check_insns;
+        loop {
+            exec_state = self.execute_metered_from_state_until_suspension(exec_state)?;
+            // The execution has terminated.
+            if exec_state.exit_code.is_ok() && exec_state.exit_code.as_ref().unwrap().is_some() {
+                break;
+            }
+            if exec_state.exit_code.is_err() {
+                return Err(exec_state.exit_code.unwrap_err());
+            }
```
shuklaayush, line 420: use `match`?

crates/vm/src/arch/interpreter.rs
```diff
@@ -405,22 +409,38 @@ where
...
+        }
+        check_termination(exec_state.exit_code)?;
+        let VmExecState { vm_state, ctx, .. } = exec_state;
+        Ok((ctx.into_segments(), vm_state))
+    }
+    pub fn execute_metered_from_state_until_suspension(
+        &self,
+        mut from_state: VmExecState<F, GuestMemory, MeteredCtx>,
```
shuklaayush, line 428: small nit - `from_state` doesn't seem the right term since it's mutated in place

crates/vm/src/arch/interpreter.rs
```diff
@@ -405,22 +409,38 @@ where
...
+            }
+            if exec_state.exit_code.is_err() {
+                return Err(exec_state.exit_code.unwrap_err());
+            }
+        }
+        check_termination(exec_state.exit_code)?;
+        let VmExecState { vm_state, ctx, .. } = exec_state;
+        Ok((ctx.into_segments(), vm_state))
+    }
+    pub fn execute_metered_from_state_until_suspension(
```
shuklaayush, line 426: can we use the same function and only do exit check based on whether `segment_suspend` is set?

crates/vm/src/arch/execution_mode/metered/ctx.rs
```diff
@@ -200,12 +208,11 @@ impl<const PAGE_BITS: usize> ExecutionCtxTrait for MeteredCtx<PAGE_BITS> {
...
     ) -> bool {
-        // E2 always runs until termination. Here we use the function as a hook called every
-        // instruction.
+        // If `segment_suspend` is set, suspend every segment. Otherwise, execute until termination.
         exec_state
             .ctx
-            .check_and_segment(instret, segment_check_insns);
-        false
+            .check_and_segment(instret, segment_check_insns)
+            && exec_state.ctx.segment_suspend
```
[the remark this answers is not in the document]
nyunyunyunyu: It cannot short circuits because `check_and_segment` is `mut`. I put `exec_state.ctx.segment_suspend` intendedly in order to show `check_and_segment` will execute anyway.

crates/vm/src/arch/interpreter.rs
nyunyunyunyu: When `segment_suspend` is not set, the loop breaks at the first time we check the condition. Why is it a problem?

crates/vm/src/arch/execution_mode/metered/ctx.rs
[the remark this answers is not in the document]
shuklaayush: oops yeah, makes sense

shuklaayush (approved):
lgtm. just add a note that the `VmExecState` returned by `execute_metered_from_state_until_suspension` is not at segment boundary but something slightly beyond that

crates/vm/src/arch/interpreter.rs
```diff
@@ -405,6 +409,24 @@ where
     ) -> Result<(Vec<Segment>, VmState<F, GuestMemory>), ExecutionError> {
         let mut exec_state = VmExecState::new(from_state, ctx);
 
+        loop {
+            exec_state = self.execute_metered_from_state_until_suspension(exec_state)?;
```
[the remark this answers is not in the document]
nyunyunyunyu: Updated

crates/vm/src/arch/execution_mode/metered/ctx.rs
```diff
@@ -106,6 +108,11 @@ impl<const PAGE_BITS: usize> MeteredCtx<PAGE_BITS> {
         self
     }
 
+    pub fn with_segment_suspend(mut self, segment_suspend: bool) -> Self {
```
jonathanpwang, line 111: nit: getset has a WithSetter macro

PR #2142, opened 2025-09-17
feat: check for segmentation after each keccak/sha instruction

stephenh-axiom-xyz:
Resolves INT-5063. Benchmark run shows no difference https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/17804008078

### DIFF
--- a/extensions/keccak256/circuit/src/execution.rs
+++ b/extensions/keccak256/circuit/src/execution.rs
@@ -200,4 +200,9 @@ unsafe fn execute_e2_impl<F: PrimeField32, CTX: MeteredExecutionCtxTrait>(
     exec_state
         .ctx
         .on_height_change(pre_compute.chip_idx as usize, height);
+
+    // We don't use this result, so for most CTX implementations this does nothing.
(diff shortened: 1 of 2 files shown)


extensions/keccak256/circuit/src/execution.rs
```diff
@@ -200,4 +200,9 @@ unsafe fn execute_e2_impl<F: PrimeField32, CTX: MeteredExecutionCtxTrait>(
     exec_state
         .ctx
         .on_height_change(pre_compute.chip_idx as usize, height);
+
+    // We don't use this result, so for most CTX implementations this does nothing.
+    // However, for MeteredCtx this forces a segment check since its implementation
+    // of should_suspend performs one as a side effect.
+    CTX::should_suspend(*instret, *pc, 0, exec_state);
```
stephenh-axiom-xyz, line 207: Note we pass 0 to `segment_check_insns` here, since `MeteredCtx` will skip the check if `instret < segment_check_insns + last_segment_check`. This should always force a segment check assuming `instret` increases monotonically.

PR #2150, opened 2025-09-26
fix: Over-align public constants in moduli_init

Avaneesh-axiom:
The `two_modular_limbs_list` constant declared by `moduli_init!` and used later in the complex setup macro was not guaranteed to be aligned to a large enough alignment. This caused an issue in some use-case.

This change aligns the public constant `two_modular_limbs_list` to the max block size of a modulus among all moduli declared to ensure no memory alignment issues arise.

Fixes INT-5135


jonathanpwang (approved):
LGTM, but hold on merging.

`limb_list_borders` doesn't need alignment because (1) it's `usize` aligned, (2) we only access it's fields statically so it likely is compiled away already.

PR #2156, opened 2025-10-07
Remove send + sync bound on executor

Schaeff:
It is not required after a small refactor.

### DIFF
--- a/crates/vm/src/arch/interpreter_preflight.rs
+++ b/crates/vm/src/arch/interpreter_preflight.rs
@@ -102,16 +102,13 @@ impl<F: Field, E> PreflightInterpretedInstance<F, E> {
     }
 
     pub fn filtered_execution_frequencies(&self) -> Vec<u32>
-    where
-        E: Send + Sync,
     {
         let base_idx = get_pc_index(self.pc_base);
         self.pc_handler
             .par_iter()
-            .enumerate()
+            .zip_eq(&self.execution_frequencies)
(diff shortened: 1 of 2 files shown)


jonathanpwang (approved):
LGTM nice!

jonathanpwang:
@Schaeff there is some clippy/fmt issue

crates/vm/src/arch/interpreter_preflight.rs
```diff
@@ -101,17 +101,13 @@ impl<F: Field, E> PreflightInterpretedInstance<F, E> {
...
-    pub fn filtered_execution_frequencies(&self) -> Vec<u32>
-    where
-        E: Send + Sync,
-    {
+    pub fn filtered_execution_frequencies(&self) -> Vec<u32> {
         let base_idx = get_pc_index(self.pc_base);
         self.pc_handler
             .par_iter()
-            .enumerate()
+            .zip_eq(&self.execution_frequencies)
```
jonathanpwang: try `self.execution_frequencies.par_iter()`

jonathanpwang:
The other two tests are failing for unrelated reasons

crates/vm/src/arch/interpreter_preflight.rs
jonathanpwang: ah I think what you need is the above plus `p3_maybe_rayon::prelude::*;` but your fix is equivalent

PR #2179, opened 2025-10-24
fix: Expose CUDA primitives

qwang98:
When building autoprecompile chips in GPU in `powdr-labs/powdr`, we need access to `histogram.cuh` in `openvm/circuit-primitives` to modify counts of periphery chips. This requires exporting include dir in `openvm/circuit-primitives` for downstream crates  (`powdr-openvm`). We believe this (and potentially other crates) to be general client extension usage.

The way we export `openvm/circuit-primitives` to `powdr-labs/powdr-openvm` is EXACTLY the same as how `stark-backend/cuda-common` is exported to `openvm`:  
- https://github.com/openvm-org/stark-backend/blob/main/crates/cuda-common/build.rs#L18-L19
- https://github.com/openvm-org/stark-backend/blob/main/crates/cuda-common/Cargo.toml#L7

### DIFF
--- a/crates/circuits/primitives/build.rs
+++ b/crates/circuits/primitives/build.rs
@@ -1,7 +1,8 @@
 #[cfg(feature = "cuda")]
-use openvm_cuda_builder::{cuda_available, CudaBuilder};
-#[cfg(feature = "cuda")]
 use std::{env, path::PathBuf};
+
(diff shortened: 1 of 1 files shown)


jonathanpwang (approved):
LGTM, thanks for figuring this out. We agree it's a re-usable lib.

PR #2182, opened 2025-10-28
feat: SHA-2 incremental hasher design

Avaneesh-axiom:
## Overview of new design
Implemented a new design for constraining the SHA-2 family of hash functions (specifically SHA-256, SHA-512, SHA-384). The new design adds incremental hasher functionality, which means we can compute the hash of a stream of bytes. More specifically, the new `Sha256`, `Sha512`, and `Sha384` structs provided in the SHA-2 guest library provide the `update(&[u8])` and `finalize() -> [u8; HASH_SIZE]` methods. We can instantiate a hasher object, `let hasher = Sha256::new()` and then call `hasher.update(data)` as many times as we want on it. The `data` parameter can be a slice of any size. When we would like to retrieve the hash, we can call `hasher.finalize()`.

### Main Idea
The `Sha256` struct in the SHA-2 guest library maintains an array of bytes that serves as the internal state of the SHA-2 hashing algorithm. This array is updated using a new opcode: `SHA256_UPDATE dst src input` which takes in one block of input and pointers to the src/dst hasher states (the guest library sets `src == dst` for updating the state in-place). The `Sha256` struct will buffer up to one block of input, and it will call `SHA2_UPDATE` when necessary to absorb the input into the state. 

The `Sha512` and `Sha384` structs are implemented similarly.

### Interoperability
The `Sha256`, `Sha512`, `Sha384` structs implement the `sha2::Digest` trait, allowing them to be used as a drop-in replacement for the popular `sha2` crate's `sha2::{Sha256, Sha512, Sha384}` hasher objects.

### Documentation
The OpenVM book, specs, and the crate docs have been updated.
Additionally, a brief justification of soundness for the main constraints has been added.

### Tests
All the SHA-2 guest library integration tests and the SHA-2 circuit unit tests pass on CI.
Both CPU and GPU trace generation is tested among these tests.

closes INT-4972 INT-5023 INT-5024 INT-5025 INT-5026


crates/circuits/sha2-air/src/air.rs
jonathanpwang: It seems the move from sha256-air to sha2-air didn't pick up the diff; could you provide a short doc summarizing the diff? links to diffview would help as well

PR #2205, opened 2025-11-07
fix: aot memory related fixes

GunaDD:
Added stable ABI to `MemoryCtx` related structs

Towards INT-5337

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
@@ -1,13 +1,10 @@
-use abi_stable::{
-    std_types::{ROption, RVec},
-    StableAbi,
-};
+use abi_stable::std_types::{ROption, RVec};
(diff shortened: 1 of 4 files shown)


crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -1,16 +1,22 @@
...
 use openvm_instructions::riscv::{RV32_NUM_REGISTERS, RV32_REGISTER_AS, RV32_REGISTER_NUM_LIMBS};
 
 use crate::{arch::SystemConfig, system::memory::dimensions::MemoryDimensions};
 
-#[derive(Clone, Debug)]
+#[repr(C)]
+#[derive(Clone, Debug, StableAbi)]
 pub struct BitSet {
-    words: Box<[u64]>,
+    words: RVec<u64>,
```
nyunyunyunyu: I think you don't need to change this. This only has 1 field so the memory layout should be fixed

PR #2207, opened 2025-11-08
perf: avoid re-computing memory merkle tree for user public values proof

jonathanpwang:
closes INT-4314


Golovanov399 (approved):
Seems correct

nyunyunyunyu (approved):

PR #2211, opened 2025-11-10
perf: general register mapping design for AOT and implementation for base ALU

GunaDD:
Closes INT-5419

### DIFF
--- a/crates/vm/src/arch/aot/pure.rs
+++ b/crates/vm/src/arch/aot/pure.rs
@@ -5,8 +5,8 @@ use openvm_stark_backend::p3_field::PrimeField32;
 
 use super::{
     common::{
-        REG_D, REG_EXEC_STATE_PTR, REG_FIRST_ARG, REG_FOURTH_ARG, REG_INSNS_PTR, REG_INSTRET_END,
-        REG_PC, REG_RETURN_VAL, REG_SECOND_ARG, REG_THIRD_ARG, sync_gpr_to_xmm, sync_xmm_to_gpr,
+        sync_gpr_to_xmm, sync_xmm_to_gpr, REG_D, REG_EXEC_STATE_PTR, REG_FIRST_ARG, REG_FOURTH_ARG,
+        REG_INSNS_PTR, REG_INSTRET_END, REG_PC, REG_RETURN_VAL, REG_SECOND_ARG, REG_THIRD_ARG,
     },
     AotInstance, AsmRunFn,
 };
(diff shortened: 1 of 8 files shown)


extensions/rv32im/circuit/src/common/mod.rs
```diff
@@ -67,4 +46,49 @@ mod aot {
         };
         format!("   pextrq {gpr}, {xmm_map_reg}, 1\n")
     }
+
+    pub(crate) fn xmm_to_gpr(
```
nyunyunyunyu: Add comments to explain this function and the return value

crates/vm/src/arch/aot/common.rs
```diff
@@ -0,0 +1,85 @@
...
+    None,         // x25
+    None,         // x26
+    None,         // x27
+    None,         // x28
+    None,         // x29
+    None,         // x30
+    None,         // x31
+];
+
+pub fn SYNC_XMM_TO_GPR() -> String {
```
nyunyunyunyu: lower case

crates/vm/src/arch/aot/pure.rs
```diff
@@ -160,28 +174,37 @@ where
...
+
+                asm_str += &Self::pop_internal_registers();
+                asm_str += &Self::pop_address_space_start();
+                asm_str += &Self::rv32_regs_to_xmm();
+                asm_str += &format!("   je asm_run_end_{pc}\n");
+
+                asm_str += &format!("   lea {REG_D}, [rip + map_pc_base]\n");
+                asm_str += &format!("   movsxd {REG_PC}, [{REG_D} + {REG_PC}]\n");
+                asm_str += &format!("   add {REG_PC}, {REG_D}\n");
+                asm_str += &format!("   jmp {REG_PC}\n");
```
Maillew: this is probably fine for now since we dont have custom extern_handlers for specific executors yet, but the original function call was intended for if an executor is known to only use a subset of registers, so you dont have to push/pop all registers

extensions/rv32im/circuit/src/load_sign_extend/aot.rs
```diff
@@ -18,14 +18,7 @@ where
     F: PrimeField32,
 {
     fn is_aot_supported(&self, inst: &Instruction<F>) -> bool {
-        let local_opcode = Rv32LoadStoreOpcode::from_usize(
-            inst.opcode
-                .local_opcode_idx(Rv32LoadStoreOpcode::CLASS_OFFSET),
-        );
-        match local_opcode {
-            Rv32LoadStoreOpcode::LOADB | Rv32LoadStoreOpcode::LOADH => true,
```
Maillew: maybe when disabling, leave the previous version as a comment? so easier to enable for later

GunaDD: okay

PR #2215, opened 2025-11-11
chore: update Rust version to 1.87.0 in Cargo.toml and rust-toolchain.toml

adust09:


jonathanpwang:
We may do this in a future release, but for now want to keep the slightly older version for backwards compatibility.

PR #2248, opened 2025-11-14
perf: try different should_suspend asm

yi-sun:


yi-sun:
No improvement on Reth benchmark (259ms E1)

PR #2274, opened 2025-11-21
perf: use blstrs for algebra/ecc bls operations

shuklaayush:
- this is a cleaner way to do it instead of patching halo2curves

[benchmark comparison](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/19581932733) - no change


Cargo.toml
jonathanpwang: if this doesn't need the halo2curves branch anymore, could you switch it back?

PR #2275, opened 2025-11-21
chore: add tracing spans for aot exec methods

shuklaayush:
- add back tracing spans for pure and metered execution mirroring interpreter implementation


jonathanpwang (approved):
test it works with openvm-prof
I think span name needs to remain the same, but perhaps you could add a label

PR #2286, opened 2025-11-30
fix: `lazy_update_boundary_heights` needs to handle uninitialized memory

jonathanpwang:
towards INT-5617

The metering update did not account for the initial 8->4 split from `handle_uninitialized_memory`


crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -271,11 +274,13 @@ impl<const PAGE_BITS: usize> MemoryCtx<PAGE_BITS> {
             // SAFETY: address_space is from 0 to len(), guaranteed to be in bounds
             let x = unsafe { *self.addr_space_access_count.get_unchecked(address_space) };
             if x > 0 {
-                // After finalize, we'll need to read it in chunk-sized units for the merkle chip
+                // After finalize, we'll need to read it in chunk-sized units for the merkle chip.
```
nyunyunyunyu: A bit concern if this will overestimate too much. Maybe we can skip AS 1 here?

nyunyunyunyu (approved):
LGTM since it will fix the prod issue

shuklaayush (approved):

PR #2314, opened 2025-12-09
feat: add parallel proof time for fixed number of devices

shuklaayush:
Resolves INT-5611

### DIFF
--- a/crates/prof/src/aggregate.rs
+++ b/crates/prof/src/aggregate.rs
@@ -523,7 +523,7 @@ impl AggregateMetrics {
     fn write_summary_markdown(&self, writer: &mut impl Write, num_parallel: usize) -> Result<()> {
         writeln!(
             writer,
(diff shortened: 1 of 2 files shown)


crates/prof/src/aggregate.rs
```diff
@@ -172,8 +179,63 @@ impl GroupedMetrics {
...
+        for (group_name, metrics) in &self.by_group {
+            if group_name.contains("keygen") {
+                continue;
+            }
+
+            let mut group_time = 0.0;
+
+            // Add serial execution time for app_proof
+            if group_name == "app_proof" {
+                group_time += sum_metric_ms(metrics, EXECUTE_METERED_TIME_LABEL);
```
jonathanpwang: is this just the time for this one segment?

what we should add here is sum of metered execution time up to (and excluding) this segment -- and no need to count both metered and E1 since we only need metered

shuklaayush: we don't have the breakdown of metered execution per segment. but i did add avg e1 instead of total e1 like we do in total parallel proof time

PR #2325, opened 2025-12-18
perf: no serial `mark_parents`

Golovanov399:
This resolves INT-5772.

The kernel used to be among the heaviest ones because it was just a linear run in one thread, now it's replaced with two (more if we include stuff from `cub/cub.cuh`) other kernels. A slight drawback is that it now requires a temporary buffer of size `5 * num_leaves` `u32`s instead of `4 * num_leaves`, which should not be a big deal compared to actual used memory for the leaves.

### DIFF
--- a/crates/vm/src/system/cuda/merkle_tree/cuda.rs
+++ b/crates/vm/src/system/cuda/merkle_tree/cuda.rs
@@ -147,7 +147,7 @@ pub mod merkle_tree {
         let num_subtrees = subtree_ptrs.len();
         let tmp_buffer = DeviceBuffer::<u32>::with_capacity(5 * num_leaves);
         let mut need_tmp_storage_bytes = 0;
-        get_prefix_scan_temp_bytes(&tmp_buffer, 2 * num_leaves, &mut need_tmp_storage_bytes)?;
+        get_prefix_scan_temp_bytes(&tmp_buffer, num_leaves, &mut need_tmp_storage_bytes)?;
         let tmp_storage = DeviceBuffer::<u8>::with_capacity(need_tmp_storage_bytes);
         let actual_heights = actual_heights.to_device().unwrap();
         CudaError::from_result(_update_merkle_tree(


crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -182,40 +184,54 @@ __global__ void prepare_for_updating(
...
-            }
-            child_ptrs[2 * i + (layer[ptr].label >> (h - 1)) % 2] = ptr;
-            ++j;
+
+    uint32_t const ptr = current_layer_ptrs[gid];
+    uint32_t const my_place = parent_ids[gid] * 2 + (layer[ptr].label >> (h - 1)) % 2;
+    uint32_t const siblings_place = my_place ^ 1;
+
+    child_ptrs[my_place] = ptr;
+    // Do we need to mark the sibling as absent?
```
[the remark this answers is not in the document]
Golovanov399: Ah this was not a question, I'll rephrase

docs/vocs/qwe.sh
jonathanpwang: is this intentional?

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -182,40 +184,54 @@ __global__ void prepare_for_updating(
...
-            auto const ptr = current_layer_ptrs[j];
-            if (layer[ptr].address_space_idx != address_space_idx ||
-                (layer[ptr].label >> h) != label) {
-                break;
-            }
-            child_ptrs[2 * i + (layer[ptr].label >> (h - 1)) % 2] = ptr;
-            ++j;
+
+    uint32_t const ptr = current_layer_ptrs[gid];
+    uint32_t const my_place = parent_ids[gid] * 2 + (layer[ptr].label >> (h - 1)) % 2;
```
jonathanpwang: % comes before + right

PR #2332, opened 2025-12-24
fix: keep track of memory accesses since last checkpoint

shuklaayush:
This pr primarily adds a `page_indices_since_checkpoint` buffer to track page accesses between checkpoints. This buffer is used to initialize the memory trace heights of the next segment

Resolves INT-5778

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
@@ -3,8 +3,8 @@ use openvm_instructions::riscv::{RV32_NUM_REGISTERS, RV32_REGISTER_AS, RV32_REGI
 
 use crate::{arch::SystemConfig, system::memory::dimensions::MemoryDimensions};
 
-/// Upper bound on memory operations per instruction. Used for buffer allocation.
-pub const MAX_MEMORY_OPS_PER_INSN: usize = 1 << 16;
+/// Upper bound on number of memory pages accessed per instruction. Used for buffer allocation.
+pub const MAX_MEM_PAGE_OPS_PER_INSN: usize = 1 << 16;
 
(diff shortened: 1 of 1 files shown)


crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -261,15 +335,16 @@ impl<const PAGE_BITS: usize> MemoryCtx<PAGE_BITS> {
             let nodes = (((1 << PAGE_BITS) - 1) + (merkle_height - PAGE_BITS)) as u32;
             // SAFETY: merkle_tree_idx is guaranteed to be in bounds
             unsafe {
-                *trace_heights.get_unchecked_mut(poseidon2_idx) += nodes * 2;
-                *trace_heights.get_unchecked_mut(merkle_tree_idx) += nodes * 2;
+                *trace_heights.get_unchecked_mut(poseidon2_idx) +=
+                    nodes * page_access_count as u32 * 2;
+                *trace_heights.get_unchecked_mut(merkle_tree_idx) +=
+                    nodes * page_access_count as u32 * 2;
```
shuklaayush: i think this was a bug earlier where the update should've been `nodes * page_access_count * 2` instead of `nodes * 2`

crates/vm/src/arch/execution_mode/metered/segment_ctx.rs
```diff
@@ -296,14 +288,30 @@ impl SegmentationCtx {
             (instret, trace_heights.to_vec())
         };
 
-        // Reset current trace heights and checkpoint
-        self.reset_trace_heights(trace_heights, &segment_heights, is_trace_height_constant);
-        self.checkpoint_instret = 0;
```
shuklaayush: moved to `initialize_segment`

PR #2333, opened 2025-12-24
fix(prof): remove debug println and add missing diff handlers

berZKerk:
Remove debug println statements that were printing unused values in get_single_summary. 

Add missing diff calculation handlers for halo2_outer and halo2_wrapper metrics in set_diff to maintain consistency with other
metric groups.


jonathanpwang:
Can you rebase once to fix the cargo audit issue?

PR #2336, opened 2025-12-30
fix(aot): execute_metered should error if exit code nonzero

jonathanpwang:


crates/vm/src/arch/aot/metered_execute.rs
```diff
@@ -341,16 +341,21 @@ where
...
-        // handle execution error
-        match vm_exec_state.exit_code {
-            Ok(_) => Ok((
-                vm_exec_state.ctx.segmentation_ctx.segments,
-                vm_exec_state.vm_state,
-            )),
-            Err(e) => Err(e),
+        let mut exec_state = VmExecState::new(from_state, ctx);
+
+        loop {
```
jonathanpwang: the loop does nothing because `execute_metered_until_suspend` never suspends in AOT, but it seemed better to just implement the general version defensively

PR #2350, opened 2026-01-12
perf: remove preprocessed trace from VmConnectorAir

876pol:
Resolves INT-5850.


crates/vm/src/system/connector/mod.rs
```diff
@@ -79,11 +79,11 @@ impl<F: Field> BaseAirWithPublicValues<F> for VmConnectorAir {
 impl<F: Field> PartitionedBaseAir<F> for VmConnectorAir {}
 impl<F: Field> BaseAir<F> for VmConnectorAir {
     fn width(&self) -> usize {
-        5
+        6
```
stephenh-axiom-xyz: Not sure why we didn't do this before, but change this (and the other width hardcode below) to `ConnectorCols::<F>::width()` please

PR #2353, opened 2026-01-14
feat(new-keccak): split keccakf into two AIRs

jonathanpwang:
Note that I removed the `cells_used` metric since with the introduction of periphery chips (both here and the new SHA-2) it is very hard to make it accurate. There are better ways to get that data, and we should add tooling for it in a separate feature.


jonathanpwang:
Lint fails due to `elliptic-curve` having old `generic-array` version. To avoid rebase conflicts, I will not resolve for now.

PR #2503, opened 2026-03-06
fix: remaining internal findings for InteractionsFoldingAir

Golovanov399:
This resolves the rest of INT-6591:
- INT-6577: received `AirId` from ProofShapeAir
- INT-6579: received `beta_tidx` from GkrInputAir by analogy with `lambda_tidx` in ConstraintsFoldingAir

Since this reuses some logic from https://github.com/openvm-org/openvm/pull/2491, this is on top of that commit.

(resolves INT-6591 INT-6577 INT-6579)


crates/recursion/src/batch_constraint/expr_eval/interactions_folding/air.rs
```diff
@@ -320,6 +324,24 @@ where
             },
             next_is_first_in_air_or_invalid * local.is_valid,
         );
+        self.air_shape_bus.lookup_key(
```
stephenh-axiom-xyz, line 369: I think you also have to constrain `air_idx` to be the same throughout the section here too

stephenh-axiom-xyz (commented):
Need to constrain `air_idx` consistency again

stephenh-axiom-xyz (approved):
LGTM after rebase

PR #2505, opened 2026-03-06
fix: change index of initial SumcheckClaimsBus send to -1

stephenh-axiom-xyz:
Resolves INT-6391. `SumcheckClaimsBus` message flow now goes like this:

- `FractionsFolderAir` sends initial round -1 value to `UnivariateSumcheckAir`
- `UnivariateSumcheckAir` sends round 0 value to `MultilinearSumcheckAir`
- `MultilinearSumcheckAir` sends round `n_max` value to `ExpressionClaimAir` (and sends/consumes rounds in between)

### DIFF
--- a/crates/recursion/src/batch_constraint/sumcheck/univariate/air.rs
+++ b/crates/recursion/src/batch_constraint/sumcheck/univariate/air.rs
@@ -260,7 +260,7 @@ where
             builder,
             local.proof_idx,
             SumcheckClaimMessage {
-                round: AB::Expr::ZERO,
+                round: AB::Expr::NEG_ONE,
                 value: local.sum_at_roots.map(Into::into),
             },
             is_last.clone(),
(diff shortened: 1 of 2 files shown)


crates/recursion/src/batch_constraint/sumcheck/univariate/air.rs
zlangley, line 272: Wouldn't it be more intuitive to change this to `AB::Expr::ONE` and then have multivariate air work on rounds `1..n`? There is no `-1` round.

stephenh-axiom-xyz: I don't think this is really a round, it seems like the AIRs use this to "receive an initial value and send the result of the round". The "-1" round is the initial value to round 0, which iirc can't be skipped

zlangley: I think it's more natural to index the claims by the rounds they are input to rather than they are output to, but won't block.

zlangley (approved):

PR #2506, opened 2026-03-06
fix: MultiLinearSumcheckAir findings

stephenh-axiom-xyz:
Resolves INT-6596, INT-6388, INT-6389.

### INT-6388: lacks same-proof cross-round `tidx` continuity

Now constrains `tidx` increment over all rows in a Proof, not just a round

```rust
let is_proof_transition =
    LoopSubAir::local_is_transition(next.is_valid, next.is_proof_start);
...
builder
    .when(is_proof_transition)
    .assert_eq(next.tidx, local.tidx + AB::Expr::from_usize(D_EF));
```

### INT-6389: `is_dummy` can suppress required interactions

`is_dummy` now forces a proof to be a single row, effectively forcing no multilinear sumcheck rounds to occur.

```rust
builder.when(local.is_dummy).assert_one(local.is_valid);
builder
    .when(local.is_dummy)
    .assert_one(local.is_proof_start);
builder
    .when(local.is_dummy)
    .assert_one(is_proof_end.clone());
```

### DIFF
--- a/crates/recursion/src/batch_constraint/sumcheck/multilinear/air.rs
+++ b/crates/recursion/src/batch_constraint/sumcheck/multilinear/air.rs
@@ -125,6 +125,8 @@ where
         let is_last_eval =
             LoopSubAir::local_is_last(local.is_valid, next.is_valid, next.is_first_eval);
 
+        let is_proof_transition =
+            LoopSubAir::local_is_transition(next.is_valid, next.is_proof_start);
         let is_proof_end =
             LoopSubAir::local_is_last(local.is_valid, next.is_valid, next.is_proof_start);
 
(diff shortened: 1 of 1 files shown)


crates/recursion/src/batch_constraint/sumcheck/multilinear/air.rs
```diff
@@ -125,9 +119,21 @@ where
         let is_last_eval =
             LoopSubAir::local_is_last(local.is_valid, next.is_valid, next.is_first_eval);
 
+        let is_proof_transition =
+            LoopSubAir::local_is_transition(next.is_valid, next.is_proof_start);
```
zlangley: This kind of sounds like the transition point between two proofs to me. How about `is_same_proof`?

zlangley (approved):
LGTM, one naming suggestion

PR #2508, opened 2026-03-06
fix: SymbolicExpressionAir findings

zlangley:
- force exact symbolic-row presence for variable-free expression trees by constraining proof-shape AirId lookup counts to VK-derived values
- reject is_present = 1 on dummy cached rows by requiring valid cached flags
- move NodeKind into symbolic_expression/air.rs so the canonical discriminants live with the audited AIR logic
- remove the unused claim_bus and eq_neg_internal_bus wiring

### DIFF
--- a/crates/recursion/src/batch_constraint/expr_eval/symbolic_expression/air.rs
+++ b/crates/recursion/src/batch_constraint/expr_eval/symbolic_expression/air.rs
@@ -12,20 +12,19 @@ use p3_field::{extension::BinomiallyExtendable, Field, PrimeCharacteristicRing};
 use p3_matrix::Matrix;
 use stark_recursion_circuit_derive::AlignedBorrow;
 use strum::{EnumCount, IntoEnumIterator};
+use strum_macros::EnumIter;
 
-use super::trace::NodeKind;
 use crate::{
     batch_constraint::{
         bus::{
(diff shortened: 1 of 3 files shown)


crates/recursion/src/proof_shape/mod.rs
```diff
@@ -43,6 +43,7 @@ mod cuda_abi;
 #[derive(Clone)]
 pub struct AirMetadata {
     is_required: bool,
+    num_air_id_lookups: usize,
```
stephenh-axiom-xyz: Could you remove this please

crates/recursion/src/system/frame.rs
```diff
@@ -20,6 +22,7 @@ use openvm_stark_sdk::config::baby_bear_poseidon2::{BabyBearPoseidon2Config, Dig
 pub struct StarkVkeyFrame {
     pub preprocessed_data: Option<VerifierSinglePreprocessedData<Digest>>,
     pub params: StarkVerifyingParams,
+    pub num_air_id_lookups: usize,
```
stephenh-axiom-xyz: Also this

stephenh-axiom-xyz (approved):
LGTM, although please remove proof shape things to avoid a conflict later

PR #2513, opened 2026-03-06
fix: TranscriptAir findings

stephenh-axiom-xyz:
Resolves INT-6586, INT-6326, INT-6327, INT-6328, INT-6457, INT-6647.

### INT-6329: `local_next_same_proof` is not tied to `proof_idx`

Inserted `NestedForLoopSubAir` to constrain `proof_idx` semantics. All above tickets are resolved as a result.

### INT-6457: `permuted` flag not constrained to transcript semantics

Removed `permuted` column, replacing it with an algebraically derived value.

```rust
let permuted =
    local_next_same_proof * not::<AB::Expr>(and(local.is_sample, not(next.is_sample)));
```

### INT-6647: when `local.is_sample == next.is_sample` count must be `CHUNK`

Added check for the above

### DIFF
--- a/crates/recursion/cuda/src/transcript/transcript.cu
+++ b/crates/recursion/cuda/src/transcript/transcript.cu
@@ -17,7 +17,6 @@ template <typename T> struct TranscriptCols {
     T is_sample;
     T mask[CHUNK];
 
-    T permuted;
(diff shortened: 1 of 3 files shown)


zlangley (approved):
LGTM

PR #2514, opened 2026-03-07
chore: update `stark-backend` to use WHIR with list decoding

jonathanpwang:


crates/recursion/src/transcript/mod.rs
```diff
@@ -195,7 +195,8 @@ impl TranscriptModule {
                     idx += 1;
                     if idx == CHUNK {
                         // If it's sample -> observe, we don't need to permute. otherwise permute
-                        permuted = !is_sample || preflight.transcript.samples()[tidx];
+                        permuted = tidx < preflight.transcript.len()
```
jonathanpwang, line 198: @stephenh-axiom-xyz @luffykai seems like cpu tracegen had some error?

PR #2518, opened 2026-03-09
fix(v2-recursion): add `is_dummy` and `is_last_layer` consistency constraints

shuklaayush:
Resolves INT-6655, INT-6656, INT-6657


stephenh-axiom-xyz (approved):
LGTM

PR #2521, opened 2026-03-09
fix: add back num_air_id_lookups

stephenh-axiom-xyz:
Resolves INT-6682.

## Overview

This reverts part of `6108e90c8` ("chore: remove vk-derivable columns in proof shape"), which incorrectly removed the per-AIR lookup count from the `ProofShapeAir` trace. That commit tried to compute the lookup multiplicity purely from VK constants inside the AIR constraints (via `num_dag_nodes` on `StarkVkeyFrame`), but this **breaks internal VK stability**: the value depends on the *child* VK's symbolic constraints, which change across recursion layers, so baking it into the AIR as a VK-derived constant causes each layer to produce a different verifying key.

The fix restores the value as a witness column (`num_air_id_lookups`) computed at trace-generation time from the child VK, keeping it out of the constraint system's VK-derived path.

---

### `recursion` — ProofShapeAir (`proof_shape/proof_shape/`)

**air.rs**: Replace VK-derived `num_dag_nodes` expression with the new witness column `num_air_id_lookups` in the two `AirShapeBus` lookup multiplicities:

```
-  local.is_present * (num_dag_nodes.clone() + TWO)
+  local.is_present * (local.num_air_id_lookups + TWO)

-  local.is_present * (num_dag_nodes + ONE)
+  local.is_present * (local.num_air_id_lookups + ONE)
```

**trace.rs**: Add `compute_air_shape_lookup_counts(vk) -> Vec<usize>` (logic moved from `frame.rs`). Called during CPU trace generation to populate `cols.num_air_id_lookups`.

**mod.rs**: Widen visibility of `trace` re-exports to `pub(crate)` so the CUDA preflight can call `compute_air_shape_lookup_counts`.

### `recursion` — System frame (`system/frame.rs`)

Remove `num_dag_nodes` from `StarkVkeyFrame` and delete the now-unused `compute_num_dag_nodes` helper. This field was the VK-stability hazard.

### `recursion` — VK metadata (`proof_shape/mod.rs`)

Remove `num_dag_nodes` from `AirMetadata` (no longer needed since the value is computed at trace-gen time, not stored on the VK frame).

### `recursion` — CUDA trace generation

- **`cuda/types.rs`** + **`cuda/include/types.h`**: Add `num_air_id_lookups` field to `TraceMetadata`.
- **`cuda/preflight.rs`**: Call `compute_air_shape_lookup_counts` and pass the per-AIR values into `TraceMetadata`.
- **`cuda/src/proof_shape/air.cu`**: Write `num_air_id_lookups` into present rows; zero it for non-present and summary rows.

### `continuations-v2` — Tests

Use a distinct `MAX_LEAF_NUM_PROOFS = 3` (instead of reusing `DEFAULT_MAX_NUM_PROOFS = 4`) for the leaf prover in `test_internal_recursive_vk_stabilization`, exercising the case where leaf and internal layers have different `MAX_NUM_PROOFS`.

---

## ProofShapeCols column change

| Column | Present row | Non-present row | Summary row |
|---|---|---|---|
| `num_air_id_lookups` | `dag_nodes + unused_vars + Σ interaction_lengths` (from child VK) | 0 | 0 |


jonathanpwang (approved):
lgtm but @Golovanov399 should also look to ensure it still satisfies the previous fix requirements

crates/recursion/src/proof_shape/proof_shape/air.rs
```diff
@@ -90,6 +90,8 @@ pub struct ProofShapeCols<F, const NUM_LIMBS: usize> {
     /// Computed as max(0, n0, n1, ...) where ni = log_height_i - l_skip for each present trace.
     pub n_max: F,
     pub is_n_max_greater: F,
+
+    pub num_air_id_lookups: F,
```
luffykai, line 94: isn't it unconstrained? the tracegen "should" compute it from vkey but we can't trust?

stephenh-axiom-xyz:
This will need to resolved on the `SymbolicExpressionAir` side

PR #2526, opened 2026-03-10
ci: rebase feature branches on top of v2

jonathanpwang:
I hope this works, but the idea is that stuff merged to develop-v2.0.0-beta will then rebase develop-v2.0.0-rc.1

On main, we do not include develop-v2.0.0-rc.1, so merge to main should set a chain of rebases main -> develop-v2.0.0-beta -> develop-v2.0.0-rc.1

### DIFF
--- a/.github/workflows/rebase-feature-branches.yml
+++ b/.github/workflows/rebase-feature-branches.yml
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - main
+      - develop-v2.0.0-beta
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/rebase-feature-branches.yml
shuklaayush: this needs to be changed to `git rebase "origin/${{ github.ref_name }}"`

shuklaayush: probably this one too

jonathanpwang: good catch

shuklaayush (approved):

PR #2529, opened 2026-03-11
fix: InitialOpenedValuesAir ensure no breaks between set flags

stephenh-axiom-xyz:
Resolves INT-6701, INT-6607.

Constraints ensure all flags from 0 to some `max_col_idx` are set by ensuring all are set on each non-terminal row. Note `stacking_indices_bus` sends the same number of messages per valid `(commit_idx, col_idx)`, and thus:

- If an extra flag is set, there will be not be a corresponding key added on the stacking side
- If a flag is left unset, there will be no way one could ensure each `(commit_idx, col_idx)` is looked up the same number of times


Golovanov399 (approved):
I am convinced

PR #2530, opened 2026-03-11
fix(recursion): fully constrain exp_bits_len in BabyBear

zlangley:
Use fixed 32-row blocks with a full 31-bit decomposition and terminal row. Enforce canonical BabyBear representatives for bit_src.

### DIFF
--- a/crates/recursion/src/primitives/exp_bits_len/air.rs
+++ b/crates/recursion/src/primitives/exp_bits_len/air.rs
@@ -3,27 +3,48 @@ use core::borrow::Borrow;
 use openvm_stark_backend::{
     interaction::InteractionBuilder, BaseAirWithPublicValues, PartitionedBaseAir,
 };
-use openvm_stark_sdk::config::baby_bear_poseidon2::F;
 use p3_air::{Air, AirBuilder, BaseAir};
-use p3_field::PrimeCharacteristicRing;
+use p3_baby_bear::BabyBear;
+use p3_field::{PrimeCharacteristicRing, PrimeField32};
 use p3_matrix::Matrix;
 use stark_recursion_circuit_derive::AlignedBorrow;
 
-use crate::primitives::bus::{ExpBitsLenBus, ExpBitsLenMessage};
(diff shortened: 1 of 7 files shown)


crates/recursion/src/primitives/exp_bits_len/air.rs
```diff
@@ -37,38 +58,153 @@ impl ExpBitsLenAir {
...
+        builder.assert_bool(local.high_bits_all_one);
+        builder.assert_bool(local_in_high_region.clone());
+        builder.when(local.is_first).assert_one(local.is_valid);
+        builder.when(local.num_bits).assert_one(local.apply_bit);
+        builder
+            .when(local.low_bits_left)
+            .assert_one(local.in_low_region);
         builder.assert_eq(
+            local.result_multiplier - AB::Expr::ONE,
+            local.apply_bit * local.bit_src_mod_2 * (local.base - AB::Expr::ONE),
```
Golovanov399, line 120: We can add something like `builder.when(not(local.apply_bit)).assert_zero(local.bit_src_mod_2)` and this constraint will become quadratic

crates/recursion/src/primitives/exp_bits_len/air.rs
```diff
@@ -37,38 +58,153 @@ impl ExpBitsLenAir {
...
+        builder
+            .when(local.is_first)
+            .assert_one(local.low_bits_are_zero);
+        builder
+            .when(local.is_first)
+            .assert_zero(local.high_bits_all_one);
+
+        builder
+            .when(local_is_last.clone())
+            .assert_one(local.is_valid);
```
Golovanov399: It's redundant because `local_is_last = local.is_valid - bool`, but I don't mind

Golovanov399 (approved):
I think the AIR should be good now.

I approve on the condition that the cuda tracegen is fixed and no further changes in the AIR (except maybe `apply_bit -> bit_src_mod_2`), otherwise please ping me to rereview

PR #2532, opened 2026-03-11
fix: MerkleVerifyAir constrains the Merkle path taken to merkle_idx

stephenh-axiom-xyz:
Resolves INT-6728. 

- Modifies `ExpBitLenAir` to provide a right-shifted `bit_src` given its original value and a number of bits
- Each Merkle path `MerkleVerifyAir` consumes a bit of `merkle_idx` (remaining bits stored in `current_idx`)
- `current_idx` is compared against `merkle_idx` using `BitShiftBus`

### DIFF
--- a/crates/recursion/src/primitives/bus.rs
+++ b/crates/recursion/src/primitives/bus.rs
@@ -30,3 +30,13 @@ pub struct ExpBitsLenMessage<T> {
 }
 
 define_typed_lookup_bus!(ExpBitsLenBus, ExpBitsLenMessage);
+
+#[repr(C)]
+#[derive(AlignedBorrow, Debug, Clone)]
+pub struct BitShiftMessage<T> {
+    pub base: T,
+    pub num_bits: T,
+    pub result: T,
+}
+
+define_typed_lookup_bus!(BitShiftBus, BitShiftMessage);
--- a/crates/recursion/src/primitives/bus.rs
+++ b/crates/recursion/src/primitives/bus.rs
@@ -33,10 +33,10 @@ define_typed_lookup_bus!(ExpBitsLenBus, ExpBitsLenMessage);
 
 #[repr(C)]
 #[derive(AlignedBorrow, Debug, Clone)]
-pub struct BitShiftMessage<T> {
-    pub base: T,
-    pub num_bits: T,
+pub struct RightShiftMessage<T> {
+    pub input: T,
+    pub shift_bits: T,
     pub result: T,
 }
 
-define_typed_lookup_bus!(BitShiftBus, BitShiftMessage);
+define_typed_lookup_bus!(RightShiftBus, RightShiftMessage);
--- a/crates/recursion/src/transcript/merkle_verify/air.rs
+++ b/crates/recursion/src/transcript/merkle_verify/air.rs
@@ -15,7 +15,7 @@ use crate::{
         CommitmentsBus, CommitmentsBusMessage, MerkleVerifyBus, MerkleVerifyBusMessage,
         Poseidon2CompressBus, Poseidon2CompressMessage,
     },
-    primitives::bus::{BitShiftBus, BitShiftMessage},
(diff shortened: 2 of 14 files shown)


crates/recursion/src/primitives/bus.rs
```diff
@@ -30,3 +30,13 @@ pub struct ExpBitsLenMessage<T> {
 }
 
 define_typed_lookup_bus!(ExpBitsLenBus, ExpBitsLenMessage);
+
+#[repr(C)]
+#[derive(AlignedBorrow, Debug, Clone)]
+pub struct BitShiftMessage<T> {
+    pub base: T,
+    pub num_bits: T,
```
zlangley: Call this `shift_size` or something?

crates/recursion/src/primitives/bus.rs
```diff
@@ -30,3 +30,13 @@ pub struct ExpBitsLenMessage<T> {
 }
 
 define_typed_lookup_bus!(ExpBitsLenBus, ExpBitsLenMessage);
+
+#[repr(C)]
+#[derive(AlignedBorrow, Debug, Clone)]
+pub struct BitShiftMessage<T> {
+    pub base: T,
```
zlangley: `input`?

crates/recursion/src/primitives/bus.rs
```diff
@@ -30,3 +30,13 @@ pub struct ExpBitsLenMessage<T> {
 }
 
 define_typed_lookup_bus!(ExpBitsLenBus, ExpBitsLenMessage);
+
+#[repr(C)]
+#[derive(AlignedBorrow, Debug, Clone)]
+pub struct BitShiftMessage<T> {
```
zlangley: Should have "Right" somewhere probably

crates/recursion/src/transcript/merkle_verify/air.rs
```diff
@@ -164,12 +175,18 @@ impl<AB: AirBuilder + InteractionBuilder> Air<AB> for MerkleVerifyAir {
...
+            local.current_idx * (AB::Expr::TWO - local.is_combining_leaves);
+        let right_child_current_idx =
+            left_child_current_idx.clone() + AB::Expr::ONE - local.is_combining_leaves;
+
         self.merkle_verify_bus.receive(
             builder,
             local.proof_idx,
             MerkleVerifyBusMessage {
                 value: local.left.map(Into::into),
                 merkle_idx: local.idx.into(),
```
zlangley: Can't we use `merkle_idx` for `current_idx`? Or just remove `merkle_idx` from the bus. Like it feels weird that `merkle_idx` is only acting as an identifier for the request but otherwise has no value?

crates/recursion/src/transcript/merkle_verify/air.rs
```diff
@@ -53,6 +56,8 @@ pub struct MerkleVerifyCols<T> {
 
     /// The merkle idx of the leaf hash root
     pub idx: T,
+    /// Merkle idx suffix after shifting right max(0, height - k) bits
+    pub current_idx: T,
```
zlangley: I am confused about the differences between `idx`, `current_idx`, and `merkle_idx`. Why do we need all three?

stephenh-axiom-xyz: `idx` is `merkle_idx`, I kept the name it originally had but I can change it. `current_idx` is `merkle_idx` shifted

crates/recursion/src/transcript/merkle_verify/air.rs
stephenh-axiom-xyz: We pass the original `merkle_idx` through the tree/path because we want to compare the final `current_idx` against it

zlangley (approved):
LGTM. I think we should change the name `merkle_idx` in `MerkleVerifyBusMessage` to emphasize that it's just a bit source and not the actual index.

PR #2539, opened 2026-03-12
docs: update stale NestedForLoopSubAir docs

zlangley:


shuklaayush:
@claude can you check if there are other changes required in crates/recursion/src/subairs/nested_for_loop/README.md to match the implementation in crates/recursion/src/subairs/nested_for_loop/air.rs?

shuklaayush:
> Note: The code also has builder.assert_bool(next_is_first) ([air.rs:77](https://github.com/openvm-org/openvm/blob/e6a6efa/crates/recursion/src/subairs/nested_for_loop/air.rs#L77)) which is redundant with assert_bool(local_is_first) (since the local constraint applies to every row). The README correctly omits this.Note: The code also has builder.assert_bool(next_is_first) ([air.rs:77](https://github.com/openvm-org/openvm/blob/e6a6efa/crates/recursion/src/subairs/nested_for_loop/air.rs#L77)) which is redundant with assert_bool(local_is_first) (since the local constraint applies to every row). The README correctly omits this.


guess we should fix this in a follow-up

shuklaayush (approved):

PR #2542, opened 2026-03-12
fix: various proof shape + stacking + transcript fixes

stephenh-axiom-xyz:
Resolves INT-6749, INT-6753, INT-6757, INT-6761, INT-6764, INT-6765. In order of commit:

#### INT-6765: TranscriptAir: Terminal FinalTranscriptStateBus export is not bound to the actual final transcript state

Asserts `pre_state` and `post_state` are equal on rows we do not permute.

#### INT-6749: ProofShapeAir: Absent AIR rows can export arbitrary cached commitments

Replace the `local.is_valid` gate in the `CachedTraceCommitBus` message mult to `local.is_present`.

#### INT-6761: UnivariateRoundAir: Univariate AIR does not carry the |D| * a_0 term forward to the a_|D| row

Constrain `local.s_0_sum_over_d == next.s_0_sum_over_d` on all rows we don't change it.

#### INT-6757: OpeningClaimsAir: Opening coefficients can be permuted across stacked columns within a commitment

Constrain that `stacked_col_idx` increments by at most 1 each row, and that for each `is_last_claim` either `commit_idx` or `stacked_col_idx` increments by 1.

#### INT-6753: ProofShapeAir: Summary row exports a forgeable n_logup

Issue was that the prover could increase `n_logup` in order to decrease `limb_to_range_check * msb_limb_zero_bits_exp`, making it easier to pass the range check. Added a range check to ensure that the MSB of the product above is 1.

### DIFF
--- a/crates/recursion/src/stacking/opening/air.rs
+++ b/crates/recursion/src/stacking/opening/air.rs
@@ -341,10 +341,8 @@ where
             .assert_zero(next.stacked_col_idx);
 
         builder
-            .when(and::<AB::Expr>(
-                not(local.is_last),
-                not::<AB::Expr>(next.commit_idx - local.commit_idx),
-            ))
+            .when(not(local.is_last))
+            .when_ne(local.commit_idx, next.commit_idx)
             .assert_bool(next.stacked_col_idx - local.stacked_col_idx);
         builder
             .when(and(local.is_last_for_claim, not(local.is_last)))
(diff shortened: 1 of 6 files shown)


crates/recursion/src/stacking/opening/air.rs
```diff
@@ -340,6 +340,19 @@ where
             .when(next.commit_idx - local.commit_idx)
             .assert_zero(next.stacked_col_idx);
 
+        builder
+            .when(and::<AB::Expr>(
+                not(local.is_last),
+                not::<AB::Expr>(next.commit_idx - local.commit_idx),
+            ))
```
zlangley: I think this reads better:
```
builder
    .when(not(local.is_last))
    .when_ne(local.commit_idx, next.commit_idx)
```

zlangley (approved):

PR #2548, opened 2026-03-13
fix: bind `sum_claim_p` and `sum_claim_q` in FractionsFolderAir to `sort_idx`

Golovanov399:
This resolves INT-6775


Golovanov399:
Discarded

PR #2550, opened 2026-03-13
feat: change new SHA2 AIRs to have v2 max constraint degree <= 3

876pol:
Resolves INT-6189.

### Summary

Reduces the maximum v2 constraint degree of the SHA2 block hasher AIR from 4 to 3 by converting two degree-4 `when_transition()` constraints into unconditional (degree-3) constraints, and fixing the `global_block_idx` padding row constraint to be wrap-around safe.

### AIR Changes

**1. Message schedule `intermed_4` constraint (line ~433):** Changed from `builder.when_transition().assert_eq(...)` to `builder.assert_eq(...)`. This removes one degree of multiplication (the transition selector), keeping the constraint at degree 3. The trace now fills dummy `intermed_4` values on the wrap-around edge (last row → first row) to satisfy the unconditional constraint.

**2. Message schedule addition (`intermed_12`) constraint (line ~477):** Similarly changed `constraint_word_addition` from being called on `builder.when_transition()` to being called on `builder` directly. Dummy `intermed_12` and `carry_or_buffer` values are filled on wrap-around rows to compensate.

**3. Global block index on padding rows (line ~276-283):** Previously constrained `global_block_idx == 0` on padding rows. Now constrains `global_block_idx` to be constant across padding rows (via a transition constraint). This removes the `when(is_digest_row) * when(is_round_row)` degree-4 interaction on the block boundary transition, since padding rows no longer need a different value from the last real block.

**4. Global block index increment (line ~269-275):** Removed the extra `.when(*next_cols.flags.is_round_row)` condition from the block-boundary increment constraint, reducing its degree by 1. The padding row constancy constraint now handles the case where the next row after a digest row is a padding row (rather than needing to exclude it).

### DIFF
--- a/extensions/sha2/circuit/cuda/src/sha2_hasher.cu
+++ b/extensions/sha2/circuit/cuda/src/sha2_hasher.cu
@@ -1031,8 +1031,8 @@ int launch_sha2_second_pass_dependencies(Fp *d_trace, size_t trace_height, size_
     auto [grid_size, block_size] = kernel_launch_params(total_blocks, 256);
     sha2_second_pass_dependencies<V>
         <<<grid_size, block_size>>>(d_trace, trace_height, total_blocks);
-    if (CHECK_KERNEL() != 0) {
-        return -1;
+    if (auto err = CHECK_KERNEL(); != 0) {
+        return err;
     }
 
     sha2_fill_wraparound<V><<<1, 1>>>(d_trace, trace_height);
--- a/extensions/sha2/circuit/cuda/src/sha2_hasher.cu
+++ b/extensions/sha2/circuit/cuda/src/sha2_hasher.cu
@@ -1031,8 +1031,8 @@ int launch_sha2_second_pass_dependencies(Fp *d_trace, size_t trace_height, size_
     auto [grid_size, block_size] = kernel_launch_params(total_blocks, 256);
     sha2_second_pass_dependencies<V>
         <<<grid_size, block_size>>>(d_trace, trace_height, total_blocks);
-    if (CHECK_KERNEL() != 0) {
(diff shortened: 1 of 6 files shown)


extensions/sha2/circuit/cuda/src/sha2_hasher.cu
```diff
@@ -997,6 +1031,11 @@ int launch_sha2_second_pass_dependencies(Fp *d_trace, size_t trace_height, size_
     auto [grid_size, block_size] = kernel_launch_params(total_blocks, 256);
     sha2_second_pass_dependencies<V>
         <<<grid_size, block_size>>>(d_trace, trace_height, total_blocks);
+    if (CHECK_KERNEL() != 0) {
+        return -1;
+    }
```
Golovanov399:
```suggestion
    if (auto err = CHECK_KERNEL(); != 0) {
        return err;
    }
```

Golovanov399 (approved):

jonathanpwang:
@876pol you should provide "proof" that this lowers the max constraint degree. Is there a test where we set SystemParams to have max degree 3 and it passes?

876pol:
When running the CUDA tests, warnings like `WARN     🚧 [warn]: Actual max constraint degree across all AIRs (3) does not match configured max constraint degree (4)` now pop up

Should I modify all the tests to have SystemParams to have max degree 3?

876pol:
I modified the tests to have SystemParams = max degree 3 in this commit: https://github.com/openvm-org/openvm/pull/2550/changes/5cbe785a9499aadc445e23e3e7329257ca557f15, and the tests passed, but I reverted the SystemParams in later commits

PR #2557, opened 2026-03-16
fix: constrain is_last in ProofShapeAir

stephenh-axiom-xyz:
Resolves INT-6817. Not convinced that the finding suggestion is correct as this AIR uses `NestedForLoopSubAir`, but this constraint is probably good to have


Golovanov399 (approved):
Seems correct

PR #2562, opened 2026-03-16
fix: constrain row_idx_flags to row_idx in UserPvsCommitAir

stephenh-axiom-xyz:
Resolves INT-6783.


shuklaayush (approved):
i didn't audit this part but looks fine

PR #2563, opened 2026-03-16
ci: only run CUDA lint when CUDA files change and use 8-cpu runner

jonathanpwang:
## Summary
- Add a `changes` detection job using `dorny/paths-filter@v3` to skip the `lint-cuda` job when no CUDA-related files (`*.cu`, `*.cuh`, `**/cuda/**`, `**/cuda*.rs`, etc.) are modified
- Request at least 8 CPUs on the GPU runner (`/cpu=8`) for faster builds
- The workflow file and CUDA cache action are also included as triggers so changes to CI itself still run the CUDA lint

## Test plan
- [ ] Open a PR that doesn't touch any CUDA files and verify `lint-cuda` is skipped
- [ ] Open a PR that touches a `.cu` or `.cuh` file and verify `lint-cuda` runs

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -26,6 +26,7 @@ jobs:
               - '**/cuda/**'
               - '**/cuda*.rs'
               - '**/cuda_abi.rs'
+              - '**/build.rs'
               - '.github/workflows/lints.yml'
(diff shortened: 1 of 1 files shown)


stephenh-axiom-xyz (approved):
LGTM except I'd add `build.rs` files

PR #2564, opened 2026-03-16
fix: constrain initial + final merkle proofs are equal up to address_height

stephenh-axiom-xyz:
Resolves INT-6766, INT-6646. Changes the following columns:

- `expected_branch_bits_offset` renamed to `is_unset`, still 1 for the entire proof iff no deferral proof is present.
- `is_untouched` changed to be 1 iff `depth <= address_height` (i.e. no dependency on `is_unset`)

We now constrain that `initial_sibling == final_sibling` when `is_untouched`. This constrains that the rest of the `DEFERRAL_AS` is left untouched during program execution.

### DIFF
--- a/crates/continuations/src/circuit/root/def_paths/trace.rs
+++ b/crates/continuations/src/circuit/root/def_paths/trace.rs
@@ -88,7 +88,7 @@ pub fn generate_proving_input<SC: StarkProtocolConfig<F = F>>(
         depth
     };
 
-    // `is_untouched` tracks the DEFERRAL_AS prefix outside the Merkle-rooted part.
+    // is_within_deferral_as tracks the DEFERRAL_AS prefix outside the Merkle-rooted part
     let untouched_cut = address_height.min(proof_len.saturating_sub(1));
 
     let mut is_right_child_bits = vec![false; num_layers];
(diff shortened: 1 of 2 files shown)


crates/continuations/src/circuit/root/def_paths/trace.rs
```diff
@@ -170,8 +144,8 @@ pub fn generate_proving_input<SC: StarkProtocolConfig<F = F>>(
         cols.is_valid = if row_idx == 0 { F::TWO } else { F::ONE };
         cols.depth = F::from_usize(row_idx);
         cols.is_skip = F::from_bool(row_idx < skip_depth);
-        cols.is_untouched = F::from_bool(is_unset && row_idx <= untouched_cut);
-        cols.expected_branch_bits_offset = F::from_bool(is_unset);
+        cols.is_untouched = F::from_bool(row_idx <= untouched_cut);
```
shuklaayush: i feel like this should be something like `is_within_deferral_as`

shuklaayush (approved):
looks good. it asserts now that the merkle paths of all untouched leaves in deferral address space do not change which implies that those circuits were never called

PR #2568, opened 2026-03-16
fix: final deferral input commit should be hash slice of commits

stephenh-axiom-xyz:
Resolves INT-6726.


crates/continuations/src/circuit/deferral/inner/input/air.rs
```diff
@@ -131,40 +125,55 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB> f
...
             InputOrMerkleCommitMessage {
-                has_verifier_pvs: local.has_verifier_pvs,
-                commit: local.current_commit,
+                has_verifier_pvs: local.has_verifier_pvs.into(),
+                commit: from_fn(|i| {
+                    is_internal * local.current_commit[i] + is_leaf.clone() * local.res_left[i]
+                }),
             },
-            local.state * (local.state - AB::Expr::ONE) * AB::F::TWO.inverse(),
+            local.is_valid * not::<AB::Expr>(is_transition),
```
shuklaayush: can probably also use `NestedForLoopSubAir::local_is_last(..)` which should be linear

crates/continuations/src/circuit/deferral/inner/input/air.rs
```diff
@@ -131,40 +125,55 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB> f
...
+         * from the previous row.
+         */
+        let is_transition = next.is_valid - next.is_first;
+
+        assert_zeros(&mut builder.when(local.is_first), local.capacity);
+        assert_array_eq(
+            &mut builder.when(is_leaf.clone() * is_transition.clone()),
+            local.res_right,
+            next.capacity,
         );
```
shuklaayush, line 156: not blocking but it feels like `res_right` and `capacity` are redundant and we can get away with just one set of columns

shuklaayush (approved):
some nits

shuklaayush (approved):

PR #2569, opened 2026-03-16
fix: verify-stark guest lib findings

stephenh-axiom-xyz:
Resolves INT-6848 and INT-6849. Note the guest `verify-stark` will now call the host function.

### DIFF
--- a/extensions/deferral/guest/src/ops.rs
+++ b/extensions/deferral/guest/src/ops.rs
@@ -41,7 +41,7 @@ pub fn deferred_compute_raw<const DEFERRAL_IDX: u16>(
     output_key_ptr: *mut u8,
     input_commit_ptr: *const u8,
 ) {
-    assert!(DEFERRAL_IDX < MAX_DEF_CIRCUITS);
+    const { assert!(DEFERRAL_IDX < MAX_DEF_CIRCUITS) };
 
     #[cfg(target_os = "zkvm")]
     deferral_call!(output_key_ptr, input_commit_ptr, DEFERRAL_IDX);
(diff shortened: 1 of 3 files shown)


guest-libs/verify-stark/circuit/src/extension.rs
```diff
@@ -83,6 +84,8 @@ pub fn get_raw_deferral_results(
     proofs
         .iter()
         .map(|proof| {
+            verify_vm_stark_proof_decoded(vk, proof)?;
```
stephenh-axiom-xyz: @jonathanpwang this is fine, right? Deferral operations are expected to be a little expensive, and this shouldn't be too bad

extensions/deferral/guest/src/ops.rs
```diff
@@ -41,7 +41,7 @@ pub fn deferred_compute_raw<const DEFERRAL_IDX: u16>(
     output_key_ptr: *mut u8,
     input_commit_ptr: *const u8,
 ) {
-    debug_assert!(DEFERRAL_IDX < MAX_DEF_CIRCUITS);
+    assert!(DEFERRAL_IDX < MAX_DEF_CIRCUITS);
```
shuklaayush: maybe this also works `const { assert!(DEFERRAL_IDX < MAX_DEF_CIRCUITS) };`

stephenh-axiom-xyz: Oh that is good to know

shuklaayush (commented):
looks fine to me. i'll leave it to @jonathanpwang for the final approval
it's unfortunate that we need to call verify twice though but maybe it's fine since this is done before execution

stephenh-axiom-xyz:
Updated to exclude the dupe engine verify

shuklaayush (approved):

PR #2570, opened 2026-03-16
docs: explain the correspondence between verifier circuit and verify()

zlangley:
The goal is to argue that the VerifierCircuit actually constrains verify(). We do not attempt to explain how each AIR's constraints work, but rather state the module-level invariants, and how those imply correctness.


jonathanpwang (commented):
@zlangley I think this should be moved to `docs/crates` (and maybe inside `recursion`) since that's where the rest of the recursion docs are.

Slightly confusing, but `repo` is mainly for repo layout / maintenance types of docs.

stephenh-axiom-xyz:
Couple comments from offline discussion:
- Some things are a little outdated as we've made some very recent changes, ex.
    - `DagCommitBus` was removed
    - `ExpBitsLenAir` also provides `RightShiftBus` lookups
    - `Eq3bShapeBus` now provides `num_interactions`
- I don't think you establish the MerkleVerifyAir lookup as a lookup, it's mentioned in WHIR but not really anywhere else

stephenh-axiom-xyz (approved):
LGTM after claude nits

PR #2571, opened 2026-03-16
fix: add num_def_circuit_proofs to DeferralAggregationPvs

stephenh-axiom-xyz:
Resolves INT-6613 and INT-6782.

#### INT-6613: `DeferralAggPvsAir`: absent child `merkle_commit` unconstrained in two-child mode

The deferral hook circuit now receives `num_def_circuit_proofs`, and constrains that there are exactly onion hash elements. This implies that there are `num_def_circuit_proofs` set Merkle tree leaves, and the remainder are constrained to be 0. Should some child `merkle_commit` not be the correct `zero_hash` value the Merkle tree root will not equal the final `merkle_commit`.

#### INT-6782: Constrain number of deferral circuit inputs in deferral hook

`num_def_circuit_proofs` is now propagated and updated throughout the deferral aggregation tree. The final onion hash must consist of exactly `num_def_circuit_proofs` elements.

### DIFF
--- a/crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
+++ b/crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
@@ -21,7 +21,7 @@ use crate::{
             DefPvsConsistencyBus, DefPvsConsistencyMessage, InputOrMerkleCommitBus,
             InputOrMerkleCommitMessage,
         },
-        DeferralAggregationPvs, DeferralCircuitPvs, DEF_CIRCUIT_PVS_AIR_ID,
+        DeferralAggregationPvs, DeferralCircuitPvs, DEF_AGG_PVS_AIR_ID, DEF_CIRCUIT_PVS_AIR_ID,
     },
     utils::digests_to_poseidon2_input,
 };
@@ -137,7 +137,7 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
                 ),
                 output: local.merkle_commit,
             },
-            is_leaf * local.is_present,
+            is_leaf.clone() * local.is_present,
         );
 
         /*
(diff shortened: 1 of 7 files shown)


crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
```diff
@@ -153,21 +153,54 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
...
 
+        /*
+         * On internal layers we receive num_def_circuit_proofs from PublicValuesBus. To save
+         * columns, we repurpose the unused local.child_pvs.input_commit[0].
+         */
+        let local_num_def_circuit_proofs = local.child_pvs.input_commit[0];
+
+        builder
+            .when(not(local.is_present))
+            .when(is_internal)
```
shuklaayush, line 165: can probably drop the `when(is_internal)`

crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
```diff
@@ -153,21 +153,54 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
...
 
-        assert_array_eq(
-            &mut builder.when(is_first * not(next.proof_idx)),
-            local.merkle_commit,
-            merkle_commit,
-        );
+        let local_num_proofs =
+            local.is_present * is_leaf.clone() + is_internal * local_num_def_circuit_proofs;
+        let next_num_proofs =
+            next.is_present * is_leaf + is_internal * next.child_pvs.input_commit[0];
```
shuklaayush, line 196: i think we should also constrain that `local.has_verifier_pvs == next.has_verifier_pvs`

crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
stephenh-axiom-xyz: We re-use `local.child_pvs.input_commit[0]` here, so when `!is_internal` this value means something else. We remedy this below by multiplying `is_leaf` by `is_present`.

shuklaayush (approved):

PR #2572, opened 2026-03-17
fix(v2-rc1): remove `block_size` from config

shuklaayush:
## Overview

With access adapters removed, every address space now uses a single fixed block size (`DEFAULT_BLOCK_SIZE = 4`). This PR removes the per-address-space `block_size` field and the associated dynamic block-size machinery, making the memory subsystem simpler and faster.

**No AIR changes.** This is purely host-side (execution, trace generation, testing).

**Bug fixes included:**

1. The metered execution path and AOT code path were using `DEFAULT_BLOCK_SIZE` (4) to compute merkle tree leaf indices, but `MemoryDimensions::address_height` is sized for `CHUNK` (8) leaves. This was introduced when `initial_block_size()` (which returned `CHUNK = 8` on `main`) was replaced with `CONST_BLOCK_SIZE = 4`. Both paths now correctly use `controller::CHUNK`.

2. The boundary AIR height estimate was `leaves` but the actual trace uses `2 * leaves` (one init + one final row per leaf). This caused segmentation underestimates that surfaced in ECC integration tests.

Resolves int-6630.

---

### `arch/config.rs`

- Removed `AddressSpaceHostConfig::block_size` field. Constructor no longer takes a block size.
- Removed `MemoryConfig::block_size_bits()`.
- Removed `MAX_CELL_BYTE_SIZE` (unused after `handle_touched_blocks` simplification).

### `system/memory/online.rs` (core simplification)

- **Removed `TracingMemory::block_size`** field (per-AS `Vec<u32>`). Metadata slots are now always `DEFAULT_BLOCK_SIZE`-sized.
- **Simplified `prev_access_time`**: no const generic, no loop. Single-slot get/set using `pointer / DEFAULT_BLOCK_SIZE`.
- **Simplified `touched_blocks`**: uses `DEFAULT_BLOCK_SIZE` directly instead of zipping with per-AS block sizes.
- **Replaced `handle_touched_blocks`** (accumulation loop with `current_timestamps`, `current_values` buffer, max-timestamp logic) with a direct parallel `.map()` in `touched_blocks_to_equipartition` — each touched block maps 1:1 to a `TimestampedValues` entry.

### `system/memory/persistent.rs`

- Extracted `group_touched_memory_by_chunk` as a `pub(crate)` function so the rechunking logic (grouping `DEFAULT_BLOCK_SIZE` blocks into `CHUNK`-sized merkle leaves) is shared between `PersistentBoundaryChip::finalize` and `MemoryController::generate_proving_ctx`.
- Added early return in `touch_range` for `len == 0`.

### `system/memory/controller/mod.rs`

- Rechunking in `generate_proving_ctx` now uses `group_touched_memory_by_chunk` instead of inline `BTreeMap` logic.

### `arch/execution_mode/metered/memory_ctx.rs` (bug fixes)

- **Fixed CHUNK mismatch**: `CHUNK` was aliased to `DEFAULT_BLOCK_SIZE` (4), but merkle tree addressing uses `controller::CHUNK` (8). Now imports and uses `controller::CHUNK` directly.
- **Fixed boundary height underestimate**: `BOUNDARY_AIR_ID += leaves` → `leaves * 2` (init + final row per leaf).
- **Fixed `num_blocks` for unaligned accesses**: uses `(ptr + size - 1) >> CHUNK_BITS - ptr >> CHUNK_BITS + 1` instead of `(size + CHUNK - 1) >> CHUNK_BITS`, which was incorrect when `ptr` is not chunk-aligned.
- Removed `chunk`, `chunk_bits`, `boundary_idx`, `merkle_tree_index` struct fields — replaced with module-level constants and `BOUNDARY_AIR_ID` / `MERKLE_AIR_ID`.
- Added doc comment on `apply_height_updates` explaining the paged overestimate strategy.

### `arch/testing/memory/` (tester simplification)

- **`MemoryTester`**: Replaced `HashMap<usize, MemoryDummyChip>` with a single `chip` field. Added `into_parts()`. `read`/`write` assert `N == DEFAULT_BLOCK_SIZE`.
- **`MemoryDummyAir`**: Removed runtime `block_size` field. Width hardcoded to `DEFAULT_BLOCK_SIZE + 4`.
- **`DeviceMemoryTester`** (CUDA): Same pattern — single `chip`, `take_chip()`, assertions on `N`.

### `system/memory/tests.rs`

- Simplified to use only `DEFAULT_BLOCK_SIZE`-aligned accesses (was exercising 1/2/4/8/16-byte widths that no longer exist).

---

### `adapters/deferral.rs`

- Updated safety comments: "minimum alignment of 1" → "block_size of DEFAULT_BLOCK_SIZE".

### `adapters/mod.rs`

- Updated safety comments to reference the fixed 4-byte memory access granularity.

### `common/mod.rs` (bug fix)

- AOT code path: `chunk_bits` now uses `CHUNK.ilog2()` (= 3) instead of `DEFAULT_BLOCK_SIZE.ilog2()` (= 2), matching the merkle tree leaf size.
- Comments updated to reflect `CHUNK = 8`.

### `base_alu/tests.rs`, `loadstore/tests.rs`

- Cleaned up stale comments referencing "min block size" and "alignment".

### DIFF
--- a/crates/vm/src/arch/config.rs
+++ b/crates/vm/src/arch/config.rs
@@ -8,9 +8,7 @@ use derive_new::new;
 use getset::{Setters, WithSetters};
 use openvm_instructions::riscv::{RV32_IMM_AS, RV32_MEMORY_AS, RV32_REGISTER_AS};
 use openvm_poseidon2_air::Poseidon2Config;
-use openvm_stark_backend::{
-    p3_field::Field, p3_util::log2_strict_usize, StarkEngine, StarkProtocolConfig, Val,
-};
+use openvm_stark_backend::{p3_field::Field, StarkEngine, StarkProtocolConfig, Val};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use super::{AnyEnum, VmChipComplex, BOUNDARY_AIR_ID, CONNECTOR_AIR_ID, PROGRAM_AIR_ID};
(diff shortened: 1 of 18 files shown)


jonathanpwang (approved):
lgtm

what's DEFAULT_BLOCK_SIZE versus CONST_BLOCK_SIZE difference?

shuklaayush:
> what's DEFAULT_BLOCK_SIZE versus CONST_BLOCK_SIZE difference?

it's the same. in hindsight, probably an unnecessary change

PR #2574, opened 2026-03-17
fix: deferral output_commit should be a true hash

stephenh-axiom-xyz:
Resolves INT-6725, INT-6727, INT-6709, INT-6784, INT-6723, INT-6724

#### INT-6784: Zero-length outputs bypass the output AIR

Output commit is now the `hash_slice` of `[def_idx, output_len, 0, ...]` and`output_raw`. The first row in `DeferralOutputAir` is for the `[def_idx, output_len, 0, ...]` (i.e. skips a write), and `DeferralOutputCommitAir` does the same.

#### INT-6727: `DeferralOutputCommitAir` doesn't handle initial `def_idx` state

Updated `DeferralOutputCommitAir` to constrain the `def_idx` and `output_len` present on the first row. The former is compared against a constant, while `output_len` is constrained to be consistent with the number of valid rows.

#### INT-6723: `DeferralOutputAir` cleanup and degree reduction

Replaced the redundant constraints with `is_transition`, which is used throughout the AIR now.

#### INT-6724: `DeferralOutputCommitAir` cleanup

Removed `next_f` and the final invalid row by calling permute on the `next` row.

### DIFF
--- a/crates/sdk/src/prover/deferral/mod.rs
+++ b/crates/sdk/src/prover/deferral/mod.rs
@@ -57,6 +57,7 @@ impl DeferralProver {
         agg_config: AggregationConfig,
         hook_params: SystemParams,
     ) -> Self {
+        assert_eq!(def_circuit_prover.get_def_idx(), 0);
         let single_circuit_prover = SingleDefCircuitProver::new(
             def_circuit_prover,
             agg_config.params.leaf,
@@ -82,6 +83,7 @@ impl DeferralProver {
         def_hook_pk: Arc<MultiStarkProvingKey<SC>>,
         internal_recursive_cached_commit: CommitBytes,
     ) -> Self {
+        assert_eq!(def_circuit_prover.get_def_idx(), 0);
         let single_circuit_prover = SingleDefCircuitProver::from_pks(
             def_circuit_prover,
             def_agg_pk.leaf_pk,
@@ -108,6 +110,10 @@ impl DeferralProver {
         mut self,
         def_circuit_prover: DP,
     ) -> Self {
+        assert_eq!(
+            def_circuit_prover.get_def_idx(),
+            self.single_circuit_provers.len()
+        );
         let leaf_params = self.single_circuit_provers[0]
             .leaf_prover
             .get_vk()
--- a/crates/sdk/src/prover/deferral/mod.rs
+++ b/crates/sdk/src/prover/deferral/mod.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use eyre::Result;
+use itertools::Itertools;
 use openvm_continuations::{
     bn254::CommitBytes,
     prover::{DeferralChildVkKind, DeferralCircuitProver},
@@ -132,7 +133,7 @@ impl DeferralProver {
         let per_circuit = self
             .single_circuit_provers
             .iter()
-            .zip(inputs)
+            .zip_eq(inputs)
             .map(|(prover, inputs)| prover.prove(inputs))
             .collect::<Result<Vec<_>>>()?;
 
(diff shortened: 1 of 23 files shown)


extensions/deferral/circuit/src/poseidon2/bus.rs
```diff
@@ -47,9 +52,8 @@ impl<T: PrimeCharacteristicRing> DeferralPoseidon2Interaction<T> {
     where
         AB: InteractionBuilder<Expr = T>,
     {
-        let key: [_; 3 * PERIPHERY_POSEIDON2_CHUNK_SIZE] = self
-            .left
-            .into_iter()
+        let key: [_; 3 * PERIPHERY_POSEIDON2_CHUNK_SIZE + 1] = once(self.is_compress)
```
shuklaayush: should we put `is_compress` at the end? it's confusing if the ordering is different here

extensions/deferral/circuit/src/output/air.rs
```diff
@@ -159,45 +161,45 @@ where
...
+        );
+        assert_array_eq(
+            &mut when_last,
+            byte_commit_to_f(&local.output_commit),
+            local.poseidon2_res,
+        );
 
+        // Constrain poseidon2_res is the running permute capacity on all non-last rows,
+        // and the compression on the last row.
+        let rhs = from_fn(|i| is_transition.clone() * local.poseidon2_res[i]);
```
shuklaayush, line 197: isn't the capacity unconstrained on first row?

stephenh-axiom-xyz: It's 0 on the first row

crates/sdk/src/prover/deferral/mod.rs
shuklaayush: unrelated but we maybe want to `debug_assert` that lengths are equal or do `zip_eq`

extensions/deferral/circuit/src/output/air.rs
```diff
@@ -233,7 +235,7 @@ where
         });
         let output_commit_and_len =
             combine_output(local.output_commit.map(Into::into), output_len_full);
-        let output_commit_and_len_chunks =
+        let output_commit_and_len_chunks: [[<AB as AirBuilder>::Expr; 4]; 10] =
```
shuklaayush: we should avoid the magic number `10` and define a new constant

extensions/deferral/circuit/src/poseidon2/trace.rs
shuklaayush, line 1: unfortunate that we need to duplicate all this

guest-libs/verify-stark/circuit/src/output/air.rs
```diff
@@ -64,83 +66,107 @@ impl<AB: AirBuilder + InteractionBuilder> Air<AB> for DeferralOutputCommitAir {
...
+         * Compute the output commit and send it on the last row. We sponge hash each
+         * valid input_vals and take the left child of the last row.
          */
-        assert_array_eq(
-            &mut builder.when_first_row(),
-            local.state,
-            [AB::Expr::ZERO; DIGEST_SIZE],
-        );
-
+        let next_not_first = not(next.is_first);
```
shuklaayush: nit: same `is_transition = next.is_valid - next.is_first` makes it a bit clearer

guest-libs/verify-stark/circuit/src/output/air.rs
```diff
@@ -64,83 +66,107 @@ impl<AB: AirBuilder + InteractionBuilder> Air<AB> for DeferralOutputCommitAir {
...
+        let next_capacity = from_fn(|i| next_not_first.clone() * local.res_right[i]);
```
shuklaayush: just `capacity` maybe?

shuklaayush (approved):
some small nits. lgtm, i didn't find any glaring issues

guest-libs/verify-stark/circuit/src/output/air.rs
stephenh-axiom-xyz: It's the capacity for the next row

PR #2579, opened 2026-03-18
fix: range check shifted top byte of deferral heap pointers + output_len

stephenh-axiom-xyz:
Resolves INT-6785. Also replaces `output_len` canonicity check with a stricter range check


shuklaayush (approved):

shuklaayush:
> Assumption about address_bits: The shift computation RV32_CELL_BITS * RV32_REGISTER_NUM_LIMBS - self.address_bits (i.e., 32 - address_bits) must produce a non-negative value and the shifted result must fit in 8 bits. This means address_bits >= 24 is required (since the top byte can be up to 255, and 255 << (32 - address_bits) must be < 256). The typical value of 29 satisfies this. If address_bits could ever be set below 24, this would silently produce incorrect range checks. This is likely fine given the system constraints but worth noting.

we should add a `debug_assert` for this though

PR #2584, opened 2026-03-18
feat(refactor): static verifier witness flow

jonathanpwang:


jonathanpwang:
addressed findings in follow-ups

PR #2586, opened 2026-03-18
refactor(static-verifier): migrate to v1 BabyBearChip

jonathanpwang:
## Summary
- Copy over v1 BabyBear chips with lazy reduction and `max_bits` tracking
- Update all consumer interfaces: `Arc`-based constructors, named `.value` field, pass-by-value, `BabyBearExt4Wire` constructors
- Add guarded debug assert infrastructure (`utils.rs`) so negative tests disable BabyBearChip `debug_assert`s and verify constraint violations through MockProver

## Test plan
- [x] `cargo check -p openvm-static-verifier` — 0 errors, 0 warnings
- [x] `cargo nextest run --cargo-profile=fast -p openvm-static-verifier` — 110/110 tests pass

### DIFF
--- a/crates/static-verifier/src/field/baby_bear/base.rs
+++ b/crates/static-verifier/src/field/baby_bear/base.rs
@@ -1,221 +1,440 @@
+use std::sync::Arc;
+
 use halo2_base::{
-    gates::{range::RangeChip, GateChip, GateInstructions, RangeInstructions},
-    utils::ScalarField,
-    AssignedValue, Context,
-    QuantumCell::Constant,
+    gates::{GateChip, GateInstructions, RangeChip, RangeInstructions},
+    halo2_proofs::{
+        arithmetic::Field as _,
+        halo2curves::{bn256::Fr, ff::PrimeField as _},
+    },
+    utils::{bigint_to_fe, biguint_to_fe, bit_length, fe_to_bigint, BigPrimeField},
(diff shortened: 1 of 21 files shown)


crates/static-verifier/src/field/baby_bear/base.rs
```diff
@@ -1,221 +1,489 @@
...
-        c: BabyBearWire,
-        d: BabyBearWire,
-    ) -> BabyBearWire {
-        let ab = self.add2(ctx, a, b);
-        let cd = self.add2(ctx, c, d);
-        self.add2(ctx, ab, cd)
+        value: usize,
+    ) -> AssignedValue<Fr> {
+        let cell = ctx.load_witness(Fr::from(value as u64));
+        let bits = bit_length(value as u64).max(1);
```
jonathanpwang, line 392: this should not be allowed since `value` is an unknown witness, will fix in followup

crates/static-verifier/src/field/baby_bear/base.rs
```diff
@@ -1,221 +1,489 @@
...
+        self.range.range_check(ctx, cell, bits);
+        cell
     }
 
-    pub fn assign_and_range_u64(&self, ctx: &mut Context<Fr>, value: u64) -> AssignedValue<Fr> {
+    pub fn assign_and_range_u64(
```
jonathanpwang, line 397: ditto

PR #2589, opened 2026-03-19
chore: add continuations, SDK, and deferral workflows

stephenh-axiom-xyz:
Resolves INT-6842. Adds:

- Continuations CUDA tests
- SDK CUDA tests
- `deferral` extension CPU tests
- `verify-stark` guest library circuit CPU + CUDA tests

### DIFF
--- a/.github/workflows/sdk.cuda.yml
+++ b/.github/workflows/sdk.cuda.yml
@@ -1,0 +1,59 @@
+name: CUDA SDK Tests
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["**"]
+    paths:
+      - "crates/sdk/**"
+      - "crates/continuations/**"
+      - "crates/recursion/**"
+      - ".github/workflows/sdk.cuda.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  cuda-backend:
+    name: sdk cuda tests
+    runs-on:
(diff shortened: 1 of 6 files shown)


.github/workflows/sdk.cuda.yml
```diff
@@ -0,0 +1,59 @@
+name: CUDA SDK Tests
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["**"]
+    paths:
+      - "crates/sdk/**"
+      - "crates/continuations/**"
```
jonathanpwang: let's only trigger it on immediate crate changes (this is the policy right now on other workflows) to keep the PR CI lighter. I want to follow up with some merge train-esque comprehensive testing on feature branches

crates/sdk/src/prover/root.rs
```diff
@@ -35,8 +35,8 @@ use crate::{
 
 cfg_if::cfg_if! {
     if #[cfg(feature = "cuda")] {
-        use openvm_continuations::prover::RootGpuProver as RootInnerProver;
-        type E = openvm_cuda_backend::BabyBearBn254Poseidon2GpuEngine;
+        use openvm_continuations::prover::RootCpuProver as RootInnerProver;
+        type E = openvm_stark_sdk::config::baby_bear_bn254_poseidon2::BabyBearBn254Poseidon2CpuEngine;
```
jonathanpwang, line 39: ?? this doesn't look right? is it temporary?

jonathanpwang (approved):
lgtm except let's trigger workflow only on more immediate crate changes

and also I don't know what this root prover change is. if it's temporary, mark it so and make a ticket

crates/sdk/src/prover/root.rs
stephenh-axiom-xyz: Temporary as `BabyBearBn254Poseidon2GpuEngine` doesn't work on some machines

PR #2598, opened 2026-03-20
fix: host verify checks Proof cached commit

stephenh-axiom-xyz:
Resolves INT-6962.

- Adds a check to `crates/verify/src/lib.rs` to force the `Proof` cached commit to be the `internal-recursive` one
- Moves `CONSTRAINT_EVAL_CACHED_INDEX` from `crates/continuations` to `crates/verify`

### DIFF
--- a/crates/verify/src/lib.rs
+++ b/crates/verify/src/lib.rs
@@ -21,8 +21,8 @@ use crate::{
     deferral::DeferralMerkleProofs,
     error::VerifyStarkError,
     pvs::{
-        DeferralPvs, VerifierBasePvs, VerifierDefPvs, VmPvs, DEF_PVS_AIR_ID, VERIFIER_PVS_AIR_ID,
-        VM_PVS_AIR_ID,
+        DeferralPvs, VerifierBasePvs, VerifierDefPvs, VmPvs, CONSTRAINT_EVAL_AIR_ID,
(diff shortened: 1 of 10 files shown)


crates/verify/src/lib.rs
```diff
@@ -195,6 +195,19 @@ pub fn verify_vm_stark_proof_pvs(
         });
     }
 
+    // Check the proof's cached commit is the internal-recursive one.
+    let proof_cached_commit = proof.inner.trace_vdata[CONSTRAINT_EVAL_AIR_ID]
+        .as_ref()
+        .unwrap()
```
jpw-axiom: I guess if in keeping with the rest of the function, you should return error if this AIR doesn't exist? or will that already be checked by the required AIR check

crates/verify/src/lib.rs
```diff
@@ -195,6 +195,19 @@ pub fn verify_vm_stark_proof_pvs(
...
+        .cached_commitments[CONSTRAINT_EVAL_CACHED_INDEX];
```
jpw-axiom: `get` and return error instead of index (which unwraps)

jpw-axiom (approved):
LGTM if you handle errors instead of unwrap

crates/verify/src/lib.rs
stephenh-axiom-xyz: This should be done in the proof shape part of `engine.verify`

stephenh-axiom-xyz:
I realized this implementation is wrong, it may be the `internal_for_leaf_cached_commit`

PR #2599, opened 2026-03-20
perf(v2-rc.1): optimize keccakf cuda kernel

shuklaayush:
### Summary
Redesign `keccakf_perm_tracegen` as a two-phase kernel to fix a memory coalescing bottleneck. The original single-kernel design assigned 24 consecutive rows per thread, causing adjacent warp threads to store 24 rows apart in the column-major trace — wasting ~97% of memory bandwidth. The new design separates state computation (thread-per-permutation) from trace writing (thread-per-row), so adjacent threads write adjacent rows and stores coalesce. **~58x kernel speedup** (9.2s → 160ms total GPU time across 138 invocations on the reth benchmark).

### Changes

**`p3_keccakf.cuh`:**
- Add `apply_round_in_place`: keccak-f round function (theta/rho/pi/chi/iota) without trace column writes, used by phase 1
- Add `RHO_PI_CYCLE_IDX`/`RHO_PI_CYCLE_ROT` constants for in-place rho/pi permutation cycle
- `generate_trace_row_for_round`: replace `state_b[5][5]` with in-place rho/pi, replace `state_c_prime[5]` with scalar `d`, in-place chi with 2 temps per row

**`keccakf_perm.cu`:**
- Phase 1 kernel (`keccakf_perm_phase1`): one thread per permutation computes all 24 keccak-f rounds, stores 25-lane u64 round-input state to a scratch buffer (~4.8 KB/permutation)
- Phase 2 kernel (`keccakf_perm_phase2`): one thread per row loads round state from scratch, writes all 2634 trace columns with coalesced stores
- Remove `initial_state[5][5]`, `fill_zero`, and prev-row global readback
- Unify valid/dummy code paths (dummy rows use zero-initialized state)

**`cuda_abi.rs` / `mod.rs`:**
- Plumb scratch buffer (`DeviceBuffer<u64>`) through FFI to the CUDA kernels
- Scratch overhead is ~1.9% of the trace matrix (4800 bytes/perm for scratch vs 252864 bytes/perm for the trace), freed after tracegen completes

[Reth benchmark comparison](https://github.com/axiom-crypto/openvm-eth/actions/runs/23360309931)

Resolves INT-6958

### DIFF
--- a/extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh
+++ b/extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh
@@ -82,6 +82,54 @@ template <typename T> struct KeccakCols {
 
 inline constexpr size_t NUM_KECCAK_COLS = sizeof(KeccakCols<uint8_t>);
 
+// Apply one keccak-f round in-place without trace writes.
+// Used by phase 1 to advance state between rounds.
+static __device__ __noinline__ void apply_round_in_place(
+    uint32_t round,
+    uint64_t current_state[5][5]
+) {
+    // Theta: C[x] = xor(A[x, 0], A[x, 1], A[x, 2], A[x, 3], A[x, 4])
+    uint64_t state_c[5];
+#pragma unroll 5
+    for (auto x = 0; x < 5; x++) {
+        state_c[x] = current_state[0][x] ^ current_state[1][x] ^ current_state[2][x] ^
+                     current_state[3][x] ^ current_state[4][x];
+    }
+
+    // Theta: A'[x, y] = A[x, y] ^ D[x], where D[x] = C[x-1] ^ ROT(C[x+1], 1)
+    for (int x = 0; x < 5; x++) {
+        uint64_t d = state_c[(x + 4) % 5] ^ ROTL64(state_c[(x + 1) % 5], 1);
+#pragma unroll 5
+        for (int y = 0; y < 5; y++) {
+            current_state[y][x] ^= d;
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


shuklaayush:
@codex review

jonathanpwang:
lint failing

@gaxiom can review, but also I think it is likely fine given both claude & codex approved

shuklaayush:
> lint failing

that's because of `cargo audit` which will be addressed in another pr

shuklaayush:
@codex review

extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh
```diff
@@ -69,6 +82,54 @@ template <typename T> struct KeccakCols {
 
 inline constexpr size_t NUM_KECCAK_COLS = sizeof(KeccakCols<uint8_t>);
 
+// Apply one keccak-f round in-place without trace writes.
+// Used by phase 1 to advance state between rounds.
+static __device__ __noinline__ void apply_round_in_place(
```
gaxiom, line 87: Seems like there is a lot of code duplication with `generate_trace_row_for_round`.  Is it possible to make "apply_round" be templated?

gaxiom (approved):
Approved, just minor thing of code duplication

extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh
shuklaayush: let me do it in a follow-up. i wasn't able to come up with a clean way to deduplicate this

PR #2604, opened 2026-03-23
chore: remove VmCommittedExe struct

stephenh-axiom-xyz:
Resolves INT-4505. Removes `VmCommittedExe`, which is currently a test-only utility.


shuklaayush (approved):
lgtm

PR #2606, opened 2026-03-23
chore: add debug_assert to eval_in_uni_assigned

stephenh-axiom-xyz:
Resolves INT-6945. It seems like `stark-backend` uses `wrapping_add_signed` though: https://github.com/openvm-org/stark-backend/blob/127b44b09f32a19f3820fe661b82f69c5eaad93a/crates/stark-backend/src/poly_common.rs#L99

### DIFF
--- a/crates/static-verifier/src/stages/stacked_reduction/mod.rs
+++ b/crates/static-verifier/src/stages/stacked_reduction/mod.rs
@@ -75,6 +75,7 @@ fn eval_in_uni_assigned(
     n: isize,
     z: BabyBearExtWire,
 ) -> BabyBearExtWire {
+    debug_assert!(n >= -(l_skip as isize));
     if n.is_negative() {
         let z_pow = ext_chip.pow_power_of_two(ctx, z, l_skip.wrapping_add_signed(n));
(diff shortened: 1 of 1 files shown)


crates/static-verifier/src/stages/stacked_reduction/mod.rs
```diff
@@ -75,6 +75,7 @@ fn eval_in_uni_assigned(
     n: isize,
     z: BabyBearExtWire,
 ) -> BabyBearExtWire {
+    debug_assert!(n >= -(l_skip as isize));
     if n.is_negative() {
         let z_pow = ext_chip.pow_power_of_two(ctx, z, l_skip.wrapping_add_signed(n));
```
Golovanov399, line 80: Please not wrapping, unchecked or simple + or something

Golovanov399 (approved):
Looks good except comment

crates/static-verifier/src/stages/stacked_reduction/mod.rs
stephenh-axiom-xyz: Discussed offline, leaving this to match `stark-backend` but will revisit

PR #2607, opened 2026-03-23
fix(v2-rc.1): load deferral as elems as felts in memory kernels

shuklaayush:
Restore DEFERRAL_AS field-element handling in CUDA memory kernels (boundary.cu, inventory.cu). Without this, initial memory values are misread as individual bytes instead of 4-byte Montgomery-encoded field elements, breaking LogUp bus interactions


stephenh-axiom-xyz (approved):
LGTM

PR #2609, opened 2026-03-23
fix: static verifier shouldn't support 0 interactions

stephenh-axiom-xyz:
Resolves INT-6947, INT-6977. Recommended to review without whitespace changes


zlangley (approved):
LGTM

Golovanov399 (approved):
Thanks!

PR #2610, opened 2026-03-23
chore: remove dupe horner eval function in static-verifier

stephenh-axiom-xyz:
Resolves INT-6973. Note the repo now uses `shared_math.rs` throughout


Golovanov399 (approved):
I approve

Note for the future: maybe move the other functions around during a refactor, whatever feels right at the moment

PR #2612, opened 2026-03-23
fix: constraint DagCommitSubAir public value and change AIR_ID constant

jonathanpwang:
closes INT-6987 INT-6986

- Constrains the root circuit's DagCommitSubAir's public values to equal the constant expected onion hash.
- Add `USER_PVS_COMMIT_AIR_ID`

~~Note that currently it seems not straightforward to computing the onion hash value in a standalone way, besides generating a dummy proof or constructing the root prover. We should add a standalone method that can be used to separate the functionalities needed for keygen versus proving.~~ Added a helper function.

### DIFF
--- a/crates/static-verifier/src/circuit.rs
+++ b/crates/static-verifier/src/circuit.rs
@@ -7,11 +7,14 @@ use halo2_base::{
     gates::circuit::builder::BaseCircuitBuilder, halo2_proofs::halo2curves::bn256::Fr, Context,
 };
 use itertools::Itertools;
-use openvm_recursion_circuit::batch_constraint::expr_eval::DagCommitPvs;
+use openvm_recursion_circuit::{
+    batch_constraint::expr_eval::DagCommitPvs,
+    system::{VerifierConfig, VerifierSubCircuit, VerifierTraceGen},
+};
 use openvm_stark_sdk::{
     config::{
         baby_bear_bn254_poseidon2::BabyBearBn254Poseidon2Config as RootConfig,
-        baby_bear_poseidon2::Digest as InnerDigest,
+        baby_bear_poseidon2::{BabyBearPoseidon2Config, Digest as InnerDigest},
     },
     openvm_stark_backend::{
         keygen::types::{MultiStarkVerifyingKey, MultiStarkVerifyingKey0},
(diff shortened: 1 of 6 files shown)


crates/static-verifier/src/circuit.rs
```diff
@@ -245,3 +268,24 @@ impl StaticVerifierCircuit {
         StaticVerifierPvs::from_slice(&pvs_fr)
     }
 }
+
+pub fn compute_dag_onion_commit(
```
jonathanpwang, line 272: @stephenh-axiom-xyz please check this is correct

crates/static-verifier/src/circuit.rs
```diff
@@ -245,3 +268,24 @@ impl StaticVerifierCircuit {
...
+    internal_recursive_vk: &MultiStarkVerifyingKey<BabyBearPoseidon2Config>,
+) -> InnerDigest {
+    // Note: the MAX_NUM_PROOFS const generic does not impact the build_cached_trace_record function
+    // used internally below, but we use the default 3
+    let verifier_circuit = VerifierSubCircuit::<3>::new_with_options(
```
stephenh-axiom-xyz: As mentioned in slack, I think this should probably be 1 to match the root circuit

crates/static-verifier/tests/real_prover_roundtrip.rs
```diff
@@ -93,7 +92,7 @@ fn real_prover_keygen_prove_verify_roundtrip() {
     let (_vk_prove, proof_prove) = FibFixture::new(1, 1, 1 << 5).keygen_and_prove(&engine);
 
     let log_heights_per_air = log_heights_per_air_from_proof(&proof_keygen);
-    let circuit = StaticVerifierCircuit::try_new(vk, [Bn254Scalar::ZERO], &log_heights_per_air)
+    let circuit = StaticVerifierCircuit::try_new(vk, InnerDigest::default(), &log_heights_per_air)
```
stephenh-axiom-xyz, line 95: I did not know there was a `Default` implementation for this lol

stephenh-axiom-xyz (approved):
LGTM, after changing `MAX_NUM_CHILDREN` to 1 + looking at the test issue Claude reported

crates/static-verifier/src/circuit.rs
```diff
@@ -218,6 +227,20 @@ impl StaticVerifierCircuit {
                 .all(|commits| commits.is_empty()),
             "RootVerifierCircuit has no cached trace"
         );
+        profiler.push("pin_dag_onion_commit", ctx.advice.len());
+        let &DagCommitPvs::<_> {
+            commit: onion_commit,
+        } = proof_wire.public_values[CONSTRAINT_EVAL_AIR_ID]
```
stephenh-axiom-xyz, line 233: I think it's a little strange that we don't do this in `public_values.rs`, but not blocking

PR #2613, opened 2026-03-23
fix(static-verifier): remove redundant reduce

shuklaayush:
`reduce_assigned_limb_to_babybear` already reduces all samples


zlangley (approved):
Ah, I did this on `feat/sdk-evm` too, but not sure when we will merge that.

PR #2614, opened 2026-03-24
fix(cuda): bn254 prover perf

gaxiom:
Closes INT-6989

### DIFF
--- a/crates/continuations/src/prover/deferral/hook/mod.rs
+++ b/crates/continuations/src/prover/deferral/hook/mod.rs
@@ -5,12 +5,9 @@ use openvm_recursion_circuit::system::{AggregationSubCircuit, VerifierConfig, Ve
 use openvm_stark_backend::{
     keygen::types::{MultiStarkProvingKey, MultiStarkVerifyingKey},
     proof::Proof,
-    prover::{
-        CommittedTraceData, DeviceDataTransporter, DeviceMultiStarkProvingKey, ProverBackend,
-    },
+    prover::{CommittedTraceData, DeviceMultiStarkProvingKey, ProverBackend},
     StarkEngine, SystemParams,
 };
-#[cfg(feature = "cuda")]
 use openvm_stark_sdk::config::baby_bear_poseidon2::{BabyBearPoseidon2CpuEngine, DuplexSponge};
 use openvm_stark_sdk::config::baby_bear_poseidon2::{Digest, EF, F};
 use p3_field::{Field, PrimeField32};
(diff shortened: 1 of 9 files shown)


crates/continuations/src/prover/deferral/hook/mod.rs
```diff
@@ -105,8 +105,12 @@ impl<
             Arc::new(verifier_circuit),
             internal_recursive_dag_commit,
         ));
-        let (pk, vk) = engine.keygen(&circuit.airs());
-        let d_pk = engine.device().transport_pk_to_device(&pk);
+        let airs = circuit.airs();
+        let (pk, vk) = keygen_for_proving_backend(&engine, &airs, || {
```
jonathanpwang: why is this needed everywhere? `engine.keygen` currently internally uses cpu anyways as the default

jonathanpwang (approved):
I don't understand why `keygen_for_proving_backend` is necessary, but I will approve first so we resolve perf regression

jonathanpwang:
claude's comments are incorrect - I don't see what the cpu keygen has to do with anything?

crates/continuations/src/prover/utils.rs
```diff
@@ -10,6 +14,40 @@ pub(crate) fn debug_checks_enabled() -> bool {
     std::env::var("OPENVM_SKIP_DEBUG") != Ok(String::from("1"))
 }
 
+pub(crate) fn keygen_for_proving_backend<SC, E, Keygen>(
+    engine: &E,
+    airs: &[AirRef<SC>],
+    cpu_keygen: Keygen,
+) -> (MultiStarkProvingKey<SC>, MultiStarkVerifyingKey<SC>)
```
zlangley: This interface is kind of clunky. Is it not possible to make the engine's keygen smarter?

crates/continuations/src/prover/deferral/hook/mod.rs
gaxiom: agree, will clean this

crates/continuations/src/prover/inner/mod.rs
```diff
@@ -115,14 +113,15 @@ impl<
...
         let child_vk_pcs_data = verifier_circuit.commit_child_vk(&engine, &child_vk);
         let circuit = Arc::new(InnerCircuit::new(
             Arc::new(verifier_circuit),
             def_hook_cached_commit.map(|d| d.into()),
         ));
-        let (pk, vk) = engine.keygen(&circuit.airs());
-        let d_pk = engine.device().transport_pk_to_device(&pk);
+        let airs = circuit.airs();
+        let (pk, vk) = engine.keygen(&airs);
+        let d_pk = transport_pk(&engine, &pk);
```
zlangley: Is this functionally any different now?

PR #2624, opened 2026-03-26
fix(v2-beta): race condition in recursion tracegen

shuklaayush:
## Summary

- Fix CUDA stream safety issue: `generate_gpu_proving_ctx` was called inside `par_iter` in 4 modules (batch_constraint, gkr, stacking, whir), causing GPU buffer allocations and H2D transfers to run on rayon worker threads instead of the main/default CUDA stream
- Split all affected sites into two phases: (1) CPU trace generation in parallel via `par_iter`, (2) H2D transfer serially on the main thread via `transport_matrix_h2d_row`
- Delete the now-unused `generate_gpu_proving_ctx` helper from `tracegen.rs`
- Inline H2D transfer in the `_ =>` fallback arms of `ModuleChip<GpuBackend>` impls for stacking and whir

### DIFF
--- a/crates/recursion/src/batch_constraint/mod.rs
+++ b/crates/recursion/src/batch_constraint/mod.rs
@@ -905,7 +905,7 @@ pub mod cuda_tracegen {
             interactions_folding::cuda::InteractionsFoldingBlobGpu,
         },
         cuda::{preflight::PreflightGpu, proof::ProofGpu, vk::VerifyingKeyGpu, GlobalCtxGpu},
-        tracegen::cuda::{generate_gpu_proving_ctx, StandardTracegenGpuCtx},
+        tracegen::cuda::StandardTracegenGpuCtx,
     };
 
     impl ModuleChip<GpuBackend> for BatchConstraintModuleChip {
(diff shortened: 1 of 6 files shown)


crates/recursion/src/batch_constraint/mod.rs
```diff
@@ -1079,23 +1079,34 @@ pub mod cuda_tracegen {
...
+                        chip.generate_trace(
                             &cpu_ctx,
                             required_heights.map(|heights| heights[chip.index()]),
                         ),
                     )
                 })
                 .collect::<Vec<_>>();
 
+            // Phase 2: H2D transfer serially on main thread
+            let indexed_cpu_traces = indexed_cpu_rm_traces
```
jonathanpwang: nit: cpu_traces isn't quite right since this already transferred to gpu

crates/recursion/src/stacking/mod.rs
```diff
@@ -610,22 +615,43 @@ mod cuda_tracegen {
...
+                            chip,
+                            &cpu_ctx,
                             required_heights.map(|heights| heights[chip.index()]),
                         ),
                     )
                 })
                 .collect::<Vec<_>>();
 
+            // Phase 2: H2D transfer serially on main thread
+            let indexed_cpu_ctxs = indexed_cpu_rm_traces
```
jonathanpwang: same nit

jonathanpwang (approved):

PR #2625, opened 2026-03-26
chore: update CLI crate with new SDK

stephenh-axiom-xyz:
Resolves INT-6956.

# PR Summary: Update CLI Crate with New SDK

## Overview

Two commits that update the CLI crate (`cargo-openvm`) to work with the restructured v2 SDK, and fix minor clippy/fmt issues. The main theme is: **CLI commands now use the SDK's internal keygen** (aggregation + root keys are computed on-demand and cached to disk) instead of relying on pre-generated keys at fixed global paths.

---

### `prove` command

- **`prove stark`**: Loads cached agg/root keys from `${target_dir}/openvm/agg.pk` (or `~/.openvm/root.pk`) if present, otherwise lets the SDK generate them. After proving, saves keys via `save_keys()` for future reuse.
- **`prove evm`**: Same key-caching pattern. Adds `--agg-pk` argument (was missing). Calls `prove_evm()` instead of `prove()`. Removes the pre-flight check that required `setup` to be run first — keys are now generated internally.
- New helper functions: `maybe_with_cached_keys()`, `save_keys()`, `target_dir_from_cargo_args()`.

### `commit` command

Previously only computed and saved `app_exe_commit`. Now:
- Computes both `app_exe_commit` and `app_vk_commit` (via new `StarkProver::app_vk_commit()`).
- Saves an `AppExecutionCommit` JSON with both values (hex-encoded with `0x` prefix).
- Saves a **baseline** file (`.baseline.json`) used by `verify stark`.
- Saves agg PK, agg VK, and root PK to disk for reuse by subsequent `prove` commands.

### `verify` command

- **`verify stark`**: Now reads a **baseline file** (`.baseline.json`) instead of an app commit file. Reads `agg.vk` from `${target_dir}/openvm/` instead of `~/.openvm/agg_stark.vk`. Error messages updated to point users to `prove stark` instead of `setup`.
- **`verify evm`**: Improved error message formatting with path display.

### `setup` command

Major simplification (115 lines removed, 53 added):
- No longer generates or persists STARK aggregation keys (`agg_stark.pk`, `agg_stark.vk`, `agg_halo2.pk`, `root.asm`). Those are now handled per-project by `prove`/`commit`.
- Now **only** generates the EVM Halo2 verifier Solidity contracts (requires `solc`).
- Versioned output directory: `~/.openvm/halo2/src/v{VERSION}/`.
- Replaces `--evm` and `--force-agg-keygen` flags with a single `--force` flag.
- Always downloads Halo2 params and requires `solc`.

### `default.rs`

Removed global-path defaults for `agg_stark.pk`, `agg_stark.vk`, `agg_halo2.pk`, `root.asm`. Remaining:
- `default_params_dir()` → `~/.openvm/params/`
- `default_root_pk_path()` → `~/.openvm/root.pk` (new)
- `default_evm_halo2_verifier_path()` → `~/.openvm/halo2/`

### `util.rs`

New path helpers:
- `get_openvm_dir()`, `get_agg_pk_path()`, `get_agg_vk_path()`, `get_app_baseline_path()`
- Removed `read_default_agg_and_halo2_pk()`.

### Dependencies

- Added `openvm-continuations` (with `root-prover` feature) and `p3-bn254` to `Cargo.toml`.
- Re-enabled `evm-verify` in default features (was temporarily disabled with a TODO).

---

## 2. CI (`cli.yml`)

Uncommmented the `pull_request` trigger and narrowed its path filter to just `crates/cli/**` and the workflow file itself (previously listed sdk, vm, toolchain, etc.).

---

## 3. Integration Tests → Shell Scripts

All 6 integration tests were refactored from inline Rust `Command` calls to delegated bash scripts in `tests/scripts/`:

| Test | Script |
|------|--------|
| `test_cli_app_e2e` | `cli_app_e2e.sh` |
| `test_cli_app_e2e_simplified` | `cli_app_e2e_simplified.sh` |
| `test_cli_stark_e2e_simplified` | `cli_stark_e2e_simplified.sh` |
| `test_cli_init_build` | `cli_init_build.sh` |
| `test_cli_run_mode_pure_default` | `cli_run_mode_pure_default.sh` |
| `test_cli_run_segment` | `cli_run_segment.sh` |
| `test_cli_run_meter` | `cli_run_meter.sh` |

Rust side now uses a single `run_script()` helper. Removed `replace_with_local_openvm()` (moved into `cli_init_build.sh` as a `perl` one-liner). Removed `tempfile`, `itertools`, `fs` imports that are no longer needed.

---

### `keygen/mod.rs`

- New `RootProvingKey` struct (serializable): holds `root_pk: Arc<MultiStarkProvingKey<RootSC>>` and `trace_heights: Vec<usize>`.
- `Halo2ProvingKey` now derives `Serialize, Deserialize` (was only `Clone`).

### `lib.rs`

New methods on `Sdk`:
- `with_agg_pk()` — inject a pre-computed aggregation proving key
- `with_root_pk()` — inject a pre-computed root proving key
- `agg_keygen()` — convenience to get `(AggProvingKey, MultiStarkVerifyingKey)`
- `root_pk()` — extract root proving key for serialization

### `prover/stark.rs`

New `app_vk_commit()` method on `StarkProver`: computes a Poseidon2 hash over the DAG commits (cached_commit + vk_pre_hash) from the leaf, internal-for-leaf, and internal-recursive layers. This is the `app_vk_commit` exposed by the `commit` CLI command.

### `types.rs`

- `AppExecutionCommit` no longer gated behind `#[cfg(feature = "evm-prove")]` — available unconditionally.
- Renamed `app_vm_commit` → `app_vk_commit` throughout.
- `hex_bytes32` module now public; serialization uses `0x` prefix; deserialization strips it.

### `solidity.rs`

Comment fix: `app_vm_commit` → `app_vk_commit` in instance layout documentation.

---

### Serialization support

Several key types now derive `Serialize, Deserialize` to enable CLI key caching:
- `StaticVerifierCircuit`
- `StaticVerifierProvingKey`
- `Halo2WrapperProvingKey`

### `Halo2ProvingPinning` serde

Custom `Serialize`/`Deserialize` impl added (the Halo2 `ProvingKey` doesn't natively support serde). Serializes PK as raw bytes + metadata. Deserializes using `ProvingKey::from_bytes::<BaseCircuitBuilder<Fr>>()` with saved `config_params`.

### Clippy fixes (commit 2)

- `keygen.rs`: Removed duplicate `use serde::{Deserialize, Serialize}` (was under `#[cfg(feature = "evm-prove")]`, now unconditional).
- `prover.rs`: Unused import cleanup.

---

## Key Behavioral Changes for Reviewers

1. **`cargo openvm setup` is no longer required for STARK proving.** Aggregation and root keys are generated on first `prove`/`commit` and cached in `${target_dir}/openvm/`. `setup` is now only needed for EVM Solidity verifier generation.

2. **Key storage moved from global to per-project.** `agg.pk` and `agg.vk` live in `${target_dir}/openvm/` instead of `~/.openvm/`. Exception: `root.pk` still uses `~/.openvm/root.pk` (shared across projects).

3. **`verify stark` now expects a baseline file** (`.baseline.json`) instead of a commit file, and reads `agg.vk` from the target directory instead of `~/.openvm/`.

4. **`app_vm_commit` renamed to `app_vk_commit`** across SDK types and Solidity interface. This is a breaking change for any external consumers of `AppExecutionCommit` JSON or the EVM proof ABI.


stephenh-axiom-xyz:
@jonathanpwang @shuklaayush have been starting to wonder if we should pair `leaf_pk` and `internal_for_leaf_pk` with `app_pk` more than `internal_recursive_pk`, as they are different per `app_pk`

.github/workflows/cli.yml
```diff
@@ -4,17 +4,11 @@ on:
   push:
     branches: ["main"]
     tags: ["v*"]
-  # pull_request:
-  #   branches: ["**"]
-  #   paths:
-  #     - "crates/circuits/primitives/**"
```
jonathanpwang, line 10: could you keep the commented stuff for now (still commented)? I want to do one cleanup to add some kind of merge queue for more complete integration test

jonathanpwang:
> @jonathanpwang @shuklaayush have been starting to wonder if we should pair `leaf_pk` and `internal_for_leaf_pk` with `app_pk` more than `internal_recursive_pk`, as they are different per `app_pk`

This seems good to also help clarify what depends on what. Let's do it in a follow-up (please make a ticket) and also come up with a good name for this body.

crates/cli/src/commands/prove.rs
```diff
@@ -194,16 +202,14 @@ impl ProveCmd {
...
-                // external agg_pk when the SDK supports it again.
                 let app_config = get_app_config(&mut app_pk, segmentation_args);
-                let sdk = Sdk::new(app_config, AggregationSystemParams::default())?
-                    .with_agg_tree_config(*agg_tree_config)
-                    .with_app_pk(app_pk);
+                let sdk = maybe_with_cached_keys(
+                    Sdk::new(app_config, AggregationSystemParams::default())?
+                        .with_agg_tree_config(*agg_tree_config)
+                        .with_app_pk(app_pk),
+                    agg_pk,
```
luffykai: if agg_pk is not set, and we don't have agg_pk in target_dir, we will still proceed but the sdk is without agg_pk right? That seems undesired, should fail earlier if we know it's missing agg_pk

crates/cli/src/util.rs
```diff
@@ -76,22 +65,39 @@ pub fn get_target_dir(target_dir: &Option<PathBuf>, manifest_path: &PathBuf) ->
 }
 
 pub fn get_target_output_dir(target_dir: &Path, profile: &str) -> PathBuf {
-    target_dir.join("openvm").join(profile).to_path_buf()
+    get_openvm_dir(target_dir).join(profile)
+}
+
+pub fn get_openvm_dir(target_dir: &Path) -> PathBuf {
+    target_dir.parent().unwrap_or(target_dir).join("openvm")
```
luffykai, line 72: maybe `openvm_target` / `target_openvm` as this is analogous to cargo's target?

crates/cli/src/commands/commit.rs
```diff
@@ -78,29 +85,77 @@ impl CommitCmd {
             mode: ExecutionMode::Pure,
         };
         let (exe, target_name_stem) = load_or_build_exe(&run_args, &self.cargo_args)?;
-        let sdk =
+        let (manifest_path, _) = get_manifest_path_and_dir(&self.cargo_args.manifest_path)?;
```
luffykai, line 88: i feel this part is still doing too much "conditional load", but since commit is going to be in prove/keygen, we can address later

crates/cli/src/commands/prove.rs
```diff
@@ -321,6 +327,47 @@ fn get_app_config(
...
+
+fn with_required_agg_pk(
+    sdk: Sdk,
+    agg_pk: &Option<PathBuf>,
+    cargo_args: &RunCargoArgs,
+) -> Result<Sdk> {
+    let agg_pk_path = resolve_agg_pk_path(agg_pk, cargo_args)?;
+    let agg_pk = read_object_from_file(&agg_pk_path).map_err(|e| {
+        eyre!(
+            "Failed to read aggregation proving key from {}: {e}\nRun 'cargo openvm commit' first to generate it",
```
luffykai: run keygen, not commit?

crates/cli/src/commands/verify.rs
```diff
@@ -174,18 +169,21 @@ impl VerifyCmd {
...
             } => {
-                let agg_vk = read_object_from_file(default_agg_stark_vk_path())
-                    .map_err(|e| {
-                        eyre::eyre!(
-                        "Failed to read aggregation STARK verifying key: {e}\nPlease run 'cargo openvm setup' first",
+                let (manifest_path, _) = get_manifest_path_and_dir(&cargo_args.manifest_path)?;
+                let target_dir = get_target_dir(&cargo_args.target_dir, &manifest_path);
+                let agg_vk_path = get_agg_vk_path(&target_dir);
+                let agg_vk = read_object_from_file(&agg_vk_path).map_err(|e| {
+                    eyre::eyre!(
```
luffykai, line 180: maybe should be a util function to just get `agg_vk`? we are also doing similar things below (calling a lot of get_

luffykai (approved):

crates/cli/src/util.rs
stephenh-axiom-xyz: Am going to leave for now, we can reconsider. I think it may be best if we have some distance from target though, as we want it to be tracked by git

crates/cli/src/commands/verify.rs
stephenh-axiom-xyz: Will keep this in mind for the cleanup

PR #2626, opened 2026-03-26
fix(v2-beta): race condition in recursion preflight

shuklaayush:
## Summary

Fixes CUDA race condition in recursion preflight where `run_preflight` was called from parallel threads via `thread::scope`, each launching `merkle_precomputation_hash_vectors` GPU kernels on different CUDA streams.

Extracted `apply_merkle_precomputation` from `run_preflight` so that:
- **CPU**: merkle precomputation runs inline within parallel `thread::scope` (pure CPU, safe)
- **CUDA**: merkle precomputation runs serially after `thread::scope` on the main thread

### DIFF
--- a/crates/recursion/src/system/mod.rs
+++ b/crates/recursion/src/system/mod.rs
@@ -620,6 +620,8 @@ impl<const MAX_NUM_PROOFS: usize> VerifierSubCircuit<MAX_NUM_PROOFS> {
     }
 
     /// Runs preflight for a single proof.
+    /// When `cuda` is enabled, merkle precomputation is excluded here and must
+    /// be applied separately via a serial call to avoid cross-thread kernel launches.
     #[tracing::instrument(name = "execute_preflight", skip_all)]
     pub fn run_preflight<TS>(
         &self,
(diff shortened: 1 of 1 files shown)


crates/recursion/src/system/mod.rs
```diff
@@ -646,6 +648,21 @@ impl<const MAX_NUM_PROOFS: usize> VerifierSubCircuit<MAX_NUM_PROOFS> {
...
+    }
+
+    /// Computes merkle precomputation (hashing opened rows) and writes results
+    /// into the preflight. When `cuda` is enabled, this launches GPU kernels on
+    /// the default stream and must not be called from parallel threads.
+    pub fn apply_merkle_precomputation(
+        proof: &Proof<BabyBearPoseidon2Config>,
+        preflight: &mut Preflight,
+    ) {
         #[cfg(feature = "cuda")]
```
jonathanpwang, line 677: this is now never used, delete

crates/recursion/src/system/mod.rs
```diff
@@ -646,6 +648,21 @@ impl<const MAX_NUM_PROOFS: usize> VerifierSubCircuit<MAX_NUM_PROOFS> {
...
+        #[cfg(not(feature = "cuda"))]
+        Self::apply_merkle_precomputation(proof, &mut preflight);
+
+        preflight
+    }
+
+    /// Computes merkle precomputation (hashing opened rows) and writes results
+    /// into the preflight. When `cuda` is enabled, this launches GPU kernels on
+    /// the default stream and must not be called from parallel threads.
+    pub fn apply_merkle_precomputation(
```
jonathanpwang, line 673: this whole thing should be `#[cfg(not(feature = "cuda"))]` now

jonathanpwang (approved):
good find

PR #2627, opened 2026-03-27
feat: use new MultiFieldChallenger in static verifier

zlangley:
Corresponds to https://github.com/openvm-org/stark-backend/pull/311

### DIFF
--- a/crates/static-verifier/src/transcript/mod.rs
+++ b/crates/static-verifier/src/transcript/mod.rs
@@ -1,16 +1,16 @@
+use core::{array, iter};
 use std::sync::OnceLock;
 
 use halo2_base::{
-    gates::{range::RangeChip, GateInstructions, RangeInstructions},
+    gates::{GateInstructions, RangeInstructions},
     halo2_proofs::arithmetic::Field,
     utils::{biguint_to_fe, fe_to_biguint},
-    AssignedValue, Context,
+    AssignedValue, Context, QuantumCell,
 };
-use itertools::Itertools;
 use num_bigint::BigUint;
 use openvm_stark_sdk::{
     config::baby_bear_bn254_poseidon2::{Bn254Scalar, Digest as RootDigest},
-    openvm_stark_backend::p3_field::PrimeField,
+    openvm_stark_backend::p3_field::{Field as P3Field, PrimeField},
 };
 
 use crate::{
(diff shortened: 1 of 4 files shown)


jonathanpwang:
@zlangley is this one ready for review?

crates/static-verifier/src/transcript/mod.rs
```diff
@@ -19,147 +19,267 @@ use crate::{
...
-    input_buffer: Vec<BabyBearWire>,
-    output_buffer: Vec<BabyBearWire>,
+fn base_baby_bear_decomp_bounds() -> &'static BaseBabyBearDecompBounds {
+    static BOUNDS: OnceLock<BaseBabyBearDecompBounds> = OnceLock::new();
+    BOUNDS.get_or_init(|| {
+        let p = BigUint::from(BABY_BEAR_MODULUS_U64);
+        let one = BigUint::from(1u64);
+        let modulus = <Bn254Scalar as P3Field>::order();
+        let modulus_minus_one = &modulus - &one;
+        let pow7 = p.pow(7);
```
jonathanpwang: should 7 be NUM_SAMPLES_PER_WORD?

crates/static-verifier/src/transcript/mod.rs
```diff
@@ -19,147 +19,267 @@ use crate::{
...
+    let output_digits = array::from_fn(|idx| {
+        let digit = ctx.load_witness(biguint_to_fe(&output_digits_big[idx]));
+        range.check_less_than_safe(ctx, digit, BABY_BEAR_MODULUS_U64);
+        BabyBearWire {
+            value: digit,
+            max_bits: BABY_BEAR_BITS,
+        }
+    });
+
+    // Load and range-check top quotient.
```
jonathanpwang: this did not range check the top quotient

jonathanpwang: you need base_baby_bear_decomp_bounds

zlangley: Ah, this is confusing. This is constrained in `constrain_base_baby_bear_decomposition`. But it's weird that some constraints are imposed here and others in `constrain_base_baby_bear_decomposition`, will clean up.

crates/static-verifier/src/transcript/mod.rs
```diff
@@ -19,147 +19,267 @@ use crate::{
...
+    top_quotient: AssignedValue<Fr>,
+    bounds: &BaseBabyBearDecompBounds,
+) {
+    let p = BigUint::from(BABY_BEAR_MODULUS_U64);
+    let one = BigUint::from(1u64);
+
+    for digit in output_digits {
+        range.check_less_than_safe(ctx, digit.value, BABY_BEAR_MODULUS_U64);
+    }
+    range.check_less_than_safe(ctx, top_quotient, bounds.top_quotient_max + 1);
```
jonathanpwang: technically we need `top_quotient_max` to be `< u64::MAX`

crates/static-verifier/src/transcript/mod.rs
```diff
@@ -19,147 +19,267 @@ use crate::{
...
+    ctx.constrain_equal(&packed, &recomposed);
+
+    // Boundary check: when top_quotient is at max, lower must not exceed the remainder.
+    let at_top_boundary = gate.is_equal(
+        ctx,
+        top_quotient,
+        QuantumCell::Constant(Fr::from(bounds.top_quotient_max)),
+    );
+    let lower_is_valid = range.is_big_less_than_safe(ctx, lower, bounds.lower_max_plus_one.clone());
+    let lower_is_invalid = gate.sub(ctx, QuantumCell::Constant(Fr::ONE), lower_is_valid);
```
jonathanpwang: there might be a `gate.not` function

crates/static-verifier/src/transcript/mod.rs
```diff
@@ -19,147 +19,267 @@ use crate::{
...
-        self.permute_state(ctx);
-
-        self.output_buffer.clear();
-        for packed in self.sponge_state {
-            let parts = self.split_state_to_babybear(ctx, packed);
-            self.output_buffer.extend(parts);
+    fn flush_observe_buf(&mut self, ctx: &mut Context<Fr>) {
+        if !self.observe_buf.is_empty() {
+            let gate = self.baby_bear.range().gate();
+            let packed = pack_base_2_31(ctx, gate, &self.observe_buf);
```
jonathanpwang, line 284: I didn't think about this in stark-backend, but `pack_base_2_31` is not injective if you allow length to vary, so throughout we are assuming that the schedule, including number of base field elements, is always fixed by the verifier (I think we have repeated this in several forms, but I think it's worth clarifying).

Note: we should add some documentation about this assumption/contract to stark-backend

crates/static-verifier/src/transcript/mod.rs
```diff
@@ -169,27 +289,30 @@ impl TranscriptChip {
...
     }
 
+    /// Absorb digest words directly into the sponge (lossless).
     pub fn observe_commit(&mut self, ctx: &mut Context<Fr>, digest: &DigestWire) {
-        for packed in &digest.elems {
-            let limbs = self.split_state_to_babybear(ctx, *packed);
-            for limb in &limbs {
-                self.observe(ctx, limb);
-            }
+        self.sample_buf.clear();
```
jonathanpwang: nit: if you wanted to exactly parallel stark-backend, add `invalidate_samples` method

crates/static-verifier/src/transcript/mod.rs
```diff
@@ -169,27 +289,30 @@ impl TranscriptChip {
...
+        self.flush_observe_buf(ctx);
+        for &elem in &digest.elems {
+            self.sponge_absorb(ctx, elem);
         }
     }
 
     pub fn sample(&mut self, ctx: &mut Context<Fr>) -> BabyBearWire {
-        if !self.input_buffer.is_empty() || self.output_buffer.is_empty() {
-            self.duplexing(ctx);
+        if let Some(val) = self.sample_buf.pop_front() {
```
jonathanpwang: to save some mental load, can we make this exactly the same as stark-backend (namely `pop` and just `Vec` instead of `VecDeque`)?

jonathanpwang (approved):
LGTM except some minor comments

crates/static-verifier/src/transcript/mod.rs
zlangley: We panic in `base_baby_bear_decomp_bounds` if it's not.

zlangley: Ah sorry, your comment is about strictly less than.

PR #2630, opened 2026-03-31
fix: resolve deferral extension CUDA audit fixes

stephenh-axiom-xyz:
Resolves INT-7151, INT-7154.

## INT-7151: Cap deferral call/output CUDA block sizes

Sets max threads per block to 256 in both the `call.cu` and `output.cu` tracegen kernels.

## INT-7154: Allocate deferral Poseidon2 counts by record capacity

`DeferralPoseidon2ChipGpu` constructor now passes in `max_trace_height` instead of `max_buffer_size`, and uses that to compute the appropriate size for the `records` and `count` buffer. Note that the maximum number of records has been lowered to `max_trace_height` from `2 * max_trace_height`, as we should always estimate that one record corresponds to one row in the poseidon2 trace.

### DIFF
--- a/extensions/deferral/circuit/src/extension/cuda.rs
+++ b/extensions/deferral/circuit/src/extension/cuda.rs
@@ -23,7 +23,7 @@ use crate::{
     call::{DeferralCallAir, DeferralCallChipGpu},
     count::{DeferralCircuitCountAir, DeferralCircuitCountChipGpu},
     output::{DeferralOutputAir, DeferralOutputChipGpu},
-    poseidon2::{poseidon2_buffer_capacity, DeferralPoseidon2Air, DeferralPoseidon2ChipGpu},
+    poseidon2::{DeferralPoseidon2Air, DeferralPoseidon2ChipGpu},
     DeferralExtension, Rv32DeferralConfig,
 };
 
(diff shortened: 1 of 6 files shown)


extensions/deferral/circuit/src/extension/cuda.rs
```diff
@@ -68,7 +61,12 @@ impl VmProverExtension<GpuBabyBearPoseidon2Engine, DenseRecordArena, DeferralExt
         inventory.add_periphery_chip(count_chip);
 
         inventory.next_air::<DeferralPoseidon2Air<CudaF>>()?;
-        let poseidon2_chip = Arc::new(DeferralPoseidon2ChipGpu::new(poseidon2_capacity, 1));
+        let max_trace_height = inventory
+            .config()
+            .segmentation_config
+            .limits
+            .max_trace_height as usize;
+        let poseidon2_chip = Arc::new(DeferralPoseidon2ChipGpu::new(max_trace_height.max(1), 1));
```
shuklaayush, line 69: huh, why are we allocating max possible capacity here? shouldn't this be coming from metered execution or something

shuklaayush: hmm okay, this is the deferral circuit which i guess has no notion of metered execution

stephenh-axiom-xyz: The other chips need a reference to this, and `extend_prover` is called at the beginning

shuklaayush (approved):

PR #2635, opened 2026-03-31
chore: match cpu/gpu program air tracegen for empty program

shuklaayush:
## Summary

- Fix GPU cached program trace generation to match CPU behavior for empty or small programs
- Replace `next_power_of_two_or_zero(num_records)` with `num_records.next_power_of_two()` so the minimum trace height is 1 instead of 0

## Motivation

The CPU path (`generate_cached_trace` in `trace.rs`) pads instructions with `while !len.is_power_of_two()`, which rounds 0 up to 1 and emits a TERMINATE padding row. The GPU path used `next_power_of_two_or_zero`, which returns 0 for empty input — producing a zero-height trace with no TERMINATE row. This is a CPU/GPU semantic mismatch.

`next_power_of_two()` returns 1 for input 0, matching the CPU behavior. The CUDA kernel already fills padding rows with TERMINATE + EXIT_CODE_FAIL, so both height and content now agree.

## Changes

- `crates/vm/src/system/cuda/program.rs`: Use `num_records.next_power_of_two()` instead of `next_power_of_two_or_zero(num_records)`, remove unused import

Closes INT-7167

### DIFF
--- a/crates/vm/src/system/cuda/program.rs
+++ b/crates/vm/src/system/cuda/program.rs
@@ -1,8 +1,6 @@
 use std::{mem::size_of, sync::Arc};
 
-use openvm_circuit::{
-    primitives::Chip, system::program::ProgramExecutionCols,
-};
+use openvm_circuit::{primitives::Chip, system::program::ProgramExecutionCols};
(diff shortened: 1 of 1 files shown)


stephenh-axiom-xyz (approved):
LGTM but pls fix lint

PR #2636, opened 2026-03-31
fix: cuda test and add a comment about a potential race condition

Golovanov399:
This resolves INT-7146

Not sure about the target branch


jonathanpwang (approved):
lgtm but let's merge mine first

PR #2639, opened 2026-03-31
feat: mod-builder + algebra + ecc CUDA tracegen

stephenh-axiom-xyz:


stephenh-axiom-xyz:
Closing for now, will revisit later

PR #2642, opened 2026-03-31
feat: add s3 download option to cargo openvm setup

stephenh-axiom-xyz:
Resolves INT-4925. Note download bucket taken from scripts.

# PR Summary: Add `--download` flag to `cargo openvm setup` for pre-built verifier artifacts

## Overview

This PR adds a `--download` option to the `setup` CLI command that lets users download pre-built EVM Halo2 verifier artifacts from a public S3 bucket instead of generating them locally. Generating the verifier contracts requires significant computation, memory, and a local `solc` installation; the download path bypasses all of that.

All changes are confined to `crates/cli/src/commands/setup.rs`.

---

### New `--download` CLI flag

A new `download: bool` field is added to `SetupCmd` (default `false`). When set, the command downloads four verifier artifacts from S3 rather than invoking `Sdk::generate_halo2_verifier_solidity`.

### `solc` check gated on local generation

The `solc`-installed check is now skipped when `--download` is used (`!check_solc_installed() && !self.download`), since `solc` is only needed for local generation.

### `download_verifier` method

A new `async fn download_verifier(versioned_verifier_dir: &PathBuf) -> Result<()>` method handles the S3 download path:

- Creates the versioned verifier directory and its `interfaces/` subdirectory.
- Connects to S3 (`openvm-public-artifacts-us-east-1` bucket, `us-east-1`, no credentials — public bucket).
- Constructs the S3 key prefix as `v{CARGO_PKG_VERSION}/halo2/src/v{OPENVM_VERSION}/`.
- Downloads four files (skipping any that already exist locally):
  - `EVM_HALO2_VERIFIER_PARENT_NAME` — parent verifier Solidity source
  - `EVM_HALO2_VERIFIER_BASE_NAME` — base verifier Solidity source
  - `EVM_VERIFIER_ARTIFACT_FILENAME` — compiled verifier bytecode artifact
  - `EVM_HALO2_VERIFIER_INTERFACE_NAME` — verifier interface (placed in `interfaces/` subdirectory)
- The S3 key for the interface file includes an `interfaces/` path segment; the other three are at the prefix root.

### New import

`EVM_VERIFIER_ARTIFACT_FILENAME` is added to the import from `openvm_sdk::fs` (the other three constants were already imported).

### `run` method refactored

The body of `run` is split into a conditional: `self.download` calls `download_verifier`, otherwise the existing local-generation path runs as before. The early-return when artifacts already exist on disk still applies to both paths.

### DIFF
--- a/crates/cli/src/commands/setup.rs
+++ b/crates/cli/src/commands/setup.rs
@@ -66,6 +66,9 @@ impl SetupCmd {
             && versioned_verifier_dir
                 .join(EVM_HALO2_VERIFIER_BASE_NAME)
                 .exists()
+            && versioned_verifier_dir
+                .join(EVM_VERIFIER_ARTIFACT_FILENAME)
+                .exists()
             && interface_dir
                 .join(EVM_HALO2_VERIFIER_INTERFACE_NAME)
                 .exists()
@@ -85,7 +88,9 @@ impl SetupCmd {
                 AggregationSystemParams::default(),
             );
 
-            println!("Generating verifier contract...");
+            println!(
+                "Generating verifier contract. Use --download to download pre-generated contract instead."
+            );
             let verifier = sdk.generate_halo2_verifier_solidity()?;
 
             println!("Writing verifier contract to {}", verifier_dir.display());
--- a/crates/cli/src/commands/setup.rs
+++ b/crates/cli/src/commands/setup.rs
@@ -12,6 +12,7 @@ use openvm_sdk::{
     fs::{
         write_evm_halo2_verifier_to_folder, EVM_HALO2_VERIFIER_BASE_NAME,
         EVM_HALO2_VERIFIER_INTERFACE_NAME, EVM_HALO2_VERIFIER_PARENT_NAME,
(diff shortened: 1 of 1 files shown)


jonathanpwang:
@claude do the s3 paths match those in release.yml?

crates/cli/src/commands/setup.rs
```diff
@@ -69,16 +77,20 @@ impl SetupCmd {
...
+        } else {
+            let sdk = Sdk::standard(
+                app_params_with_100_bits_security(MAX_APP_LOG_STACKED_HEIGHT),
+                AggregationSystemParams::default(),
+            );
 
-        println!("Generating verifier contract...");
-        let verifier = sdk.generate_halo2_verifier_solidity()?;
+            println!("Generating verifier contract...");
+            let verifier = sdk.generate_halo2_verifier_solidity()?;
```
jonathanpwang, line 94: if this does the full keygen and is expected to be slow, we can add a message that says "use --download to download pre-generated keys instead" or something along those lines

jonathanpwang (approved):
Address claude's comment and one suggestion on a display message.

PR #2644, opened 2026-03-31
chore: cargo openvm prove generates baseline file

stephenh-axiom-xyz:
Resolves INT-5362.

# PR Summary: `prove stark` writes baseline so `commit` is not required

## Overview

Previously, `cargo openvm prove stark` did not write a verification baseline file. This meant `cargo openvm verify stark` would fail unless the user had separately run `cargo openvm commit` first, since `verify stark` reads a `.baseline.json` file to check the proof against expected commits.

This PR makes `prove stark` self-sufficient by having it generate and write the baseline file itself, enabling a simpler `keygen -> prove stark -> verify stark` workflow without requiring `commit`.

---

### Baseline generation and commit logging

After generating the baseline (which was already done for its `app_exe_commit`), the STARK proving path now also:

1. Computes `app_vk_commit` from the prover.
2. Constructs an `AppExecutionCommit` and converts both the exe and vk commits to `Bn254` for display.
3. Prints both `exe commit` and `vk commit` (previously only the exe commit was printed, and as raw bytes rather than `Bn254`).
4. Converts the baseline to `VerificationBaselineJson` (the hex-wrapped format) and writes it to `<target_output_dir>/<target_name>.baseline.json` via `write_to_file_json` — the same path and format that `verify stark` expects.

This mirrors what the `commit` command already does, adapting it for the prove-without-commit flow.

### New imports

- `openvm_continuations::CommitBytes`, `openvm_sdk::types::{AppExecutionCommit, VerificationBaselineJson}`, `p3_bn254::Bn254` — for commit formatting and baseline serialization.
- `get_app_baseline_path`, `get_target_output_dir` from `crate::util` — for resolving the baseline output path.

---

### New integration test: `test_cli_stark_e2e_no_commit`

Added in `integration.rs`, this test exercises the new `keygen -> prove stark -> verify stark` flow (no `commit` step). It calls the new script `cli_stark_e2e_no_commit.sh`.

### New test script: `cli_stark_e2e_no_commit.sh`

A bash script that:

1. Runs `cargo openvm keygen` against the `tests/programs/multi` workspace.
2. Runs `cargo openvm prove stark` for the `fibonacci` example, outputting the proof to a temp directory.
3. Runs `cargo openvm verify stark` for the same example, reading the proof from the temp path.

The temp directory is cleaned up on exit. The script validates that the full STARK prove/verify cycle works without an explicit `commit` step.

### DIFF
--- a/crates/cli/src/commands/prove.rs
+++ b/crates/cli/src/commands/prove.rs
@@ -8,22 +8,24 @@ use openvm_circuit::arch::{
     },
     instructions::exe::VmExe,
 };
+use openvm_continuations::CommitBytes;
 use openvm_sdk::{
     config::{AggregationSystemParams, AggregationTreeConfig, AppConfig},
     fs::{read_object_from_file, write_object_to_file, write_to_file_json},
     keygen::AppProvingKey,
-    types::VersionedNonRootStarkProof,
+    types::{AppExecutionCommit, VerificationBaselineJson, VersionedNonRootStarkProof},
     Sdk, F,
 };
 use openvm_sdk_config::SdkVmConfig;
+use p3_bn254::Bn254;
 
 use super::{RunArgs, RunCargoArgs};
 use crate::{
     commands::build,
     input::read_to_stdin,
     util::{
-        get_agg_pk_path, get_app_pk_path, get_manifest_path_and_dir, get_single_target_name,
-        get_target_dir,
+        get_agg_pk_path, get_app_baseline_path, get_app_pk_path, get_manifest_path_and_dir,
+        get_single_target_name, get_target_dir, get_target_output_dir,
     },
 };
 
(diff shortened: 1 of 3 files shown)


crates/cli/src/commands/prove.rs
```diff
@@ -212,12 +214,30 @@ impl ProveCmd {
...
 
                 let (stark_proof, _metadata) =
                     prover.prove(read_to_stdin(&run_args.input)?, &[])?;
                 let stark_proof_bytes = VersionedNonRootStarkProof::new(stark_proof)?;
 
+                let target_dir = target_dir_from_cargo_args(cargo_args)?;
+                let target_output_dir = get_target_output_dir(&target_dir, &cargo_args.profile);
+                let target_name_path =
+                    get_single_target_name(cargo_args).unwrap_or(PathBuf::from(&target_name));
+                let baseline_path = get_app_baseline_path(&target_output_dir, target_name_path);
```
luffykai, line 236: should this function `get_app_baseline_path` should take in `cargo_args` and `target_name` and derive target_x itself? then this 4 lines become 1

luffykai (approved):
do we still need commit command at all?

crates/cli/src/commands/prove.rs
stephenh-axiom-xyz: In `commit` the intermediate values are used elsewhere

PR #2649, opened 2026-04-01
feat: make GenericSdk immutable and introduce builder

stephenh-axiom-xyz:
Resolves INT-6955.

# PR Summary

Base branch: `develop-v2.0.0-beta`

## Overview

This PR makes `GenericSdk` immutable after construction and introduces a dedicated builder for all advanced SDK initialization. The main effect is a phase split:

- `GenericSdkBuilder` owns construction-time choices such as config vs pre-generated keys, transpiler setup, aggregation tree config, deferral prover setup, root params, and Halo2 setup.
- `GenericSdk` becomes a fixed proving environment with only internal lazy caches (`OnceLock`) mutating after `build()`.

The PR also updates CLI, tests, and one benchmark to use the builder instead of mutating an already-constructed `Sdk`.

## Suggested Review Order

1. `crates/sdk/src/builder.rs`
2. `crates/sdk/src/lib.rs`
3. `crates/cli/src/commands/{prove,run,commit}.rs`
4. `crates/sdk/src/tests.rs`
5. `benchmarks/prove/src/bin/keccak_par.rs`

## `crates/sdk`: immutable `GenericSdk` + new builder

Files:

- `crates/sdk/src/builder.rs`
- `crates/sdk/src/lib.rs`

What changed:

- Added `GenericSdkBuilder` as the construction-only API.
- The builder models each stage with a single source of truth:
  - app: `app_config(...)` or `app_pk(...)`
  - aggregation: `agg_params(...)` or `agg_pk(...)`
  - root: `root_params(...)` or `root_pk(..., trace_heights)`
  - Halo2: `halo2_config(...)` or `halo2_pk(...)`
- The builder also accepts:
  - `agg_tree_config(...)`
  - `transpiler(...)` / `default_transpiler()`
  - `deferral_prover(...)`
  - `halo2_params_dir(...)`
- `build()` normalizes config/key sources into the final `GenericSdk`, validates the pre-generated-key dependency chain, and seeds the SDK’s lazy caches from any prebuilt keys.
- `GenericSdk` is now getter-only for config/state fields; mutable getters and setter-style derives were removed.
- `Sdk::new(...)` and `Sdk::new_without_transpiler(...)` now delegate to the builder.
- `root_params` is now stored on `GenericSdk` and used by `root_prover()` instead of re-deriving default root params there.

Important removals from `GenericSdk`:

- `with_deferral_prover`
- `set_transpiler` / `with_transpiler`
- `set_app_pk` / `with_app_pk`
- `with_agg_pk`
- `with_root_pk`
- `set_halo2_params_dir` / `with_halo2_params_dir`

Reviewer focus:

- Verify the builder covers all existing initialization paths that previously relied on post-construction mutation.
- Check the dependency validation for pre-generated keys, especially `halo2_pk -> root_pk -> agg_pk -> app_pk`.
- Check that root prover construction now uses stored `root_params` consistently.
- Confirm that the remaining lazy caches are only internal and do not reintroduce public mutability.

## `crates/cli`: switch callers to builder-based initialization

Files:

- `crates/cli/src/commands/prove.rs`
- `crates/cli/src/commands/run.rs`
- `crates/cli/src/commands/commit.rs`

What changed:

- `prove` now constructs the SDK through the builder for all proof modes instead of building first and then attaching keys/config.
- The app proving key still has its segmentation config adjusted in-place before SDK construction, but that happens before `build()` via `configure_app_pk(...)`.
- STARK and EVM prove paths now explicitly load aggregation/root proving keys before building the SDK, instead of calling `with_agg_pk(...)` / `with_root_pk(...)` afterward.
- `run` now has two clean construction paths:
  - `pure`: config-driven `Sdk::new(...)`
  - `meter` / `segment`: builder path seeded with an existing `app_pk`
- `commit` now incrementally builds the SDK from available proving artifacts:
  - always starts from `app_pk`
  - uses `agg_pk` if present, otherwise falls back to default aggregation params
  - uses `root_pk` if present

Reviewer focus:

- Check that each CLI mode still loads the same artifacts as before.
- Verify that the order of key loading now matches the builder’s invariants.
- Confirm there are no remaining CLI call sites that rely on mutating a built `Sdk`.

## Tests and benchmark updates

Files:

- `crates/sdk/src/tests.rs`
- `benchmarks/prove/src/bin/keccak_par.rs`

What changed:

- The SDK deferral tests now construct deferral-enabled SDKs through the builder.
- The parallel Keccak benchmark now creates per-thread SDKs from a shared `app_pk` through the builder, instead of constructing from config and setting the proving key afterward.

Reviewer focus:

- These changes are mostly migration-only. Verify they preserve the previous setup semantics while using the new construction model.

### DIFF
--- a/benchmarks/prove/src/bin/keccak_par.rs
+++ b/benchmarks/prove/src/bin/keccak_par.rs
@@ -60,7 +60,6 @@ fn main() -> eyre::Result<()> {
                 let sdk = Sdk::builder()
                     .app_pk(app_pk)
                     .agg_params(Default::default())
-                    .default_transpiler()
                     .build()?;
                 let mut prover = sdk.app_prover(exe)?;
                 let proof = prover.prove(stdin)?;
--- a/benchmarks/prove/src/bin/keccak_par.rs
+++ b/benchmarks/prove/src/bin/keccak_par.rs
@@ -50,7 +50,6 @@ fn main() -> eyre::Result<()> {
     run_with_metric_collection("OUTPUT_PATH", || -> eyre::Result<_> {
         let mut handles = vec![];
         for _ in 0..concurrency {
-            let app_config = app_config.clone();
             let app_pk = app_pk.clone();
             let app_vk = app_vk.clone();
             let exe = exe.clone();
@@ -58,9 +57,11 @@ fn main() -> eyre::Result<()> {
             let handle = std::thread::spawn(move || -> eyre::Result<_> {
                 // Sdk uses OnceLock for internal caching and is not Clone/Sync,
                 // so each thread creates its own instance with the shared app_pk.
-                let sdk = Sdk::new(app_config, Default::default())?;
-                sdk.set_app_pk(app_pk)
-                    .map_err(|_| eyre::eyre!("Error setting app pk"))?;
+                let sdk = Sdk::builder()
+                    .app_pk(app_pk)
+                    .agg_params(Default::default())
+                    .default_transpiler()
+                    .build()?;
                 let mut prover = sdk.app_prover(exe)?;
                 let proof = prover.prove(stdin)?;
                 let _ = verify_app_proof::<DefaultStarkEngine>(&app_vk.vk, memory_dims, &proof)?;
--- a/crates/sdk/src/builder.rs
+++ b/crates/sdk/src/builder.rs
@@ -1,0 +1,493 @@
+#[cfg(feature = "evm-prove")]
+use std::path::Path;
+use std::{
+    marker::PhantomData,
+    sync::{Arc, OnceLock},
+};
+
+use eyre::eyre;
+use openvm_circuit::arch::{VmBuilder, VmExecutionConfig, VmExecutor};
+use openvm_continuations::RootSC;
+use openvm_sdk_config::TranspilerConfig;
+use openvm_stark_backend::{keygen::types::MultiStarkProvingKey, StarkEngine, SystemParams};
+use openvm_stark_sdk::config::root_params_with_100_bits_security;
+#[cfg(feature = "evm-prove")]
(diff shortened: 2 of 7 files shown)


benchmarks/prove/src/bin/keccak_par.rs
```diff
@@ -50,17 +50,18 @@ fn main() -> eyre::Result<()> {
...
             let handle = std::thread::spawn(move || -> eyre::Result<_> {
                 // Sdk uses OnceLock for internal caching and is not Clone/Sync,
                 // so each thread creates its own instance with the shared app_pk.
-                let sdk = Sdk::new(app_config, Default::default())?;
-                sdk.set_app_pk(app_pk)
-                    .map_err(|_| eyre::eyre!("Error setting app pk"))?;
+                let sdk = Sdk::builder()
+                    .app_pk(app_pk)
+                    .agg_params(Default::default())
+                    .default_transpiler()
```
jonathanpwang: if it's default, could we make `default_transpiler()` automatic?

stephenh-axiom-xyz: Do we want to support `without_transpiler`?

crates/sdk/src/builder.rs
```diff
@@ -0,0 +1,505 @@
...
+    Pk(AppProvingKey<VC>),
+}
+
+#[allow(clippy::large_enum_variant)]
+enum AggSource {
+    Params(AggregationSystemParams),
+    Pk(AggProvingKey),
+}
+
+struct PrebuiltRootPk {
```
jonathanpwang: should this be in this file or in root keygen?

jonathanpwang (approved):
LGTM except it'd be nice if we could get rid of `.default_transpiler()`

PR #2650, opened 2026-04-01
ci: cache guest build artifacts

jonathanpwang:
# PR Summary: Cache guest build artifacts

## Overview

Guest program builds previously created a fresh `tempdir()` for every invocation, preventing Cargo from caching compiled artifacts (the `riscv32im` std lib, dependencies, etc.). This PR removes the temp directory usage so builds go to the workspace's default `target/` directory, which is automatically picked up by `rust-cache` in CI. It also fixes several CI workflow issues.

### 1. `benchmarks/utils/src/lib.rs` — Remove tempdir in benchmark ELF builds

- **Removed** `tempfile` dependency and `tempdir()` usage from `build_elf_with_path`.
- The function now relies on `build_guest_package` returning the output directory (which defaults to the workspace `target/` when no `target_dir` is set in `GuestOptions`).
- The `target_dir` needed by `guest_methods()` is derived by walking up two parents from the returned output path instead of being explicitly set.

### 2. `crates/toolchain/tests/src/lib.rs` — Remove tempdir in test ELF builds

- **Removed** `tempfile` dependency and `tempdir()` usage from `build_example_program_at_path_with_features`.
- Instead of manually locating the ELF via `get_dir_with_profile`, uses the output directory returned by `build_guest_package` directly (appending the example name).
- Both call sites now handle the updated `build_guest_package` return type (`Result<PathBuf, Option<i32>>`) with a match that also handles the `Err(None)` (build skipped) case.

### 3. `Cargo.lock`, `benchmarks/utils/Cargo.toml`, `crates/toolchain/tests/Cargo.toml`

- Removed `tempfile` dependency from both crates (no longer needed).

### 4. `scripts/clean-guest-builds.sh` (new file)

- Utility script to remove cached guest build artifacts (`target/riscv32im-risc0-zkvm-elf/` and `target/openvm/`) for reclaiming disk space.

### 5. `scripts/pre-push.sh`

- Added a comment pointing users to `clean-guest-builds.sh` for disk space reclamation.

### 6. `.github/workflows/extension-tests.yml`

- **Uncommented** path filters (`crates/circuits/**`, `crates/vm/**`, `crates/toolchain/**`, `Cargo.toml`) that were previously disabled. This restores proper path-based triggering so the workflow runs when upstream crate changes land, not only on `extensions/**` changes.

### 7. `.github/workflows/guest-lib-tests.yml`

- **Uncommented** path filters (same set as above, plus `crates/sdk/guest/fib/**`).
- **Flattened the matrix**: moved `runner` from a separate `platform` axis into each `crate` entry, allowing per-crate runner selection. 
- **Pairing test filter**: added `-E 'not test(fallback)'` to skip fallback tests on CPU runners, and uses `eval` to correctly expand the filter expression.

### DIFF
--- a/.github/workflows/extension-tests.yml
+++ b/.github/workflows/extension-tests.yml
@@ -7,11 +7,11 @@ on:
   pull_request:
     branches: ["**"]
     paths:
-      - "crates/circuits/**"
-      - "crates/vm/**"
-      - "crates/toolchain/**"
+#      - "crates/circuits/**"
+#      - "crates/vm/**"
+#      - "crates/toolchain/**"
       - "extensions/**"
-      - "Cargo.toml"
+#      - "Cargo.toml"
       - ".github/workflows/extension-tests.yml"
 
 concurrency:
--- a/.github/workflows/extension-tests.yml
+++ b/.github/workflows/extension-tests.yml
@@ -7,11 +7,11 @@ on:
   pull_request:
     branches: ["**"]
     paths:
-#     - "crates/circuits/**"
-#     - "crates/vm/**"
-#     - "crates/toolchain/**"
+      - "crates/circuits/**"
+      - "crates/vm/**"
+      - "crates/toolchain/**"
       - "extensions/**"
-#     - "Cargo.toml"
+      - "Cargo.toml"
       - ".github/workflows/extension-tests.yml"
 
 concurrency:
--- a/.github/workflows/guest-lib-tests.yml
+++ b/.github/workflows/guest-lib-tests.yml
@@ -7,14 +7,14 @@ on:
   pull_request:
     branches: ["**"]
     paths:
-      - "crates/circuits/**"
-      - "crates/vm/**"
-      - "crates/toolchain/**"
-      - "extensions/**"
+#      - "crates/circuits/**"
+#      - "crates/vm/**"
+#      - "crates/toolchain/**"
+#      - "extensions/**"
       - "guest-libs/**"
-      - "Cargo.toml"
+#      - "Cargo.toml"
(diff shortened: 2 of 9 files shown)


.github/workflows/extension-tests.yml
```diff
@@ -7,11 +7,11 @@ on:
   pull_request:
     branches: ["**"]
     paths:
-#     - "crates/circuits/**"
-#     - "crates/vm/**"
-#     - "crates/toolchain/**"
+      - "crates/circuits/**"
```
jonathanpwang: revert this to be commented out

.github/workflows/guest-lib-tests.yml
```diff
@@ -7,14 +7,14 @@ on:
   pull_request:
     branches: ["**"]
     paths:
-#     - "crates/circuits/**"
-#     - "crates/vm/**"
-#     - "crates/toolchain/**"
-#     - "extensions/**"
+      - "crates/circuits/**"
```
jonathanpwang: revert this to be commented out as before

jonathanpwang:
not sure why but we don't have 96cpu- runson runner, so we'll keep the ruint test as is

jonathanpwang:
@claude address review comments

zlangley (approved):
LGTM if it works. If the previous behavior was desirable in some settings (local) we can also make this configurable via an env var or something.

jonathanpwang:
> LGTM if it works. If the previous behavior was desirable in some settings (local) we can also make this configurable via an env var or something.

I forget what I ended up choosing, but there's a local cleanup script at least.

PR #2653, opened 2026-04-02
ci(v2-beta): add back e2e benchmarks

shuklaayush:
resolves int-7054


stephenh-axiom-xyz (commented):
Given how much faster benchmarks are to run now, I wonder if we could enable e2e for each existing benchmark? i.e. instead of adding `fib_e2e`, add a feature flag that would switch the existing fib benchmark from `run` to `run_evm`

shuklaayush:
doesn't it still take like 3x the time compared to just prove-stark?

jonathanpwang:
Isn't e2e supposed to run halo2? how come I don't see anything in the markdown report about halo2?

benchmarks/prove/src/lib.rs
```diff
@@ -31,17 +31,13 @@ pub struct BenchmarkCli {
...
+    /// Run full e2e proving (app → aggregation → root → halo2 wrapping)
+    #[arg(long)]
+    pub evm: bool,
+
     /// Whether to execute with additional profiling metric collection
     #[arg(long)]
     pub profiling: bool,
-    /// Directory containing KZG trusted setup files (for e2e halo2 proving)
-    #[arg(long)]
-    pub kzg_params_dir: Option<std::path::PathBuf>,
```
stephenh-axiom-xyz, line 39: Ig for benchmarks we can just always use the default path?

benchmarks/prove/src/lib.rs
```diff
@@ -147,6 +148,25 @@ pub fn run_default_app_benchmark(
...
+) -> eyre::Result<()> {
+    run_with_metric_collection("OUTPUT_PATH", || -> eyre::Result<_> {
+        let exe = VmExe::from_elf(elf, vm_config.transpiler())?;
+        let app_config = AppConfig::new(vm_config, default_bench_app_params());
+        let agg_params = AggregationSystemParams {
+            leaf: leaf_params_with_100_bits_security(),
+            internal: internal_params_with_100_bits_security(),
+        };
+        let sdk = Sdk::new(app_config, agg_params)?;
+        let _proof = sdk.prove_evm(exe, stdin, &[])?;
```
stephenh-axiom-xyz: I feel like we should verify? Or will something fail if the proof isn't valid?

stephenh-axiom-xyz (commented):
Some comments below. I am also not seeing any metrics in the `*_e2e_*` files, they seem to be empty

benchmarks/prove/src/lib.rs
shuklaayush: hmm i didn't add verify because we only have the solidity verifier and the verification implicitly has a `solc` dependency

benchmarks/prove/src/lib.rs
shuklaayush: yeah, but let me add this back

jonathanpwang (approved):

PR #2654, opened 2026-04-02
fix(v2.1): projective partialeq implementation in impl_sw_proj

shuklaayush:
## Summary

- Fix `PartialEq` for projective points in `impl_sw_proj` macro to use mathematically correct equality: two projective points `(X1:Y1:Z1)` and `(X2:Y2:Z2)` are equal iff `X1*Z2 == X2*Z1` and `Y1*Z2 == Y2*Z1`, instead of the derived field-by-field comparison which is incorrect for projective coordinates.

## Changes

- `extensions/ecc/guest/src/weierstrass.rs`: Replaced `derive(PartialEq, Eq)` with a manual `PartialEq` impl that performs cross-multiplication checks, and a manual `Eq` impl.


Tuanlinh12312:
LGTM

PR #2655, opened 2026-04-02
chore: split app-dependent and independent fields in agg pk

stephenh-axiom-xyz:
Resolves INT-7125.

## Overview

This PR splits the aggregation proving key into app-dependent and app-independent pieces, then updates the CLI to cache the app-independent recursion artifacts in `~/.openvm`.

- Local per-target artifacts are now `app.pk`, `app.vk`, and `agg_prefix.pk`.
- Global cached artifacts are now `~/.openvm/internal_recursive.pk`, and `~/.openvm/root.pk` for EVM setup.
- CLI commands that previously consumed `agg.pk` now reconstruct a full `AggProvingKey` from the local prefix key plus the cached global recursive key.
- `cargo openvm keygen` now enforces the CLI invariant that root-circuit shape stays at the default `num_public_values` / `memory_dimensions`.

## Suggested Review Order

1. `crates/sdk/src/keygen/mod.rs`, `crates/sdk/src/lib.rs`, `crates/sdk/src/prover/agg.rs`, `crates/sdk/src/builder.rs`, `crates/sdk/src/prover/deferral/mod.rs`
2. `crates/cli/src/commands/setup.rs`
3. `crates/cli/src/commands/keygen.rs`, `crates/cli/src/commands/prove.rs`, `crates/cli/src/commands/commit.rs`, `crates/cli/src/commands/verify.rs`
4. `crates/cli/src/{args.rs,default.rs,util.rs,bin/cargo-openvm.rs,commands/mod.rs}` and the updated CLI scripts

## SDK: Split `AggProvingKey` into prefix + recursive pieces

Files:
- `crates/sdk/src/keygen/mod.rs`
- `crates/sdk/src/lib.rs`
- `crates/sdk/src/builder.rs`

What changed:
- Introduces `AggPrefixProvingKey { leaf_pk, internal_for_leaf_pk }`.
- Changes `AggProvingKey` to hold:
  - `prefix_pk: AggPrefixProvingKey`
  - `internal_recursive_pk`
- Adds `Sdk::agg_prefix_pk()` and `AggProver::keygen_prefix(...)` so callers can generate only the app-dependent prefix portion.
- Updates SDK builder / prover / deferral call sites to read the new nested layout.

Review focus:
- The new prefix-only keygen path should produce exactly the material needed to rebuild the old full aggregation PK when paired with `internal_recursive_pk`.
- All code that previously read `leaf_pk` / `internal_for_leaf_pk` directly should now go through `prefix_pk`.

## CLI: `setup` now owns the global recursive cache

Files:
- `crates/cli/src/commands/setup.rs`
- `crates/cli/src/bin/cargo-openvm.rs`
- `crates/cli/src/commands/mod.rs`
- `crates/cli/src/default.rs`

What changed:
- `cargo openvm setup` is no longer feature-gated at the command level.
- `setup` always builds the default SDK and writes `~/.openvm/internal_recursive.pk`.
- `setup --evm` additionally:
  - writes `~/.openvm/root.pk`
  - downloads Halo2 params
  - either generates verifier artifacts locally or downloads them with `--download`
- `--download` only applies to verifier artifacts; the recursive/root proving keys are still generated locally.
- Existing verifier artifacts short-circuit unless `--force` is passed.
- Local verifier generation still requires `solc`; if the CLI was built without `evm-verify`, `--evm` must use `--download`.

Review focus:
- `setup` is now the single owner for machine-global recursive/root artifacts.
- STARK-only flows can use `setup` without enabling the EVM verifier feature.

## CLI: `keygen` now emits local app keys + `agg_prefix.pk`

Files:
- `crates/cli/src/commands/keygen.rs`
- `crates/cli/src/args.rs`
- `crates/cli/src/default.rs`
- `crates/cli/src/util.rs`

What changed:
- Renames the local aggregation artifact from `agg.pk` to `agg_prefix.pk`.
- `keygen` writes:
  - `app.pk`
  - `app.vk`
  - `agg_prefix.pk` unless `--app-only`
- Stops emitting `agg.vk`.
- Adds `assert_default_root_shape(...)`, which panics if `openvm.toml` changes:
  - `num_public_values`
  - derived `memory_dimensions`

Review focus:
- This enforces the CLI invariant that global recursion/root setup only supports the default root shape.
- The output-dir copy path mirrors the new artifact names.

## CLI: `prove`, `commit`, and `verify` now compose local + global keys

Files:
- `crates/cli/src/commands/prove.rs`
- `crates/cli/src/commands/commit.rs`
- `crates/cli/src/commands/verify.rs`
- `crates/cli/src/args.rs`

What changed:
- `ProvingKeyArgs` now exposes `--agg-prefix-pk` instead of `--agg-pk`.
- `prove stark`, `prove evm`, and `commit` all use `load_required_agg_pk(...)`, which:
  - loads local `agg_prefix.pk`
  - loads cached `~/.openvm/internal_recursive.pk`
  - reconstructs `AggProvingKey` in memory
- `prove evm` additionally requires cached `~/.openvm/root.pk` and points users to `cargo openvm setup --evm` if it is missing.
- `commit` now uses shared `ProvingKeyArgs` and no longer lazily generates or persists aggregation/root keys.
- `verify stark` now derives the aggregation VK from cached `internal_recursive.pk` instead of reading a local `agg.vk`.

Review focus:
- STARK proving and commit now require `setup` to run first, not just `keygen`.
- Verification is now tied to the global internal-recursive cache rather than per-target aggregation artifacts.

## CLI Scripts

Files:
- `crates/cli/tests/scripts/cli_stark_e2e_no_commit.sh`
- `crates/cli/tests/scripts/cli_stark_e2e_simplified.sh`

What changed:
- Both STARK e2e scripts now run `cargo openvm setup` before `prove` / `commit` so the shared internal-recursive cache exists.

### DIFF
--- a/crates/cli/src/commands/setup.rs
+++ b/crates/cli/src/commands/setup.rs
@@ -1,6 +1,6 @@
 use std::{
     fs::{create_dir_all, write},
-    path::PathBuf,
+    path::{Path, PathBuf},
 };
 
 use aws_config::{defaults, BehaviorVersion, Region};
@@ -10,15 +10,16 @@ use eyre::{eyre, Context, Result};
 use openvm_sdk::{
     config::AggregationSystemParams,
     fs::{
-        write_evm_halo2_verifier_to_folder, EVM_HALO2_VERIFIER_BASE_NAME,
-        EVM_HALO2_VERIFIER_INTERFACE_NAME, EVM_HALO2_VERIFIER_PARENT_NAME,
-        EVM_VERIFIER_ARTIFACT_FILENAME,
+        write_object_to_file, EVM_HALO2_VERIFIER_BASE_NAME, EVM_HALO2_VERIFIER_INTERFACE_NAME,
+        EVM_HALO2_VERIFIER_PARENT_NAME, EVM_VERIFIER_ARTIFACT_FILENAME,
     },
     Sdk, OPENVM_VERSION,
 };
-use openvm_stark_sdk::config::{app_params_with_100_bits_security, MAX_APP_LOG_STACKED_HEIGHT};
 
-use crate::default::{default_evm_halo2_verifier_path, default_params_dir};
+use crate::default::{
+    default_app_config, default_evm_halo2_verifier_path, default_internal_recursive_pk_path,
+    default_params_dir, default_root_pk_path,
+};
 
 /// The maximum value of `k` to download Halo2 KZG trusted setup parameters for. This depends on the
 /// default verifier circuit and wrapper circuit sizes.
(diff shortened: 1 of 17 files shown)


crates/sdk/src/keygen/mod.rs
```diff
@@ -34,9 +34,14 @@ pub struct AppVerifyingKey {
...
 #[derive(Clone, Serialize, Deserialize)]
-pub struct AggProvingKey {
+pub struct AggPrefixProvingKey {
     pub leaf_pk: Arc<MultiStarkProvingKey<SC>>,
     pub internal_for_leaf_pk: Arc<MultiStarkProvingKey<SC>>,
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct AggProvingKey {
+    pub prefix_pk: AggPrefixProvingKey,
```
jonathanpwang: just `prefix` ? so it's `agg_pk.prefix.`

crates/sdk/src/keygen/mod.rs
```diff
@@ -34,9 +34,14 @@ pub struct AppVerifyingKey {
...
     pub internal_recursive_pk: Arc<MultiStarkProvingKey<SC>>,
```
jonathanpwang: just `internal_recursive`

crates/cli/src/commands/setup.rs
```diff
@@ -27,52 +28,58 @@ const MAX_HALO2_VERIFIER_K_FOR_DOWNLOAD: usize = 24;
 #[derive(Parser)]
 #[command(
     name = "setup",
-    about = "Set up for generating EVM proofs. ATTENTION: this requires large amounts of computation and memory. "
+    about = "Set up OpenVM recursive proving artifacts. ATTENTION: this requires large amounts of computation and memory."
```
jonathanpwang: the ATTENTION is only for `--evm` flag now

crates/cli/src/commands/setup.rs
```diff
@@ -27,52 +28,58 @@ const MAX_HALO2_VERIFIER_K_FOR_DOWNLOAD: usize = 24;
...
 
 impl SetupCmd {
     pub async fn run(&self) -> Result<()> {
-        if !Self::check_solc_installed() && !self.download {
-            return Err(eyre!(
-                "solc is not installed, please install solc to continue"
-            ));
+        let sdk = Sdk::new(default_app_config(), AggregationSystemParams::default())?;
+
+        let internal_recursive_pk_path = PathBuf::from(default_internal_recursive_pk_path());
```
jonathanpwang, line 57: not sure if we should note that the non-evm part is always generated, never downloaded

jonathanpwang (approved):
LGTM except the small rename

crates/cli/src/commands/setup.rs
stephenh-axiom-xyz: I think it's pretty fast, also the `download` flag mentions that it's only the EVM stuff downloaded

PR #2665, opened 2026-04-06
feat: v2 reth vk

luffykai:
ref: INT-7291


jonathanpwang (commented):
This does not belong in this repo? openvm-eth is a better place. cc @shuklaayush

sin-strokes (commented):
Thevalley

PR #2666, opened 2026-04-07
chore: derive FLAMEGRAPHS_DIR from BENCH_METRICS_DIR

Ocheretovich:
Avoids duplication of the base directory path by deriving FLAMEGRAPHS_DIR from BENCH_METRICS_DIR


jonathanpwang:
Sorry, we do not accept PRs for minor fixes like this

PR #2670, opened 2026-04-07
perf(cuda): reduce stack/regs/spills in `keccakf_op` and `sha2_main` tracegen

shuklaayush:
## Summary

Reduce CUDA stack usage, register pressure, and spills in `keccakf_op` and `sha2_main` trace generation kernels, inspired by #2524.

### keccakf_op_tracegen
- Rewrote `keccakf_permutation` to operate in-place: eliminated the 200 B `temp[25]` scratch array, walk the rho/pi 24-element cycle with a single saved temp, compute chi in-place with two temps per row, and use a scalar `d` in theta. Stays `__forceinline__` so the state never has to live in local memory.
- Moved the row-fill body into a `__noinline__` helper `fill_keccakf_op_row` so the heavy working set (200 B keccak state union, `MemoryAuxColsFactory`, `BitwiseOperationLookup`, the 50-iteration `mem_helper.fill` loop) lives in the helper's own frame instead of the kernel's.

### sha2_main_tracegen
- Factored the body into a shared `__forceinline__` template `sha2_main_row_body<V>` plus a thin `__noinline__` wrapper `sha2_main_row_outlined<V>`. Uses `if constexpr (V::WORD_BITS > 32)` to route SHA-512 through the outlined wrapper while keeping SHA-256 on the inlined path (SHA-256 was already tight at 16 B stack / 255 regs, and unconditional outlining would regress it).

### ptxas -v results (sm_89)

| Kernel | Before | After |
|---|---|---|
| `keccakf_op_tracegen` | 1656 B stack / 1268 spills / 255 regs | 16 B kernel + 368 B helper / 168 spills / 24 regs |
| `sha2_main_tracegen<512>` | 1256 B stack / 1260 spills / 255 regs | 64 B kernel + 144 B helper / 144 spills / 24 regs |
| `sha2_main_tracegen<256>` | 16 B / 16 spills / 255 regs | unchanged |

[Reth benchmark comparison](https://github.com/axiom-crypto/openvm-eth/actions/runs/24105029878)

### DIFF
--- a/extensions/keccak256/circuit/cuda/include/keccakf_op.cuh
+++ b/extensions/keccak256/circuit/cuda/include/keccakf_op.cuh
@@ -49,35 +49,51 @@ __device__ __forceinline__ uint64_t rotl64(uint64_t x, int n) {
     return n ? ((x << n) | (x >> (64 - n))) : x;
 }
 
-// Compute keccak-f permutation on a 200-byte state
+// Compute keccak-f permutation on a 200-byte state.
+//
+// Uses in-place rho/pi via the 24-element permutation cycle and in-place chi
+// with two temporaries per row, so no scratch array is needed.
 __device__ __forceinline__ void keccakf_permutation(uint64_t state[25]) {
+    using keccak256::RHO_PI_CYCLE_IDX;
+    using keccak256::RHO_PI_CYCLE_ROT;
+
     for (int round = 0; round < 24; round++) {
-        // Theta
+        // Theta: C[x] = xor(A[x, 0..4])
         uint64_t c[5];
+#pragma unroll 5
         for (int x = 0; x < 5; x++) {
             c[x] = state[x] ^ state[x + 5] ^ state[x + 10] ^ state[x + 15] ^ state[x + 20];
         }
+        // A'[x, y] = A[x, y] ^ D[x] where D[x] = C[x-1] ^ ROTL(C[x+1], 1).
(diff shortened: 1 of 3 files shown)


extensions/keccak256/circuit/cuda/include/keccakf_op.cuh
```diff
@@ -49,35 +49,51 @@ __device__ __forceinline__ uint64_t rotl64(uint64_t x, int n) {
     return n ? ((x << n) | (x >> (64 - n))) : x;
 }
 
-// Compute keccak-f permutation on a 200-byte state
+// Compute keccak-f permutation on a 200-byte state.
+//
+// Uses in-place rho/pi via the 24-element permutation cycle and in-place chi
+// with two temporaries per row, so no scratch array is needed.
 __device__ __forceinline__ void keccakf_permutation(uint64_t state[25]) {
```
gaxiom, line 56: `keccakf_op.cuh:56` now carries a second copy of the Keccak round logic that already exists in `extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh:87`. It looks correct, but it does raise drift risk for future bug fixes or tuning. If you want the same performance shape with less maintenance overhead, extract a shared __forceinline__ round-body helper and keep thin __noinline__ wrappers at the call sites, similar to the pattern used in `extensions/sha2/circuit/cuda/src/sha2_main.cu:18`.

gaxiom (approved):
LGTM

PR #2674, opened 2026-04-08
fix(v2-rc1): deduplicate keccakf round code in tracegen kernel

shuklaayush:
## Deduplicate Keccak round logic in CUDA code

Addresses https://github.com/openvm-org/openvm/pull/2670#discussion_r3053633831

### Problem

`keccakf_op.cuh` carried a second copy of the Keccak-f round logic that already existed in `p3_keccakf.cuh`. Both were correct, but maintaining two copies creates drift risk for future bug fixes or tuning.

### Changes

- **`p3_keccakf.cuh`**: Extracted a shared `__forceinline__` helper `keccak256::keccakf_round_body()` that implements a single Keccak-f round on a flat `uint64_t[25]` state. Reduced `apply_round_in_place()` to a thin `__noinline__` wrapper that delegates to it.

- **`keccakf_op.cuh`**: Replaced the 50-line duplicate `keccakf_permutation()` (and its local `rotl64()` helper) with a 5-line loop over the shared round body.

- **`primitives/utils.cuh`**: Replaced the `ROTL64` macro with a proper `__forceinline__` function `rotl64()` for type safety and single-evaluation guarantees.

### Performance

No impact. All helpers are `__forceinline__`, so the compiler produces identical code. The `__noinline__` / `__forceinline__` wrapper pattern (same as `sha2_main.cu`) preserves the existing stack frame topology.

### DIFF
--- a/extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh
+++ b/extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh
@@ -29,9 +29,9 @@ __device__ __constant__ inline uint64_t RC[NUM_ROUNDS] = {
 // In-place rho/pi permutation cycle (24 elements, flat index y*5+x).
 // Follows (x,y) -> (y, 2x+3y mod 5); (0,0) is a fixed point.
 // cycle[i] receives its value from cycle[(i+1) % 24] with the listed rotation.
-__device__ __constant__ inline int RHO_PI_CYCLE_IDX[24] = {1,  6,  9,  22, 14, 20, 2,  12,
-                                                           13, 19, 23, 15, 4,  24, 21, 8,
-                                                           16, 5,  3,  18, 17, 11, 7,  10};
+__device__ __constant__ inline uint32_t RHO_PI_CYCLE_IDX[24] = {1,  6,  9,  22, 14, 20, 2,  12,
+                                                                13, 19, 23, 15, 4,  24, 21, 8,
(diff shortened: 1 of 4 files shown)


extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh
```diff
@@ -36,6 +36,50 @@ __device__ __constant__ inline uint8_t RHO_PI_CYCLE_ROT[24] = {44, 20, 61, 39, 1
                                                                8,  56, 41, 27, 14, 2,  55, 45,
                                                                36, 28, 21, 15, 10, 6,  3,  1};
 
+// Single-round keccak-f body, operating on a flat 25-element state array.
+// Marked __forceinline__ so callers control whether the round gets its own
+// stack frame (__noinline__ wrapper) or folds into a multi-round loop.
+__device__ __forceinline__ void keccakf_round_body(uint64_t *state, int round) {
```
gaxiom: int vs uint32_t mismatch. Is this intend? round can be negative?

gaxiom (approved):
LGTM - just a small nit - not blocking

extensions/keccak256/circuit/cuda/include/p3_keccakf.cuh
shuklaayush: no, it was an oversight. good catch

PR #2679, opened 2026-04-09
fix(execution): metered precompute size for keccak

jonathanpwang:


shuklaayush:
@claude can you go through all other instances of `metered_pre_compute_size` in the code and confirm that they're calculated properly

shuklaayush (approved):

PR #2689, opened 2026-04-09
chore: add sha2 bench program

luffykai:


jonathanpwang (commented):
is it better here or in `benchmarks/guest`?

shuklaayush:
we should probably also create a `benchmarks/prove/src/bin/sha2.rs`

PR #2697, opened 2026-04-10
fix(static-verifier): replace invalid use of is_big_less_than_safe

zlangley:


crates/static-verifier/src/transcript/tests.rs
```diff
@@ -26,9 +26,17 @@ use crate::{
...
+    run_mock_with_lookup_bits(expect_satisfied, 16, build);
+}
+
+fn run_mock_with_lookup_bits(
+    expect_satisfied: bool,
+    lookup_bits: usize,
+    build: impl FnOnce(&mut BaseCircuitBuilder<Fr>),
+) {
     let mut builder = BaseCircuitBuilder::from_stage(CircuitBuilderStage::Mock)
         .use_k(17)
```
jonathanpwang, line 38: this is kind of weird: hardcode 17 but modify the lookup bits. ok if intentional

jonathanpwang (approved):

crates/static-verifier/src/transcript/tests.rs
zlangley: It's just to get `transcript_decomp` to fail on the previous impl.

PR #2703, opened 2026-04-13
fix: sha2 guest library api mismatch

876pol:
Resolves INT-7422.


shuklaayush:
@claude can you check that the changes are consistent with the rust sha2 crate imported in the workspace cargo.toml

shuklaayush (approved):

PR #2706, opened 2026-04-14
fix: mark keccak256 finalize and native_keccak256 as unsafe

Tuanlinh12312:
Towards INT-7429.

 ## Summary

  - Mark `finalize_ptr`, `finalize`, and `native_keccak256` as `unsafe fn` in both `zkvm_impl.rs` and `host_impl.rs`. These functions write 32 bytes to an output pointer/slice via `copy_nonoverlapping` without runtime size checks (only `debug_assert!`), so callers must guarantee the buffer is at least 32 bytes.
  - Add `assert!` in the `tiny_keccak::Hasher` trait impl in `lib.rs` since the trait method must be safe and cannot propagate `unsafe` to callers.
  - Safe public APIs (`keccak256()`, `set_keccak256()`) are unchanged — they pass correctly-sized buffers internally.

  **TODO:** The `openvm-org/tiny-keccak` fork and `openvm-org/openvm-primitives` also call `Keccak256::finalize` / `native_keccak256` and will need matching updates (add `unsafe {}` blocks and/or `assert!` at their trait boundaries).

### DIFF
--- a/guest-libs/keccak256/src/host_impl.rs
+++ b/guest-libs/keccak256/src/host_impl.rs
@@ -24,8 +24,10 @@ impl Keccak256 {
 
     /// Finalizes the hash computation and writes the result to the output buffer.
     ///
-    /// The output buffer must be at least `KECCAK_OUTPUT_SIZE` (32) bytes.
-    pub fn finalize(self, output: &mut [u8]) {
+    /// # Safety
+    ///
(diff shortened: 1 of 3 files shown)


jonathanpwang (approved):

jonathanpwang:
@claude run `cargo +nightly fmt --all` to resolve the lint CI issue

PR #2708, opened 2026-04-15
chore: algebra+ecc clean up and mark sha2 functions as unsafe

876pol:
Resolves INT-7439, INT-7422.


shuklaayush (approved):

shuklaayush:
> The comment reads // from_be_bytes still works if len < Scalar::NUM_LIMBS, but the next line actually calls from_be_bytes_unchecked. With this PR the checked from_be_bytes explicitly does not work for short input, so the comment is now actively misleading. Cheap to fix while you're in the file.

@876pol we should fix this

PR #2712, opened 2026-04-16
feat: make rvr extension framework

gdmlcjs:
makes the basic framework for using rvr extensions in openvm

- adds an rvr feature flag
- defines `RvrExtensionCtx` struct to provide mappings between opcode/executor/air indices
- defines `VmRvrExtension` trait that extensions can implement to be registered
- updated macro so that rvr `ExtensionRegistry` can be auto-generated in `SdkVmConfig`

closes INT-7474, INT-7475, INT-7479


gdmlcjs:
Just added some updates for the Claude comments.

.github/workflows/lints.yml
```diff
@@ -65,7 +67,7 @@ jobs:
           # cargo clippy --all-targets --all --tests --features "aggregation bls12_381 bn254 build-elfs default entrypoint evm-prove evm-verify export-intrinsics export-libm function-span getrandom-unsupported halo2-compiler halo2curves heap-embedded-alloc jemalloc jemalloc-prof metrics panic-handler parallel perf-metrics rust-runtime static-verifier std test-utils" -- -D warnings
           cargo clippy --all-targets --all --tests --features "bls12_381 bn254 default entrypoint export-intrinsics export-libm function-span getrandom-unsupported halo2curves heap-embedded-alloc jemalloc jemalloc-prof metrics panic-handler parallel perf-metrics rust-runtime std test-utils" -- -D warnings
           # cargo clippy --all-targets --all --tests --no-default-features --features "mimalloc" -- -D warnings
-          cargo clippy --all-targets --all --tests --no-default-features --features "mimalloc" -- -D warnings
+          cargo clippy --all-targets --all --tests --no-default-features --features "mimalloc rvr-openvm-ext-keccak-ffi/backend-openssl" -- -D warnings
```
shuklaayush, line 70: actually, let's just delete the xkcp backend in a follow up and only keep openssl

crates/rvr/rvr-openvm-lift/src/extension.rs
```diff
@@ -1,11 +1,53 @@
...
+    pub fn resolve_opcode_air_idx(&self, opcode: VmOpcode) -> Option<u32> {
+        let executor_idx = self.resolve_opcode_executor_idx(opcode)?;
+        self.executor_idx_to_air_idx
+            .get(executor_idx)
+            .map(|air_idx| *air_idx as u32)
+    }
+
+    pub fn require_opcode_air_idx(&self, opcode: VmOpcode) -> u32 {
+        self.resolve_opcode_air_idx(opcode).unwrap_or_else(|| {
+            panic!("opcode {opcode:?} not found in rvr extension context mappings")
```
shuklaayush: not ideal, we should throw proper `Err`s instead of panicing

crates/vm/src/arch/config.rs
```diff
@@ -331,6 +331,12 @@ impl AsMut<SystemConfig> for SystemConfig {
 // Default implementation uses no init file
 impl InitFileGenerator for SystemConfig {}
 
+#[cfg(feature = "rvr")]
+impl<F: openvm_stark_backend::p3_field::PrimeField32> ::rvr_openvm_lift::VmRvrExtension<F>
```
shuklaayush: imports at top except in macros

extensions/ecc/rvr/src/lib.rs
```diff
@@ -182,21 +180,15 @@ impl EccExtension {
     }
 
     /// Create with resolved chip indices from the VM config.
-    pub fn new<E>(
-        curves_info: Vec<u32>,
-        _inventory: &ExecutorInventory<E>,
-        _executor_idx_to_air_idx: &[usize],
-        staticlib_path: PathBuf,
-    ) -> Self {
+    pub fn new(curves_info: Vec<u32>, _ctx: &RvrExtensionCtx, staticlib_path: PathBuf) -> Self {
```
shuklaayush: should we just remove the unused args like `_ctx`?

extensions/ecc/rvr/src/lib.rs
```diff
@@ -182,21 +180,15 @@ impl EccExtension {
...
         // intentionally unused here.
         Self::from_curve_ids(curves_info, staticlib_path)
     }
 
-    pub fn new_from_configs<E>(
+    pub fn new_from_configs(
         curves: Vec<CurveConfig>,
-        _inventory: &ExecutorInventory<E>,
-        _executor_idx_to_air_idx: &[usize],
+        _ctx: &RvrExtensionCtx,
```
shuklaayush: maybe we can just remove this

extensions/algebra/rvr/src/lib.rs
```diff
@@ -501,15 +498,14 @@ impl AlgebraExtension {
         }
     }
 
-    pub fn new<E>(
+    pub fn new(
         moduli: Vec<BigUint>,
         fp2_moduli: Vec<BigUint>,
-        _inventory: &ExecutorInventory<E>,
-        _executor_idx_to_air_idx: &[usize],
+        _ctx: &RvrExtensionCtx,
```
shuklaayush: maybe remove

extensions/ecc/rvr/src/lib.rs
shuklaayush: also, let's keep `ctx` as first argument in these type of functions. maybe `ctx`, `staticlib_path` and then any extra arguments

shuklaayush (approved):
nits

.github/workflows/lints.yml
gdmlcjs: added a TODO in the Cargo.toml file

PR #2716, opened 2026-04-17
[codex] Reproduce and fix mixed proof deferral stage mismatch

dghelm:
## What changed

This PR isolates the mixed-proof stage-mismatch issue on `preview-v2.0.0-beta.2` into two commits:

1. `test(sdk): reproduce mixed proof stage mismatch`
   - adds a focused SDK repro in `crates/sdk/src/tests.rs`
   - forces the aggregation tree shape
   - constructs a VM child at verifier-base stage `(2,2)` and a deferral child at `(2,1)`
   - models the pre-fix `AggProver::prove_mixed(...)` path locally and shows that the resulting mixed proof is constructed but host verification fails with `BatchConstraintError::SumClaimMismatch`
   - shows that manually lifting the deferral child with `ProofsType::Deferral` and `ChildVkKind::RecursiveSelf` to `(2,2)` makes the mixed proof host-verify

2. `fix(sdk): align deferral child before prove_mixed`
   - updates `crates/sdk/src/prover/agg.rs`
   - reads verifier-base PVS from the VM child and deferral child
   - wraps the deferral child with `ProofsType::Deferral` / `ChildVkKind::RecursiveSelf` until its `recursion_flag` matches the VM child
   - only then performs `ProofsType::Mix`

## Root cause

On beta.2, `AggProver::prove_mixed(...)` could combine a VM child that had already reached a deeper internal-recursive verifier-base stage with a deferral child that had not.

In the reproduced case:

- VM child verifier-base stage: `(internal_flag=2, recursion_flag=2)`
- deferral child verifier-base stage: `(2,1)`

That unaligned mix produces a proof object, but host verification fails on the batch-constraint path with `SumClaimMismatch`.

One nuance from the proof PVS is worth calling out explicitly: at `(2,1)`, the deferral child still has `internal_recursive_vk_commit` unset in its proof-embedded verifier-base public values. After lifting it to `(2,2)`, that proof-embedded VK commit matches the VM child, and the mixed proof host-verifies.

## Why the repro is structured this way

The test uses a local helper that mirrors the pre-fix `prove_mixed` behavior rather than calling the fixed method directly for the failing path. That keeps the repro commit stable after the runtime fix lands while still demonstrating the exact beta.2 behavior that the fix addresses.

## Validation

Passed:

- `OPENVM_SKIP_DEBUG=1 cargo test --profile fast -p openvm-sdk --features root-prover test_prove_mixed_requires_stage_aligned_deferral_child -- --nocapture`

Additional targeted check attempted:

- `OPENVM_SKIP_DEBUG=1 cargo test --profile fast -p openvm-sdk --features root-prover test_verify_stark_deferral -- --nocapture`

That second test currently panics in `p3-poseidon2-air` with `Callers expected to pad inputs to a power of two`. I did not treat that as part of this PR because it does not exercise the mixed-proof stage-alignment path introduced here.

### DIFF
--- a/crates/sdk/src/prover/agg.rs
+++ b/crates/sdk/src/prover/agg.rs
@@ -1,14 +1,19 @@
-use std::sync::Arc;
+use std::{borrow::Borrow, sync::Arc};
 
 use eyre::Result;
 use openvm_circuit::arch::ContinuationVmProof;
 use openvm_continuations::{circuit::inner::ProofsType, prover::ChildVkKind};
 use openvm_recursion_circuit::prelude::Digest;
 use openvm_stark_backend::{
-    keygen::types::MultiStarkVerifyingKey, p3_field::PrimeCharacteristicRing,
+    keygen::types::MultiStarkVerifyingKey,
+    p3_field::{PrimeCharacteristicRing, PrimeField32},
+    proof::Proof,
 };
 use openvm_stark_sdk::config::baby_bear_poseidon2::{poseidon2_compress_with_capacity, F};
-use openvm_verify_stark_host::{pvs::DeferralPvs, VmStarkProof};
+use openvm_verify_stark_host::{
+    pvs::{DeferralPvs, VerifierBasePvs, VERIFIER_PVS_AIR_ID},
+    VmStarkProof,
+};
 use tracing::info_span;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


dghelm:
Sorry, wrong origin.

PR #2720, opened 2026-04-20
fix: resolve recursion_flag mismatch in prove_mixed

stephenh-axiom-xyz:
Resolves INT-7535. Issue found and initially corrected by [Daniel Heim](https://github.com/dghelm).

## Overview

This PR fixes a mixed-aggregation edge case in `openvm-sdk` where `prove_mixed` could combine a VM proof and a deferral proof that were not at compatible internal-recursive shapes. The fix threads deferral-layer metadata through the aggregation path and wraps the VM side or deferral side before mixing when needed.

The diff is limited to `crates/sdk` and is split into:

1. prover-path changes that carry and use deferral aggregation depth metadata
2. a regression test that exercises the mismatch scenarios directly

### Files

### What changed

- `AggProver::prove_def` now returns both the final `DeferralProof` and the deferral proof's internal-recursive layer.
- `StarkProver::prove` passes that returned layer into `AggProver::prove_mixed`.
- `AggProver::prove_mixed` now conditionally normalizes the two inputs before producing the mixed proof:
  - if the VM proof still needs an extra recursive wrapper, it wraps the VM proof
  - if the deferral proof still needs an extra recursive wrapper, it wraps the deferral proof
- `AggProver` adds `wrap_def_inner` to recursively wrap a deferral proof with `ProofsType::Deferral`.
- `InternalLayerMetadata` and `DeferralProof` now derive `Clone` so the test can reuse intermediate proofs and metadata cleanly.

### Why it matters

The mixed prover path now has explicit visibility into the deferral proof's recursive-layer state instead of assuming the VM and deferral sides are already aligned. Reviewers should focus on whether the new normalization logic in `agg.rs` matches the intended recursion-flag / proof-shape invariants for mixed aggregation.

### File

- `crates/sdk/src/tests.rs`

### What changed

- Adds `test_prove_mixed_vm_def_depth_mismatch`.
- The test builds the existing verify-stark deferral flow, then exercises both mismatch directions:
  - a VM proof that has been wrapped further before mixing
  - a deferral proof that has been wrapped further before mixing
- Both resulting mixed proofs are verified against the internal-recursive VK.

### Why it matters

This gives reviewers a concrete regression test for the shape-mismatch bug instead of relying only on end-to-end SDK flows that may not naturally hit the bad case.

### DIFF
--- a/crates/sdk/src/prover/agg.rs
+++ b/crates/sdk/src/prover/agg.rs
@@ -280,9 +280,9 @@ impl AggProver {
         // The VM and deferral proofs must be at the same stage in internal recursion, i.e.
         // both have be the parent of either internal-for-leaf or internal-recursive child
         // proofs. If this is not the case, we wrap the internal-for-leaf parent proof here.
-        if metadata.internal_node_idx == 1 && def_internal_recursive_layer != 1 {
+        if metadata.internal_recursive_layer == 1 && def_internal_recursive_layer != 1 {
             vm_proof = self.wrap_proof(vm_proof, metadata)?
-        } else if def_internal_recursive_layer == 1 && metadata.internal_node_idx != 1 {
+        } else if def_internal_recursive_layer == 1 && metadata.internal_recursive_layer != 1 {
             def_inner = self.wrap_def_inner(def_inner, def_internal_recursive_layer)?
         }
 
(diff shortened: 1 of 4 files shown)


shuklaayush (approved):
lgtm

PR #2721, opened 2026-04-20
feat: rv64 sha2

876pol:
Resolves INT-7520.

### DIFF
--- a/extensions/sha2/circuit/src/sha2_chips/main_chip/air.rs
+++ b/extensions/sha2/circuit/src/sha2_chips/main_chip/air.rs
@@ -11,7 +11,7 @@ use openvm_circuit::{
 };
 use openvm_circuit_primitives::{bitwise_op_lookup::BitwiseOperationLookupBus, utils::compose};
 use openvm_instructions::riscv::{
-    RV32_CELL_BITS, RV32_MEMORY_AS, RV32_REGISTER_AS, RV32_REGISTER_NUM_LIMBS,
+    RV64_CELL_BITS, RV64_MEMORY_AS, RV64_REGISTER_AS, RV64_REGISTER_NUM_LIMBS, RV64_WORD_NUM_LIMBS,
 };
 use openvm_sha2_air::Sha2BlockHasherSubairConfig;
 use openvm_stark_backend::{
@@ -205,8 +205,8 @@ impl<C: Sha2MainChipConfig + Sha2BlockHasherSubairConfig> Sha2MainAir<C> {
             &local.mem.register_aux,
         ) {
             self.memory_bridge
-                .read::<_, _, SHA2_READ_SIZE>(
-                    MemoryAddress::new(AB::Expr::from_u32(RV32_REGISTER_AS), ptr),
+                .read::<_, _, RV64_REGISTER_NUM_LIMBS>(
+                    MemoryAddress::new(AB::Expr::from_u32(RV64_REGISTER_AS), ptr),
                     val.to_vec().try_into().unwrap_or_else(|_| panic!()), // can't unwrap because AB::Var doesn't impl Debug
                     timestamp_pp(),
                     aux,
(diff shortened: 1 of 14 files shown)


extensions/sha2/circuit/src/sha2_chips/main_chip/air.rs
```diff
@@ -205,24 +205,37 @@ impl<C: Sha2MainChipConfig + Sha2BlockHasherSubairConfig> Sha2MainAir<C> {
...
+            AB::Expr::from_usize(1 << (RV64_WORD_NUM_LIMBS * RV64_CELL_BITS - self.ptr_max_bits));
         let needs_range_check = [
-            local.instruction.dst_ptr_limbs[RV32_REGISTER_NUM_LIMBS - 1],
-            local.instruction.state_ptr_limbs[RV32_REGISTER_NUM_LIMBS - 1],
-            local.instruction.input_ptr_limbs[RV32_REGISTER_NUM_LIMBS - 1],
-            local.instruction.input_ptr_limbs[RV32_REGISTER_NUM_LIMBS - 1],
+            local.instruction.dst_ptr_limbs[RV64_WORD_NUM_LIMBS - 1],
+            local.instruction.state_ptr_limbs[RV64_WORD_NUM_LIMBS - 1],
+            local.instruction.input_ptr_limbs[RV64_WORD_NUM_LIMBS - 1],
+            local.instruction.input_ptr_limbs[RV64_WORD_NUM_LIMBS - 1],
```
shuklaayush: looks like an oversight which was present earlier as well. we should just remove the duplicate column

extensions/sha2/circuit/src/sha2_chips/tests.rs
```diff
@@ -124,20 +124,20 @@ fn set_and_execute_single_block<RA: Arena, C: Sha2Config, E: PreflightExecutor<F
...
     rng: &mut StdRng,
-    opcode: Rv32Sha2Opcode,
+    opcode: Rv64Sha2Opcode,
     message: Option<&[u8]>,
     prev_state: Option<&[u8]>,
 ) {
-    let rd = gen_pointer(rng, 4);
-    let rs1 = gen_pointer(rng, 4);
-    let rs2 = gen_pointer(rng, 4);
+    let rd = gen_pointer(rng, 8);
```
shuklaayush: let's use the default block size or some constant

extensions/sha2/circuit/src/sha2_chips/main_chip/columns.rs
```diff
@@ -46,12 +46,12 @@ pub struct Sha2InstructionCols<T> {
...
-    /// state_ptr_limbs <- \[state_reg_ptr:4\]_1
-    pub state_ptr_limbs: [T; RV32_REGISTER_NUM_LIMBS],
-    /// input_ptr_limbs <- \[input_reg_ptr:4\]_1
-    pub input_ptr_limbs: [T; RV32_REGISTER_NUM_LIMBS],
+    /// dst_ptr_limbs <- \[dst_reg_ptr:8\]_1 (upper 4 bytes constrained to zero)
+    pub dst_ptr_limbs: [T; RV64_REGISTER_NUM_LIMBS],
+    /// state_ptr_limbs <- \[state_reg_ptr:8\]_1 (upper 4 bytes constrained to zero)
+    pub state_ptr_limbs: [T; RV64_REGISTER_NUM_LIMBS],
+    /// input_ptr_limbs <- \[input_reg_ptr:8\]_1 (upper 4 bytes constrained to zero)
+    pub input_ptr_limbs: [T; RV64_REGISTER_NUM_LIMBS],
```
shuklaayush, line 54: as a general comment, i think the better approach is to not materialize the extra columns and only have "word" size columns (4) for memory addresses (since they're going to be u32)
and then in the memory bus interaction, hardcode upper byte of addresses to 0. but we can do this in a follow-up

shuklaayush (approved):

extensions/sha2/circuit/src/sha2_chips/main_chip/air.rs
876pol: i think this is because the bitwise bus code after requires the length of the list to be even. i can add a comment

PR #2725, opened 2026-04-20
feat(v2.1-rv64): update guest toolchain target to `riscv64im-unknown-none-elf`

shuklaayush:


crates/toolchain/build/src/lib.rs
```diff
@@ -268,9 +268,10 @@ pub fn cargo_command(subcmd: &str, rust_flags: &[&str]) -> Command {
     // let rust_src = get_env_var("OPENVM_RUST_SRC");
     // if !rust_src.is_empty() {
     // TODO[jpw]: only do this for custom src once we make openvm toolchain
+    // TODO(rv64-std): restore `std` and `proc_macro` when guest std works
```
jonathanpwang: maybe we just delete both this and my todo -- we don't plan for a custom target, and no-std rv64 will just be the default

shuklaayush:
@claude can you make the suggested changes and update the other todos to say re-enable when guest is updated to `no_std`

jonathanpwang (approved):
Approved. I would remove all the todos -- the no-std riscv64 is the intended target. There's no present pathway to adding std since this requires an upstream rust change and currently Rust discourages `std` implementations that are stub without more complete functionality.

PR #2729, opened 2026-04-21
fix: Poseidon2 chip tracegen handles no records

stephenh-axiom-xyz:
Resolves INT-7541. Issue found and initially corrected by [Daniel Heim](https://github.com/dghelm).

## Overview

The `Poseidon2PeripheryBaseChip` (in `crates/vm`) and the `DeferralPoseidon2Chip` (in `extensions/deferral`) both use an `AtomicBool nonempty` flag to track whether any records have been inserted. Two issues existed:

1. **Empty-trace generation was unsafe.** When `nonempty = false`, the trace-generation path still called `subchip.generate_trace(vec![])` on an empty input vector, which is not a supported input.
2. **Chip state was not fully reset after generation.** `generate_proving_ctx` cleared `self.records` but left `nonempty = true`, so a subsequent call on the same chip would see `nonempty = true` + `records.len() == 0`, compute `height = 0`, and again hit the empty-input path. Chips were effectively single-use.
3. **VM periphery only:** the `SerialReceiver::receive` path on `Poseidon2PeripheryBaseChip` in `memory/merkle/trace.rs` inserted into `records` without setting `nonempty = true`. Records added through this path were invisible to the check in step 1/2.

This PR:
- Short-circuits `generate_proving_ctx` with an empty `RowMajorMatrix` when `nonempty` is false, avoiding the `generate_trace(vec![])` call.
- Resets `nonempty = false` after trace generation (together with the existing `records.clear()`), so the chip is reusable.
- Fixes the missing `nonempty.store(true, …)` in the VM's Merkle `SerialReceiver::receive` impl.
- Adds tests asserting that each chip returns an empty trace when it has no records, and that calling `generate_proving_ctx` twice still returns an empty trace the second time.

The two chips have independent but structurally identical implementations, so the fix is applied in parallel to each.

### `crates/vm` — VM Poseidon2 periphery chip

**`crates/vm/src/system/memory/merkle/trace.rs`**
- In `impl SerialReceiver<&[F]> for Poseidon2PeripheryBaseChip::receive`, set `self.nonempty = true` after inserting/updating a record. Without this, records inserted via the Merkle path never flipped `nonempty`, so trace generation would produce an empty trace despite having records. (This is the only real correctness bug of the PR; everything else is the empty-records hardening described below.)

**`crates/vm/src/system/poseidon2/trace.rs`** — `Poseidon2PeripheryBaseChip::generate_proving_ctx`
- If `nonempty` is false, return `AirProvingContext::simple_no_pis(RowMajorMatrix::new(vec![], width))` immediately (skip `generate_trace`).
- Otherwise compute `height = next_power_of_two_or_zero(self.records.len())` and proceed as before.
- After clearing records, additionally `nonempty.store(false, …)` so the chip can be reused.

**`crates/vm/src/system/poseidon2/tests.rs`** — new test `poseidon2_periphery_empty_trace`
- Constructs a `Poseidon2PeripheryBaseChip` with no records.
- Calls `generate_proving_ctx` twice and asserts `ctx.common_main.height() == 0` both times (verifies the reset).

### `extensions/deferral` — Deferral Poseidon2 chip

**`extensions/deferral/circuit/src/poseidon2/trace.rs`** — `DeferralPoseidon2Chip::generate_proving_ctx`
- Same two changes as the VM chip: short-circuit on `!nonempty`, and reset `nonempty = false` after clearing records. (No Merkle-`SerialReceiver` counterpart here — `DeferralPoseidon2Chip::perm_and_record` already sets `nonempty`.)

**`extensions/deferral/circuit/src/poseidon2/mod.rs`**
- Exposes a new `#[cfg(test)] pub mod tests` submodule.

**`extensions/deferral/circuit/src/poseidon2/tests.rs`** (new)
- Analogous `deferral_poseidon2_empty_trace` test: asserts two back-to-back empty-records calls both yield a height-0 trace.

### DIFF
--- a/crates/vm/src/system/poseidon2/tests.rs
+++ b/crates/vm/src/system/poseidon2/tests.rs
@@ -1,10 +1,12 @@
 use std::sync::Arc;
 
+use openvm_circuit_primitives::Chip;
+use openvm_cpu_backend::CpuBackend;
 use openvm_poseidon2_air::Poseidon2Config;
 use openvm_stark_backend::{
     interaction::LookupBus,
     p3_field::{PrimeCharacteristicRing, PrimeField32},
-    prover::AirProvingContext,
+    prover::{AirProvingContext, MatrixDimensions},
     test_utils::dummy_airs::interaction::dummy_interaction_air::{
         DummyInteractionChip, DummyInteractionData,
     },
@@ -19,8 +21,8 @@ use crate::{
         testing::{TestSC, VmChipTestBuilder, POSEIDON2_DIRECT_BUS},
     },
     system::poseidon2::{
-        new_poseidon2_periphery_air, Poseidon2PeripheryChip, PERIPHERY_POSEIDON2_CHUNK_SIZE,
-        PERIPHERY_POSEIDON2_WIDTH,
+        new_poseidon2_periphery_air, Poseidon2PeripheryBaseChip, Poseidon2PeripheryChip,
+        PERIPHERY_POSEIDON2_CHUNK_SIZE, PERIPHERY_POSEIDON2_WIDTH,
     },
 };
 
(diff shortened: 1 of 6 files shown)


crates/vm/src/system/poseidon2/tests.rs
```diff
@@ -153,3 +155,16 @@ fn poseidon2_periphery_duplicate_hashes_test() {
     let mut tester = tester.build().load_periphery_ref((air, chip)).finalize();
     tester.air_ctxs.push((dummy_air, dummy_ctx));
 }
+
+#[test]
+fn poseidon2_periphery_empty_trace() {
```
Golovanov399, line 160: Maybe also makes sense to test alternating empty and nonempty tests, since we used to forget resetting `nonempty`

Golovanov399 (approved):
If no perf regression then seems good

PR #2730, opened 2026-04-21
chore: move rvr files into openvm-circuit to avoid circular dependency

gdmlcjs:
Moves the rvr files related to compiling and execution into openvm-circuit. Those rvr files previously depended on openvm-circuit and in order to enable rvr execution through the openvm pipeline, they had to be made a part of openvm-circuit to prevent circular dependencies.

closes INT-7537


crates/rvr/rvr-openvm/src/emit/context.rs
```diff
@@ -12,16 +12,19 @@ pub struct EmitContext {
     var_counter: u32,
     /// Chip index for hint_store operations (u32::MAX = no chip).
     pub hint_store_chip_idx: u32,
+    /// Address space id for public values.
+    pub public_values_as: u32,
```
shuklaayush: let's revert this. we shouldn't use variables for compile time constants. you can just duplicate the constant with a todo for now that there should be a single source of truth

crates/rvr/rvr-openvm/src/emit/project.rs
```diff
@@ -76,10 +76,28 @@ pub struct CProject {
     pub chip_widths: Option<Vec<u64>>,
     /// Compile with native debug info (`-g -fno-omit-frame-pointer`).
     pub native_debug_info: bool,
+    /// Tracer page size in log2(bytes per page).
+    pub page_bits: u32,
+    /// Merkle chunk size in log2(bytes per chunk).
+    pub chunk_bits: u32,
```
shuklaayush: let's avoid variable and use a constant value. variables should be reserved for truly runtime configurable parameters

crates/rvr/rvr-openvm/src/lib.rs
```diff
@@ -1,38 +1,39 @@
...
-    default_linker,
+    default_addr2line_cmd, default_compiler, default_compiler_command, default_dwarfdump_cmd,
+    default_linker, default_linker_or_lld, linker_exists, Compiler,
 };
+
+/// Cache stamp for backend artifacts used by rvr-native compilation.
+///
+/// This covers codegen/runtime support living in the `rvr-openvm` backend crate.
+#[must_use]
+pub fn backend_cache_stamp() -> String {
```
shuklaayush: this was useful for testing but we should probably just delete it now

crates/rvr/rvr-openvm/src/compile.rs
```diff
@@ -305,15 +311,15 @@ pub fn native_cache_stamp() -> String {
     const DEBUG_INFO_CACHE_VERSION: &str = "debug-info-v1";
 
     let mut hasher = Sha256::new();
+    hasher.update(rvr_openvm::backend_cache_stamp().as_bytes());
     for source in [
-        include_str!("../c/openvm_io.c"),
```
shuklaayush, line 309: again, let's remove all this hashing/caching stuff to limit complexity

crates/vm/src/arch/rvr/compile.rs
```diff
@@ -416,6 +422,15 @@ fn hash_cache_stamp_path(
...
 
+fn openvm_compile_constants() -> (u32, u32, u32, u32) {
+    (
+        DEFAULT_SEGMENT_CHECK_INSNS as u32,
+        DEFAULT_PAGE_BITS as u32,
+        CHUNK.ilog2(),
+        PUBLIC_VALUES_AS,
+    )
+}
+
```
shuklaayush: don't understand why we need a function for this. seems confusing

crates/vm/tests/rvr_openvm.rs
shuklaayush, line 1: i would avoid this whole file and add tests to respective extension crates instead. maybe `vm` crate can have a small `test_utils` or similar module that is shared

crates/vm/Cargo.toml
```diff
@@ -64,6 +71,46 @@ openvm-stark-backend = { workspace = true, features = ["test-utils"] }
...
+openvm-pairing-transpiler.workspace = true
+openvm-rv32im-circuit.workspace = true
+openvm-rv32im-transpiler.workspace = true
+openvm-sdk.workspace = true
+openvm-sha2-circuit.workspace = true
+openvm-sha2-transpiler.workspace = true
+openvm-toolchain-tests.workspace = true
+openvm-transpiler.workspace = true
+openvm-verify-stark-circuit.workspace = true
+openvm-verify-stark-host.workspace = true
```
shuklaayush: this is a big no and defeats the purpose of having a small, minimal shared vm crate

shuklaayush (changes requested):

crates/vm/tests/rvr_openvm.rs
gdmlcjs: I put the test utils inside the rvr directory as a separate crate.

crates/rvr/rvr-openvm-ffi-common/src/lib.rs
```diff
@@ -35,6 +35,17 @@ pub const AS_PUBLIC_VALUES: u32 = 3;
...
+// ── OpenVM memory / metered-execution layout ──────────────────────────
+// Redefined here to avoid a cycle with openvm-circuit. Checked against
+// upstream in `openvm-circuit`'s `arch::rvr::abi_consts`.
+
+pub const MEM_BITS: usize = 29;
+pub const CHUNK: usize = 8;
+/// Default, not an invariant; see TODO in `abi_consts.rs`.
+pub const DEFAULT_PAGE_BITS: usize = 6;
+/// Default, not an invariant; see TODO in `abi_consts.rs`.
+pub const DEFAULT_SEGMENT_CHECK_INSNS: u32 = 1000;
```
shuklaayush, line 47: we should figure out a way to share these later. maybe there should be a common crate somewhere. not an issue for now since we already have too many crates

crates/rvr/rvr-openvm-test-utils/Cargo.toml
shuklaayush, line 1: fine for now but we should try to minimize these extra crates and try to consolidate the code

shuklaayush (approved):
lgtm. some cleanup that can be handled later

PR #2731, opened 2026-04-21
fix: SDK merkle path generation uses right child when depth == 0

stephenh-axiom-xyz:
Resolves INT-7542.

## Overview

Fixes an off-by-one bug in `deferral_merkle_proof_from_tree` (`crates/sdk/src/prover/deferral/merkle.rs`) when generating deferral merkle proofs with `depth == 0`. Also factors out a shared test helper that constructs a deferral-enabled SDK so the new "deferrals enabled without usage" test path exercises the same configuration used by the full deferral flow.

### `crates/sdk/src/prover/deferral/merkle.rs` — the fix

`deferral_merkle_proof_from_tree` builds a merkle path from the deferral leaf up to the root, optionally skipping the first `depth` levels (which are covered by the deferral subtree).

The previous code computed the starting node as `leaf_idx >> depth`. When `depth == 0`, this returns `leaf_idx` itself, so the first sibling collected was `leaf_idx ^ 1` — the left neighbor of the deferral leaf. The proof expected here is actually the right sibling of the deferral leaf (the neighboring node at the same level).

The fix special-cases `depth == 0` to start at `leaf_idx + 1` (i.e., the right sibling), so the first collected sibling is correct. For `depth > 0`, behavior is unchanged.

```rust
let mut node_idx = if depth == 0 {
    leaf_idx + 1
} else {
    leaf_idx >> depth
};
```

**Review focus:** confirm that in the deferral memory layout, the expected first merkle path element at `depth == 0` is the right sibling (i.e., the deferral leaf lives at an even `leaf_idx` so `leaf_idx + 1` is its sibling).

### `crates/sdk/src/tests.rs` — test refactor

Extracts the deferral-enabled SDK construction into a new helper `make_deferral_enabled_sdk(...)`:

- Builds `SdkVmConfig::riscv32()`, attaches the deferral extension, bumps the deferral address-space cell count, and wires up the deferral prover on the SDK builder.
- `make_deferral_sdk` now delegates to this helper instead of inlining the same construction.
- `test_deferrals_enabled_without_usage` now uses this helper as well, so it runs with the full deferral VM inventory (previously it used the plain `AppConfig::riscv32`, which did not include the deferral extension).

**Review focus:** the behavioral change in `test_deferrals_enabled_without_usage` — it now exercises an SDK whose app VM includes deferral periphery chips but where no deferral usage occurs at runtime. This is the intended coverage for "deferrals compiled in but unused" rather than "deferrals not compiled in."

### DIFF
--- a/crates/sdk/src/prover/deferral/merkle.rs
+++ b/crates/sdk/src/prover/deferral/merkle.rs
@@ -39,8 +39,9 @@ fn deferral_merkle_proof_from_tree(
 
     // Leaf index for DEFERRAL_AS, block_id=0 in the full tree (1-indexed).
     let leaf_idx = (1u64 << overall_height) + memory_dimensions.label_to_index((DEFERRAL_AS, 0));
+    debug_assert_eq!(leaf_idx % 2, 0);
 
-    // Start at level `depth` above the leaf. When `depth == 0``, the first node in the
+    // Start at level `depth` above the leaf. When `depth == 0`, the first node in the
     // path is the right sibling of the node at `leaf_idx`.
     let mut node_idx = if depth == 0 {
         leaf_idx + 1
--- a/crates/sdk/src/prover/deferral/merkle.rs
+++ b/crates/sdk/src/prover/deferral/merkle.rs
@@ -40,8 +40,13 @@ fn deferral_merkle_proof_from_tree(
     // Leaf index for DEFERRAL_AS, block_id=0 in the full tree (1-indexed).
     let leaf_idx = (1u64 << overall_height) + memory_dimensions.label_to_index((DEFERRAL_AS, 0));
 
-    // Start at level `depth` above the leaf.
-    let mut node_idx = leaf_idx >> depth;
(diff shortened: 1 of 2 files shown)


crates/sdk/src/prover/deferral/merkle.rs
```diff
@@ -40,8 +40,13 @@ fn deferral_merkle_proof_from_tree(
     // Leaf index for DEFERRAL_AS, block_id=0 in the full tree (1-indexed).
     let leaf_idx = (1u64 << overall_height) + memory_dimensions.label_to_index((DEFERRAL_AS, 0));
 
-    // Start at level `depth` above the leaf.
-    let mut node_idx = leaf_idx >> depth;
+    // Start at level `depth` above the leaf. When `depth == 0``, the first node in the
```
shuklaayush, line 43: as claude suggested
```suggestion
    // Start at level `depth` above the leaf. When `depth == 0`, the first node in the
```

shuklaayush (approved):
lgtm

PR #2728, opened 2026-04-21
feat: rv64 ECC

Tuanlinh12312:
Resolves INT-7522. Should be merged after #2727.

PR #2734, opened 2026-04-21
feat: rv64 pairing

Tuanlinh12312:
Resolves INT-7523. Should be merged after #2728.


extensions/pairing/circuit/src/pairing_extension.rs
```diff
@@ -149,8 +149,8 @@ pub(crate) mod phantom {
             b: u32,
             c_upper: u16,
         ) -> eyre::Result<()> {
-            let rs1 = read_rv32_register(memory, a);
-            let rs2 = read_rv32_register(memory, b);
+            let rs1 = read_rv64_register(memory, a) as u32;
```
shuklaayush, line 153: should at least have a debug assert instead of silently truncating. even better if we use a shared helper that converts u64 bytes/values to u32 memory addresses with an internal debug assertion

Tuanlinh12312: added TODO

shuklaayush (approved):

PR #2737, opened 2026-04-22
feat: rv64 cuda system airs

876pol:
Resolves INT-7563.


shuklaayush:
@876pol we should revert the changes that assume `DEFAULT_BLOCK_SIZE == DIGEST_WIDTH == 8` since we're probably going to move to `u16` limbs and `DEFAULT_BLOCK_SIZE = 4` soon after this. just keep it generic on `BLOCKS_PER_CHUNK`

876pol:
I reverted the larger changes. The CUDA tracegen still takes a custom path when `BLOCKS_PER_CHUNK == 1`. Unlike the CPU code, the CUDA kernels aren't templated on block/chunk sizes, so they can't handle this case directly.

crates/vm/src/system/cuda/memory.rs
```diff
@@ -152,25 +155,54 @@ impl MemoryInventoryGPU {
...
+                })
+                .collect();
+            let merkle_words: &[u32] = unsafe {
+                std::slice::from_raw_parts(
+                    merkle_records.as_ptr() as *const u32,
+                    merkle_records.len() * MERKLE_TOUCHED_BLOCK_WIDTH,
+                )
+            };
+            self.merkle_records = Some(merkle_words.to_device_on(&self.device_ctx).unwrap());
         } else {
```
shuklaayush: it's a bit weird that it appears like this function should be generic for different block/chunk sizes but on the cuda side, the values are hardcoded

crates/vm/src/system/cuda/memory.rs
```diff
@@ -240,16 +273,24 @@ impl MemoryInventoryGPU {
...
-                values.copy_from_slice(&out_records[base + 4..base + 4 + DIGEST_WIDTH]);
+                values.copy_from_slice(
+                    &out_records
+                        [base + 2 + BLOCKS_PER_CHUNK..base + 2 + BLOCKS_PER_CHUNK + DIGEST_WIDTH],
+                );
+                let timestamp = out_records[base + 2..base + 2 + BLOCKS_PER_CHUNK]
+                    .iter()
+                    .copied()
+                    .max()
+                    .unwrap_or(0);
```
shuklaayush: can probably avoid the copy and just use `unwrap`

crates/vm/src/system/cuda/memory.rs
```diff
@@ -152,25 +155,54 @@ impl MemoryInventoryGPU {
...
+
+            let merkle_records: Vec<MemoryMerkleRecord> = records
+                .iter()
+                .map(|r| MemoryMerkleRecord {
+                    address_space: r.address_space,
+                    ptr: r.ptr,
+                    timestamp: r.timestamps[0],
+                    values: r.values,
+                })
+                .collect();
```
shuklaayush: maybe you don't need to construct this since the structs seem to have the same layout, and can just use `records` instead of `merkle_records` below

shuklaayush (approved):
approved with some comments

crates/vm/src/system/cuda/memory.rs
876pol: I think this is no longer true for when `BLOCKS_PER_CHUNK` is changed back to 2

crates/vm/src/system/cuda/memory.rs
876pol: iirc merging blocks of size 4 to size 8 on cuda was a bit annoying to do generically

876pol:
@shuklaayush When I was working on this, Claude found that the cuda kernel to merge blocks of size 4 to 8 seems to have many unnecessary copies between device and host, so this might be something that can be fixed when the merge kernel is used again.

crates/vm/src/system/cuda/memory.rs
shuklaayush: yes but this is in the `if BLOCKS_PER_CHUNK == 1` branch

crates/vm/src/system/cuda/memory.rs
shuklaayush: yeah, maybe the if should be on `CHUNK` and `DEFAULT_BLOCK_SIZE` instead and then an unreachable or some compile time panic on a default else branch

PR #2739, opened 2026-04-22
feat: rv64 extensions cleanup

Tuanlinh12312:
Resolves INT-7566.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6204,6 +6204,7 @@ dependencies = [
  "openvm-mod-circuit-builder",
  "openvm-pairing-guest",
  "openvm-riscv-adapters",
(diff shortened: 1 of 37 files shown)


shuklaayush:
@Tuanlinh12312 can you also cleanup all the air code that uses inline functions like this?
```rust
// read rs1
let rs1_data: [AB::Expr; RV64_REGISTER_NUM_LIMBS] = std::array::from_fn(|i| {
    if i < RV64_WORD_NUM_LIMBS {
        local_cols.rs1_data[i].into()
    } else {
        AB::Expr::ZERO
    }
});
```

shuklaayush (approved):
i feel there is still some cleanup remaining but i want to merge this so let's do it in a follow up

PR #2740, opened 2026-04-23
fix(v2.1-rv64): make all guests `no_std` and use portable atomics

shuklaayush:
Switches guest dependencies that need atomic CAS (`once_cell::race` in the algebra/ecc macros, `spin::Mutex` in `guest-libs/keccak256`) to `portable-atomic` so they compile on stock `riscv64im-unknown-none-elf`, fixing CI failures across the algebra/ecc/pairing/ff_derive/k256/p256/keccak256 guest tests. The stock target's spec declares `atomic-cas: false`, so rustc cfg-gates `AtomicUsize::compare_exchange` out at the language level; `-C passes=lower-atomic` can't help since it acts on LLVM IR after rustc has already rejected the call. rc.1 avoided this on `riscv32im-risc0-zkvm-elf`, which claims CAS support in its spec; we keep guest programs on the stock upstream triple and patch the dependency instead.

Switches `once_cell` (and `spin::Mutex` in `guest-libs/keccak256`) to their `portable-atomic` backend, with `portable-atomic` configured to use `critical-section` as its CAS fallback. `critical-section` is a minimal no_std mutual-exclusion interface: exactly one crate in the tree registers an `Impl` via `critical_section::set_impl!`, and every consumer calls `critical_section::with(...)`; without an impl the call panics at runtime. Off-the-shelf impls target interrupt-driven hardware and disable interrupts via CSR reads/writes, which the openvm VM can't transpile (they map to `unimp` and blow up at runtime). `openvm-platform` now registers a no-op impl, sound because the openvm guest is single-threaded, so CAS reduces to a plain load+store with zero CSR emission. A `SAFETY:` comment flags the single-thread invariant.

## Cfg conventions

Guest vs host gating now follows one rule across both code and manifests:

- `.rs` files: `cfg(openvm_intrinsics)` (set by the guest build rustflags) for openvm-specific code paths, `cfg(not(feature = "std"))` for std-availability (`no_std`), `cfg(target_os = "none")` for runtime-specific attributes (`no_main`, `openvm::entry!` registration). The `no_std` / `no_main` split matches their semantics: std availability is a package decision, entry-point registration is a runtime/target property.
- `Cargo.toml` manifests: `cfg(not(target_os = "none"))` for host-only dep gates (Cargo target predicates only accept built-in cfgs, not `openvm_intrinsics`). The `portable-atomic` CAS polyfill keeps the narrower `cfg(all(target_arch = "riscv64", target_os = "none"))` since the `critical-section` backend is arch-specific.
- `--check-cfg=cfg(openvm_intrinsics)` added to guest-build rustflags in `crates/toolchain/build/src/lib.rs` so user guest programs never see the unexpected-cfgs warning.
- `[workspace.lints.rust]` at the workspace root declares `openvm_intrinsics`; openvm-* crates inherit via `[lints] workspace = true`. Standalone sub-workspaces (tests/programs, book examples) can't inherit, so they declare it explicitly.

## Resolver and dep hygiene

- `resolver = "2"` on all standalone `[workspace]` sub-workspaces (tests/programs, book examples, cli/tests). Bare `[workspace]` defaulted to v1, which unified proc-macro features into guest builds and dragged `rand_core/std` and `getrandom/std` into the rv64 target.
- Cargo.toml host-only dep gates swapped from the stale-on-rv64 `cfg(not(target_os = "zkvm"))` to `cfg(not(target_os = "none"))`.

## Pairing hint path

Three cascading rv64-port bugs, only exercised by the `final_exp_hint` tests:

- `extensions/pairing/transpiler/src/lib.rs` used `RV32_REGISTER_NUM_LIMBS` (= 4) to compute the phantom's `a`/`b` operand addresses. On rv64 registers live at 8-byte boundaries in `RV64_REGISTER_AS`, so the phantom read zero. Changed to `RV64_REGISTER_NUM_LIMBS`.
- `guest-libs/pairing/src/{bn254,bls12_381}/pairing.rs` constructed `(P.as_ptr() as u32, P.len() as u32)` as a fat pointer. The host phantom reads at offsets `0` and `RV64_REGISTER_NUM_LIMBS` (= 8), but the `(u32, u32)` tuple laid fields at offsets 0 and 4. Changed to `(u64, u64)`.
- The adjacent `hint_buffer_chunked(ptr, (N * 12 * 2) / 4)` used the rv32 word-size divisor. Changed to `/ openvm_riscv_guest::HINT_WORD_BYTES`.

## Hint infrastructure

`hint_store_u32!` to `hint_store_u64!` in `algebra/moduli-macros`. New `openvm_riscv_guest::HINT_WORD_BYTES = 8` replaces `/ 4` and `/ 8` magic numbers. Host-side `SqrtHintSubEx` padding bumped from `.take(4)` to `.take(RV64_REGISTER_NUM_LIMBS)` to match the 8-byte guest read.

## Renames

- `rv32`/`riscv32` to `rv64`/`riscv64` across SDK: `Sdk::riscv32()` and `AppConfig::riscv32()` become `riscv64()`, struct fields `rv32i`/`rv32m` become `rv64i`/`rv64m`, `openvm_riscv32.toml` renamed to `openvm_riscv64.toml` with matching TOML keys. Historical `v1.4/` and `v1.5/` snapshots untouched.
- `Rv32Modular*`, `Rv32Weierstrass*`, `Rv32Pairing*` imports in failing guest-lib integration tests renamed to `Rv64*`.
- `crates/sdk/guest/fib/src/main.rs`: `reveal_u32` to `reveal_u64`, `fibonacci` return type and locals widened from `u32` to `u64`.

## Book examples (`examples/`)

All 7 book examples were designed as host-run programs with `openvm/std` force-enabled, which dragged std into rv64 guest builds. Per-example: removed `features = ["std"]` from the `openvm` dep, moved `std` behind an example-local `std = ["openvm/std", ...]` feature, added `default-features = false` to transitive std-leakers (`serde`, `num-bigint`, `alloy-primitives`, `openvm-ruint`, `tiny-keccak`, `hex-literal`), and added the `no_main`/`no_std`/`openvm::entry!` guest-program scaffolding. The `i256` example additionally gets `bytes` with the `extra-platforms` feature and `portable-atomic` with `critical-section` on the riscv64-none target, since `alloy-primitives` transitively pulls `bytes` which uses `compare_exchange` directly.

## ff_derive `test_batch_inversion`

The test was passing `["std"]` to enable ff's `BatchInvert` trait via a `#[cfg(feature = "std")]` block in the guest program. On rv64 stock, activating `feature = "std"` at guest level chains through `openvm/std` and tries to link std, which riscv64im-unknown-none-elf doesn't have. Dropped `["std"]` from the test harness call; the std-only inner block is compiled out (the two `BatchInverter::invert_with_*_scratch` paths still run). Generated init file renamed from `openvm_init_batch_inversion_std.rs` to `openvm_init_batch_inversion.rs` to match the new build call.

## Cleanups

- Deleted `extensions/rv32im/` (empty pre-rv64 leftover) and an untracked stale `extensions/riscv/guest/guest/` copy with the old rv32 hint-store API.
- Removed dead duplicate target-cfg block in `extensions/pairing/circuit/Cargo.toml`.
- Fixed stale `cfg(target_os = "zkvm")` gate in `examples/keccak/Cargo.toml`.
- Removed duplicate `critical_section::set_impl!` in `crates/toolchain/platform/src/heap/embedded.rs`; the canonical no-op impl now lives in `crates/toolchain/platform/src/critical_section.rs`.

resolves int-7540


crates/toolchain/platform/src/critical_section.rs
```diff
@@ -0,0 +1,19 @@
...
+//! `spin`, etc. to polyfill CAS on `riscv64im-unknown-none-elf`) satisfy its
+//! `critical-section` backend without emitting CSR interrupt-disable
+//! instructions that the openvm transpiler can't handle.
+//!
+//! SAFETY: sound while the guest stays single-threaded; revisit if that changes.
+
+use critical_section::RawRestoreState;
+
+struct SingleThreadedCriticalSection;
+critical_section::set_impl!(SingleThreadedCriticalSection);
```
jonathanpwang, line 16: Is there any risk that some other crate also uses critical section and has a different `set_impl!`?

crates/toolchain/platform/Cargo.toml
```diff
@@ -14,7 +14,11 @@ openvm-riscv-guest.workspace = true
 # This crate should have as few dependencies as possible so it can be
 # used as many places as possible to share the platform definitions.
 [target.'cfg(target_arch = "riscv64")'.dependencies]
```
jonathanpwang: For this and also the other guest lib dependencies, do you want to be defensive and make it turn on only for im and not imac? How would we use critical section for `im` and normal atomics for `imac`

examples/i256/Cargo.toml
```diff
@@ -4,16 +4,24 @@ version = "0.0.0"
...
+openvm = { git = "https://github.com/openvm-org/openvm.git" }
+alloy-primitives = { version = "1.1.2", default-features = false }
+# `bytes` is pulled in transitively by `alloy-primitives` and uses atomic CAS,
+# which stock `riscv64im-unknown-none-elf` lacks. `extra-platforms` routes CAS
+# through `portable-atomic`; the guest-target `critical-section` feature below
+# points `portable-atomic` at `openvm-platform`'s no-op impl.
+bytes = { version = "1.11.1", default-features = false, features = ["extra-platforms"] }
+
+[target.'cfg(all(target_arch = "riscv64", target_os = "none"))'.dependencies]
+portable-atomic = { version = "1", features = ["critical-section"] }
```
jonathanpwang: why is this import needed, if `openvm` imports `openvm-platform`, which should already import and handle the critical-section? 

that would be better because otherwise it means we keep needing to add this import everywhere

examples/keccak/src/main.rs
```diff
@@ -160,6 +166,7 @@ pub fn main() {
         assert_eq!(output, expected);
     }
 
+    #[cfg(feature = "std")]
     println!(
```
jonathanpwang: nit: to show how to do this with no-std, change to use `openvm::io::println(format!(...))` where `format!` doesn't need std

examples/i256/Cargo.toml
```diff
@@ -25,3 +33,6 @@ openvm = { path = "../../crates/toolchain/openvm" }
 
 [profile.release]
 lto = "thin"
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(openvm_intrinsics)'] }
```
jonathanpwang, line 37: Can we add this to `cargo openvm init`? it seems this is something any guest crate will want to make lint happy

jonathanpwang (commented):
Mostly lgtm, but I want to understand the portable-atomics dependency issue better -- it'd be much better if we could set that in openvm-platform and skip it for downstream users

jonathanpwang:
Are the Cfg conventions documented somewhere in the codebase? If not, please add (e.g., AGENTS.md or crate docs)

examples/i256/Cargo.toml
shuklaayush: i'll add it. note that this is only needed for normal cargo commands like `cargo clippy` not for `cargo openvm` commands

crates/toolchain/platform/Cargo.toml
shuklaayush: sure

crates/toolchain/platform/src/critical_section.rs
shuklaayush: i think probably not since how this is handled seems like a property of the hardware and therefore only be customized by the hardware providers

examples/i256/Cargo.toml
shuklaayush: moved to `openvm-platform`. it's just that not every guest would use these atomic functions but maybe enough do so it's better to add it here

crates/toolchain/platform/Cargo.toml
shuklaayush: okay, this isn't that straightforward since it's used by `embedded-alloc`. i feel like we can look whether we still need this later when we add A

shuklaayush:
> Mostly lgtm, but I want to understand the portable-atomics dependency issue better -- it'd be much better if we could set that in openvm-platform and skip it for downstream users

yeah, i think this should work

> Are the Cfg conventions documented somewhere in the codebase? If not, please add (e.g., AGENTS.md or crate docs)

i added it to agents.md

jonathanpwang (approved):
LGTM

PR #2741, opened 2026-04-23
feat: rv64 cuda base_alu, less_than, shift, branch_eq, branch_lt, mul, mulh, divrem, jal_lui, jalr, auipc

876pol:
Resolves INT-7550.


extensions/riscv/circuit/cuda/src/mulh.cu
shuklaayush, line 1: i feel like the templated sections of this file should be in `include/riscv/cores` for consistency

extensions/riscv/circuit/cuda/src/jal_lui.cu
```diff
@@ -1,60 +1,68 @@
...
+            (record.rd_data[RV64_WORD_NUM_LIMBS - 1] >> (RV64_CELL_BITS - 1)) == 1;
+        bw.add_range(
+            static_cast<uint32_t>(record.rd_data[RV64_WORD_NUM_LIMBS - 1]) *
+                (4u * static_cast<uint32_t>(record.is_jal) +
+                 static_cast<uint32_t>(!record.is_jal)),
+            static_cast<uint32_t>(
+                static_cast<int32_t>(record.rd_data[RV64_WORD_NUM_LIMBS - 1]) * 2 -
+                (static_cast<int32_t>(is_sign_extend) << RV64_CELL_BITS)
+            )
+        );
```
shuklaayush, line 45: this is a good trick but we should add some comments to explain why this is correct. and what it is constraining (eg. `rd[3] < 64` for jal and correct value of `is_sign_extend` based on msb of `rd[3]`)

extensions/riscv/circuit/cuda/src/jal_lui.cu
```diff
@@ -1,60 +1,68 @@
...
-        if (record.is_jal) {
-            bw.add_xor(record.rd_data[RV32_REGISTER_NUM_LIMBS - 1], ADDITIONAL_BITS);
-        }
-
-        COL_WRITE_VALUE(row, Rv32JalLuiCoreCols, is_lui, !record.is_jal);
-        COL_WRITE_VALUE(row, Rv32JalLuiCoreCols, is_jal, record.is_jal);
-        COL_WRITE_ARRAY(row, Rv32JalLuiCoreCols, rd_data, record.rd_data);
-        COL_WRITE_VALUE(row, Rv32JalLuiCoreCols, imm, record.imm);
+        bool is_sign_extend =
+            (record.rd_data[RV64_WORD_NUM_LIMBS - 1] >> (RV64_CELL_BITS - 1)) == 1;
```
shuklaayush: probably don't need the `== 1`

shuklaayush (approved):
approved with comments. let's follow the existing directory structure for cuda adapter/core code

extensions/riscv/circuit/cuda/src/mulh.cu
876pol: The current headers in `include/riscv/cores` are there because `bigint` also uses the same cuda kernels. I think `mulh` isn't used anywhere other than here.

Related: I created `divrem.cuh` and put it in `include/riscv/cores` in #2743 since it is now used by both `divrem` and `divrem_w`

shuklaayush: okay, still a bit weird

PR #2743, opened 2026-04-23
feat: rv64 cuda base_alu_w, shift_w, mul_w, divrem_w

876pol:
Resolves INT-7551.


extensions/riscv/circuit/cuda/include/riscv/cores/divrem.cuh
```diff
@@ -0,0 +1,235 @@
...
+        bool q_sign = false;
+        bool case_none = false;
+
+        uint64_t b_val = 0;
+        uint64_t c_val = 0;
+#pragma unroll
+        for (size_t i = 0; i < NUM_LIMBS; i++) {
+            b_val |= (uint64_t)record.b[i] << (i * RV64_CELL_BITS);
+            c_val |= (uint64_t)record.c[i] << (i * RV64_CELL_BITS);
+        }
```
shuklaayush, line 98: we should create helpers for these

shuklaayush (approved):
lgtm

PR #2744, opened 2026-04-23
feat: rv64 cuda load_sign_extend

Tuanlinh12312:
Resolves INT-7553.

### DIFF
--- a/extensions/riscv/circuit/src/load_sign_extend/core.rs
+++ b/extensions/riscv/circuit/src/load_sign_extend/core.rs
@@ -271,7 +271,7 @@ where
         core_record.read_data = tmp.0 .1;
         core_record.shift_amount = tmp.1;
 
-        let write_data = run_write_data_sign_extend(
+        let write_data = run_write_data_sign_extend::<NUM_CELLS, LIMB_BITS>(
             local_opcode,
             core_record.read_data,
             core_record.shift_amount as usize,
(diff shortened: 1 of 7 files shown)


extensions/riscv/circuit/src/load_sign_extend/core.rs
```diff
@@ -349,25 +349,24 @@ where
...
+pub(super) fn run_write_data_sign_extend<const NUM_CELLS: usize, const LIMB_BITS: usize>(
     opcode: Rv64LoadStoreOpcode,
     read_data: [u8; NUM_CELLS],
     shift: usize,
 ) -> [u8; NUM_CELLS] {
-    assert!(
-        NUM_CELLS > 4,
-        "sign extension must extend at least one byte"
-    );
+    debug_assert_eq!(NUM_CELLS, RV64_REGISTER_NUM_LIMBS);
```
shuklaayush: const assert

extensions/riscv/circuit/src/load_sign_extend/core.rs
```diff
@@ -349,25 +349,24 @@ where
...
+    let word_width = NUM_CELLS / 2;
+    let half_width = NUM_CELLS / 4;
```
shuklaayush, line 359: can use `const x = `

shuklaayush (approved):

extensions/riscv/circuit/src/load_sign_extend/core.rs
Tuanlinh12312: I don't think const items could reference const generics

shuklaayush: ah right, nvm then

PR #2746, opened 2026-04-23
feat: rv64 cuda bigint

876pol:
Resolves INT-7564, INT-7528.


extensions/bigint/circuit/cuda/src/bigint.cu
```diff
@@ -2,57 +2,58 @@
...
+#include "riscv/cores/blt.cuh"
+#include "riscv/cores/less_than.cuh"
+#include "riscv/cores/mul.cuh"
+#include "riscv/cores/shift.cuh"
 
 using namespace riscv;
 
 constexpr size_t INT256_NUM_LIMBS = 32;
-constexpr size_t CONST_BLOCK_SIZE = 4;
+constexpr size_t CONST_BLOCK_SIZE = 8;
```
shuklaayush, line 17: should consolidate this

extensions/bigint/circuit/src/tests.rs
```diff
@@ -832,8 +832,8 @@ fn run_shift_256_rand_test_cuda(opcode: ShiftOpcode, num_ops: usize) {
     let gpu_chip = Shift256ChipGpu::new(
         tester.range_checker(),
         tester.bitwise_op_lookup(),
-        tester.timestamp_max_bits(),
         tester.address_bits(),
+        tester.timestamp_max_bits(),
```
shuklaayush, line 836: i guess this was a bug?

shuklaayush (approved):

extensions/bigint/circuit/src/tests.rs
876pol: yeah i think that was

PR #2748, opened 2026-04-23
feat: rv64 cuda hintstore

876pol:
Resolves INT-7554.


extensions/riscv/circuit/cuda/src/hintstore.cu
```diff
@@ -212,7 +203,9 @@ extern "C" int _hintstore_tracegen(
     uint32_t timestamp_max_bits,
     cudaStream_t stream
 ) {
-    assert(width == sizeof(Rv32HintStoreCols<uint8_t>));
+    assert(height == 0 || (height & (height - 1)) == 0);
+    assert(height >= rows_used);
```
shuklaayush, line 220: i think this is true by definition on the rust side so we don't need it here

extensions/riscv/circuit/cuda/src/hintstore.cu
```diff
@@ -212,7 +203,9 @@ extern "C" int _hintstore_tracegen(
     uint32_t timestamp_max_bits,
     cudaStream_t stream
 ) {
-    assert(width == sizeof(Rv32HintStoreCols<uint8_t>));
+    assert(height == 0 || (height & (height - 1)) == 0);
```
shuklaayush, line 219: again, probably can avoid

extensions/riscv/circuit/cuda/src/hintstore.cu
shuklaayush, line 215: this is unrelated but can you do a cleanup in a follow-up pr to remove this argument and use a constant instead? we're passing `RV64_CELL_BITS as u32` as `bitwise_num_bits` from rust and it doesn't make sense to use a runtime value here. also do a sweep for other kernels so we don't pass constants as arguments

extensions/riscv/circuit/cuda/src/hintstore.cu
```diff
@@ -69,86 +72,74 @@ struct Rv32HintStore {
...
-
-// Range check for num_words (using MAX_HINT_BUFFER_WORDS_BITS)
-// These constraints only work for MAX_HINT_BUFFER_WORDS_BITS in [8, 16)
-#ifdef CUDA_DEBUG
-            assert(MAX_HINT_BUFFER_WORDS_BITS >= 8 && MAX_HINT_BUFFER_WORDS_BITS < 16);
-#endif
-
-#ifdef CUDA_DEBUG
-            assert(record.num_words <= MAX_HINT_BUFFER_WORDS);
-#endif
```
shuklaayush, line 110: we should probably keep some of these asserts

shuklaayush (approved):
approved with some comments

extensions/riscv/circuit/cuda/src/hintstore.cu
876pol: I'll restore `assert(record.num_words <= MAX_HINT_BUFFER_WORDS);` since the other two should be known at compile time.

extensions/riscv/circuit/cuda/src/hintstore.cu
876pol: I think all other extensions have similar checks, so I put these here for consistency

extensions/riscv/circuit/cuda/src/hintstore.cu
shuklaayush: btw, i think cuda also has `static_assert` for compile time asserts

PR #2749, opened 2026-04-23
feat: rv64 cuda sha2

876pol:
Resolves INT-7530.

### DIFF
--- a/extensions/sha2/circuit/cuda/include/block_hasher/variant.cuh
+++ b/extensions/sha2/circuit/cuda/include/block_hasher/variant.cuh
@@ -7,9 +7,8 @@ namespace sha2 {
 
 // Common VM constants across SHA-2 variants.
 inline constexpr size_t SHA2_REGISTER_READS = 3;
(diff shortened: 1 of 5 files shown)


extensions/sha2/circuit/cuda/include/block_hasher/variant.cuh
```diff
@@ -7,9 +7,8 @@ namespace sha2 {
 
 // Common VM constants across SHA-2 variants.
 inline constexpr size_t SHA2_REGISTER_READS = 3;
-inline constexpr size_t SHA2_READ_SIZE = 4;
-inline constexpr size_t SHA2_WRITE_SIZE = 4;
-inline constexpr size_t SHA2_MAIN_READ_SIZE = 4;
+inline constexpr size_t SHA2_READ_SIZE = 8;
+inline constexpr size_t SHA2_WRITE_SIZE = 8;
```
shuklaayush, line 11: hmm we should consolidate these constants at some point so there's a single source of truth. feels like this should just be default_block_size

shuklaayush (approved):

PR #2753, opened 2026-04-24
feat: enable rvr execution of benchmarks

gdmlcjs:
Enables running benchmarks through rvr extension. Benchmark tests do not check execution correctness and currently execution involving extensions other than RV32IM fail because `VmRvrExtension` trait implementation is not properly wired.

closes INT-7480

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5776,6 +5776,7 @@ dependencies = [
  "openvm-stark-sdk",
  "openvm-transpiler",
  "rand 0.9.2",
+ "rvr-openvm-lift",
  "serde",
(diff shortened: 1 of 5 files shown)


shuklaayush (approved):
it's okay for now since we're going to remove `aot` later but i wonder if we can share more code between executor types. maybe we should have all executors conform to some trait interface

gdmlcjs:
Yes, that would be better. Should I add a ticket and merge?

shuklaayush:
> Yes, that would be better. Should I add a ticket and merge?

yes, let's do that

PR #2755, opened 2026-04-27
fix(v2.1-rv64): convert more rv32 references to rv64

shuklaayush:
# fix(v2.1-rv64): convert more rv32 references to rv64

Follow-up cleanup that gets the lint job and the `sha2_bench` bench-prove run green on the rv64 branch, plus a few non-AOT polish items called out in review.

## Lint fixes

- `extensions/riscv/circuit/src/auipc/tests.rs` — gate `RV64_REGISTER_NUM_LIMBS` behind `#[cfg(feature = "cuda")]`; it's only used inside the CUDA test, so the CPU lint job hit `unused_imports`.
- `crates/continuations/src/tests/e2e.rs` — rename stale `Rv32DeferralBuilder` / `Rv32DeferralConfig` → `Rv64DeferralBuilder` / `Rv64DeferralConfig` (3 sites). Fixes the CUDA lint job's `E0432`.
- `crates/sdk/src/tests.rs` — `SdkVmConfig::riscv32()` → `::riscv64()` plus doc-comment.
- `benchmarks/prove/src/lib.rs` — drop unused `clap::command` import.
- `extensions/riscv/circuit/src/{mul,mul_w,jalr,mulh}/tests.rs` — error string `"failed to create Rv32IM executor"` → `"Rv64IM executor"`.

## Rename

- `Rv32ModularArithmeticOpcode` → `Rv64ModularArithmeticOpcode` (81 references across `extensions/algebra/{transpiler,circuit}`). Pure rename, no semantic change.

## Cleanup

- Remove `tmp_convert_to_u8s` and its `// TEMP[jpw]` marker from `extensions/riscv/circuit/src/adapters/mod.rs` — `pub fn` with zero in-tree callers.
- `crates/toolchain/openvm/src/io/mod.rs` — replace hardcoded `8` in `read_vec_by_len` with `openvm_platform::WORD_SIZE`.

## Bug fix

- `crates/toolchain/openvm/src/serde/serializer.rs` — `to_vec` was calling `Vec::<u64>::with_capacity(size_of_val(value))`, conflating bytes with u64 elements and over-allocating 8×. Switched to `size_of_val(value).div_ceil(size_of::<u64>())`. Heap waste only, no correctness impact.

## Benchmark ELF rebuilds

Rebuilt 16 benchmark guest ELFs as 64-bit RISC-V (the committed binaries were still 32-bit, which made `sha2_bench` fail with `Error: Not a 64-bit ELF` after the transpiler tightened its check): base64_json, bincode, bubblesort, ecrecover, fibonacci, fibonacci_iterative, fibonacci_recursive, keccak256, keccak256_iter, kitchen-sink, pairing, quicksort, rkyv, sha2_bench, sha256, sha256_iter.

Not rebuilt — `regex`, `revm_transfer`, `revm_snailtracer` — their dependency closures (`regex-automata`, `tracing-core 0.1.x`) use atomic CAS methods that the `riscv64im-unknown-none-elf` target doesn't expose. The existing `portable-atomic` polyfill in `openvm-platform` works for crates that opt in via a feature flag (`bytes`, `spin`), but neither failing crate exposes one. Tracking as a separate item.

## Verification

- `cargo +nightly fmt --all -- --check` clean.
- `cargo clippy --workspace --all-targets --tests --features "<CI feature set>" -- -D warnings` clean.
- `cargo check --workspace --all-targets --tests` clean.


Tuanlinh12312:
LGTM

Tuanlinh12312 (approved):

PR #2763, opened 2026-04-28
chore: remove air index requirement from rvr pure execution

gdmlcjs:
Previously rvr execution had to use `executor_idx_to_air_idx` information in order to construct `ExtensionRegistry`. This was a problem for pure execution which didn't need air indices so the interface diverged between rvr and aot/interpreted. Now for rvr pure execution, dummy index values of `NO_CHIP` are used instead to keep the interface consistent.

towards INT-7611

### DIFF
--- a/crates/vm/src/arch/vm.rs
+++ b/crates/vm/src/arch/vm.rs
@@ -466,6 +466,9 @@ where
         // Pure execution does not consult air indices, so we hand the extension
         // registry a placeholder filled with `NO_CHIP` sentinels sized to cover
         // every executor.
+        // TODO: Drop this placeholder by making `RvrExtensionCtx` take an
+        // `Option<Vec<usize>>` (or be generic over a trait/value) so pure
+        // execution can avoid passing `executor_idx_to_air_idx` altogether.
         let executor_idx_to_air_idx = vec![NO_CHIP as usize; self.inventory.executors.len()];
         let extensions = self.build_rvr_extensions(&executor_idx_to_air_idx);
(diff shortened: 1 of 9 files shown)


crates/vm/src/arch/vm.rs
```diff
@@ -466,12 +462,12 @@ where
...
-        &self,
-        exe: &VmExe<F>,
-        executor_idx_to_air_idx: &[usize],
-    ) -> Result<RvrPureInstance<F>, StaticProgramError> {
-        let extensions = self.build_rvr_extensions(executor_idx_to_air_idx);
+    pub fn rvr_instance(&self, exe: &VmExe<F>) -> Result<RvrPureInstance<F>, StaticProgramError> {
+        // Pure execution does not consult air indices, so we hand the extension
+        // registry a placeholder filled with `NO_CHIP` sentinels sized to cover
+        // every executor.
+        let executor_idx_to_air_idx = vec![NO_CHIP as usize; self.inventory.executors.len()];
```
shuklaayush, line 472: can you a todo somewhere to remove this hack and make `RvrExtensionCtx` either take `Option` or be generic on some trait/value so we can avoid passing `executor_idx_to_air_idx` altogether

shuklaayush (approved):
approved but please a todo so we can track this is a hacky workaround

crates/vm/src/arch/vm.rs
gdmlcjs: Yes, added the TODO

PR #2765, opened 2026-04-30
feat(v2.1-rv64): integrate `riscv64im-unknown-openvm-elf` target into build system

Tuanlinh12312:
## Summary

Integrate the built-in Rust target `riscv64im-unknown-openvm-elf` (`target_os = "openvm"`) into the OpenVM build system and guest crates, enabling `std` support for guest programs.

The target itself was added to the OpenVM Rust fork in [openvm-org/rust#3](https://github.com/openvm-org/rust/pull/3). Prebuilt toolchain tarballs are published on the fork's [Releases](https://github.com/openvm-org/rust/releases) page — install via `cargo openvm toolchain install` (this PR) or the CI helper script.

### Changes

- **Build system** (`crates/toolchain/build/src/lib.rs`):
  - Set `DEFAULT_RUSTC_TARGET = "riscv64im-unknown-openvm-elf"` (overridable via `OPENVM_TARGET`).
  - Set `DEFAULT_RUSTUP_TOOLCHAIN_NAME = "openvm-nightly-2026-01-18"` (overridable via `OPENVM_RUST_TOOLCHAIN`); guest builds invoke `cargo +<toolchain> build --target <target>` directly, using the prebuilt `std`/`core`/`alloc`/`panic_abort` rlibs that ship in the toolchain.
  - Add `ensure_openvm_toolchain_linked()` — verifies the linked toolchain has the guest target's prebuilt rlibs in its sysroot before invoking cargo.

- **Guest runtime** (`crates/toolchain/openvm/`):
  - New `pal_abi.rs`: provides `sys_write`, `sys_panic`, `sys_rand`, `sys_halt`, `sys_argc`/`sys_argv`, `sys_getenv`, `sys_read`, and `sys_alloc_aligned` symbols that the upstream `std` PAL declares. Bridges to OpenVM's custom RISC-V instructions via `openvm_riscv_guest`. `sys_rand` uses a byte-level ABI (`*mut u8, nbytes: usize`) with a `hint_buffer_bytes` helper that handles non-dword-aligned tails via a stack scratch buffer; `sys_getenv` returns `usize::MAX` per the upstream "unset" contract.
  - Simplify the `entry!` macro: a single `cfg(openvm_intrinsics)` branch emits `#[no_mangle] fn main()` for both `std` and `no_std` guests. The custom `_start` defined via `global_asm!` (now using `asm_const` for `STACK_TOP`) replaces std's default entry.

- **Platform** (`crates/toolchain/platform/`):
  - Update `target_os = "none"` cfg gates to `any(target_os = "none", target_os = "openvm")` so target-gated deps (e.g. `libm`) resolve under either form.
  - Change `fileno` constants from `u32` to `i32` to match POSIX `RawFd` / the PAL's `sys_write(fd: i32, …)` signature. Drop unused `JOURNAL` fd.

- **CLI** (`crates/cli/src/commands/toolchain.rs`):
  - New `cargo openvm toolchain {install, uninstall, list}` subcommand. `install` downloads the prebuilt tarball from openvm-org/rust Releases for the host triple, extracts it under `~/.openvm/toolchains/`, and runs `rustup toolchain link`. Supports `--version <tag>` and `--force`; idempotent re-installs skip the relink via an `is_link_to` check.

- **CI helper** (`ci/install-openvm-toolchain.sh`):
  - Bash equivalent of `cargo openvm toolchain install` for repo workflows. Linux/macOS host detection, idempotent installs (skips the download when the target's rlib dir already exists). Wired into the `extension-tests`, `continuations`, `sdk.cuda`, `guest-lib-tests`, `riscv`, and `cli` workflows.

- **Guest programs (~100 files)**:
  - Update `no_main` cfg gate to `cfg_attr(any(target_os = "none", target_os = "openvm"), no_main)`.
  - Update 45 guest `Cargo.toml`s' `unexpected_cfgs` check-cfg to include `cfg(openvm_intrinsics)` and `cfg(target_os, values("openvm"))` so stable-rustc host checks recognise the new `target_os` value.

- **`riscv-guest`** (`extensions/riscv/guest/src/io.rs`):
  - New `hint_buffer_bytes(*mut u8, nbytes)` — byte-granular wrapper over the dword-granular `hint_buffer_chunked`, with a stack scratch dword for the trailing `0..8` bytes.
  - Mark `hint_buffer_chunked` `unsafe fn` (raw pointer write) and add SAFETY blocks at the three callers in `crates/toolchain/openvm/src/io/{mod,read}.rs`; existing callers in `moduli-macros` codegen and the pairing libs were already inside `unsafe` blocks.

### Install the toolchain

```bash
cargo openvm toolchain install         # or `bash ci/install-openvm-toolchain.sh`
```

This downloads `rust-toolchain-<host>.tar.gz` from the [openvm-org/rust release](https://github.com/openvm-org/rust/releases/tag/openvm-nightly-2026-01-18) and links it as `openvm-nightly-2026-01-18`.

### Test results

- `cargo nextest run -p openvm-riscv-integration-tests --run-ignored all` — local pass.
- Guest builds: `cargo openvm build` on `examples/algebra` (moduli-macros path) and `examples/pairing` (direct `hint_buffer_chunked` callers) compile clean against the linked toolchain.
- CI: all guest-touching workflows (`extension-tests`, `continuations`, `sdk.cuda`, `guest-lib-tests`, `riscv`, `cli`) install the toolchain via the script and pass on this branch.
- Benchmarks show no regression.

- [x] Toolchain install path works end-to-end via both the CLI and the bash helper
- [x] Guest builds succeed with `cargo openvm build`
- [x] CI passes (all non-CUDA-gated checks)

Resolves INT-7635, INT-7654


crates/toolchain/openvm/src/pal_abi.rs
shuklaayush, line 1: can you use the develop v2 rc1 implementation for this and adapt to rv64. it's better if the diff is minimal and only contains the necessary changes and any fixes
https://github.com/openvm-org/openvm/blob/69373ff1d06b613025fe4e7a02d735a1609412c4/crates/toolchain/openvm/src/pal_abi.rs

crates/toolchain/build/src/lib.rs
```diff
@@ -19,18 +19,26 @@ pub use self::config::GuestOptions;
...
 
-/// Get the Rust toolchain name from environment variable or default
+/// Default rustup toolchain name. Format: `openvm-<rustup-name>` —
+/// `openvm-` prefix plus whatever rustup itself calls the upstream
+/// toolchain (`X.Y.Z` for stable, `nightly-YYYY-MM-DD` for nightly,
+/// `beta-YYYY-MM-DD` for beta). Bumped per release. Override with
+/// `OPENVM_RUST_TOOLCHAIN`.
+//
+// Keep in sync with the default `TAG` in `ci/install-openvm-toolchain.sh`.
+pub const DEFAULT_RUSTUP_TOOLCHAIN_NAME: &str = "openvm-nightly-2026-01-18";
```
jonathanpwang: why do we use nightly instead of stable? previously we only used nightly because of the need for -Zbuild-std

crates/toolchain/build/src/lib.rs
```diff
@@ -19,18 +19,26 @@ pub use self::config::GuestOptions;
...
-pub fn rustc_target_spec_dir() -> &'static Path {
-    Path::new(env!("CARGO_MANIFEST_DIR"))
+/// Default rustc target for the openvm guest. Built-in to the openvm rust
+/// fork. Override with `OPENVM_TARGET` to use a different rv64im target
+/// (e.g. an upstream `riscv64imac-unknown-none-elf` or future custom spec).
+pub const DEFAULT_RUSTC_TARGET: &str = "riscv64im-unknown-openvm-elf";
+
+/// Get the target triple from environment variable or default.
+pub fn get_rustc_target() -> String {
+    env::var("OPENVM_TARGET").unwrap_or_else(|_| DEFAULT_RUSTC_TARGET.to_string())
```
jonathanpwang: is it better to make the env var `OPENVM_RUSTC_TARGET`?

crates/toolchain/build/src/lib.rs
```diff
@@ -253,38 +261,31 @@ fn sanitized_cmd(tool: &str) -> Command {
...
 
     let mut cmd = sanitized_cmd("cargo");
-    let mut args = vec![&toolchain, subcmd, "--target", RUSTC_TARGET];
+    let mut args = vec![toolchain.as_str(), subcmd, "--target", target.as_str()];
 
     if std::env::var(BUILD_LOCKED_ENV).is_ok() {
         args.push("--locked");
     }
 
-    args.extend_from_slice(&[
```
jonathanpwang, line 274: should we selectively keep this if the toolchain is not an openvm toolchain? basically I want it to be as smooth as possible to switch between the custom openvm target versus the rust no_std riscv64 target

crates/toolchain/openvm/src/pal_abi.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/// Fill `nbytes` of `recv_buf` with random bytes from the openvm host hint stream.
+///
+/// # Safety
+///
+/// `recv_buf` must be dereferenceable for `nbytes` bytes (no alignment required).
+#[no_mangle]
+pub unsafe extern "C" fn sys_rand(recv_buf: *mut u8, nbytes: usize) {
+    hint_random(nbytes.div_ceil(HINT_WORD_BYTES));
```
jonathanpwang, line 39: was there some feature gate controlling this?

crates/cli/src/commands/toolchain.rs
jonathanpwang, line 1: this is great

crates/toolchain/openvm/src/lib.rs
```diff
@@ -129,11 +126,10 @@ _start:
     .option norelax;
     la gp, __global_pointer$;
     .option pop;
-    la sp, {0};
-    ld sp, 0(sp);
+    li sp, {STACK_TOP};
```
jonathanpwang, line 128: @claude explain this change and whether it is equivalent to before

AGENTS.md
```diff
@@ -148,7 +151,7 @@ Naming convention: `FooExecutor`, `FooFiller`, `FooCoreAir` for a chip named `Fo
 Guest programs (run inside the VM) use `#![no_main]` / `#![no_std]` with `openvm::entry!(main)`:
 
 ```rust
-#![cfg_attr(target_os = "none", no_main)]
+#![cfg_attr(all(not(feature = "std"), any(openvm_intrinsics, target_os = "openvm")), no_main)]
```
jonathanpwang, line 154: commenting here since it's where I first saw it, but this is about general guest program UX: this is now quite a huge mouthful and too hard for users to understand. I'm wondering if we should change how we handle the `std -> main` pathway so we just keep `#![no_main]` always and rely on `openvm::entry!(main)`?

we could also switch to a proc-macro like
```
#[openvm::main]
fn main {

}
```

But I recall last we tried, rust does not allow you to add `#![no_main]` inside of a macro so there's no way to hide it.

jonathanpwang (commented):
I commented on some guest UX issue.

Other than that, the main changes LGTM. However, I think we should add a separate parallel CI to test the full CLI against the `riscv64im-unknown-none-elf` target to make sure things still work smoothly there. It doesn't need to be the main recommended path, but I'd still like to make sure using this non-custom target has good support.

crates/toolchain/build/src/lib.rs
shuklaayush: i kept it same as before. stable toolchain is building and will take a couple of hours. after that, i can run and we'll find out if it works
https://github.com/openvm-org/rust/actions/runs/26241691952

AGENTS.md
shuklaayush: to be addressed in a follow up

crates/toolchain/openvm/src/pal_abi.rs
shuklaayush: do you mean a feature controlling `sys_rand`? no
https://github.com/openvm-org/openvm/blob/develop-v2.0.0-rc.1/crates/toolchain/openvm/src/pal_abi.rs#L70-L77

crates/toolchain/build/src/lib.rs
shuklaayush: okay, let me add that

crates/toolchain/openvm/src/pal_abi.rs
jonathanpwang: I meant if we implemented `hint_random` properly, but it looks like we always do via phantom instruction and deterministic randomness, so it's all good

docs/vocs/docs/pages/book/writing-apps/writing-a-program.mdx
```diff
@@ -51,6 +51,7 @@ OpenVM supports standard Rust written using the `std` library, with the followin
 
 - Standard input (e.g., from console) is not supported. Use the `read` methods [above](#handling-io) instead.
 - Standard output and standard error (e.g., `println!, eprintln!`) are supported and will _both_ print to the host standard output.
+- `std::time` is deterministic and does not expose host wall-clock time. `Instant::now()` starts at zero, and `SystemTime::now()` returns `UNIX_EPOCH`; these APIs do not panic, but they do not report elapsed VM time or real time.
```
jonathanpwang, line 54: is this done in the rust fork? we can follow up about this; we might want some other kind of behavior

shuklaayush: yes

jonathanpwang (approved):
LGTM, the `is_openvm` build is handled nicely

PR #2766, opened 2026-05-01
test: remove rvr tests

gdmlcjs:
Removes the rvr tests and instead adds rvr comparison steps in existing openvm tests in a similar way to aot. Unlike aot, metered cost execution is also run and compared for rvr and interpreted modes.

closes INT-7627


crates/rvr/rvr-openvm-tests/tests/debug_info.rs
shuklaayush, line 1: we can probably delete this as well. let's just get rid of the whole `rvr-openvm-tests` crate

shuklaayush (approved):

PR #2768, opened 2026-05-01
feat: update EF to BabyBear^5 + system params to 128 bits security

stephenh-axiom-xyz:
Resolves INT-7600.


crates/vm/src/arch/execution_mode/metered/segment_ctx.rs
```diff
@@ -7,7 +7,7 @@ use p3_baby_bear::BabyBear;
 use serde::{Deserialize, Serialize};
 
 /// Extension field size.
-const D_EF: usize = 4;
+const D_EF: usize = 5;
```
shuklaayush: can you switch this to use an imported constant

zlangley (approved):

PR #2769, opened 2026-05-01
feat: implement rvr for keccak256 extension

gdmlcjs:
Implements the rvr feature for the keccak256 extension and also adds rvr tests to CI. Now extensions don't take a `staticlib_path` argument manually and instead uses the auto-built staticlib made by a build.rs file.

closes INT-7468

### DIFF
--- a/extensions/keccak256/rvr/src/lib.rs
+++ b/extensions/keccak256/rvr/src/lib.rs
@@ -11,7 +11,7 @@ use openvm_instructions::{instruction::Instruction, LocalOpcode};
 use openvm_keccak256_transpiler::{KeccakfOpcode, XorinOpcode};
 use openvm_stark_backend::p3_field::PrimeField32;
 use rvr_openvm_ir::{ExtEmitCtx, ExtInstr, Instr, InstrAt, LiftedInstr, Reg};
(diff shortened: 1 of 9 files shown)


extensions/keccak256/rvr/src/lib.rs
```diff
@@ -90,34 +90,42 @@ pub struct KeccakExtension {
 
 impl KeccakExtension {
     /// Pure-mode constructor; chip indices are unused (`trace_chip` is no-op).
-    pub fn new_pure(asm_staticlib_path: PathBuf) -> Self {
+    pub fn new_pure() -> Self {
         Self {
             xorin_chip_idx: u32::MAX,
             keccakf_op_chip_idx: u32::MAX,
             keccakf_perm_chip_idx: u32::MAX,
```
shuklaayush:
```suggestion
            xorin_chip_idx: NO_CHIP,
            keccakf_op_chip_idx: NO_CHIP,
            keccakf_perm_chip_idx: NO_CHIP,
```

extensions/keccak256/rvr/src/lib.rs
shuklaayush, line 86: it's fine for now but i feel like maybe we should use `Option` here, and do `unwrap_or(NO_CHIP)` when we need to convert it into a u32, eg. for interfacing with c code

shuklaayush (approved):

PR #2770, opened 2026-05-02
feat: implement rvr for algebra extension

gdmlcjs:
Implements the rvr feature for the Algebra extension. The rvr side of the Algebra extension is now also split into `ModularRvrExtension` and `Fp2RvrExtension`. A notable change is to have the C code for the Algebra extension which uses `libsecp256k1` to also unconditionally contain the C code needed in the ECC extension, since they are closely related and doing so would avoid configuration dependencies.

closes INT-7470, INT-7704


.github/workflows/guest-lib-tests.yml
```diff
@@ -37,6 +37,12 @@ jobs:
           - { name: "pairing", path: "pairing", rvr: false }
           - { name: "verify-stark", path: "verify-stark/circuit", rvr: false }
           - { name: "keccak256", path: "keccak256", rvr: true }
+          # TODO: add `{ name: "ff_derive", path: "ff_derive", rvr: true }` once
+          # the algebra k256/ECC TU coupling is resolved. Today algebra's
+          # `rvr_ext_k256.c` includes `rvr_ext_ecc.h` (it folds ECC k256 ops
+          # into the same TU so libsecp256k1 stays in one TU); registering
+          # ModularRvrExtension without EccRvrExtension makes `make` fail with
+          # `'rvr_ext_ecc.h' file not found`. See ModularRvrExtension::c_sources.
```
shuklaayush: hmm yeah this seems bad

extensions/algebra/rvr/src/lib.rs
```diff
@@ -551,12 +569,22 @@ impl<F: PrimeField32> RvrExtension<F> for AlgebraExtension {
...
+        // TODO: `rvr_ext_k256.c` folds ECC k256 ops into algebra's TU so
+        // libsecp256k1 stays in a single translation unit (see the comment
+        // at the top of `rvr_ext_k256_ec.h`). It `#include`s `rvr_ext_ecc.h`,
+        // which is owned by the ECC rvr extension. So registering
+        // `ModularRvrExtension` without `EccRvrExtension` builds/links with a
+        // missing-header error. Options: (1) gate the ECC fold-in behind a
+        // `-DRVR_EXT_K256_INCLUDE_ECC` cflag set by ECC's `extra_cflags()`;
+        // (2) emit `rvr_ext_k256.c` only when at least one configured
+        // modulus matches a known k256 field; (3) move the file into ECC
+        // and lose the modular k256 fast path when ECC isn't present.
```
shuklaayush: i see. yeah this is bad. 2 sounds like the better approach but not sure how difficult it is to implement

extensions/algebra/rvr/src/lib.rs
```diff
@@ -694,6 +722,97 @@ impl AlgebraExtension {
...
+                    })),
+                    source_loc: None,
+                }))
+            }
+            None => None,
+        }
+    }
+}
+
+// ── Fp2 extension ────────────────────────────────────────────────────────────
```
shuklaayush: can we move the modular and fp2 extension into separate files?

extensions/algebra/rvr/src/lib.rs
```diff
@@ -473,34 +469,34 @@ fn secp256k1_dir() -> PathBuf {
...
+    // has a latent non-determinism: the same modulus appearing in both
+    // `moduli` and `fp2_moduli` is processed by two independent rngs (each
+    // freshly seeded here), so for primes that fall through to
+    // rejection-sampling in `find_non_qr` (anything not `p ≡ 3 (mod 4)` or
+    // `p ≡ 5 (mod 8)`), the NQR for the same prime can differ between lists.
+    // If a future Fp2 phantom ever consumes `info.non_qr_bytes`, this will
+    // diverge from what the circuit computes. Either drop NQR computation
+    // for fp2 (use a `make_fp2_moduli` that fills num_limbs+modulus_bytes
+    // only), or share rng state with the modular list.
+    let mut rng = StdRng::from_seed([0u8; 32]);
```
shuklaayush: it feels like this should be a shared constant somewhere

extensions/algebra/utils/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+                continue;
+            }
+            let non_qr = val + 2u32;
+            // To check if non_qr is a quadratic nonresidue, we compute non_qr^((p-1)/2)
+            // If the result is p-1, then non_qr is a quadratic nonresidue
+            if non_qr.modpow(&exponent, modulus) == modulus - BigUint::one() {
+                return non_qr;
+            }
+        }
+    }
```
shuklaayush, line 44: not on you but i don't understand why we need randomness here. why not just iterate and return the first nqr. @jonathanpwang maybe you know

```rust
/// Returns a non-quadratic residue in the field of prime order `modulus`.
pub fn find_non_qr(modulus: &BigUint) -> BigUint {
    if modulus % 4u32 == BigUint::from(3u8) {
        return modulus - BigUint::one();
    } else if modulus % 8u32 == BigUint::from(5u8) {
        return BigUint::from(2u8);
    }
    // p ≡ 1 mod 8: search deterministically from 3 upward.
    let exponent = (modulus - BigUint::one()) >> 1;
    let p_minus_1 = modulus - BigUint::one();
    let mut candidate = BigUint::from(3u8);
    loop {
        if candidate.modpow(&exponent, modulus) == p_minus_1 {
            return candidate;
        }
        candidate += 1u32;
    }
}
```

extensions/algebra/rvr/src/lib.rs
```diff
@@ -694,6 +722,97 @@ impl AlgebraExtension {
...
+    fn staticlib_paths(&self) -> Vec<&Path> {
+        // The algebra-ffi staticlib is owned by `ModularRvrExtension`; do not
+        // emit it again here.
+        //
+        // TODO: Reconsider having `ModularRvrExtension` owning the single
+        // staticlib used by both `ModularRvrExtension` and `Fp2RvrExtension`.
+        // Alternatives: (1) split the FFI into separate `*-modular-ffi` and
+        // `*-fp2-ffi` staticlibs so each half owns its own; (2) deduplicate
+        // staticlib paths in `ExtensionRegistry::staticlib_paths()` so both
+        // halves can return the same path safely.
```
shuklaayush: keeping them separate seems like the right thing to do since they are treated as separate extensions

shuklaayush (approved):

extensions/algebra/utils/src/lib.rs
jonathanpwang: This is the Tonelli-Shanks algorithm. I think iterating takes longer?

gdmlcjs: I think the Tonelli-Shanks algorithm is used in a different function somewhere else. This function just finds a quadratic non-residue randomly from a prime field and I thought it doesn't use iteration because it might be slow for some primes, is that right? Because someone might (intentionally) use a prime that has all small numbers as quadratic residues?

gdmlcjs:
Remarks on Claude review comments:
1. The Fp2 extension by definition only uses quadratic field extensions generated by `i` where `i^2 = -1`.
3. Outdated comment based on stale cache.
4. `include_str!` paths are always tracked by Cargo's dependency scanner.

PR #2771, opened 2026-05-02
feat: Expose AIR column names

georgwiese:
This PR introduces the `ColumnsAir` trait and makes it required by all AIRs. Via this trait, AIRs can (optionally) provide a list of column names. At powdr, we use this to generate (somewhat) human-readable autoprecompiles for OpenVM ([example](https://github.com/powdr-labs/powdr/blob/main/openvm-riscv/tests/apc_snapshots/complex/rotate.txt)).

Unfortunately, this touches all AIRs and therefore a lot of files, but the changes are small and now change to `stark-backend` are needed.

This change represents most of our [diff to OpenVM](https://github.com/openvm-org/openvm/compare/v2.0.0-beta.2...powdr-labs:openvm:v2-powdr-beta.2-reorganized?expand=1).


crates/circuits/mod-builder/src/builder.rs
```diff
@@ -317,6 +317,7 @@ impl Deref for FieldExpr {
 
 impl<F: Field> BaseAirWithPublicValues<F> for FieldExpr {}
 impl<F: Field> PartitionedBaseAir<F> for FieldExpr {}
+impl<F: Field> ColumnsAir<F> for FieldExpr {}
```
georgwiese: Note that we didn't bother implementing it for AIRs we don't build autopecompiles for. This keeps the default implementation, which returns `None`.

crates/vm/src/system/memory/merkle/air.rs
```diff
@@ -18,6 +19,11 @@ pub struct MemoryMerkleAir<const CHUNK: usize> {
 }
 
 impl<const CHUNK: usize, F: Field> PartitionedBaseAir<F> for MemoryMerkleAir<CHUNK> {}
+impl<const CHUNK: usize, F: Field> ColumnsAir<F> for MemoryMerkleAir<CHUNK> {
```
georgwiese: Note that it would be possible to reduce these types of changes using a macro, but it becomes a bit ugly, especially in the presence of const generics.

See this discussion: https://github.com/powdr-labs/openvm/pull/60#discussion_r3166995122

crates/circuits/primitives/src/range_tuple/mod.rs
```diff
@@ -79,6 +79,7 @@ impl<const N: usize> RangeTupleCheckerAir<N> {
 }
 impl<F: Field, const N: usize> BaseAirWithPublicValues<F> for RangeTupleCheckerAir<N> {}
 impl<F: Field, const N: usize> PartitionedBaseAir<F> for RangeTupleCheckerAir<N> {}
+impl<F: Field, const N: usize> ColumnsAir<F> for RangeTupleCheckerAir<N> {}
```
jonathanpwang: why didn't you add columns here?

georgwiese:
I implemented the macro per https://github.com/powdr-labs/openvm/pull/60#discussion_r3189750401, but Claude has been annoying to navigate on this large diff. I'll do a review pass myself and am marking the PR as draft until then.

crates/circuits/primitives/src/range_tuple/mod.rs
georgwiese: Done now. We don't need it though (we're only interested in instruction AIRs), and it needed a manual implementation, because it cannot be derived from `RangeTupleColsRef` (because of `tuple: &'a [T]`).

georgwiese:
OK, should be clean now!

jonathanpwang:
Please mention in PR description / docs that default implementation is always safe, since this is just for labeling purposes.

Going to wait till @shuklaayush is back to review it, but overall I feel like the new macro makes it pretty clean

shuklaayush:
@claude review

shuklaayush (approved):
looks fine

PR #2774, opened 2026-05-11
feat: implement rvr for all extensions

gdmlcjs:
Implements the rvr feature for all extensions that are left - BigInt, Sha2, ECC, Pairing, Deferral. Code for tests and CI are also updated. Changes for the Deferral extension includes additions to the VM state used in rvr execution.

Closes INT-7465


crates/vm/src/arch/rvr/io.rs
```diff
@@ -184,45 +173,72 @@ pub unsafe extern "C" fn host_deferral_call_lookup(
...
-            .try_into()
-            .unwrap();
-    if let Some(ok) = io.deferral.call_entries.get(&(def_idx, ic)) {
-        std::ptr::copy_nonoverlapping(ok.as_ptr(), output_key_out, DEFERRAL_OUTPUT_KEY_BYTES);
-        1
-    } else {
-        0
-    }
+    let io = &mut *(ctx as *mut OpenVmIoState);
+    let ic: Vec<u8> = std::slice::from_raw_parts(input_commit, DEFERRAL_COMMIT_NUM_BYTES).to_vec();
```
shuklaayush: maybe `input_commit_raw` and `input_commit`

crates/vm/src/arch/rvr/io.rs
```diff
@@ -184,45 +173,72 @@ pub unsafe extern "C" fn host_deferral_call_lookup(
...
-            .try_into()
-            .unwrap();
-    if let Some(raw) = io.deferral.output_entries.get(&oc) {
-        debug_assert_eq!(raw.len(), expected_len as usize);
-        std::ptr::copy_nonoverlapping(raw.as_ptr(), output_raw_out, raw.len());
-        1
-    } else {
-        0
-    }
+    let oc: Vec<u8> = std::slice::from_raw_parts(output_commit, DEFERRAL_COMMIT_NUM_BYTES).to_vec();
```
shuklaayush: maybe `output_commit_raw` and `output_commit`

crates/vm/src/arch/rvr/io.rs
```diff
@@ -61,7 +50,8 @@ pub struct OpenVmHostCallbacks {
     pub reveal: extern "C" fn(*mut c_void, u32, u32, u32),
     pub hint_stream_set: unsafe extern "C" fn(*mut c_void, *const u8, u32),
     pub deferral_call_lookup: unsafe extern "C" fn(*mut c_void, u32, *const u8, *mut u8) -> i32,
-    pub deferral_output_lookup: unsafe extern "C" fn(*mut c_void, *const u8, *mut u8, u32) -> i32,
+    pub deferral_output_lookup:
+        unsafe extern "C" fn(*mut c_void, u32, *const u8, *mut u8, u32) -> i32,
 }
```
shuklaayush, line 55: why do we return `i32` instead of `bool`?

extensions/bigint/rvr/src/lib.rs
```diff
@@ -331,6 +331,10 @@ fn decode_imm<F: PrimeField32>(f: F) -> i32 {
     }
 }
 
+fn default_staticlib_path() -> PathBuf {
+    PathBuf::from(env!("RVR_BIGINT_FFI_STATICLIB"))
```
shuklaayush, line 335: i wonder if there's a way to avoid passing this as an env var

crates/vm/src/arch/vm.rs
```diff
@@ -110,9 +110,13 @@ pub enum GenerationError {
...
     pub input_stream: VecDeque<Vec<F>>,
     pub hint_stream: VecDeque<F>,
-    /// Stores cached deferred operation inputs and outputs. Each idx corresponds to a
+    /// Cached deferred operation inputs and outputs. Each idx corresponds to a
     /// unique function that is constrained outside the VM in its own deferral circuit.
     pub deferrals: Vec<DeferralState>,
+    #[cfg(feature = "rvr")]
+    pub deferral_fns: Vec<rvr::io::DeferralFnPtr>,
+    #[cfg(feature = "rvr")]
+    pub deferral_hash: Option<rvr::io::DeferralHashFn>,
```
shuklaayush, line 117: can you explain why we need to add this?

crates/vm/src/arch/rvr/io.rs
gdmlcjs: Oh I didn't think of that. It should definitely be changed.

crates/vm/src/arch/rvr/io.rs
gdmlcjs: I think using `bool` with C ABI is not perfectly stable.

extensions/bigint/rvr/src/lib.rs
gdmlcjs: I thought using an env var set by Cargo would be better than choosing a path to build the staticlib in and passing it around. + We can build the Rust staticlibs at openvm build time rather than openvm run time (guest program build time).

Or are you talking about just not using `env!`?

shuklaayush: yeah, doing it in build time is good. i was just wondering if there's a way to pass the path without using environment variables

crates/vm/src/arch/vm.rs
gdmlcjs: That code was added without proper thought to implement Deferral. I was able to remove that part after the change to make rvr execution use openvm `VmState`. It's going to be in the next PR. I probably should have added a TODO there too.

extensions/bigint/rvr/src/lib.rs
gdmlcjs: There seems to be no possible way without using env vars

gdmlcjs:
Remarks on Claude comments
1: In current Rust (including the 1.91.1 version used for openvm), panics inside extern "C" callbacks do properly abort the process. The stack trace might not be properly revealed but that is a non-critical issue that has a TODO which can be dealt with later.
3, 4: To be addressed in later refactor tasks.

shuklaayush (approved):

PR #2776, opened 2026-05-12
refactor: have rvr execution use openvm state

gdmlcjs:
- Changed rvr execution to use the existing openvm `VmState` instead of defining a new state struct and copying data between the two forms. The references and pointers to the fields in `VmState` are passed to the rvr execution functions so they can be used in C code.
- The Deferral extension now uses a callback registration system to expose the Deferral related data to C code instead of piggybacking on the same mechanism of `OpenVmHostCallbacks`. This is enabled for each extension so that they can have per-extension data and state that is maintained separately.

closes INT-7572

### DIFF
--- a/crates/rvr/rvr-openvm/c/openvm_io.h
+++ b/crates/rvr/rvr-openvm/c/openvm_io.h
@@ -14,10 +14,7 @@ void openvm_reveal(uint32_t src_val, uint32_t ptr, uint32_t offset);
 /* Extension hint stream access (called by extension FFI staticlibs). */
 void ext_hint_stream_set(const uint8_t* data, uint32_t len);
 
-/* Deferral extension callbacks (called by deferral FFI staticlib). */
-int ext_deferral_call_lookup(uint32_t def_idx, const uint8_t* input_commit,
-                             uint8_t* output_key_out);
-int ext_deferral_output_lookup(uint32_t def_idx, const uint8_t* output_commit,
-                               uint8_t* output_raw_out, uint32_t expected_len);
+/* Accessor for the openvm callback ctx pointer (the registered `OpenVmIoState`). */
+void* openvm_get_io_ctx(void);
 
 #endif /* OPENVM_IO_H */
--- a/crates/vm/src/arch/deferral.rs
+++ b/crates/vm/src/arch/deferral.rs
@@ -8,11 +8,6 @@ pub type InputCommit = Vec<u8>;
 pub type OutputCommit = Vec<u8>;
 
 /// A registered deferral closure: `input_raw → output_raw`.
-///
-/// Lives here (not in the deferral extension) because `Streams<F>` carries
-/// `Vec<Arc<DeferralFn>>` for rvr-mode lazy lookups, and `Streams<F>` is in
-/// `openvm-circuit`. Methods that need the deferral Poseidon2 hasher
-/// (`execute_deferral_fn`) stay in `openvm-deferral-circuit`.
 #[allow(clippy::type_complexity)]
 pub struct DeferralFn {
     f: Box<dyn Fn(&[u8]) -> OutputRaw + Send + Sync + 'static>,
--- a/crates/vm/src/arch/deferral.rs
+++ b/crates/vm/src/arch/deferral.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt::Debug};
 
 use serde::{Deserialize, Serialize};
 
@@ -6,6 +6,33 @@ pub type InputRaw = Vec<u8>;
 pub type OutputRaw = Vec<u8>;
 pub type InputCommit = Vec<u8>;
 pub type OutputCommit = Vec<u8>;
+
+/// A registered deferral closure: `input_raw → output_raw`.
+///
+/// Lives here (not in the deferral extension) because `Streams<F>` carries
+/// `Vec<Arc<DeferralFn>>` for rvr-mode lazy lookups, and `Streams<F>` is in
+/// `openvm-circuit`. Methods that need the deferral Poseidon2 hasher
+/// (`execute_deferral_fn`) stay in `openvm-deferral-circuit`.
+#[allow(clippy::type_complexity)]
+pub struct DeferralFn {
+    f: Box<dyn Fn(&[u8]) -> OutputRaw + Send + Sync + 'static>,
+}
+
+impl Debug for DeferralFn {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DeferralFn").finish()
+    }
+}
+
+impl DeferralFn {
+    pub fn new<FN: Fn(&[u8]) -> OutputRaw + Send + Sync + 'static>(f: FN) -> Self {
+        Self { f: Box::new(f) }
+    }
+
+    pub fn call_raw(&self, input_raw: &[u8]) -> OutputRaw {
+        (self.f)(input_raw)
+    }
+}
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum InputMapVal {
--- a/crates/vm/src/arch/rvr/execute.rs
+++ b/crates/vm/src/arch/rvr/execute.rs
@@ -22,13 +22,19 @@ use super::{
         metered_periodic_check, MeteredConfig, MeteredTracerData, RvrMeteredResult,
         SegmentationState, NO_LAST_PAGE,
     },
-    metered_cost::{prepare_metered_cost, MeteredCostConfig, MeteredCostData, PureTracerData},
+    metered_cost::{
+        prepare_metered_cost, MeteredCostConfig, MeteredCostData, PureTracerData,
+        RvrMeteredCostResult,
+    },
+    pure::RvrPureResult,
     state::{
-        init_rvr_state, init_rvr_state_with_metered, init_rvr_state_with_metered_cost,
-        MeteredCostState, PureState, TracerPtr,
+        init_rvr_state, init_rvr_state_with_metered, init_rvr_state_with_metered_cost, TracerPtr,
     },
 };
 use crate::{arch::VmState, system::memory::online::GuestMemory};
+
+type RegisterFn = unsafe extern "C" fn(*const OpenVmHostCallbacks);
+type ExecuteFn = unsafe extern "C" fn(*mut c_void);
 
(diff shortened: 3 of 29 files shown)


crates/rvr/rvr-openvm/c/openvm_io.h
shuklaayush, line 12: i wonder if we should move these as callbacks that are registered by the io extension like how we do for deferrals now

crates/vm/src/arch/rvr/execute.rs
```diff
@@ -48,89 +40,63 @@ pub enum ExecuteError {
...
+pub struct RvrPureResult {
     pub state: PureState,
-    pub memory: GuardedMemory,
-    pub instret: u64,
     pub suspended: bool,
 }
 
-/// Result of metered cost execution via rvr.
+/// `suspended` is `false` for unlimited runs.
 pub struct RvrMeteredCostResult {
```
shuklaayush: should this be in `metered_cost.rs`

crates/vm/src/arch/rvr/execute.rs
```diff
@@ -141,7 +107,18 @@ pub unsafe fn register_and_execute(
...
+/// # Safety
+///
+/// - `compiled` must contain a valid rvr-compiled shared library exporting the `rv_execute` symbol.
+/// - `state_ptr` must point to a valid, mutable RV32 state struct whose tracer variant matches the
+///   one compiled into the shared library.
+pub unsafe fn rv_execute(
+    compiled: &RvrCompiled,
+    state_ptr: *mut c_void,
+) -> Result<(), ExecuteError> {
     type ExecuteFn = unsafe extern "C" fn(*mut c_void);
```
shuklaayush: let's declare types at top

extensions/deferral/rvr/src/lib.rs
```diff
@@ -172,7 +203,135 @@ impl<F: PrimeField32> RvrExtension<F> for DeferralRvrExtension {
...
+
+    let output_commit: Vec<u8> = unsafe {
+        std::slice::from_raw_parts(output_commit_raw, DEFERRAL_COMMIT_NUM_BYTES).to_vec()
+    };
+    let Some(state) = io.deferrals.get(def_idx as usize) else {
+        return 0;
+    };
+    let raw = state.get_output(&output_commit);
+    // TODO: change these panics to something better to handle across the FFI boundary.
+    assert_eq!(raw.len(), expected_len as usize);
```
shuklaayush, line 335: hmm yeah

extensions/deferral/rvr/src/lib.rs
```diff
@@ -172,7 +203,135 @@ impl<F: PrimeField32> RvrExtension<F> for DeferralRvrExtension {
...
+
     fn staticlib_path(&self) -> &Path {
         &self.staticlib_path
     }
+
+    unsafe fn register_host_callbacks(
+        &self,
+        lib: &libloading::Library,
+    ) -> Result<(), ExtensionError> {
+        type RegisterFn = unsafe extern "C" fn(*const DeferralHostCallbacks);
```
shuklaayush: types at top

extensions/deferral/rvr/c/rvr_ext_deferral.c
shuklaayush, line 1: why the _lookup suffix?

crates/vm/src/arch/deferral.rs
```diff
@@ -7,6 +7,28 @@ pub type OutputRaw = Vec<u8>;
...
+
+impl DeferralFn {
+    pub fn new<FN: Fn(&[u8]) -> OutputRaw + Send + Sync + 'static>(f: FN) -> Self {
+        Self { f: Box::new(f) }
+    }
+
+    pub fn call_raw(&self, input_raw: &[u8]) -> OutputRaw {
+        (self.f)(input_raw)
+    }
+}
```
shuklaayush, line 30: can we move this back to deferral extension and have deferral rvr import deferral circuit?

crates/vm/src/arch/rvr/state.rs
```diff
@@ -1,86 +1,112 @@
...
+            "TracerPtr dereferenced before payload was assigned"
+        );
+        unsafe { &mut *self.0 }
+    }
+}
+
+pub type PureState = RvState<Rv32, PureTracer, InstretSuspender>;
+pub type MeteredCostState = RvState<Rv32, MeteredCostMeter, InstretSuspender>;
+pub type MeteredState = RvState<Rv32, MeteredTracer, InstretSuspender>;
+
```
shuklaayush: move to top

crates/vm/src/arch/rvr/mod.rs
```diff
@@ -18,11 +19,9 @@ pub use compile::{
 };
 pub use debug::{default_addr2line_cmd, GuestDebugMap};
 pub use execute::{
-    build_callbacks, build_io_state, execute, execute_metered, execute_metered_cost,
-    execute_metered_cost_with_limit, execute_with_limit, register_and_execute, ExecuteError,
-    RvrExecutionResult, RvrLimitedResult, RvrMeteredCostLimitedResult, RvrMeteredCostResult,
+    build_callbacks, execute, execute_metered, execute_metered_cost, register_openvm_callbacks,
+    rv_execute, ExecuteError, RvrMeteredCostResult, RvrPureResult,
```
shuklaayush: can you also remove functions like `compile_with_limit` etc.?

crates/vm/src/arch/rvr/execute.rs
```diff
@@ -155,318 +132,171 @@ pub unsafe fn register_and_execute(
...
-        deferral_fns,
-        deferral_hash,
-    );
-    io_state.hint_stream = hint_stream;
-    io_state.hint_pos = 0;
+    extensions: &ExtensionRegistry<F>,
+    vm_state: &mut VmState<F, GuestMemory>,
+    state: &mut RvState<Rv32, T, S>,
+    allow_suspended: bool,
+    failure_prefix: &str,
```
shuklaayush: maybe we can `map_err` and wrap it outside in the caller instead of passing this string

crates/vm/src/arch/rvr/execute.rs
```diff
@@ -155,318 +132,171 @@ pub unsafe fn register_and_execute(
...
-    deferral_hash: Option<DeferralHashFn>,
-) -> Result<RvrLimitedResult, ExecuteError> {
-    let mut memory = GuardedMemory::new(MEM_SIZE)?;
-    let mut tracer_data = PureTracerData;
-    let mut state = init_rvr_state(exe, &mut memory);
-    state.tracer = PureTracer(&mut tracer_data);
-    state.suspender.set_target(instruction_limit);
+    extensions: &ExtensionRegistry<F>,
+    vm_state: &mut VmState<F, GuestMemory>,
+    limit: Option<u64>,
```
shuklaayush:
```suggestion
    num_insns: Option<u64>,
```

crates/vm/src/arch/rvr/execute.rs
```diff
@@ -155,318 +132,171 @@ pub unsafe fn register_and_execute(
...
+    extensions: &ExtensionRegistry<F>,
+    vm_state: &mut VmState<F, GuestMemory>,
     metered_cost_config: MeteredCostConfig,
-    instruction_limit: u64,
-    deferrals: Vec<DeferralState>,
-    deferral_fns: Vec<DeferralFnPtr>,
-    deferral_hash: Option<DeferralHashFn>,
-) -> Result<RvrMeteredCostLimitedResult, ExecuteError> {
-    let mut memory = GuardedMemory::new(MEM_SIZE)?;
+    limit: Option<u64>,
```
shuklaayush: we probably don't need a limit for metered cost

crates/vm/src/arch/rvr/execute.rs
```diff
@@ -155,318 +132,171 @@ pub unsafe fn register_and_execute(
...
 
-fn report_failure(prefix: &str, outcome: ExecuteOutcome, pc: u32, instret: u64, exit_code: u8) {
+fn report_failure(prefix: &str, status: ExecutionStatus, pc: u32, instret: u64, exit_code: u8) {
     if std::env::var_os("RVR_OPENVM_DEBUG_EXEC_FAILURE").is_some() {
         eprintln!(
-            "[rvr-openvm] {prefix}: outcome={outcome:?}, pc={pc:#x}, instret={instret}, guest_exit_code={exit_code}"
+            "[rvr-openvm] {prefix}: status={status:?}, pc={pc:#x}, instret={instret}, guest_exit_code={exit_code}"
         );
     }
 }
```
shuklaayush: why do we need this?

shuklaayush (commented):
some comments. this is a great improvement

crates/vm/src/arch/rvr/execute.rs
gdmlcjs: Removed `failure_prefix` and instead `eprintln!`ed the string in the caller

crates/rvr/rvr-openvm/c/openvm_io.h
gdmlcjs: Yes that seems right so I made a ticket to do that in a follow up PR

crates/vm/src/arch/deferral.rs
gdmlcjs: It’s a bit of a problem because deferral circuit imports deferral rvr (`DeferralExtension::extend_rvr` needs to construct `DeferralRvrExtension`). I’d like to solve this but I don’t know what the best way is.

shuklaayush:
> 2. Per-element VecDeque::pop_front in host_hint_buffer hot path (low)
> 
> io.rs:114-119 — The old code did a single std::ptr::copy_nonoverlapping over a contiguous byte buffer. The new code pops elements one at a time:
> 
> for i in 0..nbytes {
>     let byte = io.hint_stream.pop_front().unwrap().as_canonical_u32() as u8;
>     unsafe { *dst.add(i) = byte };
> }
> 
> Each pop_front involves a bounds check, index math, and potential ring-buffer wrap-around. For small buffers this is fine, but for large hints (4 KB+) the per-element overhead vs. a single memcpy could be measurable. This is an inherent tradeoff of keeping data as VecDeque<F> rather than pre-converting, and may be acceptable — worth profiling if hint-heavy programs regress.

can you rebase/cherry pick the commits from this branch over base to resolve conflicts and also address the above comment?

shuklaayush (approved):
nice, thanks!

PR #2777, opened 2026-05-12
perf: split base alu chip 

GunaDD:
Closes INT-7670


GunaDD:
Closing this PR since it makes performance worse.

PR #2782, opened 2026-05-15
perf(metered): per-AIR need_rot in segmentation memory formula

shuklaayush:
The metered segmentation memory formula hardcodes `main_cell_secondary_weight = 1.5`, which the doc comment on `DEFAULT_MAIN_CELL_SECONDARY_WEIGHT` derives assuming `need_rot = true` for every AIR (`mat_eval_bytes = (h / 2^l_skip) × (2 × w) × sizeof(EF)`). For AIRs with `need_rot = false`, `fold_ple` allocates `1 × w`, not `2 × w`, so their secondary contribution is half.

This change threads per-AIR `need_rot` from `pk.vk.params` into `SegmentationCtx` and splits the secondary term: AIRs with `need_rot = true` use weight `1.5`, AIRs with `need_rot = false` use `0.75`. Main and interaction terms unchanged.

[benchmark comparison](https://github.com/axiom-crypto/openvm-eth/actions/runs/25931043565)

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/ctx.rs
@@ -33,6 +33,7 @@ impl<const PAGE_BITS: usize> MeteredCtx<PAGE_BITS> {
         air_names: Vec<String>,
         widths: Vec<usize>,
         interactions: Vec<usize>,
+        need_rot: Vec<bool>,
         config: &SystemConfig,
     ) -> Self {
         let (trace_heights, is_trace_height_constant): (Vec<u32>, Vec<bool>) =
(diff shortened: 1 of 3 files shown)


crates/vm/src/arch/execution_mode/metered/ctx.rs
```diff
@@ -33,6 +33,7 @@ impl<const PAGE_BITS: usize> MeteredCtx<PAGE_BITS> {
         air_names: Vec<String>,
         widths: Vec<usize>,
         interactions: Vec<usize>,
+        need_rot: Vec<bool>,
```
jonathanpwang, line 36: starting to get to the point we might want a zipped vec...

jonathanpwang (approved):
LGTM

PR #2787, opened 2026-05-16
test(DON'T MERGE): develop-v2.0.0-beta

jonathanpwang:


jonathanpwang:
cuda tests fail because develop-v2 has updated the cuda stream interface

PR #2793, opened 2026-05-19
refactor: use SegmentationCtx in rvr

gdmlcjs:
- Makes rvr metered execution use the existing `SegmentationCtx` of openvm instead of its own new structs and code. This fixes the discrepancy between rvr and interpreted/aot segmentation logic and resolves the issue of rvr making too many segments. (https://github.com/axiom-crypto/openvm-eth/actions/runs/26112543464)
- Fixes the calculation of `num_insns` in segments that are used as segment boundaries. The instruction counts were recorded as multiples of `segment_check_insns` (1000) that didn't map to the actual basic block boundaries. Addresses the problem of overflowing GPU memory. (https://github.com/axiom-crypto/openvm-eth/actions/runs/26127167098)

closes INT-7835

### DIFF
--- a/crates/vm/src/arch/rvr/metered.rs
+++ b/crates/vm/src/arch/rvr/metered.rs
@@ -1,7 +1,7 @@
 //! Per-chip metered execution: page tracking and segmentation
 //! matching OpenVM's `MeteredCtx`.
 
-use std::sync::Arc;
+use std::{ffi::c_void, sync::Arc};
 
 use openvm_instructions::{
     exe::VmExe, riscv::RV32_MEMORY_AS, LocalOpcode, SystemOpcode, VmOpcode, DEFERRAL_AS,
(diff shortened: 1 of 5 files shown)


crates/vm/src/arch/rvr/metered.rs
```diff
@@ -410,186 +258,85 @@ impl SegmentationState {
...
+                    .map(|(&height, name)| format!("  {name} = {height}"))
+                    .collect::<Vec<_>>()
+                    .join("\n");
+                tracing::warn!(
+                    "Segment initialized with heights that exceed limits\n\
+                     instret={instret}\n\
+                     trace_heights=[\n{trace_heights_str}\n]"
+                );
+            }
+        }
```
shuklaayush, line 328: is it possible to reduce duplication and share more of the logic with crates/vm/src/arch/execution_mode/metered/ctx.rs?

shuklaayush: maybe we can directly use it or we have to create common helper functions over there

shuklaayush (approved):

PR #2794, opened 2026-05-19
feat(v2.1-rv64): switch memory bus blocks to u16 cells

shuklaayush:
# Switch memory bus blocks to u16 cells

PR 1 of the memory-bus-u16 split. Atomic flip of the memory-bus block from `8 u8 cells` to `4 u16 cells`. The bus interaction is now fixed-shape (`BLOCK_FE_WIDTH = 4`), and byte-shaped chips pack their u8 columns into bus cells via a single helper.

## Why

`MemoryBus` interactions previously carried 8 field-element-encoded u8s per message. RV64 cell-typed memory (registers, RV memory, public values) is naturally u16-aligned, and packing two bytes per bus cell halves the bus payload width without losing constraint power. This PR makes the bus shape a single hard constant (`BLOCK_FE_WIDTH`) so the bridge API and downstream chips no longer need a `const N` generic for it.

Follow-up PRs will use the wider cells to narrow trace columns, prune redundant range checks, reshape SHA-2 / Keccak state, and revisit deferral.

### Memory layout constants

Files: `crates/vm/src/arch/config.rs`, `crates/vm/cuda/include/system/memory/params.cuh`.

- Single source of truth: `U16_CELL_SIZE = size_of::<u16>()` (intrinsic) drives `BUS_PTR_SCALE`; `BLOCK_FE_WIDTH = 4` is the design choice; `MEMORY_BLOCK_BYTES`, `BUS_BLOCK_STRIDE`, `BUS_LEAF_STRIDE`, `DIGEST_WIDTH` all derive.
- Cross-constant equality asserts removed where the derivation makes them true by construction.
- One divisibility guard kept on `DIGEST_WIDTH.is_multiple_of(BLOCK_FE_WIDTH)`.
- CUDA-side `params.cuh` mirrors the constants and includes ASCII diagrams for the u16-celled AS layout (RV64) and the F-celled AS layout (`DEFERRAL_AS`), showing how `byte_ptr`, `cell_idx`, and `bus_ptr` relate.

### Layout reference

Terminology:

- **Cell** — one storage word in an address space (`u16` for RV64 ASes, `F` for `DEFERRAL_AS`).
- **Block** — one memory-bus message: `BLOCK_FE_WIDTH` cells = `MEMORY_BLOCK_BYTES` bytes.
- **Digest** — output of one Poseidon2 compression, `DIGEST_WIDTH` cells; also one merkle leaf.
- **Leaf** — one merkle-tree leaf = one Poseidon2 half = `DIGEST_WIDTH` cells = `BLOCKS_PER_LEAF` blocks.

<details>
<summary><b>u16-celled AS layout</b> (<code>RV64_REGISTER_AS</code>, <code>RV64_MEMORY_AS</code>, <code>PUBLIC_VALUES_AS</code>)</summary>

One merkle leaf = 16 bytes = 8 u16 cells = 2 bus blocks:

```text
  byte_ptr:     0   1   2   3   4   5   6   7   8   9   10  11  12  13  14  15
               ┌───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┐
  u8 storage:  │b0 │b1 │b2 │b3 │b4 │b5 │b6 │b7 │b8 │b9 │b10│b11│b12│b13│b14│b15│
               └───┴───┴───┴───┴───┴───┴───┴───┴───┴───┴───┴───┴───┴───┴───┴───┘
               ╰──────╯╰──────╯╰──────╯╰──────╯╰──────╯╰──────╯╰──────╯╰──────╯   u16 cells (LE)
  cell_idx:        0       1       2       3       4       5       6       7
  bus_ptr:         0       2       4       6       8      10      12      14
               ╰─────────── block 0 ──────────╯╰─────────── block 1 ──────────╯
               ╰────────────────── one merkle leaf / digest ──────────────────╯
```

- `byte_ptr = U16_CELL_SIZE * cell_idx` (coincides with `bus_ptr` for u16 ASes)
- `bus_ptr  = BUS_PTR_SCALE  * cell_idx`
- Block stride on bus: `BUS_BLOCK_STRIDE = 8  = BUS_PTR_SCALE * BLOCK_FE_WIDTH`
- Leaf stride on bus:  `BUS_LEAF_STRIDE  = 16 = BUS_PTR_SCALE * DIGEST_WIDTH`

</details>

<details>
<summary><b>F-celled AS layout</b> (<code>DEFERRAL_AS</code>)</summary>

Each cell holds one `F` element (`size_of::<F>()` bytes on host; 4 for BabyBear). The bus addresses cells directly, so cell / block / leaf counts and `bus_ptr` / `cell_idx` mappings match the u16 AS.

```text
  byte_ptr:       0       4       8       12      16      20      24      28
               ┌───────┬───────┬───────┬───────┬───────┬───────┬───────┬───────┐
  F storage:   │  F0   │  F1   │  F2   │  F3   │  F4   │  F5   │  F6   │  F7   │
               └───────┴───────┴───────┴───────┴───────┴───────┴───────┴───────┘
  cell_idx:        0       1       2       3       4       5       6       7
  bus_ptr:         0       2       4       6       8      10      12      14
               ╰─────────── block 0 ──────────╯╰─────────── block 1 ──────────╯
               ╰────────────────── one merkle leaf / digest ──────────────────╯
```

- `byte_ptr = size_of::<F>() * cell_idx` (host-memory stride; ≠ `bus_ptr`)
- `bus_ptr  = BUS_PTR_SCALE   * cell_idx` (same formula as u16 ASes)
- `BUS_BLOCK_STRIDE` and `BUS_LEAF_STRIDE` unchanged — the bus addresses cells, not bytes.

</details>

### Address-space cell types

- `RV64_REGISTER_AS`, `RV64_MEMORY_AS`, `PUBLIC_VALUES_AS` switch to u16-celled.
- `DEFERRAL_AS` stays F-celled (moves to u16 in a follow-up).

### Persistent memory and merkle

- Boundary AIR, merkle tree, merkle public-values, and CUDA inventory rewritten for 4-cell-per-block u16 leaves (`BLOCKS_PER_LEAF = DIGEST_WIDTH / BLOCK_FE_WIDTH = 2`).
- `assert_public_values_shape::<DIGEST_WIDTH>(num_public_values_bytes)` and `SystemConfig::public_values_cell_count()` formalize the public-values byte-count → cell-count contract at construction and at every read site.

### Offline-checker bridge

`crates/vm/src/system/memory/offline_checker/bridge.rs`.

- `MemoryBridge::read` / `write` drop the `const N` generic; bus payload is always `[T; BLOCK_FE_WIDTH]`.
- Unified `MemoryWriteAuxInput` lets `write` accept either an `&MemoryWriteAuxCols<V, BLOCK_FE_WIDTH>` (standard case) or `from_prev_data_exprs(&base_aux, prev_data_expr)` for chips that materialize prev_data in their own columns (Keccak, SHA-2, LoadStore).
- Added `pack_u8_block` (AB::Expr) and `pack_u8_block_value` (F-value form) so byte-shaped chips and testing helpers produce bus messages identical to what u16 chips emit natively.
- Removed `read_or_immediate` / `MemoryReadOrImmediateOperation` (an RV32 pattern superseded by `RV64_IMM_AS`).

### TracingMemory / GuestMemory byte view

- `read_bytes` / `write_bytes` (cell-type-aware, BLOCK-aligned) for raw byte access against u16-celled storage.
- Deferral-specific helpers (`timed_read_deferral`, `tracing_read_deferral`, …) for the F-celled AS.

### Chip call-sites

~50 sites across riscv, riscv-adapters, keccak256, sha2, deferral.

- Every `memory_bridge.read` / `.write` now hands the bridge a `BLOCK_FE_WIDTH`-shaped payload.
- Byte-shaped chips wrap their `[T; 8]` u8 columns with `pack_u8_block`.
- Register-read paths still zero-extend the low 32 bits to 8 bytes, then pack.
- Per-AS metered leaf-bit shift in `execution_mode/metered/memory_ctx.rs` (`U16_AS_LEAF_PTR_BITS = 4` for byte-pointer ASes, `CELL_AS_LEAF_PTR_BITS = 3` for cell-pointer ASes), so RV64 byte pointers produce the right leaf label.

### Algebra / ECC adapter cleanup

No-op semantic; finishes upstream removal.

- Dropped phantom `BLOCK_SIZE` const generic from `ModularAir`/`ModularChip`, `Fp2Air`/`Fp2Chip`, `WeierstrassAir`/`WeierstrassChip` and their `get_*_air` / `get_*_chip` helpers.
- The generic became unused after `Rv64VecHeapAdapter*` dropped its `BLOCK_SIZE` parameter; keeping it was a phantom that broke inference at call sites.

### Testing infrastructure

- `MemoryTester` / `DeviceMemoryTester` accept either `N = BLOCK_FE_WIDTH` (cell index) or `N = MEMORY_BLOCK_BYTES` (byte pointer in a u16 AS) and route accordingly; the byte path packs the result so the dummy chip's bus message matches a byte-shaped chip's exactly.
- `pack_u8_block_value` is shared between the testing helper and the CUDA boundary-chip test that previously hand-rolled the same packing.

### CUDA

- `params.cuh` constants + static_asserts kept in sync.
- Kernel memory-aux fill paths updated for the new block shape.

## Out of scope (follow-up PRs)

- Trace column narrowing now that u16-celled storage matches column width.
- Redundant range-check removal on byte-shaped chips.
- Dead bitwise-lookup cleanup.
- SHA-2 / Keccak state reshaping to natively-u16 layouts.
- Moving `DEFERRAL_AS` to u16.
- Deferral commit / output-hash changes.

## Migration notes

- `MemoryBridge::read::<N>` / `::write::<N>` calls: drop `<N>`; pass a `BLOCK_FE_WIDTH`-shaped payload. Byte-shaped data goes through `pack_u8_block` (constraint side) and `pack_u8_block_value` (witness side).
- `read_or_immediate` callers: switch to `RV64_IMM_AS`-based addressing.
- `MemoryWriteAuxCols<V, N>` with `N != BLOCK_FE_WIDTH`: use `MemoryWriteAuxInput::from_prev_data_exprs(&base_aux, prev_data)` to supply prev_data inline.
- `ModularAir`, `ModularChip`, `Fp2Air`, `Fp2Chip`, `WeierstrassAir`, `WeierstrassChip` now take one fewer const generic (`BLOCK_SIZE` dropped).

Resolves INT-7829

### DIFF
--- a/crates/vm/cuda/src/system/memory/merkle_tree.cu
+++ b/crates/vm/cuda/src/system/memory/merkle_tree.cu
@@ -25,7 +25,7 @@ __global__ void merkle_tree_init(uint8_t *__restrict__ data, digest_t *__restric
 
     Fp cells[CELLS] = {0};
 #pragma unroll
-    for (size_t i = 0; i < CELLS_OUT; ++i) {
+    for (size_t i = 0; i < DIGEST_WIDTH; ++i) {
         if constexpr (ADDR_SPACE_IDX + 1 == DEFERRAL_AS) {
             cells[i] = reinterpret_cast<Fp *>(data)[DIGEST_WIDTH * gid + i];
         } else {
--- a/crates/vm/cuda/src/system/memory/merkle_tree.cu
+++ b/crates/vm/cuda/src/system/memory/merkle_tree.cu
@@ -2,30 +2,36 @@
 #include "poseidon2.cuh"
 #include "primitives/shared_buffer.cuh"
 #include "primitives/trace_access.h"
+#include "primitives/utils.cuh"
+#include "system/memory/params.cuh"
 
 #include <cub/cub.cuh>
 
 using poseidon2::poseidon2_mix;
 
 struct alignas(32) digest_t {
-    Fp cells[CELLS_OUT];
+    Fp cells[DIGEST_WIDTH];
 };
 
 #define COPY_DIGEST(dst, src) memcpy(dst, src, sizeof(digest_t))
 
 // `ADDR_SPACE_IDX` is the address space minus `ADDR_SPACE_OFFSET` (which is 1)
+//
+// DEFERRAL_AS stores Fp cells directly.
+// Non-deferral address spaces store u16 cells in little-endian byte order.
 template <int ADDR_SPACE_IDX>
 __global__ void merkle_tree_init(uint8_t *__restrict__ data, digest_t *__restrict__ out) {
     auto gid = blockDim.x * blockIdx.x + threadIdx.x;
 
     Fp cells[CELLS] = {0};
-    // TODO: revisit when we sort out address space handling
 #pragma unroll
     for (size_t i = 0; i < CELLS_OUT; ++i) {
-        if constexpr (ADDR_SPACE_IDX < 3) {
-            cells[i] = Fp(data[CELLS_OUT * gid + i]);
+        if constexpr (ADDR_SPACE_IDX + 1 == DEFERRAL_AS) {
+            cells[i] = reinterpret_cast<Fp *>(data)[DIGEST_WIDTH * gid + i];
         } else {
-            cells[i] = reinterpret_cast<Fp *>(data)[CELLS_OUT * gid + i];
+            // u16 cells, little-endian.
+            auto byte_off = U16_CELL_SIZE * (DIGEST_WIDTH * gid + i);
+            cells[i] = Fp(u16_from_bytes_le(data + byte_off));
         }
     }
 
--- a/crates/vm/src/arch/config.rs
+++ b/crates/vm/src/arch/config.rs
@@ -20,7 +20,10 @@ use crate::{
         Arena, ChipInventoryError, ExecutorInventory, ExecutorInventoryError,
     },
     system::{
-        memory::{merkle::public_values::PUBLIC_VALUES_AS, num_memory_airs, POINTER_MAX_BITS},
+        memory::{
+            merkle::public_values::{assert_public_values_shape, PUBLIC_VALUES_AS},
+            num_memory_airs, DIGEST_WIDTH, POINTER_MAX_BITS,
+        },
         SystemChipComplex,
     },
 };
@@ -33,6 +36,65 @@ pub const DEFAULT_MAX_NUM_PUBLIC_VALUES: usize = 32;
 pub const POSEIDON2_WIDTH: usize = 16;
 /// Offset for address space indices. This is used to distinguish between different memory spaces.
 pub const ADDR_SPACE_OFFSET: u32 = 1;
+
+pub const OPENVM_DEFAULT_INIT_FILE_BASENAME: &str = "openvm_init";
+pub const OPENVM_DEFAULT_INIT_FILE_NAME: &str = "openvm_init.rs";
+
+// Memory-layout constants. Mirror the CUDA-side constants in
+// `crates/vm/cuda/include/system/memory/params.cuh` (which also contains the
+// byte/cell/block/leaf layout diagram).
+//
+// Terminology:
+//   Cell    one storage word in an address space (u16 for RV64 ASes, F for
+//           DEFERRAL_AS).
+//   Block   the unit of one memory-bus message: BLOCK_FE_WIDTH cells =
+//           MEMORY_BLOCK_BYTES bytes.
+//   Digest  the output of one Poseidon2 compression (DIGEST_WIDTH cells); also
+//           one merkle leaf.
+
+/// Host byte width of one u16-celled storage cell.
+pub const U16_CELL_SIZE: usize = size_of::<u16>();
+
+// TODO: replace with `p3_util::log2_strict_usize` once p3-util is bumped to
+// >= 0.4.3 (where it becomes `const fn`).
+pub(crate) const fn const_log2_strict_usize(value: usize) -> usize {
+    assert!(value.is_power_of_two(), "value must be a power of two");
+    value.ilog2() as usize
+}
+
+/// Normalized memory-bus pointer scale: `bus_ptr = BUS_PTR_SCALE * cell_idx`.
+/// Equals `U16_CELL_SIZE` so byte pointers in u16-celled ASes coincide with bus
+/// pointers.
+pub const BUS_PTR_SCALE: usize = U16_CELL_SIZE;
+
+/// log2 of [`BUS_PTR_SCALE`].
+pub const BUS_PTR_SCALE_BITS: usize = const_log2_strict_usize(BUS_PTR_SCALE);
+
+/// Converts address-space pointer bits to address-space cell-index bits.
+pub const fn cell_index_bits_from_pointer_max_bits(pointer_max_bits: usize) -> usize {
+    pointer_max_bits - BUS_PTR_SCALE_BITS
+}
+
+/// Converts address-space cell-index bits to address-space pointer bits.
+pub const fn pointer_max_bits_for_cell_index_bits(cell_index_bits: usize) -> usize {
+    cell_index_bits + BUS_PTR_SCALE_BITS
+}
+
+/// Cells per memory-bus block.
+pub const BLOCK_FE_WIDTH: usize = 4;
+
+/// Bytes per memory-bus block (one RV64 8-byte word pair).
+pub const MEMORY_BLOCK_BYTES: usize = BLOCK_FE_WIDTH * U16_CELL_SIZE;
+
+/// Bus-pointer delta between consecutive blocks.
+pub const BUS_BLOCK_STRIDE: usize = BUS_PTR_SCALE * BLOCK_FE_WIDTH;
+
+/// Default byte-capacity for `RV64_MEMORY_AS` in `MemoryConfig::default`.
+pub const DEFAULT_RV64_MEMORY_BYTE_CAPACITY: usize = 1 << 29;
+
+/// Number of registers in the RV64 register file.
+pub const NUM_RV64_REGISTERS: usize = 32;
+
 /// Returns a Poseidon2 config for the VM.
 pub fn vm_poseidon2_config<F: Field>() -> Poseidon2Config<F> {
     Poseidon2Config::default()
@@ -109,12 +171,6 @@ where
 {
 }
 
-pub const OPENVM_DEFAULT_INIT_FILE_BASENAME: &str = "openvm_init";
-pub const OPENVM_DEFAULT_INIT_FILE_NAME: &str = "openvm_init.rs";
-/// Default block size (in bytes) for memory bus interactions. Memory is always read/written a
-/// full block at a time even when the instruction accesses fewer bytes (e.g. RISC-V `lb`/`lh`).
-pub const DEFAULT_BLOCK_SIZE: usize = 8;
-
 /// Trait for generating a init.rs file that contains a call to moduli_init!,
 /// complex_init!, sw_init! with the supported moduli and curves.
 /// Should be implemented by all VM config structs.
@@ -154,8 +210,7 @@ pub trait AddressSpaceHostLayout {
 
     /// # Safety
     /// - This function must only be called when `value` is guaranteed to be of size `self.size()`.
-    /// - Alignment of `value` must be a multiple of the alignment of `F`.
-    /// - The field type `F` must be plain old data.
+    /// - For raw field-cell layouts, `value` must be aligned for `F` and contain a valid `F`.
     unsafe fn to_field<F: Field>(&self, value: &[u8]) -> F;
 }
 
@@ -179,10 +234,15 @@ impl Default for MemoryConfig {
     fn default() -> Self {
         let mut addr_spaces =
             Self::empty_address_space_configs((1 << 3) + ADDR_SPACE_OFFSET as usize);
-        const MAX_CELLS: usize = 1 << 29;
-        addr_spaces[RV64_REGISTER_AS as usize].num_cells = 32 * size_of::<u64>();
-        addr_spaces[RV64_MEMORY_AS as usize].num_cells = MAX_CELLS;
-        addr_spaces[PUBLIC_VALUES_AS as usize].num_cells = DEFAULT_MAX_NUM_PUBLIC_VALUES;
+        // RV64 register, memory, and public-values address spaces use u16 storage cells.
+        // Derive each capacity as a count of those cells.
+        // 32 x 8-byte registers = 256 bytes = 128 u16 cells.
+        addr_spaces[RV64_REGISTER_AS as usize].num_cells =
+            NUM_RV64_REGISTERS * size_of::<u64>() / U16_CELL_SIZE;
+        addr_spaces[RV64_MEMORY_AS as usize].num_cells =
+            DEFAULT_RV64_MEMORY_BYTE_CAPACITY / U16_CELL_SIZE;
+        addr_spaces[PUBLIC_VALUES_AS as usize].num_cells =
+            DEFAULT_MAX_NUM_PUBLIC_VALUES / U16_CELL_SIZE;
         Self::new(3, addr_spaces, POINTER_MAX_BITS, 29, 17)
     }
 }
@@ -193,11 +253,13 @@ impl MemoryConfig {
         let mut addr_spaces =
             vec![AddressSpaceHostConfig::new(0, MemoryCellType::field32()); num_addr_spaces];
         addr_spaces[RV64_IMM_AS as usize] = AddressSpaceHostConfig::new(0, MemoryCellType::Null);
-        addr_spaces[RV64_REGISTER_AS as usize] = AddressSpaceHostConfig::new(0, MemoryCellType::U8);
+        addr_spaces[RV64_REGISTER_AS as usize] =
+            AddressSpaceHostConfig::new(0, MemoryCellType::U16);
 
-        addr_spaces[RV64_MEMORY_AS as usize] = AddressSpaceHostConfig::new(0, MemoryCellType::U8);
+        addr_spaces[RV64_MEMORY_AS as usize] = AddressSpaceHostConfig::new(0, MemoryCellType::U16);
 
-        addr_spaces[PUBLIC_VALUES_AS as usize] = AddressSpaceHostConfig::new(0, MemoryCellType::U8);
+        addr_spaces[PUBLIC_VALUES_AS as usize] =
+            AddressSpaceHostConfig::new(0, MemoryCellType::U16);
 
         addr_spaces
     }
@@ -220,9 +282,9 @@ pub struct SystemConfig {
     pub max_constraint_degree: usize,
     /// Memory configuration
     pub memory_config: MemoryConfig,
-    /// Public values are stored in a special address space.
-    /// `num_public_values` indicates the number of allowed addresses in that address space.
-    pub num_public_values: usize,
+    /// Byte capacity of the user public-values address space. The `PUBLIC_VALUES_AS`
+    /// AS is u16-celled, so the underlying `num_cells` is `bytes / U16_CELL_SIZE`.
+    pub num_public_values_bytes: usize,
     /// Whether to collect detailed profiling metrics.
     /// **Warning**: this slows down the runtime.
     pub profiling: bool,
@@ -238,17 +300,19 @@ impl SystemConfig {
     pub fn new(
         max_constraint_degree: usize,
         mut memory_config: MemoryConfig,
-        num_public_values: usize,
+        num_public_values_bytes: usize,
     ) -> Self {
         assert!(
             memory_config.timestamp_max_bits <= 29,
             "Timestamp max bits must be <= 29 for LessThan to work in 31-bit field"
         );
-        memory_config.addr_spaces[PUBLIC_VALUES_AS as usize].num_cells = num_public_values;
+        assert_public_values_shape::<DIGEST_WIDTH>(num_public_values_bytes);
+        memory_config.addr_spaces[PUBLIC_VALUES_AS as usize].num_cells =
+            num_public_values_bytes / U16_CELL_SIZE;
         Self {
             max_constraint_degree,
             memory_config,
-            num_public_values,
+            num_public_values_bytes,
             profiling: false,
             segmentation_config: SegmentationConfig::default(),
         }
@@ -262,10 +326,21 @@ impl SystemConfig {
         )
     }
 
-    pub fn with_public_values(mut self, num_public_values: usize) -> Self {
-        self.num_public_values = num_public_values;
-        self.memory_config.addr_spaces[PUBLIC_VALUES_AS as usize].num_cells = num_public_values;
+    pub fn with_public_values_bytes(mut self, num_public_values_bytes: usize) -> Self {
+        assert_public_values_shape::<DIGEST_WIDTH>(num_public_values_bytes);
+        self.num_public_values_bytes = num_public_values_bytes;
+        self.memory_config.addr_spaces[PUBLIC_VALUES_AS as usize].num_cells =
+            num_public_values_bytes / U16_CELL_SIZE;
         self
+    }
+
+    /// Number of u16 cells backing the user public-values address space.
+    /// Shape validation happens at `SystemConfig` construction
+    /// (see [`assert_public_values_shape`]); this is a plain accessor.
+    #[inline]
+    pub fn public_values_cell_count(&self) -> usize {
+        debug_assert!(self.num_public_values_bytes.is_multiple_of(U16_CELL_SIZE));
+        self.num_public_values_bytes / U16_CELL_SIZE
     }
 
     pub fn with_max_segment_len(mut self, max_segment_len: usize) -> Self {
@@ -350,8 +425,10 @@ impl AddressSpaceHostConfig {
 pub enum MemoryCellType {
     Null,
     U8,
+    /// U16 cells are stored as two little-endian bytes in linear byte storage:
+    /// cell `k` is decoded from `bytes[2 * k]` and `bytes[2 * k + 1]`.
     U16,
-    /// Represented in little-endian format.
+    /// U32 cells are stored as four little-endian bytes in linear byte storage.
     U32,
     /// `size` is the size in bytes of the native field type. This should not exceed 8.
     F {
@@ -380,8 +457,7 @@ impl AddressSpaceHostLayout for MemoryCellType {
 
     /// # Safety
     /// - This function must only be called when `value` is guaranteed to be of size `self.size()`.
-    /// - Alignment of `value` must be a multiple of the alignment of `F`.
-    /// - The field type `F` must be plain old data.
+    /// - For `F` cells, `value` must be aligned for `F` and contain a valid `F`.
     ///
     /// # Panics
     /// If the value is of integer type and overflows the field.
@@ -389,8 +465,16 @@ impl AddressSpaceHostLayout for MemoryCellType {
         match self {
             Self::Null => unreachable!(),
             Self::U8 => F::from_u8(*value.get_unchecked(0)),
-            Self::U16 => F::from_u16(core::ptr::read(value.as_ptr() as *const u16)),
-            Self::U32 => F::from_u32(core::ptr::read(value.as_ptr() as *const u32)),
+            Self::U16 => F::from_u16(u16::from_le_bytes([
+                *value.get_unchecked(0),
+                *value.get_unchecked(1),
+            ])),
+            Self::U32 => F::from_u32(u32::from_le_bytes([
+                *value.get_unchecked(0),
+                *value.get_unchecked(1),
+                *value.get_unchecked(2),
+                *value.get_unchecked(3),
+            ])),
             Self::F { .. } => core::ptr::read(value.as_ptr() as *const F),
         }
     }
--- a/crates/vm/src/arch/config.rs
+++ b/crates/vm/src/arch/config.rs
@@ -62,23 +62,16 @@ pub(crate) const fn const_log2_strict_usize(value: usize) -> usize {
     value.ilog2() as usize
 }
 
-/// Normalized memory-bus pointer scale: `bus_ptr = BUS_PTR_SCALE * cell_idx`.
+/// Normalized memory-bus pointer scale: `bus_ptr = BUS_PTR_SCALE * ptr`.
 /// Equals `U16_CELL_SIZE` so byte pointers in u16-celled ASes coincide with bus
 /// pointers.
 pub const BUS_PTR_SCALE: usize = U16_CELL_SIZE;
 
+/// log2 of [`U16_CELL_SIZE`].
+pub const U16_CELL_SIZE_BITS: usize = const_log2_strict_usize(U16_CELL_SIZE);
+
 /// log2 of [`BUS_PTR_SCALE`].
-pub const BUS_PTR_SCALE_BITS: usize = const_log2_strict_usize(BUS_PTR_SCALE);
-
-/// Converts address-space pointer bits to address-space cell-index bits.
-pub const fn cell_index_bits_from_pointer_max_bits(pointer_max_bits: usize) -> usize {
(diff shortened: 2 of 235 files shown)


crates/vm/src/arch/config.rs
```diff
@@ -238,17 +300,19 @@ impl SystemConfig {
...
+        num_public_values_bytes: usize,
     ) -> Self {
         assert!(
             memory_config.timestamp_max_bits <= 29,
             "Timestamp max bits must be <= 29 for LessThan to work in 31-bit field"
         );
-        memory_config.addr_spaces[PUBLIC_VALUES_AS as usize].num_cells = num_public_values;
+        assert_public_values_shape::<DIGEST_WIDTH>(num_public_values_bytes);
+        memory_config.addr_spaces[PUBLIC_VALUES_AS as usize].num_cells =
+            num_public_values_bytes / U16_CELL_SIZE;
```
jonathanpwang: do we need div_ceil or an assert

crates/vm/src/arch/config.rs
```diff
@@ -109,12 +171,6 @@ where
 {
 }
 
-pub const OPENVM_DEFAULT_INIT_FILE_BASENAME: &str = "openvm_init";
-pub const OPENVM_DEFAULT_INIT_FILE_NAME: &str = "openvm_init.rs";
-/// Default block size (in bytes) for memory bus interactions. Memory is always read/written a
-/// full block at a time even when the instruction accesses fewer bytes (e.g. RISC-V `lb`/`lh`).
-pub const DEFAULT_BLOCK_SIZE: usize = 8;
```
jonathanpwang, line 116: why did we rename this constant once more T_T

shuklaayush: i got ai to write a base set of constants that make all this easier to understand. it might have gone a bit overboard but i feel it's much cleaner now

crates/vm/src/arch/config.rs
shuklaayush: the assert above checks that
```rust
assert!(
    num_public_values_bytes.is_multiple_of(U16_CELL_SIZE),
    "num_public_values_bytes must be a multiple of U16_CELL_SIZE"
);
```

crates/vm/src/arch/state.rs
```diff
@@ -205,11 +207,19 @@ where
         addr_space: u32,
         ptr: u32,
     ) -> [T; BLOCK_SIZE] {
-        // SAFETY:
-        // - T is stack-allocated repr(C) or repr(transparent), usually u8 or F where F is the base
-        //   field
-        // - T is the exact memory cell type for this address space, satisfying the type requirement
-        unsafe { self.memory.read(addr_space, ptr) }
+        if size_of::<T>() == 1 {
```
GunaDD: what are the possible values of `T` here ?

crates/vm/src/system/memory/offline_checker/bridge.rs
```diff
@@ -332,3 +289,36 @@ impl MemoryOfflineChecker {
...
     }
 }
+
+/// Pack `MEMORY_BLOCK_BYTES` byte expressions into `BLOCK_FE_WIDTH` bus cells.
+pub fn pack_u8_block<AB: InteractionBuilder>(
+    data: &[AB::Expr; MEMORY_BLOCK_BYTES],
+) -> [AB::Expr; BLOCK_FE_WIDTH] {
+    let mut out: [AB::Expr; BLOCK_FE_WIDTH] = std::array::from_fn(|_| AB::Expr::ZERO);
+    for i in 0..BLOCK_FE_WIDTH {
+        out[i] = data[i * 2].clone() + AB::Expr::from_u64(256) * data[i * 2 + 1].clone();
```
jonathanpwang, line 299: 2 is hard coded which doesn't really make this tied to the `BLOCK_FE_WIDTH` constant

jonathanpwang: same for below

stephenh-axiom-xyz (approved):
Am good on the deferral extension changes

crates/vm/src/system/memory/online.rs
```diff
@@ -281,18 +283,17 @@ impl GuestMemory {
         Self { memory: addr }
     }
 
-    /// Returns `[pointer:BLOCK_SIZE]_{address_space}`
+    /// Reads `BLOCK_SIZE` AS-native cells starting at `cell_idx`.
```
GunaDD: what does AS-native cells mean ?

crates/vm/src/system/memory/offline_checker/bridge.rs
shuklaayush: huh the 2 comes from the fact that this is packing two `u8` elements into one `u16`. it's independent of `BLOCK_FE_WIDTH`

extensions/riscv/circuit/src/adapters/alu.rs
```diff
@@ -127,7 +128,7 @@ impl<AB: InteractionBuilder> VmAdapterAir<AB> for Rv64BaseAluAdapterAir {
         self.memory_bridge
             .read(
                 MemoryAddress::new(local.rs2_as, local.rs2),
-                ctx.reads[1].clone(),
+                pack_u8_block::<AB>(&ctx.reads[1].clone()),
```
GunaDD, line 134: is there soundness issues with this at the moment? that is, does the `pack_u8_block` needs to be constrained too ?

crates/vm/src/system/memory/online.rs
shuklaayush: address space native i.e. either `u16` or `F`

extensions/riscv/circuit/src/adapters/alu.rs
shuklaayush: wdym, it's defined as `a + 256 *b` which implicitly constrains it as well. or maybe i misunderstood

crates/vm/src/arch/state.rs
shuklaayush: `u8` for the first path and `u16` or `F` for the else

extensions/riscv/circuit/src/adapters/alu.rs
GunaDD: ooo yea if its like that then Its fine I think

crates/vm/src/arch/state.rs
```diff
@@ -205,11 +207,19 @@ where
         addr_space: u32,
         ptr: u32,
```
GunaDD: should this also be renamed as `cell_idx` too or not ?

shuklaayush: no, because it can also be bytes. maybe it can be `ptr_or_idx` but let's do it in a follow up and not here since four other prs are stacked on top of this

extensions/keccak256/circuit/cuda/src/keccakf_op.cu
```diff
@@ -88,10 +88,9 @@ static __device__ __noinline__ void fill_keccakf_op_row(
                            << (RV64_TOTAL_BITS - pointer_max_bits);
     bitwise_lookup.add_range(scaled_limb, scaled_limb);
 
-    // Range check for postimage bytes (pairs)
-    for (size_t i = 0; i < KECCAK_WIDTH_BYTES; i += 2) {
-        bitwise_lookup.add_range(state.bytes[i], state.bytes[i + 1]);
-    }
+    // AIR bounds the pointer bytes before composing them.
+    bitwise_lookup.add_range(buffer_ptr_limbs[0], buffer_ptr_limbs[1]);
+    bitwise_lookup.add_range(buffer_ptr_limbs[2], buffer_ptr_limbs[3]);
```
GunaDD, line 93: what was this range check for postage bytes and why is it getting removed now ?

shuklaayush: you needed to decompose the `u16` postimage into bytes before sending it to the memory bus earlier. and to show that the decomposition is canonical, you needed to range check that the two limbs are indeed bytes. now that the memory bus is `u16` you don't need this

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -2,30 +2,36 @@
...
+// DEFERRAL_AS stores Fp cells directly.
+// Non-deferral address spaces store u16 cells in little-endian byte order.
 template <int ADDR_SPACE_IDX>
 __global__ void merkle_tree_init(uint8_t *__restrict__ data, digest_t *__restrict__ out) {
     auto gid = blockDim.x * blockIdx.x + threadIdx.x;
 
     Fp cells[CELLS] = {0};
-    // TODO: revisit when we sort out address space handling
 #pragma unroll
     for (size_t i = 0; i < CELLS_OUT; ++i) {
```
GunaDD: `DIGEST_WIDTH` instead of `CELLS_OUT` in the for loop ?

crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -163,15 +160,24 @@ impl<const PAGE_BITS: usize> MemoryCtx<PAGE_BITS> {
...
+        let leaf_label = ptr >> leaf_bits;
+        let end_leaf_label = end_ptr >> leaf_bits;
+        let num_leaves = end_leaf_label - leaf_label + 1;
+        let start_leaf_id = self
             .memory_dimensions
-            .label_to_index((address_space, chunk_idx)) as u32;
-        let end_block_id = start_block_id + num_blocks;
-        let start_page_id = start_block_id >> PAGE_BITS;
-        let end_page_id = ((end_block_id - 1) >> PAGE_BITS) + 1;
+            .label_to_index((address_space, leaf_label)) as u32;
```
GunaDD, line 176: is this still doing what its supposed to do ? `leaf_label` is equal to `chunk_idx / 2` but the logic of `label_to_index` function is unchanged ? or was there a change to `address_height` of memory dimensions as well ?

crates/vm/src/system/memory/merkle/public_values.rs
```diff
@@ -153,33 +174,23 @@ impl<const CHUNK: usize, F: Field> UserPublicValuesProof<CHUNK, F> {
...
     let addr_space_height = memory_dimensions.addr_space_height;
     assert_eq!(top_tree.len(), (2 << addr_space_height) - 1);
-    let num_pv_chunks: usize = num_public_values / CHUNK;
-    // This enforces the number of public values cannot be 0.
-    assert!(
-        num_pv_chunks.is_power_of_two(),
-        "pv_height must be a power of two"
-    );
-    let pv_height = log2_strict_usize(num_pv_chunks);
+    let num_pv_leaves: usize = num_public_values / DIGEST_WIDTH;
```
GunaDD, line 188: is the assertion still needed? and is `num_pv_chunks` equal to `num_pv_leaves`?

crates/vm/src/system/memory/merkle/tree.rs
```diff
@@ -99,110 +128,107 @@ impl<F: PrimeField32, const CHUNK: usize> MerkleTree<F, CHUNK> {
                     layer = new_layer
                         .into_par_iter()
                         .map(|(par_index, left, right)| {
+                            let left_node;
                             let left = if let Some(left) = left {
                                 left.0
                             } else {
-                                &self.get_node(2 * par_index)
+                                left_node = self.get_node_at_height(2 * par_index, height - 1);
```
GunaDD, line 135: why is the logic of this changed?

crates/vm/src/system/memory/tests.rs
```diff
@@ -17,16 +26,16 @@ fn test_memory_write_by_tester(tester: &mut impl TestBuilder<F>, its: usize) {
 
     // The point here is to have a lot of equal
     // and intersecting/overlapping blocks,
-    // by limiting the space of valid pointers.
-    let max_ptr = 20;
-    let value_bounds = [256, 256, 256];
+    // by limiting the space of valid cell indexes.
+    let max_cell_idx = 20;
```
GunaDD: should this be `10` instead of `20` since the cells are half the ptr? or does it not matter

crates/vm/src/arch/state.rs
```diff
@@ -192,11 +193,12 @@ where
     pub fn vm_read_slice<T: Copy + Debug>(
         &mut self,
         addr_space: u32,
-        ptr: u32,
+        cell_idx: u32,
```
jonathanpwang: we always refer to `(addr_sp, ptr)` for a VM memory address, including in the whitepaper. we shouldn't change this convention. 
this leads me to ask - why did we call it `cell_idx`? if we want to clarify we could use `cell_ptr`? I'm not sure why we renamed this. Is it just to distinguish from `byte_ptr`?

@claude do you know the answer?

crates/vm/src/arch/state.rs
```diff
@@ -219,20 +229,38 @@ where
         ptr: u32,
         data: &[T; BLOCK_SIZE],
     ) {
-        // SAFETY:
-        // - T is stack-allocated repr(C) or repr(transparent), usually u8 or F where F is the base
-        //   field
-        // - T is the exact memory cell type for this address space, satisfying the type requirement
-        unsafe { self.memory.write(addr_space, ptr, *data) }
+        if size_of::<T>() == 1 {
```
jonathanpwang: @claude why did we need to add all these if statements

jonathanpwang: @claude I don't get why we're trying to share a dispatch. This used to unambiguously mean host write with respect to VM memory state in the address space model.
If some chips like Keccak or SHA-2 are using byte semantics, then we should just make a separate `host_byte_write` function instead of trying to share functionality.

OR if the AS-native cell index function is only used in some rare case, then we should make a separate `host_cell_write` function

crates/vm/src/arch/state.rs
jonathanpwang: see my comment below on bad practice to use the same function for two totally different meanings

crates/vm/src/system/memory/offline_checker/bridge.rs
```diff
@@ -69,39 +71,66 @@ impl MemoryBridge {
...
+        P: Into<T>,
+    {
+        Self {
+            base,
+            prev_data: prev_data.map(Into::into),
+        }
+    }
+}
+
+impl<'a, T, V> From<&'a MemoryWriteAuxCols<V, BLOCK_FE_WIDTH>> for MemoryWriteAuxInput<'a, T, V>
```
jonathanpwang, line 126: is this just for some Var to Expr auto conversion?

crates/vm/src/system/memory/offline_checker/bridge.rs
```diff
@@ -157,95 +184,25 @@ impl<F: PrimeCharacteristicRing, V: Copy + Into<F>, const N: usize>
     }
 }
 
-/// Constraints and interactions for a logical memory read of `(address, data)` at time `timestamp`,
```
jonathanpwang, line 160: is this just pure cleanup?

crates/vm/src/system/memory/offline_checker/bridge.rs
jonathanpwang: sorry I guess I meant it needs that `MEMORY_BLOCK_BYTES / BLOCK_FE_WIDTH = 2`

crates/vm/src/system/memory/online.rs
```diff
@@ -175,17 +175,17 @@ impl<M: LinearMemory> AddressMap<M> {
...
+        let start = cell_idx as usize * layout.size();
         let bytes = self.get_u8_slice(addr_space, start, layout.size());
         layout.to_field(bytes)
     }
 
     /// # Safety
     /// - `T` **must** be the correct type for a single memory cell for `addr_space`
     /// - Assumes `addr_space` is within the configured memory and not out of bounds
-    pub unsafe fn get<T: Copy>(&self, (addr_space, ptr): Address) -> T {
+    pub unsafe fn get<T: Copy>(&self, (addr_space, index): Address) -> T {
```
jonathanpwang: I really do not like this (imo arbitrary) renaming. There needs to be a really good reason for us to move away from `(addr_sp, ptr)` notation because that is a global convention we have been using everywhere, and it was already independent conceptually from any notion of bytes.

In my mind, `addr_sp, ptr` always refers to VM memory according to the OpenVM ISA semantics.

any RISC-V byte pointer should be denoted by `byte_ptr` or something else.

crates/vm/src/system/memory/online.rs
```diff
@@ -401,78 +446,149 @@ impl TracingMemory {
...
 
-    /// Returns the previous access timestamp and updates the metadata slot.
-    /// Block size is always `DEFAULT_BLOCK_SIZE`, so this is a single-slot read/write.
     #[inline(always)]
-    fn prev_access_time(&mut self, address_space: usize, pointer: usize) -> u32 {
-        let idx = pointer / DEFAULT_BLOCK_SIZE;
+    fn prev_access_time(&mut self, address_space: usize, start_cell_idx: usize) -> u32 {
+        let idx = start_cell_idx / BLOCK_FE_WIDTH;
+        // SAFETY: address_space is validated during instruction decoding
+        let meta_page = unsafe { self.meta.get_unchecked_mut(address_space) };
```
jonathanpwang, line 479: @claude why was this added when it wasn't there before

crates/vm/cuda/src/system/memory/merkle_tree.cu
shuklaayush: yeah, good catch

crates/vm/src/system/memory/tests.rs
shuklaayush: let me change it to `10` but it probably doesn't matter since we just need a small value

crates/vm/src/system/memory/offline_checker/bridge.rs
shuklaayush: yes, i believe this was only ever used for the native address space

crates/vm/src/system/memory/offline_checker/bridge.rs
shuklaayush: yes, because sometimes you have `prev_data` as an packed expression of `u8` bytes. see for example
https://github.com/openvm-org/openvm/blob/075b593d1f98128c723733d45ff0acf38a44bb66/extensions/keccak256/circuit/src/keccakf_op/air.rs#L130-L133

crates/vm/src/arch/state.rs
shuklaayush: okay, i'll split it into separate functions. fwiw there was an earlier ai review that asked it to be combined to deduplicate

crates/vm/src/system/memory/online.rs
shuklaayush: the vm still operates on byte pointers and the memory itself is byte addressed from the ISA perspective. the `u16` memory storage is an implementation detail of how the circuit/hardware is constructed
i intentionally changed these to `index` to make the distinction that this is not a "pointer" to memory in the traditionally sense. it was straightforward earlier because the memory was byte addressed and also stored as bytes

but okay, since this is confusing, then let me just change these back to `ptr`

crates/vm/src/system/memory/merkle/tree.rs
shuklaayush: yeah, this is kinda unrelated. you're operating on the merkle tree's leaves here so you already know the height and don't have to do the `self.height - index.ilog2()` calculation in `get_node()` every time so i added another helper for this
the function still does the same thing

crates/vm/src/system/memory/merkle/public_values.rs
shuklaayush: i moved all public values assertions to this function which should be called during construction
https://github.com/openvm-org/openvm/blob/d30973fd5a2fa9b6a654e52b15321c3d868021f9/crates/vm/src/system/memory/merkle/public_values.rs#L20-L39

> and is num_pv_chunks equal to num_pv_leaves

yes, in the sense that it represents the number of merkle leaves in the public value address space

extensions/riscv/circuit/src/base_alu/core.rs
```diff
@@ -139,6 +139,14 @@ where
...
         }
 
+        // Memory bus checks only packed u16 values; ADD/SUB read bytes need separate bounds.
+        let add_or_sub = cols.opcode_add_flag + cols.opcode_sub_flag;
+        for i in 0..NUM_LIMBS {
+            self.bus
+                .send_range(b[i], c[i])
+                .eval(builder, add_or_sub.clone());
+        }
+
```
GunaDD, line 149: should we put the range checks that constraints the input in the adapters or here ?

crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
shuklaayush: yes, `label_to_index` just converts a memory cell (leaf label) in a particular address space to its index in the global memory merkle tree. `leaf_label` is now half because you are packing twice the number of bytes in a leaf

extensions/riscv/circuit/src/base_alu/core.rs
shuklaayush: it's usually in adapters but i added this one here because it's conditional on whether the opcode is only add or sub. let me think about this. maybe we can just do it unconditionally and move it to the adapter

extensions/riscv/circuit/src/hintstore/mod.rs
```diff
@@ -4,8 +4,8 @@ use openvm_circuit::{
     arch::*,
     system::memory::{
```
GunaDD, line 5: why is everything in `mod.rs` for hintstore? for other riscv chips the contents of this file would be split into `core.rs` and `execution.rs`.

extensions/riscv/circuit/src/hintstore/mod.rs
```diff
@@ -194,7 +194,7 @@ impl<AB: InteractionBuilder> Air<AB> for Rv64HintStoreAir {
         self.memory_bridge
             .write(
                 MemoryAddress::new(AB::F::from_u32(RV64_MEMORY_AS), mem_ptr.clone()),
-                local_cols.data,
+                pack_u8_block::<AB>(&local_cols.data.map(Into::into)),
                 timestamp_pp(),
                 &local_cols.write_aux,
             )
```
GunaDD, line 209: are we missing range checks here ?

extensions/riscv/circuit/src/hintstore/mod.rs
shuklaayush: good q but it's outside the scope of this pr. let's stick to the changes here

extensions/riscv/circuit/src/hintstore/mod.rs
```diff
@@ -247,13 +244,6 @@ impl<AB: InteractionBuilder> Air<AB> for Rv64HintStoreAir {
...
             .eval(builder, is_start.clone());
 
-        // Checking that hint is bytes
-        for i in 0..RV64_REGISTER_NUM_LIMBS / 2 {
-            self.bitwise_operation_lookup_bus
-                .send_range(local_cols.data[2 * i], local_cols.data[(2 * i) + 1])
-                .eval(builder, is_valid.clone());
-        }
-
```
GunaDD, line 256: do we still need to check whether `local_cols.data` is `u16` though?

extensions/riscv/circuit/src/hintstore/mod.rs
```diff
@@ -592,11 +582,6 @@ impl<F: PrimeField32> TraceFiller<F> for Rv64HintStoreFiller {
                     .rchunks_exact_mut(width)
                     .zip(record.var.iter().enumerate().rev())
                     .for_each(|(row, (idx, var))| {
-                        for pair in var.data.chunks_exact(2) {
-                            self.bitwise_lookup_chip
-                                .request_range(pair[0] as u32, pair[1] as u32);
-                        }
-
                         let cols: &mut Rv64HintStoreCols<F> = row.borrow_mut();
```
GunaDD, line 618: same here

extensions/riscv/circuit/src/less_than/core.rs
```diff
@@ -127,6 +127,13 @@ where
             )
             .eval(builder, is_valid.clone());
 
+        // Memory bus checks only packed u16 values; these non-MSB read bytes need separate bounds.
```
GunaDD: why is the MSB not range checked ? also, this made me think of an optimization idea: does it suffice to range check the even bits since the memory bus already constrains that b_2i + 2^8 * b_2i+1 is u16 ?

extensions/riscv/circuit/src/loadstore/tests.rs
```diff
@@ -142,10 +143,11 @@ fn set_and_execute<RA: Arena, E: PreflightExecutor<F, RA>>(
         * RV64_REGISTER_NUM_LIMBS;
 
     let is_load = [LOADD, LOADWU, LOADHU, LOADBU].contains(&opcode);
+    // Store tests choose writable u16-celled address spaces.
     let mem_as = mem_as.unwrap_or(if is_load {
         2
     } else {
-        *[2, 3, 4].choose(rng).unwrap()
+        *[2, 3].choose(rng).unwrap()
     });
```
GunaDD, line 151: is it dangerous that we are removing the test for AS 4?

extensions/riscv/circuit/src/mul/cuda.rs
```diff
@@ -3,7 +3,8 @@ use std::{mem::size_of, sync::Arc};
 use derive_new::new;
 use openvm_circuit::{arch::DenseRecordArena, utils::next_power_of_two_or_zero};
 use openvm_circuit_primitives::{
-    range_tuple::RangeTupleCheckerChipGPU, var_range::VariableRangeCheckerChipGPU, Chip,
```
GunaDD, line 6: also, I think we should not forget to update the cuda tracegen for the missing range checks as well

extensions/riscv/circuit/src/loadstore/tests.rs
shuklaayush: no, support for writing to AS 4 is going to be removed anyway

extensions/sha2/circuit/src/sha2_chips/main_chip/air.rs
GunaDD, line 231: should we range check the entire `data` colums here ?

GunaDD, line 241: should the entire ptr limbs be range checked ?

GunaDD (approved):
finished reviewing, other than the comments and issues flagged by the AI audit, LGTM. but we might also think if we should fix it in this PR since some of the things mentioned by the AI will be invalid once the columns changed to u16 too.

extensions/riscv/circuit/src/hintstore/mod.rs
```diff
@@ -247,13 +244,6 @@ impl<AB: InteractionBuilder> Air<AB> for Rv64HintStoreAir {
             )
             .eval(builder, is_start.clone());
 
-        // Checking that hint is bytes
```
jonathanpwang, line 250: I believe this is still needed? unless you have some future plans. as it is now, certainly you still need to make sure the hint is bytes

jonathanpwang (approved):
lgtm besides all existing comments. I did not check the range check changes carefully nor the deferral chips. I think we can do an audit after the changes settle down.

extensions/riscv/circuit/src/hintstore/mod.rs
shuklaayush: not `u16`, but for this pr we do need to check that they're bytes

extensions/riscv/circuit/src/less_than/core.rs
shuklaayush: added the check. i don't think the optimization idea is sound

extensions/riscv/circuit/src/base_alu/core.rs
shuklaayush: kept in core because the range check is conditional

PR #2795, opened 2026-05-20
refactor: move rvr compile stage into instance construction

gdmlcjs:
moves the rvr compilation stage for metered and metered cost out of execution and into instance construction

closes INT-7626


shuklaayush (approved):
lgtm

PR #2799, opened 2026-05-20
feat: split for root prover benchmarking

Aalanli:
Need these changes so https://github.com/axiom-crypto/openvm-eth/pull/609 becomes possible

Resolves INT-7744

### DIFF
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -470,8 +470,8 @@ where
     }
 
     #[cfg(feature = "root-prover")]
-    /// Constructs an [`EvmProver`] for the given executable, generating prerequisite keys lazily.
-    pub fn evm_prover(
+    /// Constructs an [`EvmProver`] for the given executable with only the root prover
+    pub fn evm_prover_without_halo2(
         &self,
         app_exe: impl Into<ExecutableFormat>,
     ) -> Result<EvmProver<E, VB>, SdkError> {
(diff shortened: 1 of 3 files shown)


crates/sdk/src/prover/evm.rs
shuklaayush, line 28: can you add a comment here for the new method?

stephenh-axiom-xyz (commented):
This probably isn't the right split, as a public SDK function shouldn't expose an intermediate form of the root proof. Maybe smth like:
- STARK prove + root ctx generation + test verify in `prove_stark_and_generate_root_ctx`
- Root prove in `prove_unwrapped_from_ctx`
- Keep `prove_unwrapped` as a wrapper for both

EDIT: As discussed offline we may want to keep additional internal-recursive wraps in the bench, so the split then becomes:
- `prove_unwrapped` - whole proof flow up to root (as it currently is)
- `prove_unwrapped_from_stark` - wraps + root prove from the stark_prover output

crates/sdk/src/prover/agg.rs
```diff
@@ -44,6 +47,44 @@ pub struct InternalLayerMetadata {
     pub proofs_type: ProofsType,
 }
 
+impl Encode for InternalLayerMetadata {
```
stephenh-axiom-xyz: nit: pls move these to the bottom of the file

crates/sdk/src/lib.rs
```diff
@@ -485,11 +485,26 @@ where
...
+    #[cfg(feature = "root-prover")]
+    /// Constructs an [`EvmProver`] for the given executable, generating prerequisite keys lazily.
+    pub fn evm_prover(
+        &self,
+        app_exe: impl Into<ExecutableFormat>,
+    ) -> Result<EvmProver<E, VB>, SdkError> {
+        #[allow(unused_mut)]
+        let mut evm_prover = self.evm_prover_without_halo2(app_exe)?;
+        #[cfg(feature = "evm-prove")]
+        {
```
stephenh-axiom-xyz, line 503: nit: don't need these brackets

crates/sdk/src/lib.rs
```diff
@@ -470,8 +470,8 @@ where
     }
 
     #[cfg(feature = "root-prover")]
-    /// Constructs an [`EvmProver`] for the given executable, generating prerequisite keys lazily.
```
stephenh-axiom-xyz, line 473: It still generates prerequisite keys lazily

stephenh-axiom-xyz (approved):
LGTM 🚀

crates/sdk/src/lib.rs
Aalanli: If I don't it says that attributes on expressions is experimental.

PR #2800, opened 2026-05-20
refactor: use enum for air index

gdmlcjs:
Air indices are now represented as an enum in rvr code. The `AirIndex` enum has `Uninitialized` and `NoChip` variants that replace the previous `NO_CHIP = u32::MAX`. `AirIndex::Uninitialized` is only used in pure execution where air indices don't matter and causes a panic in rvr metered and metered cost execution.

closes INT-7611

### DIFF
--- a/crates/rvr/rvr-openvm-lift/src/extension.rs
+++ b/crates/rvr/rvr-openvm-lift/src/extension.rs
@@ -9,39 +9,45 @@ use openvm_instructions::{instruction::Instruction, LocalOpcode, VmOpcode};
 use openvm_stark_backend::p3_field::PrimeField32;
 use rvr_openvm_ir::LiftedInstr;
 
-/// Index of an AIR in the trace.
+/// Real AIR index in the trace.
 ///
-/// - `Uninitialized` is a placeholder used only during pure execution where air indices are
-///   irrelevant. Metered execution panics when it encounters this variant.
-/// - `NoChip` marks opcodes/instructions that do not contribute to any chip's trace (e.g.
-///   `TERMINATE`).
-/// - `Chip(u32)` is a real AIR index.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-pub enum AirIndex {
-    Uninitialized,
-    NoChip,
-    Chip(u32),
-}
+/// `Option<AirIndex>` represents an extension's dynamic tracing metadata: `None`
+/// in pure mode (AIR metadata was not requested), `Some(idx)` in metered modes.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct AirIndex(u32);
 
 impl AirIndex {
     #[inline]
-    pub fn next(self) -> AirIndex {
-        match self {
-            AirIndex::Chip(n) => AirIndex::Chip(n + 1),
-            other => other,
-        }
+    pub const fn new(idx: u32) -> Self {
+        Self(idx)
     }
 
-    /// Lower to the `u32` chip index baked into generated C source. Both
-    /// `NoChip` and `Uninitialized` become `u32::MAX` (the `NO_CHIP` sentinel
-    /// on the C side, where `trace_chip` is a no-op for that value).
     #[inline]
-    pub fn to_c_chip_idx(self) -> u32 {
-        match self {
-            AirIndex::Chip(n) => n,
-            AirIndex::NoChip | AirIndex::Uninitialized => u32::MAX,
-        }
+    pub const fn as_u32(self) -> u32 {
+        self.0
     }
+
+    #[inline]
+    pub fn next(self) -> AirIndex {
+        AirIndex(self.0 + 1)
+    }
+}
+
+/// PC-to-chip mapping entry. `NoChip` means the instruction at this PC
+/// intentionally does not contribute to any chip's trace (e.g. `TERMINATE` or
+/// unmapped slots).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum TraceChipIndex {
+    Chip(AirIndex),
+    NoChip,
+}
+
+/// Lower an extension's `Option<AirIndex>` to the `u32` chip index baked into
+/// generated C source. `None` (pure mode) becomes `u32::MAX` (the `NO_CHIP`
+/// sentinel on the C side, where `trace_chip` is a no-op for that value).
+#[inline]
+pub fn air_index_to_c(idx: Option<AirIndex>) -> u32 {
+    idx.map_or(u32::MAX, AirIndex::as_u32)
 }
 
 /// Data needed by extension crates to resolve opcode/chip metadata when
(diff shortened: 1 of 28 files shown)


crates/rvr/rvr-openvm-lift/src/extension.rs
```diff
@@ -5,15 +5,44 @@ use std::{
...
+///   irrelevant. Metered execution panics when it encounters this variant.
+/// - `NoChip` marks opcodes/instructions that do not contribute to any chip's trace (e.g.
+///   `TERMINATE`).
+/// - `Chip(u32)` is a real AIR index.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum AirIndex {
+    Uninitialized,
+    NoChip,
+    Chip(u32),
+}
```
shuklaayush: can we do something like this instead
```rust
pub struct AirIndex(u32);

pub enum TraceChipIndex {
    Chip(AirIndex),
    NoChip,
}

impl TraceChipIndex {
    pub fn to_c_chip_idx(self) -> u32 {
        match self {
            TraceChipIndex::Chip(idx) => idx.as_u32(),
            TraceChipIndex::NoChip => u32::MAX,
        }
    }
}

impl AirIndex {
    pub fn next(self) -> Self {
        Self(self.0 + 1)
    }
}
```

with the following semantics:
```rust
Option<AirIndex>        // absent because mode/context has no AIR metadata
TraceChipIndex::NoChip  // intentionally no chip row
AirIndex                // definitely a real AIR index
```

crates/vm/src/arch/rvr/compile.rs
```diff
@@ -53,14 +54,43 @@ pub enum CompileError {
...
     /// increments `cost` by a single constant instead of loading from the
     /// `chip_widths` array at runtime.
     pub chip_widths: Option<Vec<u64>>,
 }
 
+pub fn build_pc_to_chip<F, E>(
+    exe: &VmExe<F>,
+    inventory: &ExecutorInventory<E>,
+    executor_idx_to_air_idx: &[usize],
+) -> Vec<AirIndex>
```
shuklaayush: this should return `Result<Vec<_>, CompileError>` for cases when an unknown instruction is encountered

shuklaayush (commented):
i don't like having both `Uninitialized` and `NoChip` in the enum. it makes it confusing. i asked codex for suggestions and here's what it said

```
Use these meanings:

pub struct AirIndex(u32);          // definitely a real AIR index

pub enum TraceChipIndex {
    Chip(AirIndex),                // this instruction contributes to a chip
    NoChip,                        // this instruction intentionally has no chip
}

Then use Option<AirIndex> only for pure-mode extension metadata:

None            // pure mode: AIR metadata was not requested
Some(AirIndex)  // metered mode: real AIR index

So the model becomes:

AirIndex              // real index only
TraceChipIndex        // PC-to-chip mapping: chip or no chip
Option<AirIndex>      // extension dynamic tracing: absent in pure mode

The core helper should become:

pub fn opcode_air_idx(
    ctx: Option<&RvrExtensionCtx>,
    opcode: impl LocalOpcode,
) -> Result<Option<AirIndex>, ExtensionError> {
    let opcode = opcode.global_opcode();
    let Some(ctx) = ctx else {
        return Ok(None);
    };

    let executor_idx = ctx
        .resolve_opcode_executor_idx(opcode)
        .ok_or(ExtensionError::UnknownOpcode(opcode))?;

    let raw = ctx.executor_idx_to_air_idx.get(executor_idx).ok_or(
        ExtensionError::ExecutorIndexOutOfBounds {
            opcode,
            executor_idx,
        },
    )?;

    Ok(Some(AirIndex::new(*raw as u32)))
}

Then extension code like SHA/Keccak stores:

sha256_main_chip_idx: Option<AirIndex>,
sha256_block_hasher_chip_idx: Option<AirIndex>,

and computes adjacent chips explicitly:

let sha256_block_hasher_chip_idx = sha256_main_chip_idx.map(AirIndex::next);

For C lowering:

fn air_index_to_c(idx: Option<AirIndex>) -> u32 {
    idx.map_or(u32::MAX, AirIndex::as_u32)
}

That makes pure-mode behavior explicit at the call site, instead of smuggling it through an Uninitialized variant.

For pc_to_chip, use TraceChipIndex, not Option<AirIndex>:

pub struct ChipMapping {
    pub pc_to_chip: Vec<TraceChipIndex>,
    pub chip_widths: Option<Vec<u64>>,
}

build_pc_to_chip returns:

TraceChipIndex::Chip(AirIndex::new(...))
TraceChipIndex::NoChip

And crates/rvr/rvr-openvm/src/emit/project.rs:186 should assume metered mode has a mapping:

fn chip_idx_for_pc(&self, pc: u32) -> TraceChipIndex {
    let mapping = self
        .pc_to_chip
        .as_ref()
        .expect("pc_to_chip must be set for metered rvr codegen");

    let Some(offset) = pc.checked_sub(self.pc_base) else {
        return TraceChipIndex::NoChip;
    };

    mapping
        .get((offset / 4) as usize)
        .copied()
        .unwrap_or(TraceChipIndex::NoChip)
}

And validate once in compile setup:

match opts.tracer_mode {
    TracerMode::Pure => {}
    TracerMode::Metered | TracerMode::MeteredCost => {
        let chips = opts.chips.ok_or(CompileError::InvalidOptions(
            "metered rvr compile requires ChipMapping",
        ))?;
        project.pc_to_chip = Some(chips.pc_to_chip.clone());
        project.pc_base = exe.program.pc_base;
        project.chip_widths = chips.chip_widths.clone();
    }
}

This is the cleanest because every absence has one meaning:

NoChip means intentional no trace row.

None means pure mode did not ask for AIR metadata.

Missing pc_to_chip in metered mode is a construction bug, caught at setup, not encoded as a fake chip index
```

crates/rvr/rvr-openvm-lift/src/extension.rs
gdmlcjs: yes I think that might be clearer

crates/vm/src/arch/rvr/compile.rs
```diff
@@ -161,10 +200,16 @@ fn compile_impl<F: PrimeField32>(
...
-    if let Some(chips) = opts.chips {
-        project.pc_to_chip = Some(chips.pc_to_chip.clone());
-        project.pc_base = exe.program.pc_base;
-        project.chip_widths = chips.chip_widths.clone();
+    match opts.tracer_mode {
+        TracerMode::Pure => {}
+        TracerMode::Metered | TracerMode::MeteredCost => {
+            let chips = opts
+                .chips
+                .expect("metered rvr compile requires ChipMapping");
```
shuklaayush: `CompileError`? in general, it's better to return errors wherever possible instead of panics

gdmlcjs: I though it would be okay in this case because this error would be made by the code in this file, and not from the public interface being misused. And so adding an `InvalidOptions` variant didn't feel like a good name, but do you have another name suggestion? Or maybe I'm just overthinking?

shuklaayush: yep, invalid options with a string inside should be fine

gdmlcjs: okay thanks!

shuklaayush (approved):

PR #2801, opened 2026-05-20
feat(v2.1-rv64): switch to u16 limbs in riscv airs

shuklaayush:
PR 2 of the memory-bus-u16 split. This migrates selected RV64 chips from byte-shaped register/memory columns to u16-shaped columns after #2794 made the memory bus operate on fixed `BLOCK_FE_WIDTH = 4` u16 cells.

The scope is intentionally split into two classes:

- For equality/comparison-style chips, `hintstore`, and the new u16 adapters, the main change is the limb shape: `u8` limbs become `u16` limbs, byte-pair bitwise range checks become per-limb `VariableRangeChecker` checks, and the core logic remains essentially the same.
- For control-flow instructions that previously used byte-level decompositions for PC/result arithmetic (`auipc`, `jalr`, `jal_lui`), the constraints were reworked more substantially because the old carry structure was tied to 8-bit limbs.

Subsequent PRs build on the u16 adapter variants and bigint limb constants introduced here.

## Why

#2794 made the memory bus fixed-shape over packed u16 cells. Some chips still materialized 8 byte columns per RV64 register value, packed them into u16 bus cells at the memory interaction, and carried `BitwiseOperationLookup` only to prove those byte limbs were canonical.

For chips that do not inspect byte-level structure, storing u16 limbs directly halves the affected register-value trace width and removes those byte-pair lookup dependencies. The bus sees the native u16 payload, while the AIR uses `VariableRangeChecker` to prove each materialized u16 limb is bounded.

For PC/result arithmetic, the old byte AIRs also avoided field-modulus aliasing by carrying through small limbs. The u16 rewrites preserve that property with fewer, wider limbs; they do not replace 32-bit arithmetic with one unconstrained field equality.

### Mostly Mechanical Limb-Shape Migrations

These chips keep the same core argument, but change the limb type/width:

- `branch_eq`: compares u16 limbs directly. Equality inverse logic is unchanged.
- `branch_lt`: signed/unsigned comparison logic is unchanged, with `LIMB_BITS = 16` and u16 MSB handling.
- `less_than`: same comparison migration, wired through the u16 ALU adapter.
- `hintstore`: stores `mem_ptr`, `rem_words`, and hint data as u16-shaped columns. `rem_words` fits in one u16 cell, and pointer bounds are checked through the high u16 limb.
- Bigint `BranchEqual256`, `BranchLessThan256`, and `LessThan256`: inherit the u16 shape with `INT256_NUM_U16_LIMBS = 16`.

The new u16 adapters follow the same pattern: register pointer values and heap/register payloads are represented as u16 cells directly, with range checks moving from byte-pair lookup to `VariableRangeChecker`.

Byte-shaped bigint chips stay byte-shaped:

- `BaseAlu256`, `Mul256`, `Shift256`, and `Sll256` still use `INT256_NUM_U8_LIMBS = 32` because their AIRs rely on byte-level boolean, carry, or shift structure.

### Constraint Rewrites

These chips needed more than a limb-size substitution:

- `auipc`: the old AIR decomposed the PC and immediate into byte limbs and carried the addition byte-by-byte. The new AIR uses u16 limbs instead: it derives/range-checks `pc_low` from `from_pc` and a witnessed `pc_high`, then constrains `rd = from_pc + (imm << 8)` with low-u16 and high-u16 carry bits. This keeps each arithmetic relation below the field modulus while dropping the extra byte-limb PC columns.
- `jalr`: the old AIR used low-32-bit byte limbs for `rs1`, `to_pc`, and `rd = pc + 4`. The new AIR keeps u16 limbs, derives the low rd limb from `from_pc`, range-checks the high rd limb against `PC_BITS`, and carries `rs1 + sign_extend(imm)` over two u16 limbs while separately enforcing the cleared target low bit.
- `jal_lui`: JAL derives the low rd limb from `from_pc + 4` and the witnessed high u16 limb, then range-checks the high limb against `PC_BITS`. LUI uses a u16-specific split because `imm << 12` crosses a u16 boundary: `rd[0] = imm_low_4 << 12` and `imm = imm_low_4 + (rd[1] << 4)`.

The old byte-limb form would still be possible, but the u16 form uses fewer witness columns/range checks while preserving limb-by-limb integer soundness in the BabyBear field.

### New u16 Adapter Variants

- `Rv64BaseAluU16Adapter`: reads/writes one `[u16; BLOCK_FE_WIDTH]` register block directly and handles immediates as u16-low-limb plus sign bit.
- `Rv64VecHeapU16Adapter`: u16-cell heap reads/writes with u16-shaped register pointer values.
- `Rv64VecHeapBranchU16Adapter`: read-only branch variant for u16-shaped heap operands.
- `VecToFlatAluAdapter` gains a `BLOCK_VALUE_WIDTH` parameter so both byte-shaped and u16-shaped heap adapters can share the flattening logic.

The legacy byte-shaped adapters remain for chips that still need byte-level constraints.

### Wiring

- RISC-V and bigint extension wiring now threads `VariableRangeChecker` into the u16-shaped chips.
- `BitwiseOperationLookup` remains wired only for byte-shaped chips that still need byte-level constraints.
- Rv64Io no longer creates a bitwise lookup solely for `hintstore`.

### CUDA Parity

CUDA tracegen mirrors the Rust changes:

- migrated core records use `uint16_t` arrays where applicable;
- kernels use `VariableRangeChecker.add_count(limb, LIMB_BITS)` for u16 limbs;
- bitwise lookup parameters are removed from migrated kernels;
- bigint CUDA wiring separates byte-shaped chips from u16-shaped chips.

### Helper Cleanup

- `INT256_NUM_U8_LIMBS` and `INT256_NUM_U16_LIMBS` make bigint limb width explicit.
- Shared RV64 helpers cover byte/u16 conversion, low-32-bit pointer splitting, pointer-bound range-check values, and zero-extension to one u16 memory block.
- Byte-view memory helpers use explicit `read_bytes` / `write_bytes` where the access is intentionally byte-addressed.


extensions/riscv/circuit/src/base_alu/core.rs
GunaDD, line 32: why is the `NUM_LIMBS` constant never changed in this PR? I thought the goal of using `u16` memory was so that the number of columns get halved.

extensions/riscv/circuit/src/branch_eq/core.rs
```diff
@@ -39,7 +35,6 @@ pub struct BranchEqualCoreCols<T, const NUM_LIMBS: usize> {
 #[derive(Copy, Clone, Debug, derive_new::new, ColumnsAir)]
 #[columns_via(BranchEqualCoreCols<u8, NUM_LIMBS>)]
 pub struct BranchEqualCoreAir<const NUM_LIMBS: usize> {
-    pub bitwise_lookup_bus: BitwiseOperationLookupBus,
```
GunaDD, line 42: I guess we will need the lookup bus back once the missing range checks are merged

extensions/riscv/circuit/src/base_alu/core.rs
GunaDD: I think the change will be in `base_alu/mod.rs`

extensions/riscv/circuit/src/branch_eq/tests.rs
```diff
@@ -54,13 +47,24 @@ use crate::{
...
+#[inline]
+fn u16_array_to_bytes_le(arr: &[u16; BLOCK_FE_WIDTH]) -> [u8; RV64_REGISTER_NUM_LIMBS] {
+    let mut out = [0u8; RV64_REGISTER_NUM_LIMBS];
+    for (i, &v) in arr.iter().enumerate() {
+        let [lo, hi] = v.to_le_bytes();
+        out[2 * i] = lo;
+        out[2 * i + 1] = hi;
+    }
+    out
+}
```
GunaDD: is there no builtin std function for this? and is it better if this function be placed at another file?

extensions/riscv/circuit/src/branch_eq/tests.rs
```diff
@@ -280,10 +261,10 @@ fn rv64_beq_wrong_cmp_negative_test() {
 fn rv64_beq_zero_inv_marker_negative_test() {
     run_negative_branch_eq_test(
         BranchEqualOpcode::BEQ,
-        [0, 0, 7, 0, 0, 0, 0, 0],
-        [0, 0, 0, 7, 0, 0, 0, 0],
+        [0, 7, 0, 0],
+        [0, 0, 7, 0],
```
GunaDD: I don't think this conversion is correct? shouldn't the second one be `[0, 256*7, 0, 0]` ?

extensions/riscv/circuit/src/branch_eq/tests.rs
```diff
@@ -304,17 +285,17 @@ fn rv64_beq_invalid_inv_marker_negative_test() {
 fn rv64_bne_wrong_cmp_negative_test() {
     run_negative_branch_eq_test(
         BranchEqualOpcode::BNE,
-        [0, 0, 7, 0, 0, 0, 0, 0],
-        [0, 0, 0, 7, 0, 0, 0, 0],
+        [0, 7, 0, 0],
+        [0, 0, 7, 0],
         Some(false),
```
GunaDD, line 281: same here

extensions/riscv/circuit/src/branch_eq/tests.rs
```diff
@@ -325,10 +306,10 @@ fn rv64_bne_wrong_cmp_negative_test() {
 fn rv64_bne_zero_inv_marker_negative_test() {
     run_negative_branch_eq_test(
         BranchEqualOpcode::BNE,
-        [0, 0, 7, 0, 0, 0, 0, 0],
-        [0, 0, 0, 7, 0, 0, 0, 0],
+        [0, 7, 0, 0],
+        [0, 0, 7, 0],
         Some(false),
-        Some([0, 0, 0, 0, 0, 0, 0, 0]),
+        Some([0, 0, 0, 0]),
```
GunaDD: same here

extensions/riscv/circuit/src/branch_eq/tests.rs
```diff
@@ -355,10 +336,10 @@ fn rv64_bne_invalid_inv_marker_negative_test() {
 fn execute_roundtrip_sanity_test() {
     let mut rng = create_seeded_rng();
     let mut tester = VmChipTestBuilder::default();
-    let (mut harness, _bitwise) = create_harness(&mut tester);
+    let mut harness = create_harness(&mut tester);
 
-    let x = [19, 4, 179, 60, 201, 77, 1, 240];
-    let y = [19, 32, 180, 60, 201, 77, 1, 240];
+    let x = [19, 4179, 60201, 7700];
+    let y = [19, 4180, 60201, 7700];
```
GunaDD: I don't think this conversion is correct ?

GunaDD (changes requested):
Two questions from me:
1. Why is many of the riscv chips still use `NUM_LIMBS = 8` and is not changed to 4? 
2. Why do we need to make a change in design in the execution to use `u16` type?

extensions/riscv/circuit/src/base_alu/core.rs
shuklaayush: we can't do it in base alu yet because it also does logical operations xor/and/or which need `u8` limbs. we'd need to split add/sub and xor/and/or into separate airs

shuklaayush:
> Why is many of the riscv chips still use NUM_LIMBS = 8 and is not changed to 4?

some chips like basealu can't be migrated to `u16` limbs because they also do xor which requires a `u8 x u8` lookup (and can't be changed to `u16 x u16` because of table size restrictions). i think i migrated all which i felt were possible

> Why do we need to make a change in design in the execution to use u16 type?

it depends on which one you're referring to but for eg. in this branch opcode, you need to do the calculations in `u16` limbs. so i just changed it everywhere the air was changed to be consistent

https://github.com/openvm-org/openvm/blob/9effefa3d5d7aac59768613412872d0c235b26da/extensions/riscv/circuit/src/branch_eq/core.rs#L229

extensions/riscv/circuit/src/base_alu/core.rs
GunaDD: got it

extensions/riscv/circuit/src/adapters/branch.rs
```diff
@@ -154,14 +154,14 @@ where
...
-        let rs1 = tracing_read(
+        let rs1 = tracing_read_u16::<BLOCK_FE_WIDTH>(
             memory,
             RV64_REGISTER_AS,
             a.as_canonical_u32(),
             &mut record.reads_aux[0].prev_timestamp,
         );
         record.rs2_ptr = b.as_canonical_u32();
-        let rs2 = tracing_read(
+        let rs2 = tracing_read_u16::<BLOCK_FE_WIDTH>(
```
GunaDD, line 170: I think this change will not be needed if we go back to `u8` record ?

extensions/riscv/circuit/src/auipc/core.rs
```diff
@@ -26,26 +19,27 @@ use openvm_stark_backend::{
...
-    pub imm_limbs: [T; RV64_WORD_NUM_LIMBS - 1],
-    // The limbs of the PC except the most significant and the least significant limbs
-    pub pc_limbs: [T; RV64_WORD_NUM_LIMBS - 2],
-    pub rd_data: [T; RV64_WORD_NUM_LIMBS],
+    // Low byte of imm.
+    pub imm_low_8: T,
+    // High 16 bits of imm.
+    pub imm_high_16: T,
+    // Low 32 bits of rd as u16 limbs.
+    pub rd_data: [T; RV64_LOW32_U16_LIMBS],
```
GunaDD: is the `pc_limbs` not needed anymore ?

shuklaayush: let me take a look at this again. i think there was some issue with just translating the pc limbs from `u8` to `u16` so i changed the constraints a bit. i did convince myself that the changes are correct but i'll take a look again

extensions/riscv/circuit/src/mul/core.rs
GunaDD, line 102: I don't think `mul` needs to do the `xor` check. Can it use `u16` columns then ? @shuklaayush

extensions/riscv/circuit/src/divrem/core.rs
GunaDD, line 324: similarly to `mul`, I don't thin `divrem` uses the xor check. can we use the `u16` columns here too then? @shuklaayush

GunaDD (approved):
Finished reviewing this PR. My comments would be:
1. We don't need to change the memory type of the record to `u16`. Maybe it's better to keep it `u8` to not make many extra changes.
2. Looks like the `mul` and `divrem` doesn't use the `bitwise_xor` lookup bus and hence can be made `u16`. 
3. Is dropping the `pc_limbs` okay in the columns of `auipc`?
4. Some other minor comments in the PR. 

Other than that LGTM.

extensions/riscv/circuit/src/mul/core.rs
shuklaayush: iirc for mul, it can't be `u16` because then the carry would overflow the field which is 31 bits. for `u8` the carry is at max 16-bits while for `u16` it would be 32-bits

extensions/riscv/circuit/src/divrem/core.rs
shuklaayush: same as mul, can't be made `u16` because carry will overflow

extensions/riscv/circuit/src/adapters/branch.rs
shuklaayush: not really since the read data has to be of `[u16; N]` form so the conversion has to be done at some point. this seems fine since otherwise we'd need to duplicate the `u8` to `u16` conversions in every adapter

extensions/riscv/circuit/src/branch_eq/core.rs
shuklaayush: no, because we don't need to check that the limbs are `u16`. this is enforced implicitly by the bus

extensions/riscv/circuit/src/branch_eq/tests.rs
shuklaayush: i think it didn't matter but i changed the conversions to be runtime to minimize the diff

extensions/riscv/circuit/src/branch_eq/tests.rs
shuklaayush: moved it to the adapters file. i don't think there's an std function for this. we can use `transmute` but it's unsafe

shuklaayush:
> We don't need to change the memory type of the record to u16. Maybe it's better to keep it u8 to not make many extra changes.

i thought about this. i think it's better to use `u16` records/limbs where the columns are switched to `u16` just to keep things consistent. though ultimately, we should only store `u32`/`u64` records and no limbs but that's left as a follow up

> Looks like the mul and divrem doesn't use the bitwise_xor lookup bus and hence can be made u16.

unfortunately no because of the carry

> Is dropping the pc_limbs okay in the columns of auipc?

i updated the description that auipc, jalr and jal_lui changes are not trivial column updates and we should properly review them/formally verify them later

PR #2802, opened 2026-05-21
feat(v2.1-rv64): switch to u16 limbs in keccak/sha2 airs

shuklaayush:
PR 3 of the memory-bus-u16 split, stacked on **#2794** (disjoint from PR 2). Shrinks SHA-2 and Keccak chip trace columns that are only consumed as packed u16 cells through the memory bus.

## Why

#2794 still stores SHA-2 main-chip state / message bytes and Keccak state / pointer columns byte-shaped, packing them at memory-bus call sites via `pack_u8_block`. Because these columns are never inspected byte-by-byte by the AIR (the constraints operate on packed u16 values, or on byte-XOR buffers that intentionally stay byte-shaped), they can be stored u16-shaped directly — halving the affected trace columns and removing one `pack_u8_block` per row.

This is the second of the four cell-narrowing PRs (after PR 2 for RV64 control-flow) and is disjoint from it: the only files in this PR live under `extensions/sha2`, `extensions/keccak256`, and `crates/circuits/sha2-air`, plus one helper addition in `extensions/riscv/circuit/src/adapters/mod.rs`.

### SHA-2 section

Files: `extensions/sha2/circuit/src/sha2_chips/{main_chip,block_hasher_chip}/`, `crates/circuits/sha2-air/src/{air,columns,trace}.rs`, `extensions/sha2/circuit/cuda/`.

Main chip:

- `prev_state`, `new_state` are u16 cells.
- `message_bytes` reshapes to `message_u16s` (byte pairs packed at the call site).
- Pointer limbs `dst_ptr_limbs`, `state_ptr_limbs`, `input_ptr_limbs` reshape to u16 where the AIR only consumes them as packed pointer cells; the high u16 cell gets the scaled 16-bit range check.

Block hasher:

- `prev_hash`, `final_hash` are u16 cells.

`sha2_bus`:

- `MESSAGE_1` / `MESSAGE_2` payloads are packed to u16 on both sender and receiver.

Stays byte-shaped (8-bit bounded by construction):

- `carry_a`, `carry_e`, `carry_or_buffer` in `sha2-air`.

### Keccak section

Files: `extensions/keccak256/circuit/src/{keccakf_op,xorin}/`, `extensions/keccak256/circuit/cuda/`.

`keccakf_op`:

- `preimage`, `postimage` reshape to u16 cells.
- `buffer_ptr_limbs` reshape to u16; pointer composition uses the scaled 16-bit bound.
- `keccakf_state_bus` payload between `keccakf_op` and `keccakf_perm` is packed to u16.

`xorin`:

- Pointer limbs u16-shaped (same treatment as `keccakf_op`).
- Byte-XOR buffers (`preimage_buffer_bytes`, `input_bytes`, `postimage_buffer_bytes`) stay byte-shaped — those columns still carry byte-level XOR data and the AIR inspects them directly.

CUDA tracegen mirrors the new shapes; `keccakf_op` drops the `bitwise_lookup` parameter from its kernel and ABI signatures now that pointer bytes no longer need byte-level checks. The call chip still wires `bitwise_lu`.

### Adapter helper (cross-cut)

`extensions/riscv/circuit/src/adapters/mod.rs`.

- Add `expand_to_rv64_block`: zero-pads `N` u16 limbs to one RV64 bus block (`BLOCK_FE_WIDTH` cells). Used by `keccakf_op` and `xorin` for pointer reads against u16-celled `RV64_REGISTER_AS`.
- `BLOCK_FE_WIDTH` imported into the adapters module to support the new helper.

The existing `expand_to_rv64_register` (8 u8 limbs, byte-shaped) is kept for byte-shaped chips and the legacy paths.

## Migration notes

- Chips passing SHA-2 main-chip state through the bus should hand the bus u16 cells directly (no `pack_u8_block`).
- Pointer reads against u16-celled `RV64_REGISTER_AS` from chips storing only the low 32 bits of the register should use `expand_to_rv64_block` (BLOCK_FE_WIDTH cells, zero-padded) instead of `expand_to_rv64_register` (8 u8s).
- Keccak pointer composition should use the scaled 16-bit bound on the high u16 cell; the previous byte-pair range check on `bitwise_lookup_bus` is removed for `keccakf_op`.

resolves int-7832, int-7833


extensions/keccak256/circuit/src/keccakf_op/air.rs
```diff
@@ -67,43 +67,34 @@ impl<AB: InteractionBuilder> Air<AB> for KeccakfOpAir {
...
-            .send_range(local.buffer_ptr_limbs[0], local.buffer_ptr_limbs[1])
-            .eval(builder, is_valid);
-        self.bitwise_lookup_bus
-            .send_range(local.buffer_ptr_limbs[2], local.buffer_ptr_limbs[3])
+        self.range_bus
+            .range_check(
+                scale_rv64_ptr_high_u16_expr::<AB::Expr, _>(
+                    local.buffer_ptr_limbs[RV64_PTR_U16_LIMBS - 1],
+                    self.ptr_max_bits,
+                ),
```
GunaDD: might be a nit but I feel like this helper function makes it harder to see what is happening in the code

extensions/keccak256/circuit/src/xorin/air.rs
```diff
@@ -38,6 +38,9 @@ pub struct XorinVmAir {
     pub memory_bridge: MemoryBridge,
     /// Bus to send 8-bit XOR requests to.
     pub bitwise_lookup_bus: BitwiseOperationLookupBus,
+    /// Used to range-check the u16 high cells of `buffer_ptr` and `input_ptr`
+    /// after scaling.
```
GunaDD: this comment is not true due to the missing constraints, all limbs of the pointers have to be range checked

extensions/keccak256/circuit/src/xorin/columns.rs
GunaDD, line 48: I think the columns here can be divided by 2.

shuklaayush: no, because xor requires columns to be `u8`. xor is done using a lookup table which only works for `u8` limbs. the size of the lookup table scales as $4^{limb\\_bits}$ which makes it impractical for 16-bit limbs

extensions/keccak256/circuit/src/xorin/air.rs
```diff
@@ -167,30 +169,25 @@ impl XorinVmAir {
...
-                .send_range(pair[0] * limb_shift, pair[1] * limb_shift)
+        for top_cell in [
+            instruction.buffer_ptr_limbs[RV64_PTR_U16_LIMBS - 1],
+            instruction.input_ptr_limbs[RV64_PTR_U16_LIMBS - 1],
+        ] {
+            self.range_bus
+                .range_check(
+                    scale_rv64_ptr_high_u16_expr::<AB::Expr, _>(top_cell, self.ptr_max_bits),
+                    RV64_U16_LIMB_BITS,
+                )
```
GunaDD, line 182: missing range checks on the pointers

extensions/keccak256/circuit/src/xorin/execution.rs
```diff
@@ -166,7 +166,7 @@ unsafe fn execute_e12_impl<F: PrimeField32, CTX: ExecutionCtxTrait, const IS_E1:
         MEMORY_BLOCK_BYTES
     );
 
-    // SAFETY: RV64_MEMORY_AS is memory address space of type u8
+    // Byte-view reads from u16-celled RV64 memory.
```
GunaDD: I think this comment shouldn't be changed ?

extensions/keccak256/circuit/src/xorin/mod.rs
```diff
@@ -18,6 +20,8 @@ pub struct XorinVmExecutor {
 // number of bits = 8
 pub struct XorinVmFiller {
     pub bitwise_lookup_chip: SharedBitwiseOperationLookupChip<8>,
+    /// Used to range-check the u16 high cells of `buffer_ptr` and `input_ptr`.
```
GunaDD: this comment is incorrect (its used not only to range check the high cells)

extensions/keccak256/circuit/src/keccakf_op/air.rs
shuklaayush: i see what you're saying but i still feel like it's better to have it to consolidate the logic. for eg. if later we want to expand the address space, then it's easier to make changes in one place instead of all the chips

extensions/keccak256/circuit/src/xorin/air.rs
shuklaayush: i'll take a look at all these tomorrow but just fyi that some columns which are `u16` and directly read from memory bus don't require a range check because the bus interaction implicitly range checks them to be `u16`. this is similar to how they didn't have a `u8` range check earlier

extensions/keccak256/circuit/src/xorin/trace.rs
```diff
@@ -233,9 +234,9 @@ impl<F: PrimeField32> TraceFiller<F> for XorinVmFiller {
         trace_row.instruction.input_reg_ptr = F::from_u32(record.rs1_ptr);
         trace_row.instruction.len_reg_ptr = F::from_u32(record.rs2_ptr);
         trace_row.instruction.buffer_ptr = F::from_u32(record.buffer);
-        trace_row.instruction.buffer_ptr_limbs = record.buffer.to_le_bytes().map(F::from_u8);
+        trace_row.instruction.buffer_ptr_limbs = u32_to_le_u16_limbs(record.buffer);
         trace_row.instruction.input_ptr = F::from_u32(record.input);
-        trace_row.instruction.input_ptr_limbs = record.input.to_le_bytes().map(F::from_u8);
+        trace_row.instruction.input_ptr_limbs = u32_to_le_u16_limbs(record.input);
```
GunaDD: I think these changes are unnecessary

shuklaayush: huh, you're converting a u32 into two `u16` limbs while earlier you were converting it into four `u8` limbs

crates/circuits/sha2-air/src/config.rs
```diff
@@ -53,6 +53,8 @@ pub trait Sha2BlockHasherSubairConfig: Send + Sync + Clone {
 
     /// Number of 16-bit limbs in a SHA word
     const WORD_U16S: usize = Self::WORD_BITS / 16;
+    /// Bit width of one u16 limb.
+    const WORD_U16_BITS: usize = Self::WORD_BITS / Self::WORD_U16S;
```
GunaDD: so `WORD_U16_BITS` is just `WORD_U16_BITS = WORD_BITS / (WORD_BITS / 16) = 16`

extensions/keccak256/circuit/src/keccakf_op/air.rs
```diff
@@ -31,12 +30,13 @@ use crate::keccakf_op::columns::{KeccakfOpCols, NUM_KECCAKF_OP_COLS};
 pub struct KeccakfOpAir {
     pub execution_bridge: ExecutionBridge,
     pub memory_bridge: MemoryBridge,
-    pub bitwise_lookup_bus: BitwiseOperationLookupBus,
     /// Direct bus with keccakf pre- or post-state. Bus message is
     /// ```text
     /// is_post || timestamp || state_u16_limbs
     /// ```
     pub keccakf_state_bus: PermutationCheckBus,
+    /// Range-checks the u16 high cell of `buffer_ptr` after scaling.
```
GunaDD, line 38: this comment might be outdated if we need to fix the range checks to check all lims of buffer ptr

extensions/keccak256/circuit/src/keccakf_op/trace.rs
```diff
@@ -242,17 +253,10 @@ impl<F: PrimeField32> TraceFiller<F> for KeccakfOpChip<F> {
...
-                    1u32 << (RV64_CELL_BITS * RV64_WORD_NUM_LIMBS - self.pointer_max_bits) as u32;
-                let scaled_limb = (ptr_bytes[RV64_WORD_NUM_LIMBS - 1] as u32) * limb_shift;
-                self.bitwise_lookup_chip
-                    .request_range(scaled_limb, scaled_limb);
-
-                // AIR bounds the pointer bytes before composing them.
-                self.bitwise_lookup_chip
-                    .request_range(ptr_bytes[0] as u32, ptr_bytes[1] as u32);
-                self.bitwise_lookup_chip
-                    .request_range(ptr_bytes[2] as u32, ptr_bytes[3] as u32);
```
GunaDD, line 255: do we still need to range check the buffer ptr bytes here for soundness ?

GunaDD (approved):
overall LGTM, there is probably some minor changes needed when the fixes in PR1 come in and also some minor changes to the preflight if we decide to stick to u8 instead of u16 but nothing major.

extensions/keccak256/circuit/src/keccakf_op/air.rs
shuklaayush: we don't since that is constrained implicitly by the bus interaction

extensions/keccak256/circuit/src/keccakf_op/trace.rs
shuklaayush: no, because they're `u16`

extensions/keccak256/circuit/src/xorin/air.rs
shuklaayush: they don't

crates/circuits/sha2-air/src/config.rs
shuklaayush: removed it

PR #2803, opened 2026-05-21
feat: add memory bounds check for rvr

mansur20478:
Memory read and write functions now have an optional `check_bounds` invocation before accessing the memory. `check_bounds` checks that the access lies within the VM's addressable memory region and aborts otherwise. The same is applied for `openvm_io.h` functions that work with the user IO address space in data memory.

To turn off protected mode, add the `openvm-cli/unprotected` feature.

Mirrors the interpreter's `check_bounds` and `panic_oob` functions, and `unprotected` Cargo feature.

closes INT-7702


crates/rvr/rvr-openvm/c/openvm_state.h
```diff
@@ -85,28 +104,48 @@ static __attribute__((always_inline)) inline void reg_write(RvState* restrict st
 /* ── Per-width memory reads ──────────────────────────────────────── */
 
 static __attribute__((always_inline)) inline uint8_t rd_mem_u8(RvState* restrict state, uint32_t addr) {
+  #ifndef OPENVM_UNPROTECTED
```
shuklaayush: better to have the `ifndef` inside `check_mem_bounds` and make it noop otherwise 

but actually, let's just avoid `ifndef` altogether. we don't use any preprocessor directives in this code since they can make compilation slower. let's define an externally statically linked `check_mem_bounds` function that is noop for unprotected and does bounds checking for protected. we can have two separate header or c files

crates/rvr/rvr-openvm/c/openvm_state.h
```diff
@@ -85,28 +104,48 @@ static __attribute__((always_inline)) inline void reg_write(RvState* restrict st
...
+  check_mem_bounds(state, addr, sizeof(uint8_t));
+  #endif
+
   return *mem_ptr(state, addr);
 }
 
 static __attribute__((always_inline)) inline int8_t rd_mem_i8(RvState* restrict state, uint32_t addr) {
+  #ifndef OPENVM_UNPROTECTED
+  check_mem_bounds(state, addr, sizeof(int8_t));
```
shuklaayush: either inline checks or create separate `check_mem_bounds_xx` functions and just use
```
if (unlikely(start > mem_size - sizeof(int8_t))) abort_oob(...);
```

crates/rvr/rvr-openvm/c/openvm_state.h
```diff
@@ -135,12 +186,20 @@ static __attribute__((always_inline)) inline void wr_mem_u32(RvState* restrict s
  * so these lower to a single memcpy. */
 static __attribute__((always_inline)) inline void rd_mem_u32_range(RvState* restrict state, uint32_t base_addr,
```
shuklaayush, line 139: this probably is the only one that needs both checks in the if statement

crates/rvr/rvr-openvm/c/openvm_state.h
```diff
@@ -74,6 +76,23 @@ static __attribute__((always_inline)) inline uint8_t* mem_ptr(RvState* restrict
...
+
+/* Cold error path: out-of-lined so the inlined check stays tight. */
+static __attribute__((noinline, cold)) void abort_oob(uint32_t start, size_t size, size_t mem_size) {
+    fprintf(stderr, "Memory access out of bounds: start=%u size=%zu memory_size=%zu\n",
+            start, size, mem_size);
+    fflush(stderr);
+    abort();
+}
+
+static __attribute__((always_inline)) inline void check_mem_bounds(RvState* restrict state, uint32_t start, size_t size) {
```
shuklaayush: remove state

crates/rvr/rvr-openvm/c/openvm_state.h
```diff
@@ -74,6 +76,23 @@ static __attribute__((always_inline)) inline uint8_t* mem_ptr(RvState* restrict
...
+    const size_t mem_size = (size_t)MEMORY_MASK + 1u;
+    if (unlikely(start > mem_size || size > mem_size - start)) {
+        abort_oob(start, size, mem_size);
+    }
```
shuklaayush: add a ` __builtin_assume((start & ~(uint32_t)MEMORY_MASK) == 0u);` after `if`. maybe it helps

shuklaayush (changes requested):

shuklaayush:
can you also add a test somewhere under `feature = "rvr"` with `should_panic`

crates/rvr/rvr-openvm/src/emit/project.rs
```diff
@@ -237,6 +237,29 @@ impl CProject {
...
+            &bounds_prot_path,
+            include_str!("../../c/openvm_check_mem_bounds_protected.c"),
+        )?;
+        let bounds_unprot_path = self
+            .output_dir
+            .join("openvm_check_mem_bounds_unprotected.c");
+        fs::write(
+            &bounds_unprot_path,
+            include_str!("../../c/openvm_check_mem_bounds_unprotected.c"),
+        )?;
```
gdmlcjs: I think it would be better if we did the specialization in this file like how it's done for the tracer header. We can do that with a single `#[cfg(feature = "unprotected")]` branch instead of passing an argument to the Makefile.

gdmlcjs (commented):
I think you should also add checking for the public values address space.

gdmlcjs (approved):
LGTM! Only the lint failure seems like a problem

shuklaayush (approved):

PR #2804, opened 2026-05-21
Avoid rewriting cached aggregation keys during EVM setup

1sgtpepper:
## Summary

- reuse existing aggregation STARK and Halo2 proving keys when `cargo openvm setup --evm` is run without `--force-agg-keygen`
- only write aggregation key files after the command actually regenerates them
- keep verifier ASM and Solidity verifier generation working from the reused SDK keys

## Changes

- Split the cached-key and keygen paths in `crates/cli/src/commands/setup.rs`.
- On the cached-key path, load the existing aggregation and Halo2 proving keys into the SDK and deserialize the existing aggregation verifying key as a validity check.
- On the keygen path, generate the aggregation keys and write the STARK proving key, STARK verifying key, and Halo2 proving key immediately after generation.
- Leave root verifier ASM and EVM verifier contract generation on the shared path.

## Verification

- `git diff --check`
- `cargo check -p cargo-openvm`

Fixes #2356

### DIFF
--- a/crates/cli/src/commands/setup.rs
+++ b/crates/cli/src/commands/setup.rs
@@ -109,7 +109,7 @@ impl SetupCmd {
             // halo2 keygen does not depend on the app config
             let sdk = Sdk::standard();
 
-            let agg_vk = if !self.force_agg_keygen
+            if !self.force_agg_keygen
                 && PathBuf::from(&default_agg_stark_pk_path).exists()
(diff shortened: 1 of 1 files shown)


1sgtpepper:
Closing this in favor of #2813, which fixes the same issue with a smaller ordering-only change. Thanks for taking this over.

PR #2805, opened 2026-05-21
feat(v2.1-rv64): switch to u16 limbs in riscv adapters and modular airs 

shuklaayush:
PR 4 of the memory-bus-u16 split, stacked on **PR 2**. Shrinks adapter trace columns that only reconstruct packed memory-bus payloads, and migrates the algebra modular `is_eq` chip to u16 limbs.

This PR sits on PR 2 (not #2794) because the PR 2 squash extended the shared vec-heap u16 adapter family and dropped the `INT256_NUM_U8_LIMBS` from `riscv/circuit::adapters`; PR 4 builds on those changes.

## Why

#2794 already reshaped these adapters' memory-aux columns and bus payloads to packed u16 blocks, but the **trace columns** that reconstruct those payloads were still byte-shaped:

- `alu_w` / `mul_w` stored `rs1_high` / `rs2_high` as 4 u8 limbs even though the bus consumes them as 2 u16 cells.
- `loadstore` stored `rs1_data` as 8 u8 limbs even though pointer arithmetic only needs the low 32 bits as 2 u16 cells.
- The byte-shaped vec-heap adapters (`vec_heap`, `vec_heap_branch`, `eq_mod`) reconstructed register pointers through byte limbs and bitwise byte-pair range checks.

All of these can store the column as u16 cells directly and replace the high-byte range check with a scaled 16-bit range check on the high u16 cell, removing the `BitwiseOperationLookup` traffic.

Separately, algebra's modular `is_eq` chip operates on packed u16 operands and only needs limb-level range checks at `LIMB_BITS = 16`, so it can migrate to the new u16 adapter introduced in PR 2.

### RISC-V W-adapters and loadstore

Files: `extensions/riscv/circuit/src/adapters/{alu_w,mul_w,loadstore,mod}.rs`, `extensions/riscv/circuit/src/loadstore/{execution,tests}.rs`, `extensions/riscv/circuit/src/{base_alu_w,divrem_w,mul_w,shift_w}/tests.rs`, `extensions/riscv/circuit/cuda/include/riscv/adapters/{alu_w,mul_w,loadstore}.cuh`.

- `alu_w` / `mul_w`: `rs1_high`, `rs2_high` are packed as two u16 cells; bus reads zero-extend the lower 32 bits and pull u16-shaped upper cells directly from the trace.
- `loadstore`: `rs1_data` shrinks to `RS1_DATA_U16S = 2` u16 cells (the low 32 bits of the rs1 register). Bus reads zero-extend the upper half on the wire.
- STORE-to-`DEFERRAL_AS` is explicitly unsupported in this adapter and the now-dead `timed_write_deferral` path is removed — #2794 already dropped that opcode, and the remaining call would have instantiated `TracingMemory::write::<F, 8>`, which violates `BLOCK_FE_WIDTH`. The F-celled path needs a separate two-message bus shape if it is ever reintroduced.
- Shared helpers handle u8-pair packing, low-32-bit u16 composition, high-u16 pointer scaling, and 16-bit immediate sign-extension.
- W-suffix tests (`base_alu_w`, `divrem_w`, `mul_w`, `shift_w`) updated to prank u16 cells where the column is migrated.

### Shared heap-data adapters

Files: `extensions/riscv-adapters/src/{vec_heap,vec_heap_branch,eq_mod,eq_mod_u16,lib}.rs`, `extensions/riscv-adapters/cuda/include/riscv-adapters/{constants,vec_heap,vec_heap_branch}.cuh`.

- `vec_heap`, `vec_heap_branch`, and `eq_mod` materialize register pointer values (`rs_val` / `rd_val`) as u16 cells.
- Pointer bounds use `VariableRangeChecker` on the high u16 cell instead of byte-pair `BitwiseOperationLookup` send_range.
- Adds the byte-shaped vec-heap adapter `READ_SIZE` / `WRITE_SIZE` const generics so byte-shaped 256-bit chips (`BaseAlu256`, `Mul256`, `Shift256`) and u16-shaped 256-bit chips (`LessThan256`) can share the same wrapper without losing the block-size information needed for `MemoryWriteAuxCols` and pointer arithmetic.
- Adds `Rv64IsEqualModU16Adapter*` for the new u16-shaped modular `is_eq`.
- CUDA mirrors the new column shapes and packing helpers.

### Algebra modular `is_eq`

Files: `extensions/algebra/circuit/src/modular_chip/{is_eq,addsub,muldiv,mod,tests}.rs`, `extensions/algebra/circuit/src/extension/{modular,hybrid,fp2}.rs`, `extensions/algebra/circuit/src/{execution,fields,preflight,lib}.rs`, `extensions/algebra/circuit/src/fp2_chip/{addsub,mod,muldiv,tests}.rs`.

- `LIMB_BITS` 8 → 16 where the chip's limbs are only consumed as packed u16 values.
- Wires through the new `Rv64IsEqualModU16` adapter variant.
- Core uses `VariableRangeChecker` for `b_lt_diff - 1` and `c_lt_diff - 1` instead of byte-pair bitwise checks.
- CPU, hybrid, and CUDA registration paths all use `Rv64IsEqualModU16Adapter*`.
- Negative tests that previously pranked `[F; 4]` byte arrays now prank `[F; 2]` u16 cells where the column is migrated; positive coverage preserved.
- Mod-builder generated AIRs are untouched.

`is_eq` executor memory reads were already adjusted in #2794 for `MEMORY_BLOCK_BYTES`; PR 4 does not rework that API.

### Cascade through dependents

Files: `extensions/bigint/circuit/src/{lib,extension/mod,tests}.rs`, `extensions/bigint/circuit/cuda/src/bigint.cu`, `extensions/ecc/circuit/src/weierstrass_chip/{add_ne,double}/mod.rs`.

- Bigint type aliases and ecc weierstrass call sites pick up the new `READ_SIZE` / `WRITE_SIZE` generic positions on `Rv64VecHeapAdapter*`. Byte-shaped 256-bit chips pass `MEMORY_BLOCK_BYTES`; u16-shaped chips pass `BLOCK_FE_WIDTH`.
- Bigint pulls `INT256_NUM_U8_LIMBS` from the bigint crate root (per PR 2's reviewer feedback) instead of `riscv/circuit::adapters`.

## Migration notes

- `Rv64VecHeapAdapter*` now takes 5 const generics (`NUM_READS, BLOCKS_PER_READ, BLOCKS_PER_WRITE, READ_SIZE, WRITE_SIZE`). Byte-shaped 256-bit chips: pass `MEMORY_BLOCK_BYTES` for both sizes. u16-shaped chips: use `Rv64VecHeapU16Adapter`.
- `alu_w` / `mul_w` adapter callsites that materialized `rs_high` bytes should pull u16-shaped upper cells from the new column slot.
- `loadstore` `rs1_data` columns shrink from 8 u8s to 2 u16 cells; trace-fillers that previously wrote 8 byte cells should write 2 u16 cells (low 32 bits) and let the AIR zero-extend on the bus.
- `is_eq` modular chip callers should switch from `Rv64IsEqualModAdapter*` to `Rv64IsEqualModU16Adapter*` and pass `LANE_SIZE = BLOCK_FE_WIDTH` (= 4 u16 cells per heap-read block).
- STORE-to-`DEFERRAL_AS` is no longer routed through `Rv64LoadStoreAdapter`; the F-celled deferral-store path was already removed in #2794 and any reintroduction needs a dedicated two-message bus shape.

resolves int-7830, int-7831


extensions/riscv/circuit/src/adapters/loadstore.rs
```diff
@@ -84,7 +85,8 @@ impl<AB: InteractionBuilder> VmAdapterInterface<AB::Expr> for Rv64LoadStoreAdapt
 pub struct Rv64LoadStoreAdapterCols<T> {
     pub from_state: ExecutionState<T>,
     pub rs1_ptr: T,
-    pub rs1_data: [T; RV64_WORD_NUM_LIMBS],
+    /// Low 32 bits of the rs1 register, packed as two u16 cells.
```
GunaDD, line 88: so just to check, if we want to add support for memory pointer up to 2^64 this column has to be expanded right ?

shuklaayush: yes, but supporting 64-bit addresses will lead to issues in other parts of the system atm. for eg. in execution we allocate all guest memory up front and we can't allocate 2^64 bytes. which is why we want to start with 32-bit addresses (2^32 bytes = 4GB which is still practical). to support wider addresses like 64-bit, we'll probably have to move to paged memory for execution (which has performance implications)

extensions/riscv-adapters/src/vec_heap.rs
```diff
@@ -137,43 +138,27 @@ impl<
             self.memory_bridge
                 .read(
                     MemoryAddress::new(AB::F::from_u32(RV64_REGISTER_AS), ptr),
-                    pack_u8_block::<AB>(&expand_to_rv64_register(&val)),
+                    expand_to_rv64_block(&val),
```
GunaDD, line 144: I noticed that since the PR1 of the memory bus change, we are now fixing the memory reads to have block size equal to 4 instead of having it as a generic as before. Is there any downsides to this? One thing is that we have to use this `expand_to_rv64_block` function in some places. Is there any safety issue to fixing the block size to 4 ?

extensions/riscv-adapters/src/vec_heap_branch_u16.rs
```diff
@@ -112,19 +110,19 @@ impl<AB: InteractionBuilder, const NUM_READS: usize, const BLOCKS_PER_READ: usiz
...
-        let limb_shift = AB::F::from_u32(1 << (RV64_LOW32_BITS - self.address_bits));
+        // Each materialized pointer is stored as two u16 cells. Bound the high
+        // cell so `lo + 2^16 * hi < 2^address_bits`.
         for val in cols.rs_val.iter() {
             self.range_bus
-                .range_check(val[1] * limb_shift, RV64_U16_LIMB_BITS)
+                .range_check(
+                    scale_rv64_ptr_high_u16(val[1], self.address_bits),
+                    RV64_U16_LIMB_BITS,
+                )
```
GunaDD, line 123: Just to confirm, the code before and after this change here does the same thing right ? It's just written in a way using the helper function. But I feel like the old way to write it is easier to read & understand.

crates/vm/cuda/include/system/memory/offline_checker.cuh
```diff
@@ -43,21 +45,27 @@ using MemoryWriteBytesAuxRecord = MemoryWriteAuxRecord<uint8_t, NUM_LIMBS>;
...
+
 template <typename T, size_t NUM_U16_CELLS>
 __device__ inline void pack_u32_to_u16_cells_le(T (&out)[NUM_U16_CELLS], uint32_t value) {
     static_assert(NUM_U16_CELLS == sizeof(uint32_t) / sizeof(uint16_t));
 #pragma unroll
     for (size_t i = 0; i < NUM_U16_CELLS; i++) {
-        out[i] = T((value >> ((8 * sizeof(uint16_t)) * i)) & 0xffffu);
+        out[i] = T(uint16_t(value >> ((8 * sizeof(uint16_t)) * i)));
     }
 }
```
GunaDD: was this something left out from PR1 ?

extensions/bigint/circuit/src/extension/mod.rs
```diff
@@ -324,7 +324,7 @@ where
         inventory.next_air::<Rv64BaseAlu256Air>()?;
         let alu = Rv64BaseAlu256Chip::new(
             BaseAluFiller::new(
-                Rv64VecHeapAdapterFiller::new(pointer_max_bits, bitwise_lu.clone()),
+                Rv64VecHeapAdapterFiller::new(pointer_max_bits, range_checker.clone()),
```
GunaDD: was this remaining stuff from the first PR ?

extensions/ecc/circuit/src/extension/hybrid.rs
```diff
@@ -137,7 +130,6 @@ impl VmProverExtension<GpuBabyBearPoseidon2Engine, DenseRecordArena, Weierstrass
                     config.clone(),
                     mem_helper.clone(),
                     range_checker.clone(),
```
GunaDD, line 132: was this remaining stuff from the first PR ?

extensions/algebra/circuit/src/extension/modular.rs
```diff
@@ -309,7 +291,6 @@ impl<SC: StarkProtocolConfig> VmCircuitExtension<SC> for ModularExtension {
                     memory_bridge,
                     config.clone(),
                     range_checker_bus,
-                    bitwise_lu,
                     pointer_max_bits,
                     start_offset,
                 );
```
GunaDD, line 296: I saw the `get_modular_addsub_air` is not changed to using the u16 vecheap adapter chips. Is this intentional or should we change it to use the u16 columns for the adapter? Also, the mildiv and other things might also need an update if it is not intentional.

extensions/algebra/circuit/src/fp2_chip/addsub.rs
```diff
@@ -71,18 +69,12 @@ pub fn get_fp2_addsub_air<const BLOCKS: usize>(
...
 ) -> Fp2Air<BLOCKS> {
     let (expr, local_opcode_idx, opcode_flag_idx) = gen_base_expr(config, range_checker_bus);
     Fp2Air::new(
-        Rv64VecHeapAdapterAir::new(
-            exec_bridge,
-            mem_bridge,
-            bitwise_lookup_bus,
-            pointer_max_bits,
-        ),
+        Rv64VecHeapAdapterAir::new(exec_bridge, mem_bridge, range_checker_bus, pointer_max_bits),
```
GunaDD, line 77: similar comment here, should we use the u16 vec heap adapter air ?

extensions/algebra/circuit/src/fp2_chip/muldiv.rs
```diff
@@ -111,18 +109,12 @@ pub fn get_fp2_muldiv_air<const BLOCKS: usize>(
...
 ) -> Fp2Air<BLOCKS> {
     let (expr, local_opcode_idx, opcode_flag_idx) = gen_base_expr(config, range_checker_bus);
     Fp2Air::new(
-        Rv64VecHeapAdapterAir::new(
-            exec_bridge,
-            mem_bridge,
-            bitwise_lookup_bus,
-            pointer_max_bits,
-        ),
+        Rv64VecHeapAdapterAir::new(exec_bridge, mem_bridge, range_checker_bus, pointer_max_bits),
```
GunaDD, line 117: same comment here too about the u16 vecheapadapter

extensions/algebra/circuit/src/modular_chip/addsub.rs
```diff
@@ -68,18 +66,12 @@ pub fn get_modular_addsub_air<const BLOCKS: usize>(
...
 ) -> ModularAir<BLOCKS> {
     let (expr, local_opcode_idx, opcode_flag_idx) = gen_base_expr(config, range_checker_bus);
     ModularAir::new(
-        Rv64VecHeapAdapterAir::new(
-            exec_bridge,
-            mem_bridge,
-            bitwise_lookup_bus,
-            pointer_max_bits,
-        ),
+        Rv64VecHeapAdapterAir::new(exec_bridge, mem_bridge, range_checker_bus, pointer_max_bits),
```
GunaDD, line 74: same comment here about the u16 vecheapadapter

extensions/algebra/circuit/src/modular_chip/is_eq.rs
```diff
@@ -289,8 +291,8 @@ where
 #[derive(AlignedBytesBorrow, Debug)]
 pub struct ModularIsEqualRecord<const READ_LIMBS: usize> {
     pub is_setup: bool,
-    pub b: [u8; READ_LIMBS],
-    pub c: [u8; READ_LIMBS],
+    pub b: [u16; READ_LIMBS],
+    pub c: [u16; READ_LIMBS],
 }
```
GunaDD, line 294: I think we would want to undo the change of type to u16 for the records here.

extensions/algebra/circuit/src/modular_chip/is_eq.rs
```diff
@@ -314,8 +316,8 @@ pub struct ModularIsEqualFiller<
 > {
     adapter: A,
     pub offset: usize,
-    pub modulus_limbs: [u8; READ_LIMBS],
-    pub bitwise_lookup_chip: SharedBitwiseOperationLookupChip<LIMB_BITS>,
+    pub modulus_limbs: [u16; READ_LIMBS],
+    pub range_checker_chip: SharedVariableRangeCheckerChip,
 }
```
GunaDD, line 319: same here and other places in this file.

GunaDD (commented):
Finished the review on this PR. The main comments specific for this PR would be:
* Looks like some of the algebra chips adapter can use the u16 columns version ? So we should probably use that version (i.e. wire that version in the extend_prover) to save columns if possible. 
* Other than that, similar comments to previous PR about wanting to revert the type of the preflight records back to u8 from u16.
* Also ensure the CUDA paths are up to date (synced) with the CPU version. 
* Other minor comments in the PR.

extensions/algebra/circuit/src/modular_chip/is_eq.rs
shuklaayush: i feel it's easier if it's `u16`. there's no reason for it to be `u8` if the columns are `u16`

extensions/riscv-adapters/src/vec_heap.rs
shuklaayush: no, there shouldn't be any issue. we in essence already use a fixed block size after the memory access adapters were removed. these prs just make it more explicit and avoid the unnecessary flexibility altogether
the `expand_to_rv64_block` is used not because of the fixed block size but because sometimes we want to read only a `u32` value while the standard word size in rv64 is `u64`

extensions/algebra/circuit/src/extension/modular.rs
shuklaayush: yes, because mod builder internally uses `u8` limbs and i didn't want to touch that

extensions/algebra/circuit/src/fp2_chip/addsub.rs
shuklaayush: no, because mod builder requires `u8` limbs

extensions/riscv-adapters/src/vec_heap_branch_u16.rs
shuklaayush: yes it's the same constraint. yeah the function name is not the best but i feel it's better than having to duplicate this everywhere
```rust
let limb_shift = AB::F::from_u32(1 << (RV64_LOW32_BITS - self.address_bits));
```

PR #2806, opened 2026-05-22
fix(v2.1-rvr): use clang-only rvr native builds

shuklaayush:
## Summary

- Standardize all RVR native FFI builds on clang (default `clang-22`). Non-clang `CC` values are now ignored and an explicit error points users at `RVR_CC=clang-22` / `clang-22` install instructions instead of silently falling back to GCC.
- Drop the XKCP keccak backend and its submodule; keep only the OpenSSL asm backend. Removes the `backend-openssl` / `backend-xkcp` cargo features from `rvr-openvm-ext-keccak-ffi`.
- Extract shared RVR build-script logic into a new `rvr-openvm-build` crate (`default_compiler_command`, `ensure_clang_compiler`, `command_exists`, `build_rust_staticlib`) and reuse it from every `extensions/*/rvr/build.rs` and from `rvr-openvm`'s toolchain module.
- Replace `submodules: recursive` checkouts in `lints.yml` and `riscv.yml` with targeted `git submodule update --init --depth 1` for just the submodules each job needs; mark the secp256k1 submodule shallow.

resolves int-7568

### DIFF
--- a/.github/workflows/base-tests.cuda.yml
+++ b/.github/workflows/base-tests.cuda.yml
@@ -34,6 +34,10 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         env:
           RUSTUP_PERMIT_COPY_RENAME: "1"
+      - name: Configure sccache startup timeout
+        run: |
+          mkdir -p ~/.config/sccache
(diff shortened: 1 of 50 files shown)


gdmlcjs (approved):
LGTM!

PR #2808, opened 2026-05-22
feat(v2.1-rv64): switch to u16 limbs in deferral airs

shuklaayush:
PR 5 of the memory-bus-u16 split, stacked on **#2794**. Shrinks deferral trace columns whose constraints only operate on packed u16 values, and doubles the sponge absorption rate of the deferral output chip from one digest per row to two.

## Why

#2794 still stores deferral commit / output-key columns byte-shaped while packing the memory-bus payloads. Shrinking those columns to u16 cells halves the affected trace width and lets the canonicity sub-AIR use per-u16-cell 16-bit range checks instead of byte-pair `BitwiseOperationLookup` checks.

Separately, the deferral output chip absorbs only `DIGEST_SIZE = 8` bytes per Poseidon row even though the sponge state can hold `2 * DIGEST_SIZE = 16` bytes. Doubling the absorption rate halves the number of output rows for the same byte stream.

### Trace columns to u16 cells

Files: `extensions/deferral/circuit/src/{call,output,canonicity}/`, `extensions/deferral/circuit/src/utils.rs`, `extensions/deferral/circuit/cuda/`.

- Call core: `input_commit`, `output_commit`, `output_len` are u16 cells.
- Output core: `output_commit`, `output_len` are u16 cells.
- Canonicity sub-AIR operates over u16 cells; upper-bound checks become per-u16-cell 16-bit range checks (instead of byte-pair bitwise checks).
- Deferral `rd_val` / `rs_val` register-pointer columns are u16-shaped.
- Reuses #2794's `split_byte_memory_ops` / `split_f_memory_ops` and `byte_memory_op_chunk` / `f_memory_op_chunk` helpers; no new `split_memory_ops`-style helper is introduced.
- CUDA tracegen mirrors the new column shapes.

### Sponge absorption rate

Files: `extensions/deferral/circuit/src/output/{air,trace,execution,tests}.rs`, `extensions/deferral/circuit/src/def_fn.rs`, `extensions/deferral/circuit/src/extension/mod.rs`, `extensions/deferral/circuit/cuda/src/output.cu`.

- `SPONGE_BYTES_PER_ROW = 2 * DIGEST_SIZE = 16` (was `DIGEST_SIZE = 8`).
- `sponge_inputs` cells hold u16 values on data rows, each carrying two output bytes; data-row `sponge_inputs` receive 16-bit range checks. The init row remains `[deferral_idx, output_len, 0, ...]` and is exempt.
- Two memory-bus writes per output row instead of one — the row now consumes 16 bytes of output, not 8.
- `def_fn::hash_output_raw` packs byte pairs into each sponge cell before absorbing.
- Bitwise-lookup wiring removed from the output chip; the call chip still uses bitwise lookup for non-canonical commit checks.

### Output length contract

- `output_len` remains a byte count.
- Raw deferral outputs must be a multiple of 16 bytes; non-16-byte-aligned outputs are invalid with no implicit padding.
- Host-side failure-path test asserts the panic when a non-16-aligned raw output is submitted.

## Migration notes

- Hosts producing deferral output must align raw outputs to a multiple of 16 bytes.
- Output-chip configs that previously passed `bitwise_lu` should drop the argument.

resolves int-7834


extensions/deferral/circuit/src/canonicity/air.rs
```diff
@@ -67,18 +71,22 @@ where
 }
 
 impl CanonicitySubAir {
+    /// Constrain canonicity of the `F_NUM_U16S` LE u16-cell limbs in `x`.
+    /// `x` is given LE; the sub-AIR consumes MSL-first, so we reverse here.
     pub fn assert_canonicity<AB: InteractionBuilder>(
         &self,
         builder: &mut AB,
-        x: &[AB::Var],
+        x: [AB::Expr; F_NUM_U16S],
```
stephenh-axiom-xyz: nit: Is it possible to use a slice of `Into<AB::Expr>` or smth

guest-libs/verify-stark/circuit/src/lib.rs
```diff
@@ -103,6 +105,9 @@ impl<SC: StarkProtocolConfig<F = F>, S: AggregationSubCircuit> Circuit<SC>
             output_commit_bus,
             def_idx: self.def_idx,
         };
+        let output_range_air = RangeCheckerAir::<U16_BITS> {
```
stephenh-axiom-xyz, line 119: Given you converted the RISC-V chips to do 16-bit limb addition/multiplication, I think you could change the range checker in `recursion` to be 16-bits instead of adding it here?

extensions/deferral/circuit/src/call/air.rs
```diff
@@ -61,17 +60,37 @@ pub struct DeferralCallReads<B, F> {
...
     // Deferral address space accumulators after incorporating the current deferral call
     pub new_input_acc: [F; DIGEST_SIZE],
     pub new_output_acc: [F; DIGEST_SIZE],
 }
 
+// Byte-shaped sibling of DeferralCallReads used for the tracegen record.
+// The trace filler packs byte pairs into the u16-shaped columns.
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct DeferralCallReadsBytes<F> {
```
stephenh-axiom-xyz, line 78: Why not initialize the `DeferralCallReads<u16, F>` during execution?

extensions/deferral/circuit/src/call/air.rs
```diff
@@ -50,8 +49,8 @@ use crate::{
 #[repr(C)]
 #[derive(AlignedBorrow, StructReflection, Clone, Copy, Debug)]
 pub struct DeferralCallReads<B, F> {
```
stephenh-axiom-xyz: nit: `B` stands for bytes

extensions/deferral/circuit/src/call/air.rs
```diff
@@ -89,7 +108,7 @@ pub struct DeferralCallCoreCols<T> {
 pub struct DeferralCallCoreAir {
     pub count_bus: DeferralCircuitCountBus,
     pub poseidon2_bus: DeferralPoseidon2Bus,
-    pub bitwise_bus: BitwiseOperationLookupBus,
+    pub range_bus: VariableRangeCheckerBus,
```
stephenh-axiom-xyz, line 114: Did we not update `BitwiseOperationLookupAir` to handle u16 operations? Saves an interaction per range check if we did

extensions/deferral/circuit/src/call/tests.rs
```diff
@@ -121,7 +110,7 @@ fn deferral_fns(num_deferrals: usize) -> Vec<Arc<DeferralFn>> {
                     .iter()
                     .fold(idx as u8, |acc, byte| acc.wrapping_add(*byte));
                 let num_chunks = (seed as usize % 4) + 1;
-                let len = num_chunks * DIGEST_SIZE;
+                let len = num_chunks * SPONGE_BYTES_PER_ROW;
```
stephenh-axiom-xyz, line 108: Why is the length changed here? The input should be in u16s now no?

extensions/deferral/circuit/src/output/air.rs
```diff
@@ -57,26 +59,26 @@ pub struct DeferralOutputCols<T> {
...
     pub output_commit_and_len_aux: [MemoryReadAuxCols<T>; OUTPUT_TOTAL_MEMORY_OPS],
 
-    // Auxiliary columns to ensure the canonicity of each F byte decomposition in
-    // output_commit.
+    // Auxiliary columns to ensure canonicity of output_commit cells.
     pub output_commit_lt_aux: [CanonicityAuxCols<T>; DIGEST_SIZE],
 
     // First row sponge input is [deferral_idx, output_len, 0, ...].
-    // Later rows sponge and write the next DIGEST_SIZE output bytes.
+    // Later rows sponge and write the next SPONGE_BYTES_PER_ROW output bytes.
```
stephenh-axiom-xyz: Why aren't we measuring this in u16s?

extensions/deferral/circuit/src/def_fn.rs
```diff
@@ -78,7 +81,10 @@ fn hash_output_raw<F: VmField>(
     deferral_idx: u32,
     output_ref: &[u8],
```
stephenh-axiom-xyz, line 67: Same comment, why not u16?

stephenh-axiom-xyz (commented):
Seems to be a lot of places that still use `u8` instead of `u16`, I've pointed some out in comments

extensions/deferral/circuit/src/call/air.rs
shuklaayush: `BitwiseOperationLookupAir` can't handle `u16` operations since the table size would need to be `2^32`

guest-libs/verify-stark/circuit/src/lib.rs
shuklaayush: hmm yeah, i was trying to not touch recursion but i guess maybe just bumping the recursion range checker to be 16-bits isn't that big of a deal

crates/continuations/programs/examples/multiple.elf
GunaDD, line 1: was this accidentally included ?

crates/continuations/src/circuit/root/commit/air.rs
```diff
@@ -42,9 +45,7 @@ impl UserPvsCommitAir {
     ) -> Self {
         // Each leaf consumes `DIGEST_SIZE` public values, which are compressed with zeros
         // to compute the leaf hash. We require at least one leaf, and a full binary tree.
-        debug_assert!(num_user_pvs >= DIGEST_SIZE);
-        debug_assert!(num_user_pvs.is_multiple_of(DIGEST_SIZE));
-        debug_assert!((num_user_pvs / DIGEST_SIZE).is_power_of_two());
+        assert_user_pvs_shape(num_user_pvs);
         let encoder = Encoder::new(num_user_pvs / DIGEST_SIZE, MAX_ENCODER_DEGREE, true);
```
GunaDD: I guess this is unrelated to the u16 memory bus change

crates/continuations/src/circuit/root/mod.rs
```diff
@@ -39,6 +39,31 @@ pub struct RootCircuit<S: AggregationSubCircuit> {
...
+
+impl<S: AggregationSubCircuit> RootCircuit<S> {
+    pub fn new(
+        verifier_circuit: Arc<S>,
+        internal_recursive_vk_commit: VkCommitBytes,
+        def_hook_commit: Option<CommitBytes>,
+        memory_dimensions: MemoryDimensions,
+        num_user_pvs: usize,
+    ) -> Self {
+        assert_user_pvs_shape(num_user_pvs);
```
GunaDD, line 65: why was this assertion added now ? was this always needed or is this a new requirement

crates/continuations/programs/examples/multiple.elf
shuklaayush: no, the output in the program changed so this had to be re-compiled

crates/continuations/src/circuit/root/mod.rs
shuklaayush: it just consolidates all these checks into a single function
```
debug_assert!(num_user_pvs >= DIGEST_SIZE);
debug_assert!(num_user_pvs.is_multiple_of(DIGEST_SIZE));
debug_assert!((num_user_pvs / DIGEST_SIZE).is_power_of_two());
```

PR #2811, opened 2026-05-23
Return an error when Cargo.toml is missing

1sgtpepper:
## Summary

`find_manifest_dir` currently panics if it walks up to the filesystem root without finding a `Cargo.toml`, because it unwraps `Path::parent()`.

This changes that path to return a normal `eyre` error instead, matching the function's `Result` return type.

## Testing

- `cargo test -p cargo-openvm --lib find_manifest_dir --no-default-features`
- `cargo check -p cargo-openvm --no-default-features`
- `git diff --check`
- `rustfmt --check crates/cli/src/util.rs`

### DIFF
--- a/crates/cli/src/util.rs
+++ b/crates/cli/src/util.rs
@@ -47,7 +47,11 @@ pub fn find_manifest_dir(mut current_dir: PathBuf) -> Result<PathBuf> {
     while !current_dir.join("Cargo.toml").exists() {
         current_dir = current_dir
             .parent()
-            .expect("Could not find Cargo.toml in current directory or any parent directory")
+            .ok_or_else(|| {
+                eyre::eyre!(
+                    "Could not find Cargo.toml in current directory or any parent directory"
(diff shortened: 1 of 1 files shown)


jonathanpwang:
Is this problem still present on `develop-v2.0.0-rc.2`?

1sgtpepper:
No, it looks like this is already fixed on `develop-v2.0.0-rc.2`.

I checked `crates/cli/src/util.rs` on that branch and `find_manifest_dir` already uses `parent().ok_or_else(...)?` there, so the missing-`Cargo.toml` case returns an `eyre` error instead of panicking. This PR is against `main`, where the `expect(...)` is still present.

Should I close this, or would you prefer this change to still land on `main`?

jonathanpwang:
Closing for now. Working development changes are on top of `develop-v2.0.0-rc.2` for the next release. Thanks for bringing up the issue.

PR #2814, opened 2026-05-26
fix(v2.1-rvr): make rvr execution and preflight consistent

shuklaayush:
Three correctness fixes in the RVR backend so that its per-segment VM state is byte-identical to the Rust preflight executor's, plus a test-coverage fix that closes the gap that hid all three bugs from `check_rvr_equivalence`.

### `test(v2.1-rvr): cover all guest address spaces in pure RVR equivalence check`

- `check_rvr_equivalence` (and its AOT sibling) only walked AS=1 (registers) byte-by-byte, silently missing any divergence in RV32 main memory (AS=2), public values (AS=3), and deferral (AS=4). All three correctness fixes below live in AS=2 or AS=4 and would have surfaced on the first `air_test`-style run if the check had walked every address space.

Extracted the closure into a `check_vm_state_eq(lhs, rhs) -> eyre::Result<()>` free function shared by both the RVR and AOT equivalence checks, replacing the AS=1-only loop with a slice-level diff over every `LinearMemory`. Short-circuits at the first mismatch and reports `(AS, byte offset, lhs, rhs)`. Microseconds on typical test VM configs.

### `fix(v2.1-rvr): use block-boundary instret in on_periodic_check`

- `SegmentationState::on_periodic_check` was bumping `segmentation_ctx.instret` by a full `segment_check_insns` interval up-front, then incrementing `tracer.check_counter` by the same delta on the way out. The anchor and the countdown ended up ahead of the actual VM by exactly `remaining_counter`, so the next interval inherited an inconsistent baseline.

In termination paths this could let the segmenter seal a non-terminal block as the final segment. The callback now:
- computes the actual block-boundary instret directly: `prev_anchor + (segment_check_insns - remaining_counter)`,
- writes that back as the new anchor,
- resets `check_counter` to a full fresh interval rather than incrementing.

This matches the Rust metered executor's behavior at the same point.

### `fix(v2.1-rvr): write zero output for modular/fp2 setup ops`

Mod-builder evaluates `SymbolicExpr` inputs **modulo the configured prime**. For `SETUP_ADDSUB` / `SETUP_MULDIV` and their Fp2 counterparts, the compute formula resolves to `Input(0)`, which during setup is the modulus `p` itself — so the variable is `p % p = 0`. The VM writes 32 zero bytes (64 for Fp2) to `rd`.

`rvr_ext_mod_setup` and `rvr_ext_fp2_setup` were copying `rs1`'s bytes (the modulus) to `rd`. Those bytes then leaked into the guest's stack as register-loaded values, propagating downstream as a memory divergence between RVR and preflight at later segment boundaries.

The FFI now traces the `rs1`/`rs2` reads (still required for chip metering) but writes zero bytes to `rd`.

### `fix(v2.1-rvr): sync DEFERRAL_AS in rvr_ext_deferral_call`

The deferral CALL FFI in RVR only traced AS=4 access for metering and never updated the `(input_acc, output_acc)` accumulator bytes. The Rust preflight executor (`DeferralCallExecutor::execute_e12_impl`) hashes each `(old_acc, commit)` pair via poseidon2 and writes the new accumulator F's to DEFERRAL_AS. Every deferral CALL therefore left RVR's AS=4 a hash-round behind preflight, producing a memory divergence that cascaded through subsequent CALLs.

Plumbed a `(*mut F, len_in_F_units)` alias of DEFERRAL_AS through `OpenVmIoState` (via a new `deferral_memory_ptr` helper in `bridge.rs` with a debug-mode alignment check on the `u8 → F` cast) and registered a `DeferralCompressFn` poseidon2 closure on the host side. `host_deferral_call_lookup` now hashes the accumulators and writes the new F bytes into AS=4 in F-element units that exactly match preflight's `vm_write::<F, BLOCK_SIZE>` layout. `F::from_u32` is bijective with the perm output for `MontyField31`, so the stored bytes are byte-identical to what the preflight executor writes.

resolves int-7974

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/segment_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/segment_ctx.rs
@@ -614,3 +614,36 @@ impl SegmentationCtx {
         );
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn small_segmentation_ctx() -> SegmentationCtx {
+        let config = SegmentationConfig {
+            limits: SegmentationLimits::new(4, usize::MAX, usize::MAX),
+            ..SegmentationConfig::default()
+        };
+        SegmentationCtx::new(
+            vec!["air".to_string()],
+            vec![1],
+            vec![0],
+            vec![false],
+            config,
+        )
+    }
+
+    #[test]
+    fn test_check_and_segment_uses_last_safe_checkpoint() {
+        let mut ctx = small_segmentation_ctx();
+        ctx.update_checkpoint(10, &[2]);
+
+        let mut trace_heights = vec![8];
+        assert!(ctx.check_and_segment(15, &mut trace_heights, &[false]));
+
+        assert_eq!(ctx.segments.len(), 1);
+        assert_eq!(ctx.segments[0].instret_start, 0);
+        assert_eq!(ctx.segments[0].num_insns, 10);
+        assert_eq!(ctx.segments[0].trace_heights, vec![2]);
+    }
+}
--- a/crates/vm/src/arch/rvr/bridge.rs
+++ b/crates/vm/src/arch/rvr/bridge.rs
@@ -5,7 +5,12 @@
 //! `Streams<F>` and the host RNG are borrowed directly into [`OpenVmIoState`]
 //! and never converted upfront.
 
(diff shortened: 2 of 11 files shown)


gdmlcjs (approved):
A lot of critical fixes... Looks good to me!

PR #2815, opened 2026-05-26
fix: constrain deferral circuit aggregation padding and depth

stephenh-axiom-xyz:
Resolves INT-7918 and INT-7920.

## Overview

This PR tightens deferral aggregation around padded Merkle subtrees.

The main changes are:

- `DeferralAggregationPvs` now carries `merkle_depth`, in addition to `merkle_commit` and `num_def_circuit_proofs`.
- `DeferralAggPvsAir` constrains aggregation depth and canonical padding hashes when the second child is absent.
- `DeferralHookPvsAir` constrains the final decommitment tree size from `merkle_depth`, and range-checks the depth against `MAX_DEF_AGG_MERKLE_DEPTH`.
- `MerkleRootMessage` optionally carries the Merkle trace row count so the hook can bind the final aggregation root to the expected tree size.
- Verifier trace generation now supports external power-check inputs, mirroring existing external range-check plumbing.

### Purpose

The deferral aggregation public values now include:

| Public value | Meaning |
| --- | --- |
| `merkle_commit` | Root commit for the current deferral aggregation subtree. |
| `num_def_circuit_proofs` | Number of real deferral circuit proofs contained in the subtree. |
| `merkle_depth` | Log2 of the padded Merkle leaf count represented by the subtree. |

The AIR constrains:

- Leaf children still hash `input_commit || output_commit` into `merkle_commit`.
- Internal children receive both `num_def_circuit_proofs` and `merkle_depth` from the child aggregation PVS.
- One-row wrapper proofs preserve the child `merkle_commit`, `num_def_circuit_proofs`, and `merkle_depth`.
- Two-row aggregation proofs set parent `num_def_circuit_proofs` to the sum of present children.
- Two-row aggregation proofs set parent `merkle_depth = child_merkle_depth + 1`.
- If both children are present, their child depths must be equal.
- If the second child is absent, its `merkle_commit` columns encode a depth selector; the AIR uses that selector to choose the canonical zero hash at the matching child depth.

### Example Trace Shapes

`DeferralAggPvsCols` did not add physical columns. For internal rows, unused `child_pvs.input_commit` entries are repurposed; for padded second rows, the start of `merkle_commit` stores encoder flags instead of a digest.

| Case | Row | `is_present` | `has_verifier_pvs` | `child_pvs.input_commit[0]` | `child_pvs.input_commit[1]` | `merkle_commit` |
| --- | ---: | ---: | ---: | --- | --- | --- |
| Leaf child | 0 | `1` | `0` | child input digest word | child input digest word | `H(input_commit, output_commit)` |
| Internal child | 0 | `1` | `1` | child `num_def_circuit_proofs` | child `merkle_depth` | child aggregation root |
| Present sibling | 1 | `1` | same as row 0 | same meaning as row 0 | same meaning as row 0 | sibling root |
| Padded sibling | 1 | `0` | same as row 0 | `0` if internal | `0` if internal | depth encoder flags in the first encoder-width words |

For a padded two-row proof:

| Child depth | Parent depth | Encoded padding choice | Right child used in Poseidon lookup |
| ---: | ---: | --- | --- |
| `0` | `1` | flag `0` | `zero_hash(1)` |
| `d` | `d + 1` | flag `d` | `zero_hash(d + 1)` |

This prevents a prover from using an arbitrary value as the missing sibling. The padded sibling must select one of the precomputed canonical zero hashes, and the selected hash must match the parent depth relation.

### Bus Interactions

```text
Child proof public values
        |
        | PublicValuesBus
        v
DeferralAggPvsAir
  - receives child aggregation num_def_circuit_proofs
  - receives child aggregation merkle_depth
  - receives leaf output commits
  - receives input-or-merkle commits
        |
        | Poseidon2CompressBus
        v
Hashes leaf commits and parent aggregation commits
```

### Purpose

The hook layer binds the final aggregation PVS to the Merkle decommitment trace.

`DeferralHookPvsCols` adds:

| Column | Meaning |
| --- | --- |
| `num_merkle_leaves` | Witnessed value constrained to equal `2^def_pvs.merkle_depth`. |

The AIR constrains:

- `num_merkle_leaves = 2^merkle_depth` through `PowerCheckerBus`.
- `MAX_DEF_AGG_MERKLE_DEPTH - merkle_depth` is externally range-checked.
- The `MerkleRootBus` receive includes `num_rows_or_zero = 2 * num_merkle_leaves`.

The row-count relation matters because this Merkle tree trace has `2 * num_leaves` rows: one row for each leaf-pair hash, one row for each internal hash up to the root, and one final terminal row.

### Bus Interactions

```text
DeferralHookPvsAir
  |
  | PowerCheckerBus: (log = merkle_depth, exp = num_merkle_leaves)
  | RangeCheckerBus: MAX_DEF_AGG_MERKLE_DEPTH - merkle_depth
  |
  v
MerkleRootBus receive:
  merkle_root       = def_pvs.merkle_commit
  idx               = 0
  num_rows_or_zero  = 2 * num_merkle_leaves

MerkleTreeSubAir send:
  merkle_root       = computed root
  idx               = 0
  num_rows_or_zero  = actual trace row count, when enabled
```

This prevents arbitrary unchecked wrapping: increasing `merkle_depth` requires a matching power-check result and a matching Merkle decommitment trace size.

## Shared Merkle Tree Sub-AIR

`MerkleRootMessage` now includes `num_rows_or_zero`.

`MerkleTreeSubAir` adds `send_num_rows`:

| Caller | `send_num_rows` | Bus value |
| --- | ---: | --- |
| Deferral hook Merkle tree | `true` | actual `num_rows` |
| Root/user public value commits | `false` | `0` |
| verify-stark user public value commits | `false` | `0` |

This keeps existing Merkle root users compatible while allowing the deferral hook to enforce the tree size.

## Tracegen And Prover Plumbing

Deferral inner tracegen now computes and publishes `merkle_depth`:

- Wrapper proofs copy the child depth.
- Two-child proofs increment the child depth by one.
- Padded second rows store depth encoder flags in `merkle_commit`.
- Public `merkle_commit` for padded siblings still uses the actual canonical zero hash.

Deferral hook tracegen now:

- Computes `num_merkle_leaves = 1 << merkle_depth`.
- Emits external range-check input `MAX_DEF_AGG_MERKLE_DEPTH - merkle_depth`.
- Emits external power-check input `merkle_depth`.

`VerifierExternalData` now carries `power_check_inputs`, and CPU/GPU verifier tracegen feeds those inputs into the power checker generators.

## Primitive Encoder Change

`Encoder` methods now require `AirBuilder<F: Field, Var: Copy>` instead of `InteractionBuilder`.

This lets `DeferralAggPvsAir` use `Encoder` as a plain sub-AIR for canonical padding selectors without requiring interaction-builder-specific APIs.

## Tests

Continuation CUDA tests now check `merkle_depth` next to existing `num_def_circuit_proofs` checks:

- Deferral leaf prover tests assert `merkle_depth = log2(next_power_of_two(num_children))`.
- Deferral aggregation tests assert the same relation.

## Review Guide

Suggested review order:

1. Review `DeferralAggPvsAir` first, especially the depth relation and padded sibling selector.
2. Review hook constraints next, especially `PowerCheckerBus`, `RangeCheckerBus`, and `MerkleRootMessage.num_rows_or_zero`.
3. Review `MerkleTreeSubAir` to confirm only the hook sends row counts.
4. Review tracegen to confirm witnesses match the new AIR semantics.
5. Review tests for the new `merkle_depth` assertions.

### DIFF
--- a/crates/continuations/src/circuit/deferral/hook/verifier/air.rs
+++ b/crates/continuations/src/circuit/deferral/hook/verifier/air.rs
@@ -2,9 +2,14 @@ use std::{array::from_fn, borrow::Borrow};
 
 use openvm_circuit::arch::POSEIDON2_WIDTH;
 use openvm_circuit_primitives::{ColumnsAir, StructReflection, StructReflectionHelper, SubAir};
-use openvm_recursion_circuit::bus::{
-    CachedCommitBus, CachedCommitBusMessage, Poseidon2CompressBus, Poseidon2CompressMessage,
-    PreHashBus, PreHashMessage, PublicValuesBus, PublicValuesBusMessage,
+use openvm_recursion_circuit::{
+    bus::{
+        CachedCommitBus, CachedCommitBusMessage, Poseidon2CompressBus, Poseidon2CompressMessage,
+        PreHashBus, PreHashMessage, PublicValuesBus, PublicValuesBusMessage,
+    },
+    primitives::bus::{
+        PowerCheckerBus, PowerCheckerBusMessage, RangeCheckerBus, RangeCheckerBusMessage,
+    },
 };
 use openvm_recursion_circuit_derive::AlignedBorrow;
 use openvm_stark_backend::{
@@ -25,7 +30,7 @@ use crate::{
             hook::bus::{
                 DefCircuitCommitBus, DefCircuitCommitMessage, OnionResultBus, OnionResultMessage,
             },
-            DeferralAggregationPvs, DEF_AGG_PVS_AIR_ID,
+            DeferralAggregationPvs, DEF_AGG_PVS_AIR_ID, MAX_DEF_AGG_MERKLE_DEPTH,
         },
         root::NUM_DIGESTS_IN_VM_COMMIT,
         subair::{HashSliceCtx, HashSliceSubAir, MerkleRootBus, MerkleRootMessage},
@@ -40,6 +45,8 @@ use crate::{
 pub struct DeferralHookPvsCols<F> {
     pub verifier_pvs: VerifierBasePvs<F>,
     pub def_pvs: DeferralAggregationPvs<F>,
+
+    pub num_merkle_leaves: F,
 
     pub intermediate_vk_states: [[F; POSEIDON2_WIDTH]; NUM_DIGESTS_IN_VM_COMMIT - 1],
     pub def_circuit_commit: [F; DIGEST_SIZE],
@@ -60,6 +67,8 @@ pub struct DeferralHookPvsAir {
     pub pre_hash_bus: PreHashBus,
     pub poseidon2_compress_bus: Poseidon2CompressBus,
     pub hash_slice_subair: HashSliceSubAir,
+    pub pow_checker_bus: PowerCheckerBus,
+    pub range_checker_bus: RangeCheckerBus,
 
     pub def_circuit_commit_bus: DefCircuitCommitBus,
     pub merkle_root_bus: MerkleRootBus,
@@ -77,6 +86,8 @@ impl DeferralHookPvsAir {
         pre_hash_bus: PreHashBus,
         poseidon2_compress_bus: Poseidon2CompressBus,
         hash_slice_subair: HashSliceSubAir,
+        pow_checker_bus: PowerCheckerBus,
+        range_checker_bus: RangeCheckerBus,
         def_circuit_commit_bus: DefCircuitCommitBus,
         merkle_root_bus: MerkleRootBus,
         onion_res_bus: OnionResultBus,
@@ -89,6 +100,8 @@ impl DeferralHookPvsAir {
             pre_hash_bus,
             poseidon2_compress_bus,
             hash_slice_subair,
+            pow_checker_bus,
+            range_checker_bus,
             def_circuit_commit_bus,
             merkle_root_bus,
             onion_res_bus,
@@ -247,11 +260,30 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
          * and performing two onion hashes. Both steps are done in different AIRs, but we
          * must receive these commits to constrain consistency.
          */
+        self.pow_checker_bus.lookup_key(
+            builder,
+            PowerCheckerBusMessage {
+                log: local.def_pvs.merkle_depth,
+                exp: local.num_merkle_leaves,
+            },
+            AB::F::ONE,
+        );
+
+        self.range_checker_bus.lookup_key(
+            builder,
+            RangeCheckerBusMessage {
+                value: AB::Expr::from_usize(MAX_DEF_AGG_MERKLE_DEPTH) - local.def_pvs.merkle_depth,
+                max_bits: AB::Expr::from_usize(8),
+            },
+            AB::F::ONE,
+        );
+
         self.merkle_root_bus.receive(
             builder,
             MerkleRootMessage {
                 merkle_root: local.def_pvs.merkle_commit.map(Into::into),
                 idx: AB::Expr::ZERO,
+                num_rows_or_zero: local.num_merkle_leaves * AB::Expr::TWO,
             },
             AB::F::ONE,
         );
--- a/crates/continuations/src/circuit/deferral/hook/verifier/air.rs
+++ b/crates/continuations/src/circuit/deferral/hook/verifier/air.rs
@@ -258,7 +258,8 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
         /*
          * The input_onion and output_onion are computed by decommitting def_pvs.merkle_commit
          * and performing two onion hashes. Both steps are done in different AIRs, but we
-         * must receive these commits to constrain consistency.
+         * must receive these commits to constrain consistency. We range check merkle_depth to
+         * ensure it's less than or equal to MAX_DEF_AGG_MERKLE_DEPTH.
          */
         self.pow_checker_bus.lookup_key(
             builder,
--- a/crates/continuations/src/circuit/deferral/hook/verifier/air.rs
+++ b/crates/continuations/src/circuit/deferral/hook/verifier/air.rs
@@ -257,9 +257,10 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
 
         /*
          * The input_onion and output_onion are computed by decommitting def_pvs.merkle_commit
-         * and performing two onion hashes. Both steps are done in different AIRs, but we
-         * must receive these commits to constrain consistency. We range check merkle_depth to
-         * ensure it's less than or equal to MAX_DEF_AGG_MERKLE_DEPTH.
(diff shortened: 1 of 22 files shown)


crates/continuations/src/circuit/deferral/mod.rs
```diff
@@ -13,6 +13,8 @@ pub const DEF_AGG_PVS_AIR_ID: usize = 1;
 
 pub const DEF_HOOK_PVS_AIR_ID: usize = 0;
 
+const MAX_DEF_AGG_MERKLE_DEPTH: usize = 20;
```
shuklaayush, line 17: is there any rationale for this value?

crates/continuations/src/circuit/deferral/hook/verifier/air.rs
```diff
@@ -247,11 +260,30 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
...
+
+        self.range_checker_bus.lookup_key(
+            builder,
+            RangeCheckerBusMessage {
+                value: AB::Expr::from_usize(MAX_DEF_AGG_MERKLE_DEPTH) - local.def_pvs.merkle_depth,
+                max_bits: AB::Expr::from_usize(8),
+            },
+            AB::F::ONE,
+        );
+
```
shuklaayush, line 282: a comment here would be useful that this is just checking that the subtraction doesn't wrap since the 8 seems arbitrary

crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
```diff
@@ -47,6 +54,37 @@ pub struct DeferralAggPvsAir {
     pub poseidon2_bus: Poseidon2CompressBus,
     pub input_or_merkle_commit_bus: InputOrMerkleCommitBus,
     pub def_pvs_consistency_bus: DefPvsConsistencyBus,
+
+    pub encoder: Encoder,
+    pub zero_hashes: [CommitBytes; MAX_DEF_AGG_MERKLE_DEPTH + 1],
```
shuklaayush, line 58: lots of columns

crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
```diff
@@ -47,6 +54,37 @@ pub struct DeferralAggPvsAir {
...
+        Self {
+            public_values_bus,
+            poseidon2_bus,
+            input_or_merkle_commit_bus,
+            def_pvs_consistency_bus,
+            encoder,
+            zero_hashes: from_fn(|_| {
+                zero_hash = poseidon2_compress_with_capacity(zero_hash, zero_hash).0;
+                zero_hash.into()
+            }),
```
shuklaayush: nit: would be easier to read if this is a separate funciton

crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
```diff
@@ -164,11 +202,11 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
          * columns, we repurpose the unused local.child_pvs.input_commit[0].
          */
         let local_num_def_circuit_proofs = local.child_pvs.input_commit[0];
+        let local_merkle_depth = local.child_pvs.input_commit[1];
```
shuklaayush, line 201: add a clarifying comment that the column is being repurposed like for the above

crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
```diff
@@ -199,20 +249,61 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
...
+         * If there are two rows, then the second (next) is either present or not. If present
+         * the parent merkle_hash should be the compression of both children's merkle_commit.
+         * If not the second merkle_commit contains encoder flags for the child merkle depth,
+         * which is used to select the correct zero hash depth.
+         */
+        let flags = next
+            .merkle_commit
+            .into_iter()
+            .take(self.encoder.width())
+            .collect_vec();
```
shuklaayush, line 277: let's document these column reuse tricks so it's much easier to follow

crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
```diff
@@ -199,20 +249,61 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
...
+            builder
+                .when(not(next.is_present))
+                .when(is_current.clone())
+                .assert_eq(merkle_depth, AB::Expr::from_usize(i + 1));
+
+            let current_zero_hash: [AB::F; DIGEST_SIZE] = self.zero_hashes[i].into();
+            for j in 0..DIGEST_SIZE {
+                zero_hash[j] += is_current.clone() * current_zero_hash[j];
+            }
+        }
```
shuklaayush, line 291: we can probably move the merkle depth constraint outside the loop
```rust
  // 1. select the canonical zero-hash for the encoded depth
  let mut zero_hash = [AB::Expr::ZERO; DIGEST_SIZE];
  for (i, zh) in self.zero_hashes.iter().enumerate() {
      let is_current = self.encoder.get_flag_expr::<AB>(i, &flags);
      let zh: [AB::F; DIGEST_SIZE] = (*zh).into();
      for j in 0..DIGEST_SIZE {
          zero_hash[j] += is_current.clone() * zh[j];
      }
  }

  // 2. the encoded depth must match the parent depth — one assert, not 21
  let encoded_parent_depth = self.encoder.flag_with_val::<AB>(
      &flags,
      &(0..=MAX_DEF_AGG_MERKLE_DEPTH).map(|i| (i, i + 1)).collect_vec(),
  );
  builder
      .when(not(next.is_present))
      .assert_eq(merkle_depth, encoded_parent_depth);
```

shuklaayush (commented):
some comments to improve clarity

crates/continuations/src/circuit/deferral/mod.rs
stephenh-axiom-xyz: Will add this as a comment, but this is the default max trace height of the deferral hook circuit

crates/continuations/src/circuit/deferral/hook/verifier/air.rs
```diff
@@ -258,7 +258,8 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
         /*
          * The input_onion and output_onion are computed by decommitting def_pvs.merkle_commit
          * and performing two onion hashes. Both steps are done in different AIRs, but we
-         * must receive these commits to constrain consistency.
+         * must receive these commits to constrain consistency. We range check merkle_depth to
+         * ensure it's less than or equal to MAX_DEF_AGG_MERKLE_DEPTH.
```
shuklaayush: i feel like it's still better to clarify that this range check works for constraining the less that or equal relation because `merkle_depth` is constrained to `[0, 31]` (or whatever) because of the earlier power checker lookup

shuklaayush (approved):

PR #2817, opened 2026-05-26
fix: ensure verify crate and root continuations layer match

stephenh-axiom-xyz:
Resolves INT-7919.

## Overview

This PR aligns the Rust STARK verifier with the root-layer continuation AIR semantics for deferral-enabled proofs.

The main security fixes are:

- The native verifier now requires and verifies deferral Merkle proofs whenever deferrals are enabled, including the `deferral_flag == 0` case.
- Native deferral Merkle proof verification now checks that the deferral address space is unchanged outside the claimed aggregated subtree.
- Native verification and the root deferral-path AIR now reject deferral depths larger than the deferral address-space height.
- Root verifier AIRs now explicitly constrain `def_hook_commit` to be unset when `deferral_flag != 2`, matching the native verifier.
- Verification baselines now record the expected number of raw user public values, and the native verifier rejects proofs whose user public value proof length differs.

The branch also expands the SDK CUDA workflow trigger paths so changes in continuations and recursion run that workflow.

### Baseline Schema

`VerificationBaseline` now includes `num_user_pvs`, the expected number of raw user public values for the fixed VM configuration and executable. The SDK fills this from `vm.config().num_public_values` when generating a baseline, and `VerificationBaselineJson` now serializes/deserializes the field alongside `app_exe_commit`, `memory_dimensions`, and verifier key commits.

Affected files:

- `crates/verify/src/vk.rs`
- `crates/sdk/src/prover/app.rs`
- `crates/sdk/src/types.rs`

### Native Verifier Check

`verify_vm_stark_proof_pvs` now compares `proof.user_pvs_proof.public_values.len()` with `vk.baseline.num_user_pvs`.

| Case | `user_pvs_proof.public_values.len()` | `baseline.num_user_pvs` | Result |
| --- | ---: | ---: | --- |
| Matching baseline | `N` | `N` | Continue with normal verification. |
| Too short | `< N` | `N` | `UserPvsLengthMismatch`. |
| Too long | `> N` | `N` | `UserPvsLengthMismatch`. |

This supplements the existing Merkle proof verification: the Merkle proof checks commitment consistency, while the baseline length check ensures the proof exposes the exact number of user public values expected by the committed VM configuration.

### Native Deferral Verification

`verify_vm_stark_proof_pvs` now extracts and verifies `deferral_merkle_proofs` outside the per-flag branch whenever `expected_def_hook_commit` is set.

Before this change:

- `deferral_flag == 2` required deferral Merkle proofs.
- `deferral_flag == 0` only checked that deferral public values were unset.

After this change:

| `expected_def_hook_commit` | `deferral_flag` | Required checks |
| --- | ---: | --- |
| `Some` | `0` | `def_hook_commit`, accumulators, and depth are unset; deferral Merkle proofs verify the unset accumulator path and unchanged deferral address space. |
| `Some` | `2` | `def_hook_commit` matches expected value; deferral Merkle proofs verify accumulator paths and unchanged deferral address space outside the aggregate. |
| `Some` | other | rejected. |
| `None` | any deferral PVS/proofs present | rejected. |

This closes the native verifier discrepancy where a proof could set `deferral_flag == 0`, omit deferral Merkle proofs, and skip the root-layer unchanged-address-space check that the AIR requires.

### `DeferralMerkleProofs::verify`

`DeferralMerkleProofs::verify` now performs two additional checks:

- `depth <= memory_dimensions.address_height`
- `initial_merkle_proof[i] == final_merkle_proof[i]` for `i in 0..memory_dimensions.address_height`

The sibling equality check matches `DeferralAccMerklePathsAir`'s invariant that the part of `DEFERRAL_AS` outside the aggregated deferral subtree is left untouched for the duration of execution.

New error variants were added for these checks:

- `DeferralDepthTooLarge { depth, address_height }`
- `DeferralMerkleProofSiblingMismatch`

The depth bound is deliberately the address-space height, not the full memory Merkle path height. Deferral aggregation depth describes a subtree inside `DEFERRAL_AS`; allowing it to extend past the address-space boundary would skip levels that the AIR expects to compare as unchanged.

### Verification Baseline Docs

The `expected_def_hook_commit` comment now describes the actual semantics:

- `Some` means deferral public values are enabled.
- The proof may have `deferral_flag == 0` or `deferral_flag == 2`.
- Valid deferral Merkle proofs are required in both cases.
- `None` means no deferral public values may be present.

### DeferralAccMerklePathsAir

`crates/continuations/src/circuit/root/def_paths/air.rs` now constrains received deferral depths to be less than or equal to the deferral address-space height.

No columns were added or removed, and the `DeferralAccPathMessage` bus shape is unchanged. The change factors the receive multiplicity into `is_next_start_depth` and adds:

- while inside `DEFERRAL_AS`, the existing path equality constraints still apply;
- after leaving `DEFERRAL_AS`, no new `DeferralAccPathMessage` may be received.

Example row behavior:

| Region | `is_within_deferral_as` | `is_next_start_depth` | Meaning |
| --- | ---: | ---: | --- |
| Before/inside `DEFERRAL_AS` | `1` | `0` or `1` | A deferral subtree may start while still inside the address-space path. |
| First level after `DEFERRAL_AS` | `0` | `0` | A new deferral subtree start is rejected past `address_height`. |
| Later path levels | `0` | `0` | Continue normal path verification to the memory root. |

The trace generator now uses the same bound and asserts `depth in 1..=address_height` when the accumulator is set.

### AIRs Changed

- `crates/continuations/src/circuit/root/verifier/air.rs`
- `guest-libs/verify-stark/circuit/src/verifier/air.rs`

Both AIRs now constrain `def_hook_commit` to be zero in the no-deferral branch.

No columns were added or removed.

### Constraint Summary

The root verifier AIR already constrained:

- `deferral_flag` is either `0` or `2`.
- If `deferral_flag == 2`, `def_hook_commit == expected_def_hook_commit`.
- If `deferral_flag != 2`, deferral depth and accumulator hashes are unset.
- A `DeferralAccPathMessage` is sent in both cases, so `DeferralAccMerklePathsAir` checks the relevant memory paths.

This PR adds the missing no-deferral constraint:

- If `deferral_flag != 2`, `def_hook_commit == 0`.

Example accepted/rejected public value shapes:

| Case | `deferral_flag` | `def_hook_commit` | `depth` | `initial_acc_hash` / `final_acc_hash` |
| --- | ---: | --- | ---: | --- |
| No deferral | `0` | zero | `0` | zero |
| Deferral present | `2` | expected deferral hook commit | nonzero aggregate depth | committed accumulator roots |
| Invalid no-deferral | `0` | nonzero | `0` | zero |
| Invalid flag | `1` or other | any | any | any |

### Existing Bus Interaction

No bus interface changes were made, but the unchanged root-layer bus flow is important to this fix:

```text
RootVerifierPvsAir / DeferredVerifyPvsAir
  |
  | sends DeferralAccPathMessage
  |   initial_acc_hash
  |   final_acc_hash
  |   depth
  |   is_unset = 1 when deferral_flag == 0
  |
  v
DeferralAccMerklePathsAir
  |
  | receives DeferralMerkleRootsMessage
  |   initial_root
  |   final_root
  |
  v
Checks initial/final accumulator paths and unchanged DEFERRAL_AS prefix
```

The native verifier changes mirror this AIR behavior directly.

## CI

`.github/workflows/sdk.cuda.yml` now runs on PR changes under:

- `crates/continuations/**`
- `crates/recursion/**`

This ensures SDK CUDA coverage is triggered for continuation and recursion changes that affect SDK proving flows.

## Review Guide

Suggested review order:

1. Review `crates/verify/src/lib.rs` and the baseline JSON plumbing to confirm the user public value length check uses the VM-configured `num_public_values`.
2. Review `crates/verify/src/lib.rs` to confirm deferral Merkle proofs are required for both `deferral_flag == 0` and `deferral_flag == 2`.
3. Review `crates/verify/src/deferral.rs` and `crates/continuations/src/circuit/root/def_paths/air.rs` to confirm the depth bound is `address_height` and the `DEFERRAL_AS` sibling equality matches the root AIR semantics.
4. Review both root verifier AIR copies to confirm `def_hook_commit` is unset in the no-deferral branch.
5. Review the workflow path additions.

### DIFF
--- a/crates/verify/src/deferral.rs
+++ b/crates/verify/src/deferral.rs
@@ -64,10 +64,10 @@ impl DeferralMerkleProofs<F> {
             });
         }
 
-        if depth > overall_height {
+        if depth > memory_dimensions.address_height {
             return Err(VerifyStarkError::DeferralDepthTooLarge {
                 depth,
-                overall_height,
+                address_height: memory_dimensions.address_height,
             });
         }
 
--- a/crates/verify/src/deferral.rs
+++ b/crates/verify/src/deferral.rs
@@ -133,7 +133,7 @@ fn merkle_path_root(
         if i < depth {
             continue;
         }
-        let is_right = depth == 0 && i == 0 || (idx_prefix >> i) & 1 == 1;
+        let is_right = (depth == 0 && i == 0) || (idx_prefix >> i) & 1 == 1;
         node = if is_right {
             poseidon2_compress_with_capacity(*sibling, node).0
         } else {
--- a/crates/verify/src/deferral.rs
+++ b/crates/verify/src/deferral.rs
@@ -64,6 +64,23 @@ impl DeferralMerkleProofs<F> {
             });
         }
 
+        if depth > overall_height {
+            return Err(VerifyStarkError::DeferralDepthTooLarge {
+                depth,
+                overall_height,
+            });
+        }
+
+        for i in 0..memory_dimensions.address_height {
+            if self.initial_merkle_proof[i] != self.final_merkle_proof[i] {
+                return Err(VerifyStarkError::DeferralMerkleProofSiblingMismatch {
+                    depth: i,
+                    initial: self.initial_merkle_proof[i],
+                    final_: self.final_merkle_proof[i],
+                });
+            }
+        }
+
         let is_unset = depth == 0;
         let skip_depth = if is_unset { 0 } else { depth };
         let idx_prefix =
--- a/crates/verify/src/deferral.rs
+++ b/crates/verify/src/deferral.rs
@@ -82,7 +82,6 @@ impl DeferralMerkleProofs<F> {
         }
 
         let is_unset = depth == 0;
-        let skip_depth = if is_unset { 0 } else { depth };
         let idx_prefix =
             usize::try_from(memory_dimensions.label_to_index((DEFERRAL_AS, 0))).unwrap();
 
@@ -98,12 +97,8 @@ impl DeferralMerkleProofs<F> {
             final_acc_hash
         };
 
-        let computed_initial_root = merkle_path_root(
-            initial_leaf,
-            &self.initial_merkle_proof,
-            idx_prefix,
-            skip_depth,
-        );
+        let computed_initial_root =
+            merkle_path_root(initial_leaf, &self.initial_merkle_proof, idx_prefix, depth);
         if computed_initial_root != initial_root {
             return Err(VerifyStarkError::DeferralInitialRootMismatch {
                 expected: initial_root,
@@ -112,7 +107,7 @@ impl DeferralMerkleProofs<F> {
         }
 
         let computed_final_root =
-            merkle_path_root(final_leaf, &self.final_merkle_proof, idx_prefix, skip_depth);
+            merkle_path_root(final_leaf, &self.final_merkle_proof, idx_prefix, depth);
         if computed_final_root != final_root {
             return Err(VerifyStarkError::DeferralFinalRootMismatch {
                 expected: final_root,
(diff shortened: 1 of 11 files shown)


crates/verify/src/deferral.rs
```diff
@@ -107,20 +119,21 @@ impl DeferralMerkleProofs<F> {
...
+    depth: usize,
 ) -> [F; DIGEST_SIZE] {
     let mut node = leaf;
     for (i, sibling) in proof.iter().enumerate() {
-        if i < skip_depth {
+        if i < depth {
             continue;
         }
-        let is_right = (idx_prefix >> i) & 1 == 1;
+        let is_right = depth == 0 && i == 0 || (idx_prefix >> i) & 1 == 1;
```
shuklaayush: would be easier to read with parenthesis

crates/verify/src/deferral.rs
```diff
@@ -64,8 +64,24 @@ impl DeferralMerkleProofs<F> {
...
+            return Err(VerifyStarkError::DeferralDepthTooLarge {
+                depth,
+                address_height: memory_dimensions.address_height,
+            });
+        }
+
+        for i in 0..memory_dimensions.address_height {
+            if self.initial_merkle_proof[i] != self.final_merkle_proof[i] {
+                return Err(VerifyStarkError::DeferralMerkleProofSiblingMismatch {
+                    depth: i,
```
shuklaayush, line 77: `depth` is a bit ambiguous here since it can refer to the deferral tree depth as well

shuklaayush (approved):
lgtm. maybe we should also add some negative tests for the cases that this pr is fixing

crates/verify/src/deferral.rs
stephenh-axiom-xyz: That is not really a concept in the `verify` crate

stephenh-axiom-xyz:
@shuklaayush it's a little tricky to test for specific verifier checks at the moment, am going to defer testing to avoid making larger refactors

PR #2819, opened 2026-05-27
chore: update app max constraint degree to 3 + hook params

stephenh-axiom-xyz:
Resolves INT-7419

### DIFF
--- a/crates/continuations/src/tests/mod.rs
+++ b/crates/continuations/src/tests/mod.rs
@@ -96,6 +96,12 @@ pub(in crate::tests) fn root_system_params() -> SystemParams {
     root_params_with_100_bits_security()
 }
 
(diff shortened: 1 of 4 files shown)


crates/continuations/src/tests/mod.rs
```diff
@@ -96,6 +96,12 @@ pub(in crate::tests) fn root_system_params() -> SystemParams {
     root_params_with_100_bits_security()
 }
 
+#[cfg(all(feature = "cuda", feature = "root-prover"))]
+pub(in crate::tests) fn hook_system_params() -> SystemParams {
+    use openvm_stark_sdk::config::hook_params_with_100_bits_security;
```
shuklaayush, line 101: i would move this to top for consistency

shuklaayush (approved):

PR #2820, opened 2026-05-27
perf(v2.1-rvr): isolate metered segment-check to cold block helper

shuklaayush:
A perf pass on RVR metered execution. The main change isolates the rare segment-check callback into a cold per-block helper so the hot block stays frameless.

## Cold per-block segment-check helper

Hot metered RVR blocks were paying a stack-frame cost because the rare `on_check` callback could fire from the same C function. Generated asm showed every metered block — even single-instruction ones — getting an entry prologue and stack spills to preserve guest-register parameters across the possible callback.

Hot blocks now only test `check_counter < block_insn_count` inline. On underflow they musttail-jump to a cold per-block `block_0xPC_checkpoint(...)` helper that runs the segment-check callback, suspends/exits if needed, or musttails back to the hot block with a refreshed counter. Same semantics; the hot path is frameless again.

## Cleanup landing in the same patch

- **`uses_page_tracking()` IR predicate.** Only blocks that can touch memory emit AS_MEMORY page-tracking locals. ~59% of blocks in the reth benchmark didn't need them. Extension emitters default to `true`; `HintNonQrInstr` opts out, plain host-only phantoms (`HintInput`, `HintRandom`, `PrintStr`) don't trigger page locals.
- **Metered block ABI hoist.** `check_counter` (`_cc`) and `trace_heights` (`_th`) are passed as block parameters, removing `state->tracer` loads from every block. Metered mode uses 8 hot guest registers instead of 10 to fit the new parameters.
- **`CompileOptions::keep_artifacts`.** Retains the generated RVR C tempdir on success and logs the path. Useful for codegen / asm audits.
- Per-width fast traced memory helpers and clang-format / formatting cleanups across the FFI C/Rust crates.

results in a modest ~100ms (out of 1.7s) improvement in metered execution of [reth benchmark](https://github.com/axiom-crypto/openvm-eth/actions/runs/26514539244)

on my laptop, the improvement is much more significant (~20%)


crates/rvr/rvr-openvm-ir/src/ext.rs
```diff
@@ -44,6 +69,15 @@ pub trait ExtInstr: std::fmt::Debug + Send + Sync {
...
     }
 
+    /// Whether this instruction may touch guest memory or otherwise need
+    /// metered page tracking.
+    ///
+    /// The conservative default is `true` because most opaque extension FFIs
+    /// read or write guest memory through `state`.
+    fn uses_page_tracking(&self) -> bool {
+        true
+    }
```
gdmlcjs: Could we rename this to something that doesn't need the tracing context, like `accesses_memory`?

gdmlcjs (approved):
Big improvement to perf. Looks good to me

crates/rvr/rvr-openvm-ir/src/ext.rs
shuklaayush: yeah, good call

PR #2821, opened 2026-05-27
refactor: import openvm constants instead of redefining

mansur20478:
In `rvr`, some constants are redefined or set as variables. Resolved some of the dependency issues (e.g. circular dependency) to import the constants instead.

Related constants:
1) `WORD_SIZE`: imported from `openvm_platform::WORD_SIZE`

2) `AS_MEMORY`: imported from `openvm_instructions::riscv::RV32_MEMORY_AS`

3) `AS_REGISTER`: imported from `openvm_instructions::riscv::RV32_REGISTER_AS`

4) `AS_PUBLIC_VALUES`: imported from `openvm_instructions::PUBLIC_VALUES_AS` (moved to `openvm_instructions` from `openvm-circuit`. Is it right choice????)

5) `DEFERRAL_AS`: imported from `openvm_instructions::DEFERRAL_AS`

6) `MAX_BLOCK_INSNS` (`rvr-openvm-lift/src/cfg.rs`): was `let`, now `const`

The following ones are kept redefined:
1) `CHUNK`, `DEFERRAL_DIGEST_SIZE`: logically are from `openvm-stark-sdk`. `openvm-circuit` and `openvm-recursion-circuit` already redefine CHUNK. (can not import from them due to circular dependency). 

2) `DEFAULT_PAGE_BITS`, `DEFAULT_SEGMENT_CHECK_INSNS`: logically are from `openvm-circuit::arch::execution_mode::metered::{ctx, segment_ctx}`. These are host-side metered-execution defaults. (can not import from them due to circular dependency)

closes INT-7571

### DIFF
--- a/crates/rvr/rvr-openvm-lift/src/cfg.rs
+++ b/crates/rvr/rvr-openvm-lift/src/cfg.rs
@@ -1061,7 +1061,7 @@ fn build_block_list(
 ) -> Vec<Block> {
     // Max block size; used to flush periodically so the segmentation check in
     // metered mode (which fires at block boundaries) stays granular enough.
-    let max_block_insns = rvr_openvm_ext_ffi_common::DEFAULT_SEGMENT_CHECK_INSNS;
+    const MAX_BLOCK_INSNS: u32 = rvr_openvm_ext_ffi_common::DEFAULT_SEGMENT_CHECK_INSNS;
     let mut blocks: Vec<Block> = Vec::new();
 
     // Accumulate body instructions for the current block.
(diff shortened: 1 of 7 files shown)


crates/rvr/rvr-openvm-lift/src/cfg.rs
```diff
@@ -1061,7 +1061,7 @@ fn build_block_list(
 ) -> Vec<Block> {
     // Max block size; used to flush periodically so the segmentation check in
     // metered mode (which fires at block boundaries) stays granular enough.
-    let max_block_insns = rvr_openvm_ext_ffi_common::DEFAULT_SEGMENT_CHECK_INSNS;
+    const MAX_BLOCK_INSNS: u32 = rvr_openvm_ext_ffi_common::DEFAULT_SEGMENT_CHECK_INSNS;
```
shuklaayush, line 1064: seems unnecessary to redeclare

shuklaayush (approved):
lgtm. i think something that we should do as a follow up is clean up some of this `DEFAULT_*` terminology
before rvr, we used to have the flexibility of changing some parameters at runtime but now a lot of those need to be baked in at compile time. therefore it's better if we just rename things from say `DEFAULT_SEGMENT_CHECK_INSNS` to `SEGMENT_CHECK_INSNS`. same for `DEFAULT_PAGE_BITS`

PR #2822, opened 2026-05-27
feat: constrain AIRs are sorted by air_idx after trace height in recursion

stephenh-axiom-xyz:
Towards INT-7977.

## Overview

This PR tightens recursion proof-shape ordering. `ProofShapeAir` already required rows to be sorted by non-increasing trace height; it now also requires deterministic AIR-index ordering among rows with equal effective height.

The enforced order is lexicographic by:

```text
(-rank, air_idx)
```

where:

```text
rank = log_height + is_present
```

This gives absent AIRs rank `0`, present height-1 AIRs rank `1`, present height-2 AIRs rank `2`, etc. Equal-height/equal-rank rows must appear with strictly increasing `air_idx`.

Base used for this summary: `origin/develop-v2.0.0-rc.2`.

## Recursion: `ProofShapeAir`

Files:

- `crates/recursion/src/proof_shape/proof_shape/air.rs`
- `crates/recursion/src/proof_shape/proof_shape/trace.rs`
- `crates/recursion/cuda/src/proof_shape/air.cu`
- `crates/recursion/src/proof_shape/mod.rs`

### New Column

`ProofShapeCols` adds:

```rust
pub is_height_equal_to_next: F
```

This is a boolean selector for valid non-summary transitions:

- `1`: current row and next row have equal effective height; enforce `air_idx < next_air_idx`.
- `0`: current row must have strictly larger effective height rank than the next row.

Example row patterns:

| Row kind | `is_present` | `log_height` | `height` | `rank` | next row | `is_height_equal_to_next` |
| --- | ---: | ---: | ---: | ---: | --- | ---: |
| present height 8 | 1 | 3 | 8 | 4 | present height 8, larger `air_idx` | 1 |
| present height 8 | 1 | 3 | 8 | 4 | present height 4 | 0 |
| present height 1 | 1 | 0 | 1 | 1 | absent | 0 |
| absent | 0 | 0 | 0 | 0 | absent, larger `air_idx` | 1 |
| last row before summary | varies | varies | varies | varies | summary row | 0 |

### Ordering Constraints

The AIR now borrows the next row's variable columns to decode `next_air_idx`.

For each valid transition where `next` is not the summary row:

1. `is_height_equal_to_next` is boolean.
2. If `is_height_equal_to_next = 1`:
   - constrain `local.height == next.height`;
   - range-check `next_air_idx - air_idx - 1` as 8 bits.
3. If `is_height_equal_to_next = 0`:
   - range-check `local_rank - next_rank - 1` as 5 bits.

This preserves the existing height sort while ruling out ambiguous ordering among equal-height rows.

### Range-Check Bus Flow

The change adds/changes `RangeCheckerBus` lookups from `ProofShapeAir`:

```text
ProofShapeAir
  |-- equal-height transition:
  |     lookup RangeCheckerBus(value = next_air_idx - air_idx - 1, max_bits = 8)
  |        balanced by ProofShapeModule RangeCheckerAir<8>
  |
  |-- strict-height transition:
        lookup RangeCheckerBus(value = local_rank - next_rank - 1, max_bits = 5)
           balanced by global PowerCheckerAir<2, 32> range side
```

`ProofShapeModule::airs` now asserts `per_air.len() <= 2^8`, matching the 8-bit AIR-index gap check.

### Trace Generation

CPU and CUDA tracegen now populate `is_height_equal_to_next` and add the corresponding range-check counts:

- equal effective height: add the AIR-index gap count;
- strict effective-height drop: add the rank-drop count;
- summary transition: set the selector to `0`.

The CUDA proof-shape column layout is updated to include the new column, keeping GPU tracegen aligned with the Rust AIR layout.

## CI

File:

- `.github/workflows/sdk.cuda.yml`

The CUDA SDK workflow trigger now includes:

```text
crates/continuations/**
crates/recursion/**
```

This makes recursion/continuations changes run the CUDA SDK workflow instead of only SDK-path changes.

## Suggested Review Order

1. Review `ProofShapeAir` ordering semantics and the `rank` definition.
2. Check CPU tracegen counts against the two AIR branches.
3. Check CUDA struct layout and CUDA tracegen counts mirror CPU tracegen.
4. Confirm the `per_air.len() <= 2^8` assumption is acceptable for recursion proof shapes.
5. Review the workflow path trigger addition.

### DIFF
--- a/crates/recursion/src/proof_shape/mod.rs
+++ b/crates/recursion/src/proof_shape/mod.rs
@@ -88,6 +88,11 @@ impl ProofShapeModule {
         bus_inventory: BusInventory,
         continuations_enabled: bool,
     ) -> Self {
+        assert!(
+            mvk.per_air.len() <= 1 << 8,
+            "recursion circuit only supports child verifying keys with at most 256 AIRs"
+        );
+
         let idx_encoder = Arc::new(Encoder::new(mvk.per_air.len(), 2, true));
 
         let (min_cached_idx, min_cached) = mvk
@@ -230,7 +235,6 @@ impl AirModule for ProofShapeModule {
     }
 
     fn airs<SC: StarkProtocolConfig<F = F>>(&self) -> Vec<AirRef<SC>> {
-        assert!(self.per_air.len() <= 1 << 8);
         let proof_shape_air = ProofShapeAir::<4, 8> {
             per_air: self.per_air.clone(),
             l_skip: self.l_skip,
--- a/crates/recursion/src/proof_shape/mod.rs
+++ b/crates/recursion/src/proof_shape/mod.rs
@@ -230,6 +230,7 @@ impl AirModule for ProofShapeModule {
     }
 
     fn airs<SC: StarkProtocolConfig<F = F>>(&self) -> Vec<AirRef<SC>> {
+        assert!(self.per_air.len() <= 1 << 8);
         let proof_shape_air = ProofShapeAir::<4, 8> {
             per_air: self.per_air.clone(),
             l_skip: self.l_skip,
--- a/crates/recursion/src/proof_shape/proof_shape/air.rs
+++ b/crates/recursion/src/proof_shape/proof_shape/air.rs
@@ -47,9 +47,8 @@ pub struct ProofShapeCols<F, const NUM_LIMBS: usize> {
     pub is_last: F,
 
     pub sorted_idx: F,
-    /// Represents log2 trace height when `is_present`.
-    ///
-    /// Has a special use on summary row (when `is_last`).
+
+    /// Log2 trace height when `is_present`; has a special use on summary row (when `is_last`).
     pub log_height: F,
     /// When `is_present`, constrained to equal `log_height - l_skip < 0 ? 1 : 0`.
     pub n_sign_bit: F,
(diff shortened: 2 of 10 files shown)


crates/recursion/src/proof_shape/proof_shape/air.rs
```diff
@@ -385,14 +396,42 @@ where
...
+        builder.assert_bool(local.is_height_equal_to_next);
+        builder
+            .when(local.is_height_equal_to_next)
+            .assert_one(is_valid_transition.clone());
+        builder
+            .when(local.is_height_equal_to_next)
+            .assert_eq(local.height, next.height);
+
+        // If is_height_equal_to_next == 0, ensure local.height > next.height by constraining
+        // local_rank > next_rank. By definition rank == 0 for non-present AIRs and rank > 0
```
Golovanov399, line 412: Is it enforced that `rank == 0` for non-present AIRs?

stephenh-axiom-xyz: It is enforced that `log_height` and `is_present` are 0

Golovanov399 (approved):

crates/recursion/src/proof_shape/mod.rs
```diff
@@ -230,6 +230,7 @@ impl AirModule for ProofShapeModule {
     }
 
     fn airs<SC: StarkProtocolConfig<F = F>>(&self) -> Vec<AirRef<SC>> {
+        assert!(self.per_air.len() <= 1 << 8);
```
zlangley: We should add to some verifier circuit documentation somewhere that this constraint exists.

PR #2823, opened 2026-05-27
refactor: derive MAX_MEM_PAGES_PER_INSN from a PAGE_BITS value

mansur20478:
`MAX_MEM_PAGES_PER_INSN ` is a worst-case number of pages a single instruction can touch. The worst-case unique pages per instruction (`HINT_BUFFER`) is `MAX_HINT_BUFFER_WORDS * WORD_SIZE` bytes divided by page size. One page covers `CHUNK * 2^PAGE_BITS` bytes.

So the formula is:
`MAX_MEM_PAGES_PER_INSN = div_ceil(MAX_HINT_BUFFER_WORDS * WORD_SIZE, CHUNK * 2^PAGE_BITS) + 1`

`+1` misalignment.

closes INT-7462


crates/rvr/rvr-openvm/src/constants.rs
```diff
@@ -5,20 +5,36 @@
...
 
+/// Worst-case AS_MEMORY pages a single instruction can touch.
+///
+/// Bound is set by `HINT_BUFFER`, which writes up to
+/// `MAX_HINT_BUFFER_WORDS * WORD_SIZE` contiguous bytes. One AS_MEMORY page
+/// covers `CHUNK * 2^PAGE_BITS` bytes. The `+1` covers worst-case
+/// misalignment of the range across page boundaries; the second `+1` is
+/// safety slack to keep parity with the prior hardcoded value under default
+/// `PAGE_BITS` in case an executor touches more pages than this is
+/// accounted for.
```
shuklaayush: > the second `+1` is safety slack to keep parity with the prior hardcoded value under default `PAGE_BITS` in case an executor touches more pages than this is accounted for

can you explain what this means?

mansur20478: Originally, `MAX_MEM_PAGES_PER_INSN` was equal to `10`.

However, when I write the formula: `MAX_MEM_PAGES_PER_INSN = div_ceil(MAX_HINT_BUFFER_WORDS * WORD_SIZE, CHUNK * 2^PAGE_BITS) + 1`, it would be `div_ceil(1023 * 4, 8 * 2^6) + 1 => 9`, which is less than initial `10`. So, I thought it might be better to add another `+1` in case the formula didn't account for something. Should I remove this `+1`?

shuklaayush: maybe @gdmlcjs can answer. i feel we should remove it if it's unnecessary and/or we can't justify it.
also, are we sure that hint_buffer is the worst offender?

mansur20478: yes, hint_buffer is likely to be the worst offender. there are runner ups, such as ecc instructions that can take 6 pages in the worst case, but not close to hint_buffer?

i asked @gdmlcjs about the original value. the `+1` seems to be unnecessary, I will remove it.

gdmlcjs (approved):

shuklaayush (approved):

PR #2825, opened 2026-05-28
chore: move keygen outside metrics block

Aalanli:
As in the title.
Resolves: INT-8001


stephenh-axiom-xyz:
Keygen is already not included in summary metrics, closing

PR #2826, opened 2026-05-28
feat: persist compiled artifacts in rvr 

mansur20478:
Add save and load compiled artifacts feature in `rvr` mode. The feature consist of having the ability to save compilation artifacts on disk and load them into the sdk to execute (part 1). Reusing of the persisted artifacts whenever possible instead of recompiling based on some metadata (part 2) will be done in separate PR.

This PR is related to the part 1. The following methods were added:
1) `Sdk::load_compiled_pure`, `Sdk::load_compiled_metered`, `Sdk::load_compiled_metered_cost` and related methods for loading pure, metered and metered cost `.so` files

2) `RvrPureInstance::save`, `RvrMeteredInstance::save`, `RvrMeteredCostInstance::save` and related methods for saving `.so` file

towards INT-7843

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -88,6 +88,14 @@ jobs:
           export RUST_MIN_STACK=8388608
           CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS=true cargo nextest run --release --features parallel,root-prover,evm-verify
 
+      - name: Run openvm-sdk crate tests (RVR)
+        if: ${{ matrix.mode == 'default' }}
+        working-directory: crates/sdk
+        run: |
+          export RUST_BACKTRACE=1
+          export RUST_MIN_STACK=8388608
+          CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS=true cargo nextest run --release --features parallel,root-prover,evm-verify,rvr
+
       - name: Run openvm-sdk crate tests (AOT)
         if: ${{ matrix.mode == 'aot' }}
         working-directory: crates/sdk
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -25,7 +25,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        mode: ["default", "ignored", "aot"]
+        mode: ["default", "ignored", "aot", "rvr"]
 
     runs-on:
       - runs-on=${{ github.run_id }}-sdk-${{ github.run_attempt }}-${{ strategy.job-index }}/family=m7a.24xlarge/image=ubuntu24-full-x64/disk=large/extras=s3-cache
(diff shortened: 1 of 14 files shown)


crates/vm/src/arch/rvr/metered_cost.rs
```diff
@@ -74,6 +74,13 @@ impl TracerPayload for PureTracerData {
 pub type PureTracer = TracerPtr<PureTracerData>;
 
 impl<F: PrimeField32> RvrMeteredCostInstance<F> {
+    /// Persist the compiled shared library into `dir`. Returns the path to
+    /// the copied artifact. The user must re-supply `exe`, `executor_idx_to_air_idx`,
+    /// and `widths` when loading.
+    pub fn save(&self, dir: &std::path::Path) -> Result<std::path::PathBuf, super::CompileError> {
```
shuklaayush: imports at top

shuklaayush:
can you enable the `rvr` feature in ci's sdk tests so that we can confirm that they pass? 
also i think currently both pure and metered have the same name for the compiled `.so`. can you add a suffix to them so that the save doesn't overwrite by default

.github/workflows/sdk.yml
```diff
@@ -88,6 +88,14 @@ jobs:
           export RUST_MIN_STACK=8388608
           CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS=true cargo nextest run --release --features parallel,root-prover,evm-verify
 
+      - name: Run openvm-sdk crate tests (RVR)
```
shuklaayush: can you create a matrix instead of adding a duplicate step. see how it's done in the extension tests ci for example

crates/vm/src/arch/rvr/compile.rs
```diff
@@ -74,6 +89,10 @@ impl RvrCompiled {
         })?;
         Ok(dest_lib)
     }
+
+    pub fn save_artifact(&self, dest_dir: &Path) -> Result<PathBuf, CompileError> {
```
shuklaayush: do we still need this?

crates/vm/src/arch/rvr/metered.rs
```diff
@@ -413,7 +413,7 @@ impl<F: PrimeField32, S> RvrMeteredInstanceWith<F, S> {
     /// the copied artifact. The user must re-supply `exe`, `executor_idx_to_air_idx`,
     /// and any mode-specific data when loading.
     pub fn save(&self, dir: &std::path::Path) -> Result<std::path::PathBuf, super::CompileError> {
-        self.compiled.save_artifact(dir)
+        self.compiled.save_artifact_with_suffix(dir, "-metered")
```
shuklaayush: maybe the hyphen can be added automatically. also i feel like `save_artifact` should just take a path and all the filename creation with suffix or whatever should be a separate function. so the flow is
```
let filename = ...
save_artifact(dir.join(filename));
```

shuklaayush (commented):
some nits. also see claude's comments

shuklaayush (approved):

gdmlcjs (approved):
LGTM!

PR #2828, opened 2026-05-29
chore: change kitchen-sink benchmark instance type

stephenh-axiom-xyz:
Prover is much faster now, not really a need kitchen-sink to be on a `g6e` machine


gaxiom (approved):
Is it for test purpose?

stephenh-axiom-xyz:
Due to the lack of `g6e` machines

PR #2833, opened 2026-06-01
feat: reuse compiled artifacts in rvr

mansur20478:
Part 2 of the original ticket split into two. Part 1: [link](https://github.com/openvm-org/openvm/pull/2826) 

Part 2 consist of reusing persisted artifacts whenever possible instead of recompiling. The idea is to compile in rvr-mode till getting `.c` files, and instead of invoking `make`, check whether already existing `.so` are fit for reuse. Artifact fingerprints must match for `.so` file to be reused. During compilation

Artifact fingerprint is a SHA-256 fingerprint of:
1) `native_debug_info` flag
2) `toolchain.compiler`, `toolchain.linker`, `toolchain.make`, `toolchain.host_os`
3) `host_cpu_features`: affects `-march=native` (unnecessary?)
4) `make_args`: make variables except EXT_LIBS and EXT_SRCS
5) name + content of every file under generated C `project_dir` in sorted order. The folder includes external static libs and external source files too.

New methods:
`compile_pure_cached`, `compile_metered_cached`, `compile_metered_cost_cached` that take `cache_dir` as an additional argument - a place on disk for storing artifacts. Invokable from sdk.

The following loading functions do not perform validation:
`load_compiled_pure`, `load_compiled_metered`, `load_compiled_metered_cost` methods don't have `.so` validation.

Modified `build.rs` of extensions to rebuild rvr FFI staticlibs on Cargo.lock and ffi-common source changes.

resolves INT-7843


crates/vm/src/arch/rvr/artifact_cache.rs
```diff
@@ -0,0 +1,115 @@
...
+    h.update(toolchain.make.as_bytes());
+    h.update(b"\0");
+    h.update(toolchain.host_os.as_bytes());
+    h.update(b"\0");
+    h.update(host_cpu_features().as_bytes());
+    h.update(b"\0");
+    for arg in make_args {
+        h.update(arg.as_bytes());
+        h.update(b"\0");
+    }
```
gdmlcjs, line 47: I think this contains `EXT_LIBS`, the absolute paths to the FFI staticlibs that get written into a random temporary directory. Could we exclude those and have only the contents in the fingerprinting process?

gdmlcjs (changes requested):
No requests other than this

gdmlcjs (approved):
LGTM!

mansur20478:
@claude provide Code Review for this PR

gdmlcjs (commented):
- In crates/rvr/rvr-openvm/c/Makefile, there are some configs that can be set with environment variables using `?=`. They are not contained in the fingerprint so we should better add them in the fingerprint or change `?=` to `=`.
- There were some changes in the base branch: removed `lto=false` for macOS, changed method names from `compile_pure` to `compile`. Could you update those changes? A rebase might be helpful.
- I'm not sure, but instead of having two versions of each function (with `cache_dir: Option<&Path>` and without) it might be simpler to have just one version (with `cache_dir: Option<&Path>`). Could you look into that?

PR #2834, opened 2026-06-02
refactor: cleanup rvr callbacks

gdmlcjs:
- Moved callbacks for IO functions from the global context into the local scope of each extension. They are divided into RV32I phantom instructions and RV32IO instructions.
- Changed the implementation of the Deferral functions to be mostly in C to reduce the indirection through FFI boundaries. The only FFI call is through the registered callback function.
- Simplified staticlib path resolution for extensions.

closes INT-7838


extensions/rv32im/rvr/Cargo.toml
```diff
@@ -8,13 +8,33 @@ homepage.workspace = true
...
+
+libloading = { workspace = true, optional = true }
+rand = { workspace = true, optional = true }
+
+[dev-dependencies]
+p3-baby-bear.workspace = true
 
 [features]
 default = []
+rvr = [
```
shuklaayush, line 29: this probably needs to enable `openvm-circuit/rvr`

crates/rvr/rvr-openvm-lift/src/opcode.rs
```diff
@@ -455,27 +448,6 @@ fn lift_phantom<F: PrimeField32>(insn: &Instruction<F>, pc: u32) -> LiftedInstr
...
-
-            // HintRandom — fill hint_stream with [a]_1 random words
-            Rv32Phantom::HintRandom => {
-                let num_words_reg = decode_reg(insn.a);
-                body(pc, Instr::HintRandom { num_words_reg })
-            }
-        };
-    }
-
     // Unknown phantom — treat as nop (forward compatible).
```
shuklaayush: seems like a footgun. i feel like we should emit a warning or error here

extensions/deferral/rvr/c/rvr_ext_deferral.c
```diff
@@ -25,14 +43,79 @@ void register_deferral_callbacks(const DeferralHostCallbacks* cb) {
...
+  uint32_t key_words[OUTPUT_KEY_WORDS];
+  rd_mem_u32_range_traced(state, input_ptr, key_words, OUTPUT_KEY_WORDS);
+
+  /* output_len is the u64 LE stored right after the commit. */
+  uint64_t output_len = (uint64_t)key_words[COMMIT_WORDS] |
+                        ((uint64_t)key_words[COMMIT_WORDS + 1] << 32);
+
+  /* Look up the raw output into a heap buffer sized to output_len. The
+   * output_commit is the leading DEFERRAL_COMMIT_NUM_BYTES of output_key. */
+  uint8_t* output_raw = (uint8_t*)malloc((size_t)output_len);
```
shuklaayush, line 98: maybe we want to assert that output_len > 0 or something

extensions/rv32im/rvr/c/rv32i_phantom_callbacks.c
```diff
@@ -0,0 +1,37 @@
...
+ * HINT_INPUT, PRINT_STR, HINT_RANDOM phantoms and the extension hint-stream
+ * setter.
+ */
+
+#include "rv32i_phantom_callbacks.h"
+
+#include "openvm_io.h"
+
+typedef struct {
+  void (*hint_input)(void* ctx);
```
shuklaayush, line 12: i wonder if we can keep a pointer to `ctx` here instead of calling `openvm_get_io_ctx` every time

gdmlcjs: I think it might be a bit complicated to pass the pointer to each extension to store. How about using `extern void* g_io_ctx` instead of `openvm_get_io_ctx()` so the extensions can access `ctx` directly?

shuklaayush: in that case, let's just keep it how it is. `extern void* g_io_ctx` doesn't really help much. i was wondering if we could avoid an opaque call altogether

shuklaayush (approved):
merging this but we should fix what claude is mentioning in a follow up

PR #2835, opened 2026-06-02
fix: SDK verify_app_proof checks number of user pvs

stephenh-axiom-xyz:
Resolves INT-8046.

## Overview

This branch makes SDK app proof verification reject proofs with the wrong number of user public values.

`verify_app_proof` now takes `num_user_pvs` and checks it against `proof.user_public_values.public_values.len()` before verifying the user-PVS commitment. This prevents callers from verifying against only memory dimensions while silently accepting a proof with an unexpected user-PVS shape.

## SDK Changes

Files:

- `crates/sdk/src/prover/app.rs`
- `crates/sdk/src/keygen/mod.rs`

`AppVerifyingKey` now stores `num_user_pvs`, derived from `SystemConfig::num_public_values` during keygen.

`verify_app_proof` now:

- accepts `num_user_pvs`,
- returns `SdkError`,
- fails with a clear error if the proof has the wrong number of user public values,
- continues to verify segments and the user-PVS commitment as before.

## Call Sites

Files:

- `crates/cli/src/commands/verify.rs`
- `crates/sdk/examples/sdk_app.rs`
- `benchmarks/prove/src/lib.rs`
- `benchmarks/prove/src/bin/keccak_par.rs`

Callers now pass `app_vk.num_user_pvs` into `verify_app_proof`. The benchmark paths also use `app_vk.memory_dimensions` instead of recomputing memory dimensions from the VM config.

### DIFF
--- a/crates/sdk/src/prover/app.rs
+++ b/crates/sdk/src/prover/app.rs
@@ -21,6 +21,7 @@ use openvm_stark_sdk::config::baby_bear_poseidon2::Digest;
 use tracing::instrument;
 
 use crate::{
+    keygen::AppVerifyingKey,
     prover::vm::{new_local_prover, types::VmProvingKey},
     util::check_max_constraint_degrees,
     SdkError, StdIn, F, SC,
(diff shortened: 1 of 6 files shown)


crates/sdk/src/prover/app.rs
```diff
@@ -158,20 +163,32 @@ where
 pub fn verify_app_proof<E: StarkEngine<SC = SC>>(
     vk: &MultiStarkVerifyingKey<SC>,
     memory_dimensions: MemoryDimensions,
+    num_user_pvs: usize,
```
shuklaayush, line 181: why not just pass `AppVerifyingKey` here?

shuklaayush (approved):
lgtm

PR #2838, opened 2026-06-02
fix: prevent DeferralAggregationPvs values from overflowing

stephenh-axiom-xyz:
Resolves INT-8053.

## Overview

This branch prevents `DeferralAggregationPvs::{merkle_depth, num_def_circuit_proofs}` from being accepted only modulo the BabyBear field in the deferral inner aggregation circuit.

The main change is a new 8-bit range-check lookup in `DeferralAggPvsAir` on:

```text
MAX_DEF_AGG_MERKLE_DEPTH - merkle_depth
```

Together with the existing recursive depth constraints, this bounds the canonical aggregation Merkle depth by `MAX_DEF_AGG_MERKLE_DEPTH`. Since the number of deferral circuit proofs is inductively bounded by `2^merkle_depth`, this also prevents `num_def_circuit_proofs` from wrapping in `F`.

Base branch: `fix/canonical-deferral-agg-padding`

Commit in this branch:

- `fix: prevent DeferralAggregationPvs values from overflowing`

## `crates/continuations`: Deferral Inner Aggregation PVS AIR

Files:

- `crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs`
- `crates/continuations/src/circuit/deferral/inner/mod.rs`

### What Changed

`DeferralAggPvsAir` now depends on the shared recursion `RangeCheckerBus`.

On the first row of the AIR, it performs a lookup:

```text
value = MAX_DEF_AGG_MERKLE_DEPTH - merkle_depth
max_bits = 8
```

This records that `MAX_DEF_AGG_MERKLE_DEPTH - merkle_depth` is in `[0, 256)`.

### Constraint Intent

Before this branch, the aggregation public values were related by field equalities:

```text
parent_num_def_circuit_proofs = left_num_proofs + right_num_proofs
parent_merkle_depth = child_merkle_depth + 1
```

Those equalities are over `F`, so a maliciously shaped recursive aggregation could satisfy them with wrapped field values. For example, a child depth of `p - 1` would produce parent depth `0` modulo `p`.

The new range check rules out this wraparound in valid finite recursive aggregation proofs. The proof tree still relies on the existing child-depth recurrence:

- wrapper proofs copy the child depth,
- non-wrapper aggregation proofs set parent depth to child depth plus one,
- present sibling aggregation proofs must have equal child depths,
- one-child padded aggregation proofs use the existing depth encoder to select the padding hash depth.

### AIR Columns

No trace columns were added or removed. The change is a new bus interaction, not a column-layout change.

Existing row shapes are unchanged:

| Case | Rows | `local.is_present` | `next.is_present` | `has_verifier_pvs` | Public `merkle_depth` behavior | New range lookup |
| --- | ---: | ---: | ---: | ---: | --- | --- |
| Wrapper | 1 | `1` | N/A | leaf or aggregation child | Copies child depth | `MAX - copied_depth` on first row |
| Two present children | 2 | `1` | `1` | leaf or aggregation children | `child_depth + 1`; aggregation siblings have equal child depth | `MAX - parent_depth` on first row |
| One child plus padding | 2 | `1` | `0` | leaf or aggregation child | `child_depth + 1`; padding zero hash selected by existing depth encoder | `MAX - parent_depth` on first row |

## `crates/continuations`: Tracegen Plumbing

Files:

- `crates/continuations/src/circuit/deferral/inner/def_pvs/trace.rs`
- `crates/continuations/src/circuit/deferral/inner/trace.rs`
- `crates/continuations/src/prover/deferral/inner/trace.rs`

### What Changed

`def_pvs::generate_proving_ctx` now returns:

```rust
DeferralAggPvsTraceCtx {
    proving_ctx,
    range_check_inputs,
}
```

The range-check input is computed from the same `DeferralAggregationPvs` public values generated for the AIR:

```text
MAX_DEF_AGG_MERKLE_DEPTH - pvs.merkle_depth
```

The deferral inner pre-context carries these `range_check_inputs`, and the prover passes them through `VerifierExternalData` to the recursive verifier subcircuit’s existing 8-bit range checker.

### DIFF
--- a/crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
+++ b/crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
@@ -11,6 +11,7 @@ use openvm_recursion_circuit::{
         Poseidon2CompressBus, Poseidon2CompressMessage, PublicValuesBus, PublicValuesBusMessage,
     },
     prelude::DIGEST_SIZE,
+    primitives::bus::{RangeCheckerBus, RangeCheckerBusMessage},
 };
 use openvm_recursion_circuit_derive::AlignedBorrow;
 use openvm_stark_backend::{
@@ -51,6 +52,7 @@ pub struct DeferralAggPvsCols<F> {
 pub struct DeferralAggPvsAir {
     pub public_values_bus: PublicValuesBus,
     pub poseidon2_bus: Poseidon2CompressBus,
+    pub range_bus: RangeCheckerBus,
     pub input_or_merkle_commit_bus: InputOrMerkleCommitBus,
     pub def_pvs_consistency_bus: DefPvsConsistencyBus,
 
@@ -66,6 +68,7 @@ impl DeferralAggPvsAir {
     pub fn new(
         public_values_bus: PublicValuesBus,
         poseidon2_bus: Poseidon2CompressBus,
+        range_bus: RangeCheckerBus,
         input_or_merkle_commit_bus: InputOrMerkleCommitBus,
         def_pvs_consistency_bus: DefPvsConsistencyBus,
     ) -> Self {
@@ -74,6 +77,7 @@ impl DeferralAggPvsAir {
         Self {
             public_values_bus,
             poseidon2_bus,
+            range_bus,
             input_or_merkle_commit_bus,
             def_pvs_consistency_bus,
             encoder,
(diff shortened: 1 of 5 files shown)


shuklaayush (approved):
lgtm. seems like a straightforward fix

PR #2839, opened 2026-06-02
fix: enforce CommitBytes canonical representation

stephenh-axiom-xyz:
Resolves INT-8066.

## Overview

This PR enforces a canonical byte representation for `CommitBytes` in `openvm-continuations`.
Previously, arbitrary 32-byte values could be accepted and later converted into the same BabyBear
digest limbs by reduction modulo `p^8`. The change makes `CommitBytes` reject non-canonical
encodings at construction/conversion boundaries.

Base reviewed: `origin/develop-v2.0.0-rc.2`

### `CommitBytes` canonicality

- `CommitBytes::new` now asserts that input bytes round-trip through
  `bytes_to_u32_digest` and `u32_digest_to_bytes`.
- `From<[u32; DIGEST_SIZE]> for CommitBytes` now asserts every limb is less than the
  BabyBear modulus before encoding.
- `From<[F; DIGEST_SIZE]> for CommitBytes` delegates through the checked `u32` limb path.
- `From<CommitBytes> for [F; DIGEST_SIZE]` asserts canonicality before converting bytes
  into field limbs.
- The `root-prover` `From<Bn254> for CommitBytes` path now uses `CommitBytes::new`, so it
  observes the same canonicality invariant.

### API cleanup

- Removed the unused `CommitBytes::reverse` mutator. Since `CommitBytes` has private storage,
  removing this keeps the canonicality invariant tied to the checked constructors/conversions.

### DIFF
--- a/crates/continuations/src/commit_bytes.rs
+++ b/crates/continuations/src/commit_bytes.rs
@@ -7,23 +7,24 @@ use p3_field::{PrimeCharacteristicRing, PrimeField32};
 
 pub const COMMIT_NUM_BYTES: usize = 32;
 
-/// Wrapper for an array of big-endian bytes, representing an unsigned big integer. Each commit can
-/// be converted to a Bn254 using the trivial identification as natural numbers or into a `u32`
-/// digest by decomposing the big integer base-`F::MODULUS`.
+/// Wrapper for the canonical big-endian byte representation of a BabyBear digest interpreted as
+/// an unsigned integer in base `F::MODULUS`. Each commit can be converted to a Bn254 using the
+/// trivial identification as natural numbers or into a `u32` digest by decomposing the big integer
+/// base-`F::MODULUS`.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct CommitBytes([u8; COMMIT_NUM_BYTES]);
 
 impl CommitBytes {
     pub fn new(bytes: [u8; COMMIT_NUM_BYTES]) -> Self {
+        assert!(
+            u32_digest_to_bytes(&bytes_to_u32_digest(&bytes)) == bytes,
+            "non-canonical CommitBytes for BabyBear digest"
+        );
         Self(bytes)
     }
 
     pub fn as_slice(&self) -> &[u8; COMMIT_NUM_BYTES] {
         &self.0
-    }
-
-    pub fn reverse(&mut self) {
-        self.0.reverse();
     }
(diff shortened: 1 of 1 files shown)


crates/continuations/src/commit_bytes.rs
```diff
@@ -7,24 +7,25 @@ use p3_field::{PrimeCharacteristicRing, PrimeField32};
...
+/// base-`F::MODULUS`.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct CommitBytes([u8; COMMIT_NUM_BYTES]);
 
 impl CommitBytes {
     pub fn new(bytes: [u8; COMMIT_NUM_BYTES]) -> Self {
+        assert!(
+            u32_digest_to_bytes(&bytes_to_u32_digest(&bytes)) == bytes,
+            "non-canonical CommitBytes for BabyBear digest"
+        );
```
shuklaayush, line 22: this is correct but i feel like we can also just do something like `bytes_to_biguint(bytes) < p^8`

crates/continuations/src/commit_bytes.rs
```diff
@@ -44,18 +45,26 @@ impl<F: PrimeCharacteristicRing> From<VkCommitBytes> for VkCommit<F> {
...
+        Self::from(value.map(|x| x.as_canonical_u32()))
     }
 }
 
 impl From<[u32; DIGEST_SIZE]> for CommitBytes {
     fn from(value: [u32; DIGEST_SIZE]) -> Self {
+        assert!(
+            value.iter().all(|&digit| digit < F::ORDER_U32),
+            "non-canonical BabyBear digest limb"
+        );
```
shuklaayush, line 57: maybe we should implement `TryFrom` instead

shuklaayush (approved):
couple of nits

crates/continuations/src/commit_bytes.rs
stephenh-axiom-xyz: There was an explicit reason I didn't do this that I can't remember lol, but I think it had smth to do with `CommitBytes` being valid for both the `BabyBear` and `Bn254` digest? Either way I think we should leave it for later, as the blast radius is large

crates/continuations/src/commit_bytes.rs
stephenh-axiom-xyz: I feel like both are a little gross lol

PR #2840, opened 2026-06-03
perf: gpu memory optimization by omitting the bottom rows of the merkle tree

GunaDD:
Motivation: A memory optimization of the GPU merkle tree is needed (once we go to 2^32 address space size) because currently how the MemoryMerkleTree works is that if we have N bytes then we will have N/16 leaves (and N/8 nodes in total) and each leaf is stored as 8 field elements. This means we need 4N bytes of VRAM to store the memory merkle tree in the GPU. With N = 2^32 (once we go to 2^32 AS size), 4N is roughy 17 GB which will OOM the GPU (our current limit is ~15 G). 

Optimization idea of this PR: Don't store the last OMITTED_BOTTOM_LEVELS levels of the MemoryMerkleSubTree in the GPU memory and only computing it when needed for address spaces with large sizes (AS2 and deferral AS). This is done by recomputing (i.e. re-computing the poseidon2 hash) from the raw memory. This saves the GPU memory required for the Merkle tree itself by a factor of 8.   

Reth benchmark results: https://github.com/axiom-crypto/openvm-eth/actions/runs/26959247863

The summary of the result is proving time increased by 0.77s but the generate mem proving ctxs went down from 3.84 G to 2.09 G and set initial memory went down from 3.63 G to 1.88 G. Note that it didn't went down by a factor of 8 because there are other things that uses the memory. The other things here include the boundarychipgpu and initial_memory buffer (see the MemoryInventoryGPU struct for details). 

Closes INT-8079

### DIFF
--- a/crates/vm/cuda/include/system/memory/params.cuh
+++ b/crates/vm/cuda/include/system/memory/params.cuh
@@ -60,6 +60,9 @@ inline constexpr size_t BLOCK_FE_WIDTH = MEMORY_BLOCK_BYTES / U16_CELL_SIZE;
 // Blocks per merkle leaf.
 inline constexpr size_t BLOCKS_PER_LEAF = DIGEST_WIDTH / BLOCK_FE_WIDTH;
 
+// Number of bottom merkle levels omitted from large GPU subtree buffers.
+inline constexpr size_t OMITTED_BOTTOM_LEVELS = 3;
+
 // Address space ID for the F-celled deferral AS. Matches
 // `openvm_instructions::DEFERRAL_AS`.
 inline constexpr uint32_t DEFERRAL_AS = 4;
--- a/crates/vm/cuda/src/system/memory/merkle_tree.cu
+++ b/crates/vm/cuda/src/system/memory/merkle_tree.cu
@@ -326,7 +326,6 @@ __device__ __forceinline__ bool virtual_node_exists(
 __device__ __forceinline__ size_t stored_node_index(
     uint32_t const subtree_height,
     uint32_t const actual_height,
-    uint8_t const layout,
     uint32_t const node_height,
     size_t const label
 ) {
@@ -359,7 +358,7 @@ __device__ void load_virtual_node(
         return;
     }
 
-    auto const idx = stored_node_index(subtree_height, actual_height, layout, node_height, label);
+    auto const idx = stored_node_index(subtree_height, actual_height, node_height, label);
     COPY_DIGEST(out, &subtree[idx]);
 }
 
@@ -375,7 +374,7 @@ __device__ void store_virtual_node(
     if (layout == OMIT_BOTTOM_LEVELS && node_height < OMITTED_BOTTOM_LEVELS) {
         return;
     }
-    auto const idx = stored_node_index(subtree_height, actual_height, layout, node_height, label);
+    auto const idx = stored_node_index(subtree_height, actual_height, node_height, label);
     COPY_DIGEST(&subtree[idx], value);
 }
 
--- a/crates/vm/cuda/src/system/memory/merkle_tree.cu
+++ b/crates/vm/cuda/src/system/memory/merkle_tree.cu
@@ -15,6 +15,32 @@ struct alignas(32) digest_t {
 
 #define COPY_DIGEST(dst, src) memcpy(dst, src, sizeof(digest_t))
 
+enum MemoryMerkleSubTreeLayout : uint8_t {
+    FULL = 0,
+    OMIT_BOTTOM_LEVELS = 1,
+};
+
+__device__ __forceinline__ void hash_raw_memory_leaf(
+    uint8_t const *__restrict__ data,
+    uint32_t const addr_space_idx,
+    size_t const leaf_label,
+    digest_t *out
+) {
+    Fp cells[CELLS] = {0};
+#pragma unroll
+    for (size_t i = 0; i < DIGEST_WIDTH; ++i) {
+        if (addr_space_idx + 1 == DEFERRAL_AS) {
+            cells[i] = reinterpret_cast<Fp const *>(data)[DIGEST_WIDTH * leaf_label + i];
+        } else {
+            auto byte_off = U16_CELL_SIZE * (DIGEST_WIDTH * leaf_label + i);
+            cells[i] = Fp(u16_from_bytes_le(data + byte_off));
+        }
+    }
+
+    poseidon2_mix(cells);
+    COPY_DIGEST(out, cells);
+}
+
 // `ADDR_SPACE_IDX` is the address space minus `ADDR_SPACE_OFFSET` (which is 1)
 //
 // DEFERRAL_AS stores Fp cells directly.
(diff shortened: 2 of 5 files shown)


GunaDD:
reth benchmark: https://github.com/axiom-crypto/openvm-eth/actions/runs/26959247863

proving time increased by 0.77s but the generate mem proving ctxs went down from 3.84 G to 2.09 G and set initial memory went down from 3.63 G to 1.88 G

crates/vm/cuda/include/system/memory/params.cuh
```diff
@@ -60,6 +60,9 @@ inline constexpr size_t BLOCK_FE_WIDTH = MEMORY_BLOCK_BYTES / U16_CELL_SIZE;
 // Blocks per merkle leaf.
 inline constexpr size_t BLOCKS_PER_LEAF = DIGEST_WIDTH / BLOCK_FE_WIDTH;
 
+// Number of bottom merkle levels omitted from large GPU subtree buffers.
+inline constexpr size_t OMITTED_BOTTOM_LEVELS = 3;
```
shuklaayush, line 64: i feel like this is conceptually similar to paging so let's call this `MERKLE_PAGE_BITS` instead. see how metered execution also does paging and [this old doc](https://hackmd.io/@axiom/Hyda2v3Xex) which has some diagrams

crates/vm/src/system/cuda/merkle_tree/cuda.rs
shuklaayush, line 162: can you add a `tracing::instrument` call here so we can measure the time

crates/vm/src/system/cuda/merkle_tree/mod.rs
shuklaayush: we can probably just remove this

crates/vm/src/system/cuda/merkle_tree/mod.rs
```diff
@@ -48,9 +57,48 @@ pub struct MemoryMerkleSubTree {
...
+    /// SAFETY / lifetime: under the `OmitBottomLevels` layout the omitted bottom levels are never
+    /// materialized into `buf`; they are recomputed on demand from this buffer during
+    /// [`MemoryMerkleTree::update_with_touched_blocks`] (see `recompute_omitted_node` in
+    /// `merkle_tree.cu`). The pointed-to buffer must therefore stay alive and unmodified from the
+    /// `build_async` call all the way until that update completes — not merely until the build
+    /// kernel runs. The compiler cannot enforce this through the erased pointer, so the caller is
+    /// responsible: in production the owning `DeviceBuffer`s are held in
+    /// `MemoryInventoryGPU::initial_memory` and only dropped after `update_with_touched_blocks` is
+    /// done.
+    initial_data_ptr: usize,
```
shuklaayush: can this be `Arc<DeviceBuffer<u8>>` or `Option<Arc<DeviceBuffer<u8>>>` instead for proper lifetime tracking?

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -15,29 +15,90 @@ struct alignas(32) digest_t {
...
+// `ADDR_SPACE_IDX` is the address space minus `ADDR_SPACE_OFFSET` (which is 1)
+//
+// DEFERRAL_AS stores Fp cells directly.
+// Non-deferral address spaces store u16 cells in little-endian byte order.
+template <int ADDR_SPACE_IDX>
+__global__ void merkle_tree_init(uint8_t *__restrict__ data, digest_t *__restrict__ out) {
+    auto gid = blockDim.x * blockIdx.x + threadIdx.x;
+
+    digest_t digest;
+    hash_raw_memory_leaf(data, ADDR_SPACE_IDX, gid, &digest);
```
shuklaayush, line 53: wonder if we can avoid some of this hashing for leaves that are all zeros

shuklaayush (commented):
dropped some comments. i'll leave it to @Golovanov399 and/or @gaxiom to review the cuda changes in detail

Golovanov399:
Should it be rebased?

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -15,29 +15,90 @@ struct alignas(32) digest_t {
...
+    // that is, node_height = 0 for the leafs
+    size_t const first_leaf = label << node_height;
+
+    for (size_t i = 0; i < num_leaves; ++i) {
+        hash_raw_memory_leaf(data, addr_space_idx, first_leaf + i, &layer[i]);
+    }
+
+    // cells is a 2-to-1 compression buffer
+    Fp cells[CELLS];
+    for (size_t width = num_leaves / 2; width > 0; width /= 2) {
```
Golovanov399, line 84: You either don't need `1 << OMITTED_BOTTOM_LEVELS` in `layer` (you only use half of this memory, yet it probably affects register slippage or something), or you may want to parallelize computation on each of the omitted levels (probably not worth it). Or maybe you want to delegate this all to the same SM and do some device synchronize and whatnot, but it can be postponed

crates/vm/cuda/include/system/memory/params.cuh
GunaDD: hmmm what does `PAGE` mean there actually

crates/vm/src/system/cuda/merkle_tree/mod.rs
GunaDD: should we do it though ?

crates/vm/cuda/src/system/memory/merkle_tree.cu
GunaDD: I think we can by checking if data is zero or not and storing zero hash if it were. But can we just do it as a follow up since its orthogonal to was this PR is doing ?

crates/vm/cuda/include/system/memory/params.cuh
shuklaayush: essentially what you're doing is saying that instead of storing the commitment to individual memory addresses, you'll only store a commitment to a collection of memory addresses (a page)
this is similar to how [virtual memory](https://en.wikipedia.org/wiki/Virtual_memory#Paged_virtual_memory) is divided into pages (not exactly but close enough)

crates/vm/src/system/cuda/merkle_tree/mod.rs
shuklaayush: yeah, let's just get rid of it

crates/vm/cuda/src/system/memory/merkle_tree.cu
shuklaayush: cool, but please make a ticket so we can track it

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -297,33 +401,48 @@ __global__ void update_merkle_layer(
...
         layer_height - 1,
-        actual_subtree_heights[address_space_idx],
-        2 * parent_label
+        2 * parent_label,
+        &old_left_digest
     );
-    auto const old_right_digest = layer_value_on_height(
-        subtree_layer,
+    digest_t old_right_digest;
+    load_virtual_node(
```
Golovanov399, line 431: Noting that you may save some recomputing on the lowest levels here by somehow reusing what you need from `old_left_digest` for `old_right_digest`

Golovanov399 (commented):
I left some comments, in principle I understand if not redundantly recomputing stuff twice for close addresses is out of the scope of this task, but I think it's worth doing in long term and if it's not in this PR then I ask you to make a corresponding ticket for the future

crates/vm/cuda/src/system/memory/merkle_tree.cu
GunaDD: Done.

crates/vm/src/system/cuda/merkle_tree/mod.rs
```diff
@@ -48,9 +57,49 @@ pub struct MemoryMerkleSubTree {
...
+    /// while the subtree is alive. Under the `OmitBottomLevels` layout the omitted bottom levels
+    /// are never materialized into `buf`; they are recomputed on demand from this buffer during
+    /// [`MemoryMerkleTree::update_with_touched_blocks`] (see `recompute_omitted_node` in
+    /// `merkle_tree.cu`), so the buffer must stay alive until that update completes.
+    ///
+    /// NOTE: this only fixes host-side ownership. The buffer is also consumed by GPU kernels
+    /// enqueued on the stream, so it must additionally outlive those kernels — that ordering is
+    /// guaranteed by the `stream.synchronize()` in [`MemoryMerkleTree::drop_subtrees`], not by the
+    /// `Arc`. Drop the subtrees (which releases these handles) only after that sync.
+    initial_data: Option<Arc<DeviceBuffer<u8>>>,
```
shuklaayush: comments like this are too verbose and borderline slop. see if you can trim them down

crates/vm/cuda/src/system/memory/merkle_tree.cu
GunaDD: I think this is not true because if you load the left subtree, it will only compute its subtree which is independent of the right subtree.

crates/vm/cuda/src/system/memory/merkle_tree.cu
GunaDD: I think we can postpone this

crates/vm/cuda/include/system/memory/params.cuh
GunaDD: I honestly think `OMITTED_BOTTOM_LEVELS` is the better name here

Golovanov399 (approved):

PR #2841, opened 2026-06-04
perf(v2.1-rv64): pair byte range checks

shuklaayush:
## Summary

Reduce RV64 loadstore and load-sign-extend interaction counts by using the existing 8-bit `BitwiseOperationLookup` for byte-limb range checks. These checks are still required because the u16 memory bus only bounds packed u16 cells, but pairing bytes through the bitwise lookup matches the older byte-oriented pattern and avoids one interaction per checked byte.

## Changes

- Switch `LoadStoreCoreAir` read/previous byte checks from per-byte `VariableRangeChecker` lookups to paired bitwise range lookups.
- Switch `LoadSignExtendCoreAir` shifted-read byte checks to paired bitwise range lookups while keeping the separate sign-bit tail range check.
- Thread the bitwise lookup chip/bus through CPU and CUDA loadstore/load-sign-extend wiring.
- Update CPU and CUDA tests to include the bitwise lookup periphery.

[reth benchmark comparison](https://github.com/axiom-crypto/openvm-eth/actions/runs/26968844374#summary-79578787595)
144 -> 120 segments 🤯

### DIFF
--- a/extensions/riscv/circuit/cuda/src/load_sign_extend.cu
+++ b/extensions/riscv/circuit/cuda/src/load_sign_extend.cu
@@ -36,11 +36,15 @@ template <size_t NUM_CELLS> struct LoadSignExtendCoreRecord {
 
 template <size_t NUM_CELLS> struct LoadSignExtendCore {
     VariableRangeChecker range_checker;
+    BitwiseOperationLookup bitwise_lookup;
 
     template <typename T> using Cols = LoadSignExtendCoreCols<T, NUM_CELLS>;
 
-    __device__ LoadSignExtendCore(VariableRangeChecker range_checker)
-        : range_checker(range_checker) {}
+    __device__ LoadSignExtendCore(
+        VariableRangeChecker range_checker,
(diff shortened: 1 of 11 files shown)


extensions/riscv/circuit/cuda/src/load_sign_extend.cu
```diff
@@ -36,11 +36,15 @@ template <size_t NUM_CELLS> struct LoadSignExtendCoreRecord {
 
 template <size_t NUM_CELLS> struct LoadSignExtendCore {
     VariableRangeChecker range_checker;
```
GunaDD, line 38: is the variablerangechecker still needed or is this dead code ?

GunaDD (approved):
LGTM

extensions/riscv/circuit/cuda/src/load_sign_extend.cu
shuklaayush: yes i believe its required in the msb range check

PR #2843, opened 2026-06-04
feat: add verify-stark deferral path prover constructor in SDK

stephenh-axiom-xyz:
Resolves INT-8084.

## Overview

This PR adds an SDK path for constructing a deferral-enabled `DeferralPathProver` backed by the
`verify-stark` circuit. The new constructor lets one SDK recursively verify its own VM STARK proofs:
the same SDK can prove an app execution, prove a `verify-stark` program over that proof, and then
prove `verify-stark` over the previous `verify-stark` proof.

The core idea is to derive the self-referential deferral path from a cheap dummy deferral circuit,
then replace only the leaf deferral circuit prover with the real `verify-stark` prover while
reusing the derived deferral aggregation path.

### Dummy Deferral Circuit VK Seed

- Adds `circuit::deferral::dummy`.
- Introduces a crate-private `EmptyAirWithPvs`, a minimal one-column AIR with
  `DeferralCircuitPvs`-shaped public values.
- Adds `dummy_deferral_circuit_vk::<E>(system_params)`, which keygens only the dummy VK needed to
  seed fixed-point deferral path derivation.

### Test Cleanup

- Replaces the test-local duplicate `EmptyAirWithPvs` implementation with a re-export of the new
  shared dummy AIR.

### Verify-Stark Deferral Path Constructor

- Moves `openvm-verify-stark-circuit` from a dev-dependency to a normal SDK dependency.
- Adds `prover::deferral::verify_stark`.
- Implements `DeferralPathProver::verify_stark(...)`.

The constructor:

- derives fixed-point deferral path artifacts from the dummy deferral circuit;
- builds a real `DeferredVerify{Cpu,Gpu}CircuitProver`;
- bakes in the derived `def_hook_commit` so verify-stark can accept deferral-carrying child proofs;
- returns a usable `DeferralPathProver` with the derived `AggProver` reused rather than regenerated.

### Builder API

- Replaces the builder's single `Option<DeferralProver>` storage with a `DeferralSource` enum.
- Keeps `deferral_prover(DeferralProver)` working as before by wrapping it into a path prover during
  `build`.
- Adds `deferral_path_prover(DeferralPathProver)` for callers that already constructed the full
  deferral path, such as through `DeferralPathProver::verify_stark`.
- Both entry points share one builder slot, so callers cannot set both.

### SDK Tests

- Adds `make_verify_stark_path_sdk`, which constructs an SDK using
  `DeferralPathProver::verify_stark`.
- Adds `test_verify_stark_path_sdk_can_verify_own_proofs`.

The new test uses one SDK to:

1. prove the Fibonacci program;
2. prove `verify-stark` over the Fibonacci proof;
3. prove `verify-stark` over the first `verify-stark` proof;
4. verify each produced STARK proof with `Sdk::verify_proof`.

It also reuses the decoded/converted `verify-stark` executable across both verify-stark proof
generations.

## Review Notes

- The fixed-point derivation assumes the deferral aggregation path and hook commitments depend only
  on aggregation and hook params, not on the concrete deferral circuit being proven.
- `DeferralPathProver::verify_stark` takes explicit `memory_dimensions` and `num_user_pvs` so the
  caller controls the verify-stark circuit shape without passing a full `SystemConfig`.
- The dummy AIR remains crate-private to `openvm-continuations`; SDK code only uses the public dummy
  VK helper.

### DIFF
--- a/crates/continuations/src/circuit/deferral/dummy.rs
+++ b/crates/continuations/src/circuit/deferral/dummy.rs
@@ -45,6 +45,7 @@ impl<AB: AirBuilder + AirBuilderWithPublicValues> Air<AB> for EmptyAirWithPvs {
             .map(|pv| (*pv).into())
             .collect::<Vec<AB::Expr>>();
         for pv in pvs {
+            // Trivial asserts to ensure pvs get placed in the DAG
             builder.assert_eq(pv.clone(), pv);
         }
     }
--- a/crates/continuations/src/circuit/deferral/dummy.rs
+++ b/crates/continuations/src/circuit/deferral/dummy.rs
@@ -1,0 +1,60 @@
+use std::sync::Arc;
+
+use openvm_circuit_primitives::ColumnsAir;
+use openvm_stark_backend::{
+    keygen::types::MultiStarkVerifyingKey, p3_matrix::Matrix, AirRef, PartitionedBaseAir,
+    StarkEngine, SystemParams,
+};
+use p3_air::{Air, AirBuilder, AirBuilderWithPublicValues, BaseAir, BaseAirWithPublicValues};
+
+use super::DeferralCircuitPvs;
+use crate::SC;
+
+/// Minimal AIR with a single zero column and deferral-circuit public values.
+///
+/// This is used only as a cheap verifying-key seed for deferral path fixed-point derivation; it is
+/// not intended to be proven.
+pub(crate) struct EmptyAirWithPvs(pub(crate) usize);
+
+// No columns provided: this dummy AIR has a single anonymous column and no matching `Cols` struct.
+impl ColumnsAir for EmptyAirWithPvs {}
+
+impl<F> BaseAir<F> for EmptyAirWithPvs {
+    fn width(&self) -> usize {
+        1
+    }
+}
+
+impl<F> BaseAirWithPublicValues<F> for EmptyAirWithPvs {
+    fn num_public_values(&self) -> usize {
+        self.0
+    }
+}
+
+impl<F> PartitionedBaseAir<F> for EmptyAirWithPvs {}
+
+impl<AB: AirBuilder + AirBuilderWithPublicValues> Air<AB> for EmptyAirWithPvs {
+    fn eval(&self, builder: &mut AB) {
+        let main = builder.main();
+        let local = main.row_slice(0).unwrap();
+        builder.assert_zero(local[0].clone());
+
+        let pvs = builder
+            .public_values()
+            .iter()
+            .map(|pv| (*pv).into())
+            .collect::<Vec<AB::Expr>>();
+        for pv in pvs {
+            builder.assert_eq(pv.clone(), pv);
+        }
+    }
+}
+
+pub fn dummy_deferral_circuit_vk<E>(system_params: SystemParams) -> Arc<MultiStarkVerifyingKey<SC>>
+where
+    E: StarkEngine<SC = SC>,
+{
+    let engine = E::new(system_params);
+    let dummy_air = Arc::new(EmptyAirWithPvs(DeferralCircuitPvs::<u8>::width())) as AirRef<SC>;
+    Arc::new(engine.keygen(&[dummy_air]).1)
+}
--- a/crates/sdk/src/builder.rs
+++ b/crates/sdk/src/builder.rs
@@ -42,6 +42,12 @@ enum AggSource {
     Pk(AggProvingKey),
 }
 
+#[allow(clippy::large_enum_variant)]
+enum DeferralSource {
+    Prover(DeferralProver),
+    PathProver(DeferralPathProver),
+}
+
 #[cfg(feature = "root-prover")]
 enum RootSource {
     Params(SystemParams),
@@ -73,7 +79,7 @@ where
     root_source: Option<RootSource>,
     agg_tree_config: Option<AggregationTreeConfig>,
     transpiler: Option<Transpiler<F>>,
-    deferral_prover: Option<DeferralProver>,
+    deferral_source: Option<DeferralSource>,
     #[cfg(feature = "evm-prove")]
     halo2_source: Option<Halo2Source>,
     #[cfg(feature = "evm-prove")]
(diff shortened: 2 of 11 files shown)


crates/sdk/src/tests.rs
```diff
@@ -115,40 +115,74 @@ fn make_deferral_enabled_sdk(
...
+fn make_verify_stark_path_sdk(
     app_params: SystemParams,
     agg_params: AggregationSystemParams,
-) -> Result<(Sdk, StdIn, DeferralInput)> {
-    let fib_vk = VmStarkVerifyingKey {
-        mvk: fib_sdk.agg_vk().as_ref().clone(),
-        baseline: fib_baseline,
+) -> Result<Sdk> {
+    let mut vm_config = openvm_sdk_config::SdkVmConfig::riscv32();
+    vm_config.system.config.memory_config.addr_spaces[DEFERRAL_AS as usize].num_cells = 1 << 25;
```
shuklaayush, line 123: probably can be extracted to some make deferral config helper

crates/sdk/src/builder.rs
```diff
@@ -460,9 +473,23 @@ where
     /// VM config.
     pub fn deferral_prover(mut self, deferral_prover: DeferralProver) -> Self {
         Self::set_once(
-            &mut self.deferral_prover,
-            "deferral_prover",
-            deferral_prover,
+            &mut self.deferral_source,
+            "deferral_source",
```
shuklaayush, line 473: unrelated but i feel like the more idiomatic pattern here is to have set_once return error and do a map_err here. rn, we'll get an assertion failure that `deferral_source` is already set but it doesn't really tell you where the error is coming from

crates/continuations/src/circuit/deferral/dummy.rs
```diff
@@ -0,0 +1,60 @@
...
+        let local = main.row_slice(0).unwrap();
+        builder.assert_zero(local[0].clone());
+
+        let pvs = builder
+            .public_values()
+            .iter()
+            .map(|pv| (*pv).into())
+            .collect::<Vec<AB::Expr>>();
+        for pv in pvs {
+            builder.assert_eq(pv.clone(), pv);
```
shuklaayush, line 49: not sure what's happening here. i realize this is dummy but maybe better to have a comment explaining why we do this

crates/sdk/src/prover/deferral/verify_stark.rs
```diff
@@ -0,0 +1,140 @@
...
+            .internal_recursive_prover
+            .get_self_vk_pcs_data()
+            .expect("internal-recursive prover must expose its self vk pcs data")
+            .commitment
+            .into();
+        let path_prover = Self::new(Arc::new(deferral_prover), agg_prover.clone());
+
+        DeferralPathArtifacts {
+            def_hook_commit: path_prover.def_hook_commit(),
+            ir_vk,
```
shuklaayush: it seems like `def_hook_commit` only needs `path_prover.agg_prover`. maybe we can move it to `AggProver` and avoid the `path_prover` construction altogether

crates/sdk/src/prover/deferral/verify_stark.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        let agg_config = AggregationConfig {
+            params: agg_params.clone(),
+        };
+        let deferral_prover = DeferralProver::new(dummy, agg_config.clone(), hook_params.clone());
+
+        let deferral_tree_config = AggregationTreeConfig {
+            num_children_leaf: 2,
+            num_children_internal: 2,
+        };
```
shuklaayush: it's better to have this as a constant somewhere

crates/sdk/src/prover/deferral/verify_stark.rs
```diff
@@ -0,0 +1,140 @@
...
+        memory_dimensions: MemoryDimensions,
+        num_user_pvs: usize,
+    ) -> Self {
+        let DeferralPathArtifacts {
+            def_hook_commit,
+            ir_vk,
+            ir_cached_commit,
+            agg_prover,
+        } = Self::fixed_point_artifacts(agg_params, &hook_params);
+
```
shuklaayush: should we also return `vk_pre_hash` and `cached_commit` here and assert that they're consistent with the deferral prover ones?

crates/sdk/src/prover/deferral/verify_stark.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// with the given params can recursively verify the VM STARK proofs it produces, including its
+    /// own deferral-carrying proofs.
+    ///
+    /// The deferral-enabled internal-recursive vk and the self-referential `def_hook_commit` are
+    /// derived internally from a dummy deferral circuit.
+    pub fn verify_stark(
+        agg_params: &AggregationSystemParams,
+        hook_params: SystemParams,
+        memory_dimensions: MemoryDimensions,
+        num_user_pvs: usize,
```
shuklaayush: i wonder if we should just pass `vm_config` here

shuklaayush (approved):
approving to unblock but i'm not a fan of this "dummy" pattern in general. i feel like what need is just something like `MultiStarkVkeyFrame` (something which describes a circuit shape) and not this dummy air with keygen

crates/sdk/src/prover/deferral/verify_stark.rs
stephenh-axiom-xyz: The `verify_stark` circuit is good for any `vm_config` given a fixed `memory_dimensions` and `num_user_pvs`

crates/sdk/src/builder.rs
stephenh-axiom-xyz: Perhaps something to think about for another time. IMO it makes actual construction awkward (as you have to unwrap or map or smth per setter)

crates/sdk/src/tests.rs
stephenh-axiom-xyz: We do this in a lot of tests, am going to leave for now to ensure blast radius is small

stephenh-axiom-xyz:
Agreed that the dummy pattern is non-ideal, but changing away from it will be a change with a large blast radius (and shouldn't be in this PR)

PR #2844, opened 2026-06-04
feat: collapse user public values to be dense in verify-stark guest

stephenh-axiom-xyz:
Resolves INT-8085.

## Overview

This commit changes the verify-stark guest API output for `user_public_values` from the sparse byte-expanded encoding to the original dense byte list.

Previously, each user public value byte was encoded as a BabyBear field element and then serialized as four little-endian bytes, so only byte positions `0, 4, 8, ...` carried data and the remaining lanes were expected to be zero. The guest library now removes those zero lanes before returning `ProofOutput::user_public_values`.

## `openvm-verify-stark-guest`

- Updates `verify_stark_unchecked` to post-process the deferred output's user public values.
- Adds `collapse_user_public_values`, which:
  - requires the expanded output length to be a multiple of 4,
  - keeps the first byte of each 4-byte field serialization,
  - panics if any of the upper three bytes are non-zero.

Reviewer focus:
- The kept byte lane is `0, 4, 8, ...`, matching the producer's little-endian `u32` field serialization.
- Non-zero bytes in lanes `1..=3` of each group are treated as malformed output.

## SDK Tests And Fixture

- Updates `make_verify_stark_inputs` in `crates/sdk/src/tests.rs` so SDK tests pass compact expected `user_public_values` into the verify-stark guest program.
- Rebuilds the checked-in `crates/sdk/programs/examples/verify-stark.elf` so the fixture includes the new guest-library behavior.

Reviewer focus:
- The SDK helper mirrors the guest-library lane handling by asserting the upper three bytes of each 4-byte group are zero and passing only the first byte to the guest.
- The ELF fixture changed and now has executable mode in this commit.


crates/sdk/src/tests.rs
```diff
@@ -158,7 +158,17 @@ fn make_verify_stark_inputs(
...
+    let expanded_user_public_values = &output_raw[64..];
+    assert_eq!(expanded_user_public_values.len() % 4, 0);
+    let user_public_values = expanded_user_public_values
+        .chunks_exact(4)
+        .map(|bytes| {
+            assert_eq!(&bytes[1..], &[0; 3]);
+            bytes[0]
+        })
+        .collect::<Vec<_>>();
+
```
shuklaayush: maybe this can be extracted into a helper too like in the guest

shuklaayush (approved):
approving to unblock, but i wonder if the better way is to export the dense public values in `output_raw_from_proof` instead

stephenh-axiom-xyz:
We shouldn't change `output_raw_from_proof` because the output of the actual deferral function is still comprised of field elements. This is purely a verify-stark guest library change for convenience

PR #2845, opened 2026-06-04
fix: resolve openvm v2.0 audit issues

gdmlcjs:
- made Keccak `finalize` panic for output buffers that are longer that 32 bytes instead of silently truncating
- added an assertion to ensure the zkvm implementation of `xorin` only handles input lengths up to the Keccak rate
- added assertion `ptr_max_bits >= 24` to Sha2 and Keccak

closes INT-8048, INT-8049, INT-8051

### DIFF
--- a/guest-libs/keccak256/src/host_impl.rs
+++ b/guest-libs/keccak256/src/host_impl.rs
@@ -26,11 +26,12 @@ impl Keccak256 {
     ///
     /// # Safety
     ///
-    /// `output` must be at least `KECCAK_OUTPUT_SIZE` (32) bytes long.
+    /// `output` must be exactly `KECCAK_OUTPUT_SIZE` (32) bytes long.
     pub unsafe fn finalize(self, output: &mut [u8]) {
-        debug_assert!(
-            output.len() >= super::KECCAK_OUTPUT_SIZE,
-            "output buffer too small"
+        debug_assert_eq!(
+            output.len(),
+            KECCAK_OUTPUT_SIZE,
+            "output must be exactly 32 bytes"
         );
         self.inner.finalize(output);
     }
--- a/guest-libs/keccak256/src/lib.rs
+++ b/guest-libs/keccak256/src/lib.rs
@@ -22,11 +22,12 @@ impl tiny_keccak::Hasher for Keccak256 {
 
     #[inline(always)]
     fn finalize(self, output: &mut [u8]) {
-        assert!(
-            output.len() >= KECCAK_OUTPUT_SIZE,
-            "output buffer too small"
+        assert_eq!(
+            output.len(),
+            KECCAK_OUTPUT_SIZE,
+            "output must be exactly 32 bytes"
         );
-        // SAFETY: output is at least KECCAK_OUTPUT_SIZE bytes (checked above).
+        // SAFETY: output is exactly KECCAK_OUTPUT_SIZE bytes (checked above).
         unsafe { Keccak256::finalize(self, output) };
     }
 }
--- a/guest-libs/keccak256/src/zkvm_impl.rs
+++ b/guest-libs/keccak256/src/zkvm_impl.rs
@@ -102,14 +102,15 @@ impl Keccak256 {
     ///
     /// # Safety
     ///
-    /// `output` must be at least `KECCAK_OUTPUT_SIZE` (32) bytes long.
+    /// `output` must be exactly `KECCAK_OUTPUT_SIZE` (32) bytes long.
     #[inline(always)]
     pub unsafe fn finalize(mut self, output: &mut [u8]) {
-        debug_assert!(
-            output.len() >= KECCAK_OUTPUT_SIZE,
-            "output buffer too small"
+        debug_assert_eq!(
+            output.len(),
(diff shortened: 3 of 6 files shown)


guest-libs/keccak256/src/lib.rs
```diff
@@ -23,10 +23,10 @@ impl tiny_keccak::Hasher for Keccak256 {
     #[inline(always)]
     fn finalize(self, output: &mut [u8]) {
         assert!(
```
shuklaayush: `assert_eq`?

guest-libs/keccak256/src/zkvm_impl.rs
```diff
@@ -102,14 +102,14 @@ impl Keccak256 {
     ///
     /// # Safety
     ///
-    /// `output` must be at least `KECCAK_OUTPUT_SIZE` (32) bytes long.
+    /// `output` must be exactly `KECCAK_OUTPUT_SIZE` (32) bytes long.
     #[inline(always)]
     pub unsafe fn finalize(mut self, output: &mut [u8]) {
         debug_assert!(
```
shuklaayush: `debug_assert_eq`?

guest-libs/keccak256/src/host_impl.rs
```diff
@@ -26,11 +26,11 @@ impl Keccak256 {
     ///
     /// # Safety
     ///
-    /// `output` must be at least `KECCAK_OUTPUT_SIZE` (32) bytes long.
+    /// `output` must be exactly `KECCAK_OUTPUT_SIZE` (32) bytes long.
     pub unsafe fn finalize(self, output: &mut [u8]) {
         debug_assert!(
```
shuklaayush: `debug_assert_eq`

crates/sdk/src/keygen/mod.rs
```diff
@@ -59,16 +61,16 @@ where
...
+            let system_config = config.app_vm_config.as_ref();
+            let mut keygen_builder = MultiStarkKeygenBuilder::new(app_engine.config().clone());
+            for (air_id, air) in config.app_vm_config.create_airs()?.into_airs().enumerate() {
+                if system_config.is_required_air_id(air_id) {
+                    keygen_builder.add_required_air(air as AirRef<_>);
+                } else {
+                    keygen_builder.add_air(air as AirRef<_>);
+                }
+            }
+            let vm_pk = keygen_builder.generate_pk().unwrap();
```
shuklaayush: maybe we should move this to `AirInventory::keygen` and reuse also in vm
```rust
impl<SC: StarkProtocolConfig> AirInventory<SC> {
    /// Consumes the inventory and generates the proving key, marking system AIRs that
    /// must be present in every valid proof (see [SystemConfig::is_required_air_id])
    /// as required.
    pub fn keygen(self, config: SC) -> MultiStarkProvingKey<SC> {
        let system_config = self.config.clone();
        let mut builder = MultiStarkKeygenBuilder::new(config);
        for (air_id, air) in self.into_airs().enumerate() {
            if system_config.is_required_air_id(air_id) {
                builder.add_required_air(air as AirRef<SC>);
            } else {
                builder.add_air(air as AirRef<SC>);
            }
        }
        builder.generate_pk().unwrap()
    }
}
```

guest-libs/keccak256/src/zkvm_impl.rs
shuklaayush, line 99: can you check if we can squeeze out more bytes instead of limiting to 32 bytes without causing any soundness issues? something like

```rust
/// Finalizes the hash and squeezes `len` bytes of output.
///
/// # Safety
///
/// `output` must be valid for writes of `len` bytes.
unsafe fn finalize_ptr(&mut self, output: *mut u8, len: usize) {
    // Keccak padding (pad10*1)
    self.state.0[self.idx] ^= 0x01;
    self.state.0[KECCAK_RATE - 1] ^= 0x80;
    // Sponge squeeze
    let mut written = 0;
    while written < len {
        self.keccakf();
        let n = min(len - written, KECCAK_RATE);
        core::ptr::copy_nonoverlapping(self.state.0.as_ptr(), output.add(written), n);
        written += n;
    }
}

...

pub fn finalize(mut self, output: &mut [u8]) {
    // SAFETY: output is valid for output.len() bytes.
    unsafe { self.finalize_ptr(output.as_mut_ptr(), output.len()) }
}

pub unsafe extern "C" fn native_keccak256(bytes: *const u8, len: usize, output: *mut u8) {
    let mut hasher = KECCAK256_HASHER.lock();
    hasher.update_ptr(bytes, len);
    hasher.finalize_ptr(output, KECCAK_OUTPUT_SIZE); // C ABI contract: exactly 32
    hasher.reset();
}
```

shuklaayush (commented):
some comments. also change the base branch to rc2 and also let's split the sdk and keccak changes into separate prs for easier external review

guest-libs/keccak256/src/zkvm_impl.rs
gdmlcjs: This change is going to be made to the tiny-keccak patch, but for our guest library it seems safer to only allow 32 bytes and follow convention.

shuklaayush (approved):

PR #2848, opened 2026-06-05
fix(static-verifier): require reduced BabyBear transcript inputs

zlangley:
Introduce reduced wire for BabyBear values absorbed by the transcript and hash.

### DIFF
--- a/crates/static-verifier/src/field/baby_bear/base.rs
+++ b/crates/static-verifier/src/field/baby_bear/base.rs
@@ -17,6 +17,7 @@ use openvm_stark_sdk::{
     p3_baby_bear::BabyBear,
 };
 
+use super::BABY_BEAR_MODULUS_U64;
 use crate::utils::{guarded_debug_assert, guarded_debug_assert_eq};
 
 pub(crate) const BABYBEAR_MAX_BITS: usize = 31;
@@ -37,6 +38,34 @@ pub struct BabyBearWire {
     pub value: AssignedValue<Fr>,
     /// The value is guaranteed to be less than 2^max_bits.
     pub max_bits: usize,
+}
+
+/// A BabyBear wire constrained to its canonical representative in `[0, p)`.
+///
+/// This type marks values that are safe to absorb into BabyBear-domain transcript
+/// and hash inputs. Arithmetic may still use the underlying `BabyBearWire`; converting
+/// via `BabyBearWire::from` only drops this type-level evidence and does not add or
+/// remove constraints.
+#[derive(Copy, Clone, Debug)]
+pub struct ReducedBabyBearWire(BabyBearWire);
+
+impl ReducedBabyBearWire {
+    pub fn value(&self) -> AssignedValue<Fr> {
+        self.0.value
+    }
+}
+
+impl From<ReducedBabyBearWire> for BabyBearWire {
+    /// Drops the canonicality evidence and returns the underlying arithmetic wire.
+    fn from(wire: ReducedBabyBearWire) -> Self {
+        wire.0
+    }
+}
+
+impl From<&ReducedBabyBearWire> for BabyBearWire {
+    fn from(wire: &ReducedBabyBearWire) -> Self {
+        (*wire).into()
+    }
 }
 
 impl BabyBearWire {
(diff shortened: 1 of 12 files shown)


crates/static-verifier/src/field/baby_bear/base.rs
```diff
@@ -39,6 +40,26 @@ pub struct BabyBearWire {
...
+/// This type marks values that are safe to absorb into BabyBear-domain transcript
+/// and hash inputs. Arithmetic may still use the underlying `BabyBearWire`; calling
+/// `forget_reduced` only drops this type-level evidence and does not add or
+/// remove constraints.
+#[derive(Copy, Clone, Debug)]
+pub struct ReducedBabyBearWire(BabyBearWire);
+
+impl ReducedBabyBearWire {
+    /// Drops the canonicality evidence and returns the underlying arithmetic wire.
+    pub fn forget_reduced(self) -> BabyBearWire {
```
jonathanpwang: slightly more canonical to impl `From`

crates/static-verifier/src/field/baby_bear/base.rs
```diff
@@ -104,6 +146,16 @@ impl BabyBearChip {
         wire
     }
 
+    /// Loads a canonical BabyBear constant and returns it with reduced type evidence.
+    pub fn load_reduced_constant(
```
jonathanpwang, line 158: technically any constant is reduced, but maybe it's too annoying to make `load_constant` always return `ReducedBabyBearWire`?

jonathanpwang: not sure how often it'd come up, but it's a bit better if `load_constant` returned `ReducedBabyBearWire`, and then we called `.into()` when needed

jonathanpwang: if the code diff is too large, we can keep the current

crates/static-verifier/src/stages/batch_constraints/mod.rs
```diff
@@ -853,18 +877,43 @@ pub(crate) fn constrain_batch_constraints_verification(
...
+            assert!(
+                openings.len().is_multiple_of(2),
+                "rotated opening vector must be even",
+            );
+            for claim in openings.chunks_exact(2) {
+                transcript.observe_ext(ctx, &claim[0]);
+                transcript.observe_ext(ctx, &claim[1]);
+            }
+        } else {
+            let zero = ext_chip.load_reduced_constant(ctx, RootEF::ZERO);
```
jonathanpwang: small optimization: since zero is special, we should `load_zero` once outside the loop and re-use it.

crates/static-verifier/src/stages/batch_constraints/mod.rs
```diff
@@ -853,18 +877,43 @@ pub(crate) fn constrain_batch_constraints_verification(
...
+                assert!(
+                    claims.len().is_multiple_of(2),
+                    "rotated opening vector must be even",
+                );
+                for claim in claims.chunks_exact(2) {
+                    transcript.observe_ext(ctx, &claim[0]);
+                    transcript.observe_ext(ctx, &claim[1]);
+                }
+            } else {
+                let zero = ext_chip.load_reduced_constant(ctx, RootEF::ZERO);
```
jonathanpwang: ditto. (note also `load_zero` saves a little compared to `load_constant`)

jonathanpwang (commented):
can you provide a profiling comparison on cells before and after this change?

otherwise lgtm: the main comment is it's preferred to impl `From` and then you can also use `impl Into<BabyBearWire>` to potentially clean up some code.

crates/static-verifier/src/field/baby_bear/base.rs
zlangley: Yeah, making `load_constant` return `ReducedBabyBearWire` required adding `forget_reduced()` or `into()` in many places.

crates/static-verifier/src/stages/batch_constraints/mod.rs
zlangley: `BabyBearChip.load_constant` calls `ctx.load_zero` if it's zero.

crates/static-verifier/src/field/baby_bear/base.rs
zlangley: I had this in an earlier revision but wasn't sure I liked it. Will add it back.

shuklaayush:
fyi the base here should be `develop-v2.0.0-rc.2`

zlangley:
> can you provide a profiling comparison on cells before and after this change?

```
before: 25,692,295
after: 25,962,573
delta: +270,278
```

All from `load_proof_wire`.

jonathanpwang (approved):

PR #2855, opened 2026-06-07
fix: preserve AUIPC AOT immediates

### LINKED ISSUE #2851: RV32IM AUIPC AOT truncates transpiled immediates
## Summary

RV32IM AUIPC AOT truncates the transpiled AUIPC immediate before computing the destination value.

The interpreter/circuit path uses:

```text
rd = pc + (imm << RV32_CELL_BITS)
```

The AOT path converts the immediate through `i16`, losing valid upper bits.

## Affected Version

```text
openvm-org/openvm main
```

## Reproduction

Regression test:

```bash
cargo test -p openvm-rv32im-circuit --features aot aot_auipc_preserves_full_24_bit_transpiled_immediate -- --nocapture
```

x86_64 Actions run:

```text
```

Case:

```text
pc = 0
imm = 0x10000
expected rd = 16777216
```

## Actual Behavior

AOT emits:

```text
mov edi, 0
pinsrd xmm0, edi, 1
```

The test fails because AOT materializes `0` instead of `16777216`.

## Expected Behavior

AOT should preserve the same immediate width and semantics as the interpreter/circuit path.

## Source References

```text
extensions/rv32im/circuit/src/auipc/execution.rs
```

The AOT path narrows through:

```text
c_i24 as i16
```

## Impact

Valid AUIPC instructions can write different register values under AOT than under the interpreter/circuit path.

## Fix Direction

Remove the `i16` narrowing and use the same immediate interpretation as the non-AOT execution path.

1sgtpepper:
Fixes #2851.

## Summary

RV32IM AUIPC AOT now uses the raw 24-bit transpiled immediate when computing `pc + (imm << 8)`, instead of narrowing the immediate through `i16`.

## Testing

- `OPENVM_FAST_TEST=1 OPENVM_SKIP_DEBUG=1 cargo test -p openvm-rv32im-circuit --features aot aot_uses_full_24_bit_auipc_immediate`
- `cargo clippy -p openvm-rv32im-circuit --features aot --all-targets -- -D warnings`


docs/vocs/docs/pages/specs/openvm/isa.mdx
```diff
@@ -373,7 +373,7 @@ For branch instructions, we fix `d = e = 1`. For jump instructions, we fix `d =
...
-| AUIPC_RV32 | `a,_,c,1,_,_`    | `[a:4]_1 = decompose(pc) + (decompose(c) << 8)`. Here `i32(c)` must be in `[0, 2^24)`.                                                                                                                                                                                                                                                                                     |
+| AUIPC_RV32 | `a,_,c,1,_,_`    | `[a:4]_1 = decompose(pc) + (decompose(c) << 8)`. Here `c` is the raw 24-bit transpiled immediate and must be in `[0, 2^24)`.                                                                                                                                                                                                                                                |
```
GunaDD: Let's keep the previous description of the openvm isa.

GunaDD (commented):
I agree with you that the issue is valid and the fix is correct. Thank you for this! Could you undo the change in the `isa.mdx` file? Other than that, LGTM.

shuklaayush (approved):

PR #2856, opened 2026-06-07
fix: reject dead AOT PC slots

### LINKED ISSUE #2852: RV32IM AOT dispatch does not reject dead PC slots
## Summary

RV32IM AOT can execute a PC slot that the interpreter treats as unreachable when a `Program` contains an explicit `None` instruction slot.

`Program::new_without_debug_infos_with_option` can represent dead PC slots. The interpreter installs an unreachable handler for them, but AOT builds dispatch state from `program.enumerate_by_pc()`, which drops `None` slots.

## Affected Version

```text
openvm-org/openvm main
```

## Reproduction

Regression test:

```bash
cargo test -p openvm-rv32im-circuit --features aot aot_dispatch_rejects_dead_pc_slot -- --nocapture
```

x86_64 Actions run:

```text
```

The test starts execution at a PC whose instruction slot is `None`.

## Actual Behavior

The interpreter rejects the dead PC slot.

AOT does not reject the same PC. The Actions run fails with:

```text
AOT must reject execution at the same dead PC slot
```

## Expected Behavior

AOT should match interpreter behavior for dead PC slots: either install unreachable dispatch entries or reject sparse programs before AOT generation.

## Source References

```text
crates/vm/src/arch/interpreter.rs
crates/vm/src/arch/aot/pure.rs
crates/vm/src/arch/aot/metered_execute.rs
crates/toolchain/instructions/src/program.rs
```

## Impact

The same `Program` has different execution behavior under interpreter and AOT backends.

## Fix Direction

Preserve `None` slots when building the AOT dispatch table, or reject sparse programs before AOT compilation.

1sgtpepper:
Fixes #2852.

## Summary

AOT dispatch now emits fallback handlers for padded PC slots before `pc_base` and for explicit `None` program slots. This keeps pure and metered AOT behavior aligned with the interpreter for unreachable PCs instead of dispatching through a dead table slot.

## Testing

- `OPENVM_FAST_TEST=1 OPENVM_SKIP_DEBUG=1 cargo test -p openvm-rv32im-circuit --features aot test_aot_dispatch_rejects_dead_pc_slots`
- `cargo clippy -p openvm-circuit -p openvm-rv32im-circuit --features aot --all-targets -- -D warnings`

### DIFF
--- a/crates/vm/src/arch/aot/metered_execute.rs
+++ b/crates/vm/src/arch/aot/metered_execute.rs
@@ -1,6 +1,6 @@
 use std::{ffi::c_void, mem::offset_of};
 
-use openvm_instructions::exe::VmExe;
+use openvm_instructions::{exe::VmExe, program::DEFAULT_PC_STEP};
 use openvm_stark_backend::p3_field::PrimeField32;
 
 use super::{common::*, AotInstance};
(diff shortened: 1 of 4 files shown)


GunaDD:
LGTM, thanks for this!

PR #2859, opened 2026-06-08
feat(v2-rc.2): emit metered execution metrics

shuklaayush:
- Adds metered segment and per-AIR metrics for rows, cells, memory, padding, and interactions.
- Emits opcode execution counts from preflight execution frequencies, labeled by AIR and opcode.
- Keeps exact opcode trace-cell attribution behind `perf-metrics`.
- Defines interaction cells as `rows * interactions`; emits interaction memory overhead separately because it is added once per segment.

### DIFF
--- a/crates/vm/src/arch/interpreter_preflight.rs
+++ b/crates/vm/src/arch/interpreter_preflight.rs
@@ -199,22 +199,27 @@ impl<F: PrimeField32, E> PreflightInterpretedInstance<F, E> {
         unsafe {
             *self.execution_frequencies.get_unchecked_mut(pc_idx) += 1;
         };
-        // SAFETY: the `executor_idx` comes from ExecutorInventory, which ensures that
-        // `executor_idx` is within bounds
+        tracing::trace!("pc: {pc:#x} | {:?}", pc_entry.insn);
+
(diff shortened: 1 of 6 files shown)


crates/vm/src/arch/interpreter_preflight.rs
```diff
@@ -119,6 +121,33 @@ impl<F: Field, E> PreflightInterpretedInstance<F, E> {
 }
 
 impl<F: PrimeField32, E> PreflightInterpretedInstance<F, E> {
+    #[cfg(feature = "metrics")]
+    pub fn opcode_counts_by_air<RA>(&self) -> BTreeMap<(usize, String), u64>
```
jonathanpwang, line 125: I guess this is not too heavy since it's called just once

jonathanpwang (approved):
lgtm

crates/vm/src/arch/interpreter_preflight.rs
shuklaayush: yes, once per segment. the heavy lifting is done by `execution_frequencies` which is already accumulated to fill the `ProgramAir` trace

PR #2860, opened 2026-06-08
fix(static-verifier): bound signed_div_mod quotient for negative dividends

zlangley:
`signed_div_mod` range-checked `div = floor(a / b)` against `bound = floor(2^a_num_bits / b)`, justified by `|a // b| = |a| // b`. That identity is false for negative `a` with a nonzero remainder, where `|floor(a / b)| = ceil(|a| / b)` can exceed the bound, making `shifted_div = div + bound` negative and the range check unsatisfiable (e.g. reducing `neg(one)`, which carries `value == -1`, `max_bits == 1`).

Compute `bound = ceil((2^a_num_bits - 1) / b)` so the symmetric shift covers both signs, and correct the proof-of-correctness comment.

### DIFF
--- a/crates/static-verifier/src/field/baby_bear/base.rs
+++ b/crates/static-verifier/src/field/baby_bear/base.rs
@@ -6,7 +6,7 @@ use halo2_base::{
         arithmetic::Field as _,
         halo2curves::{bn256::Fr, ff::PrimeField as _},
     },
-    utils::{bigint_to_fe, biguint_to_fe, bit_length, fe_to_bigint, BigPrimeField},
+    utils::{bigint_to_fe, biguint_to_fe, bit_length, fe_to_bigint, modulus, BigPrimeField},
     AssignedValue, Context, QuantumCell,
 };
 use itertools::Itertools;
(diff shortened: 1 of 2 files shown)


crates/static-verifier/src/field/baby_bear/base.rs
```diff
@@ -423,8 +423,9 @@ impl BabyBearChip {
 /// * a_num_bits: number of bits needed to represent the absolute value of `a`
 ///
 /// ## Assumptions
-/// * `a_max_bits < F::CAPACITY = F::NUM_BITS - RESERVED_HIGH_BITS`
-///   * Unsafe behavior if `a_max_bits >= F::CAPACITY`
+/// * `a_num_bits <= F::CAPACITY - RESERVED_HIGH_BITS`
+///   * This provides the headroom for the wrap-free bound asserted in the body (`(4 * bound + 2) *
```
jonathanpwang: mention that the first bullet point implies the second one specifically for `Fr` and babybear. <br>for generic `F` it involves something like `2^CAPACITY + 6 BabyBear::ORDER_U32 - 8 <= p`

jonathanpwang (approved):
LGTM except the comment

PR #2861, opened 2026-06-08
feat(v2-rc.2): use backend segment memory metering

shuklaayush:
follow up to https://github.com/openvm-org/stark-backend/pull/352

resolves int-8130


jonathanpwang (approved):
less code, good.

PR #2862, opened 2026-06-09
fix: constrain canonical deferral hook location in main agg tree

stephenh-axiom-xyz:
Resolves INT-8115.

## Overview

This PR makes each deferral proof carry and prove its canonical position in the deferral memory subtree.

The first commit adds `def_idx` public values to the deferral-circuit aggregation tree and constrains them to stay consistent while proofs for the same deferral circuit are aggregated. The second commit adds `node_idx` to the main recursive deferral public values, derives it from `def_idx` at the hook layer, updates it during main deferral aggregation, and requires the final/root deferral proof to be at `node_idx = 0`.

The effective supported deferral circuit index range is now `0..512`.

## Continuations: deferral aggregation PVS

Files:

- `crates/continuations/src/circuit/deferral/mod.rs`
- `crates/continuations/src/circuit/deferral/inner/def_pvs/{air,trace}.rs`

### Public values

`DeferralCircuitPvs` now includes:

```rust
pub def_idx: F
```

`DeferralAggregationPvs` now includes:

```rust
pub def_idx: F
```

This index identifies which deferral circuit the subtree belongs to.

### `DeferralAggPvsAir`

The AIR now receives each child's `def_idx` from `PublicValuesBus`:

```text
leaf child:
  DeferralCircuitPvs.air[DEF_CIRCUIT_PVS_AIR_ID]
    pv[2 * DIGEST_SIZE] = child def_idx

internal child:
  DeferralAggregationPvs.air[DEF_AGG_PVS_AIR_ID]
    pv[DIGEST_SIZE + 2] = child def_idx
```

It constrains every present row's child `def_idx` to equal the produced aggregation `def_idx`:

```text
present child def_idx == parent DeferralAggregationPvs.def_idx
```

Example rows:

| case | row | `is_present` | `has_verifier_pvs` | child source | child `def_idx` | output `def_idx` |
|---|---:|---:|---:|---|---:|---:|
| leaf wrapper | 0 | 1 | 0 | `DeferralCircuitPvs` | `k` | `k` |
| two leaf proofs | 0 | 1 | 0 | `DeferralCircuitPvs` | `k` | `k` |
| two leaf proofs | 1 | 1 | 0 | `DeferralCircuitPvs` | `k` | `k` |
| leaf + padding | 0 | 1 | 0 | `DeferralCircuitPvs` | `k` | `k` |
| leaf + padding | 1 | 0 | 0 | padding | unconstrained | `k` |
| internal wrapper | 0 | 1 | 1 | `DeferralAggregationPvs` | `k` | `k` |

Bus diagram:

```text
DeferralCircuit proof PVS
  └─ PublicValuesBus: input_commit, output_commit, def_idx
       └─ DeferralAggPvsAir

DeferralAggregation proof PVS
  └─ PublicValuesBus: merkle_commit, num proofs, merkle_depth, def_idx
       └─ DeferralAggPvsAir

DeferralAggPvsAir
  └─ public values: merkle_commit, num_def_circuit_proofs, merkle_depth, def_idx
```

## Verify-stark deferral circuit

Files:

- `guest-libs/verify-stark/circuit/src/{lib,trace}.rs`
- `guest-libs/verify-stark/circuit/src/verifier/{air,trace}.rs`
- `guest-libs/verify-stark/circuit/src/prover/trace.rs`

The deferred verify circuit now stores a configured `def_idx`. `DeferredVerifyPvsAir` constrains:

```text
DeferralCircuitPvs.def_idx == configured def_idx
```

The trace generator writes that same constant into `DeferralCircuitPvs`.

This is the base constraint that prevents a deferral proof from claiming another circuit index before entering the deferral aggregation tree.

## Continuations: main deferral PVS `node_idx`

Files:

- `crates/verify/src/pvs.rs`
- `crates/continuations/src/circuit/deferral/hook/verifier/{air,trace}.rs`
- `crates/continuations/src/circuit/inner/def_pvs/{air,trace}.rs`
- `crates/continuations/src/circuit/inner/{mod,trace}.rs`
- `crates/continuations/src/prover/inner/trace.rs`
- `crates/continuations/src/circuit/mod.rs`

### Public values

`DeferralPvs` now includes:

```rust
pub node_idx: F
```

This is the node index of the deferral accumulator subtree at the current `depth`.

### Hook layer

`DeferralHookPvsAir` constrains:

```text
depth == 1
node_idx == DeferralAggregationPvs.def_idx
```

So each hook proof starts as the depth-1 subtree for its deferral circuit index.

### Main inner `DeferralPvsAir`

The AIR now constrains how `node_idx` behaves while deferral hook proofs are recursively aggregated.

One-row case:

```text
parent PVS == child PVS
node_idx == 0
```

This means one-row wrapping is only valid once the subtree is already rooted at index 0.

Two-row case:

```text
child depths are equal
parent depth = child depth + 1
if both children are present:
  right_child.node_idx = left_child.node_idx + 1
parent.node_idx = (local_child.node_idx - single_present_is_right) / 2
range_check(parent.node_idx, 8 bits)
```

Example rows:

| case | row | `is_present` | `single_present_is_right` | child `node_idx` | parent `node_idx` |
|---|---:|---:|---:|---:|---:|
| root wrapper | 0 | 1 | 0 | 0 | 0 |
| left + right present | 0 | 1 | 0 | 2n | n |
| left + right present | 1 | 1 | 0 | 2n + 1 | n |
| present left, absent right | 0 | 1 | 0 | 2n | n |
| present left, absent right | 1 | 0 | 0 | unconstrained hash subtree | n |
| absent left, present right | 0 | 1 | 1 | 2n + 1 | n |
| absent left, present right | 1 | 0 | 1 | absent subtree | n |

Range-check bus interaction:

```text
DeferralPvsAir
  └─ RangeCheckerBus: value = parent node_idx, max_bits = 8
       └─ VerifierSubCircuit range checker trace
```

The trace plumbing now carries these range-check inputs through `SubCircuitTraceData` into `VerifierExternalData`.

## Root and host verification

Files:

- `crates/continuations/src/circuit/root/verifier/air.rs`
- `guest-libs/verify-stark/circuit/src/verifier/air.rs`
- `crates/verify/src/{lib,error}.rs`

Both root verifier AIRs now require:

```text
final DeferralPvs.node_idx == 0
```

The host verifier also rejects final proofs with nonzero `node_idx`:

```rust
VerifyStarkError::DefNodeIdxNonZero { actual }
```

This matches the current deferral Merkle path logic, which verifies a deferral accumulator subtree rooted at the canonical leftmost position for `(DEFERRAL_AS, 0)`.

## SDK and guest limits

Files:

- `extensions/deferral/guest/src/lib.rs`

The SDK now initializes absent per-circuit deferral PVS with:

```text
node_idx = def_idx
depth = 1
```

When two absent subtrees are combined, it updates:

```text
parent.node_idx = left.node_idx / 2
```

The guest/transpiler-visible maximum deferral circuit count is reduced:

```text
MAX_DEF_CIRCUITS: 1024 -> 512
```

This matches the main deferral aggregation range check: parent `node_idx` is 8-bit, so depth-1 child `def_idx` must be below 512.

## Review checklist

- Check that `def_idx` is constrained at every deferral circuit source, especially custom implementations of `DeferralCircuitProver`.
- Check that `DeferralAggPvsAir` only aggregates proofs with the same `def_idx`.
- Check that `DeferralHookPvsAir` correctly maps `def_idx` into depth-1 `node_idx`.
- Check that `DeferralPvsAir` enforces canonical parent indices and rejects non-root one-row wrapping.
- Check that the range-check lookup for parent `node_idx` is backed by trace inputs in both CPU and CUDA trace paths.
- Check that root and host verification consistently require final `node_idx == 0`.

### DIFF
--- a/crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
+++ b/crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
@@ -158,6 +158,17 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
             );
         }
 
+        self.public_values_bus.receive(
+            builder,
+            local.proof_idx,
+            PublicValuesBusMessage {
+                air_idx,
+                pv_idx: AB::Expr::from_usize(2 * DIGEST_SIZE),
+                value: local.child_pvs.def_idx.into(),
+            },
+            is_present_leaf.clone(),
+        );
+
         self.input_or_merkle_commit_bus.receive(
             builder,
             local.proof_idx,
@@ -213,9 +224,9 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
         );
 
         /*
-         * On internal layers we receive num_def_circuit_proofs and merkle_depth from
+         * On internal layers we receive num_def_circuit_proofs, merkle_depth, and def_idx from
          * PublicValuesBus. To save columns, we repurpose the first two elements of the unused
-         * local.child_pvs.input_commit.
+         * local.child_pvs.input_commit for num_def_circuit_proofs and merkle_depth.
          */
         let local_num_def_circuit_proofs = local.child_pvs.input_commit[0];
         let local_merkle_depth = local.child_pvs.input_commit[1];
@@ -246,16 +257,29 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
             is_internal * local.is_present,
         );
 
+        self.public_values_bus.receive(
+            builder,
+            local.proof_idx,
+            PublicValuesBusMessage {
+                air_idx: AB::Expr::from_usize(DEF_AGG_PVS_AIR_ID),
+                pv_idx: AB::Expr::from_usize(DIGEST_SIZE + 2),
+                value: local.child_pvs.def_idx.into(),
+            },
+            is_internal * local.is_present,
+        );
+
         /*
          * If there is only one row, then this proof is a wrapper and its public values should
          * match those of its child's. Otherwise, constrain that num_def_circuit_proofs is the
          * sum of the present children's and that merkle_depth is the child merkle_depth + 1.
          * If both children are present, we constrain that their merkle_depth values are equal.
+         * Every present child's def_idx should match this proof's.
          */
         let &DeferralAggregationPvs::<_> {
             merkle_commit,
             num_def_circuit_proofs,
             merkle_depth,
+            def_idx,
         } = builder.public_values().borrow();
 
         let is_first = not(local.proof_idx);
@@ -283,6 +307,9 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
             .when_transition()
             .when(next.is_present)
             .assert_eq(local_merkle_depth, next_merkle_depth);
+        builder
+            .when(local.is_present)
+            .assert_eq(local.child_pvs.def_idx, def_idx);
 
         /*
          * MAX_DEF_AGG_MERKLE_DEPTH - merkle_depth is range checked to be in [0, 256). This,
--- a/crates/continuations/src/circuit/inner/def_pvs/air.rs
+++ b/crates/continuations/src/circuit/inner/def_pvs/air.rs
@@ -19,7 +19,7 @@ use openvm_verify_stark_host::pvs::{
     DeferralPvs, CONSTRAINT_EVAL_AIR_ID, CONSTRAINT_EVAL_CACHED_INDEX, DEF_PVS_AIR_ID,
 };
 use p3_air::{Air, AirBuilder, AirBuilderWithPublicValues, BaseAir};
-use p3_field::{Field, PrimeCharacteristicRing};
+use p3_field::PrimeCharacteristicRing;
 use p3_matrix::Matrix;
 
 use crate::{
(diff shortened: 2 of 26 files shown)


crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
```diff
@@ -158,6 +158,17 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB>
             );
         }
 
+        self.public_values_bus.receive(
+            builder,
+            local.proof_idx,
+            PublicValuesBusMessage {
+                air_idx,
+                pv_idx: AB::Expr::from_usize(2 * DIGEST_SIZE),
```
shuklaayush, line 166: i wonder if there's a better way to tie these offsets to the public values structs. plonky3 used to have this macro that generates an index mapping so you could do `DeferralAggregationPvs_indices.def_idx`

crates/continuations/src/circuit/inner/def_pvs/air.rs
```diff
@@ -274,11 +285,31 @@ impl<AB: AirBuilder + InteractionBuilder + AirBuilderWithPublicValues> Air<AB> f
...
+            is_first_of_two_rows,
         );
 
         builder
             .when(has_two_rows)
             .assert_one(depth.into() - local.child_pvs.depth);
+
+        let canonical_parent_node_idx =
+            (local.child_pvs.node_idx - local.single_present_is_right) * AB::F::TWO.inverse();
+
```
shuklaayush, line 297: i would write it as this instead since multiplication by inv(2) is ambiguous if left part is odd
```
child = 2 * parent + side_bit
```

shuklaayush (approved):
looks fine

crates/continuations/src/circuit/deferral/inner/def_pvs/air.rs
stephenh-axiom-xyz: I found that you can just do all of these as a slice later on, but am leaving for blast radius

PR #2863, opened 2026-06-09
fix: non-root def hook proofs should always have siblings

stephenh-axiom-xyz:
Resolves INT-8157.

## Overview

This PR fixes SDK deferral aggregation when the configured deferral circuits do not naturally form a complete binary tree. `DeferralProver::prove` now pads per-circuit deferral outputs to a power-of-two length with absent deferral proofs, and `AggProver::prove_def` now requires power-of-two input so every non-singleton aggregation round has explicit sibling data.

The main behavioral change is that non-root deferral hook proofs are no longer allowed to advance through an odd aggregation round without a sibling. Missing siblings are represented by absent `DeferralPvs`, preserving the expected accumulator tree shape.

### SDK deferral proof construction

Files:

Changes:

- Adds `absent_deferral_pvs`, a shared helper for constructing absent deferral public values.
- Keeps absent proofs for configured-but-unused deferral circuits tied to that circuit's actual commit.
- Pads the returned per-circuit proof vector to `next_power_of_two()` with absent proofs for synthetic padding indices.
- Uses a zero digest as the circuit commit for padding entries beyond the configured deferral circuits.

Review focus:

- Confirm the absent-proof accumulator hashes match the expected deferral tree semantics.
- Confirm padding entries are only used to complete the aggregation tree and do not get exposed as real configured deferral circuits.

### SDK deferral aggregation

Files:

Changes:

- `AggProver::prove_def` now asserts that deferral aggregation input is non-empty and power-of-two sized.
- `reduce_def_round` now requires every non-singleton round to have an even number of proofs.
- A singleton round is handled only by wrapping a single present proof.
- The previous trailing-singleton behavior was removed, so absent proofs cannot be silently passed through and present proofs cannot be aggregated at a non-root position without sibling context.

Review focus:

- Confirm the new assertions match all callers' expected input shape.
- Confirm present/absent pair handling still passes the absent sibling PVS with the correct `is_right` value.
- Confirm absent/absent folding still updates hashes, depth, and node index consistently.

### Regression coverage

Files:

- `crates/sdk/src/tests.rs`
- `crates/sdk/programs/examples/verify-many.rs`
- `crates/sdk/programs/examples/verify-many.elf`

Changes:

- Adds helpers for constructing deferral-enabled SDKs with multiple verify-stark deferral circuits.
- Adds `verify-many`, a guest program that calls `verify_stark` for deferral indices `0`, `1`, `3`, and `4`.
- Adds `test_verify_many_deferrals`, covering five configured deferral circuits where one configured circuit is absent and the prover must also pad to a power-of-two tree.
- The test verifies the resulting proof carries deferral Merkle proofs and passes `Sdk::verify_proof`.

Review focus:

- Confirm the chosen present indices exercise both configured absent siblings and synthetic padding.
- Confirm the committed `verify-many.elf` matches the new guest source and test expectations.

## Notes

- All changes are contained in `openvm-sdk` and its SDK example programs.
- This branch is one commit on top of `develop-v2.0.0-rc.2`: `9b49e342f fix: non-root def hook proofs should always have siblings`.


crates/sdk/src/prover/agg.rs
```diff
@@ -372,74 +374,71 @@ impl AggProver {
...
-                            node_idx: pvs0.node_idx.halve(),
-                        })
-                    }
-                };
-                next.push(combined);
+    if proofs.len() == 1 {
+        // A singleton round can only happen when the entire round has one present input proof.
+        let DeferralProof::Present(p) = proofs.pop().unwrap() else {
+            panic!("singleton deferral round must contain a present proof");
+        };
```
shuklaayush, line 385: does this mean that there should always be a deferral proof when deferrals are enabled?

crates/sdk/src/prover/agg.rs
```diff
@@ -372,74 +374,71 @@ impl AggProver {
...
+                DeferralProof::Present(prover.agg_prove::<E>(
+                    &[p],
+                    kind,
+                    ProofsType::Deferral,
+                    Some((pvs, true)),
+                )?)
             }
-        }
+            (DeferralProof::Absent(pvs0), DeferralProof::Absent(pvs1)) => {
+                debug_assert_eq!(pvs0.depth, pvs1.depth);
```
shuklaayush, line 424: should we check `pvs1.node_idx == pvs0.node_idx + 1`?

crates/sdk/src/prover/agg.rs
```diff
@@ -372,74 +374,71 @@ impl AggProver {
 }
 
 fn reduce_def_round<const N: usize>(
-    proofs: Vec<DeferralProof>,
+    mut proofs: Vec<DeferralProof>,
```
shuklaayush: this doesn't seem necessary. just avoid the `pop` below

shuklaayush (approved):
changes seem fine but i wonder if we should add more tests for other scenarios like all absent, power of two counts etc.

crates/sdk/src/prover/agg.rs
stephenh-axiom-xyz: No, it means that if a deferral proof can only have no sibling if it is being wrapped (i.e. to get to internal-recursive layer or smth). There is a test to show that you don't have to use deferral when deferrals are enabled

PR #2868, opened 2026-06-10
fix(static-verifier): phase3 nits

zlangley:
This tightens several static verifier invariants and trims redundant Halo2 cells in the verifier circuit.

- Require `SafeBool` for BabyBear field select APIs
- Add explicit shape/width checks for BabyBear reduction, base-2^31 packing, GKR claims, WHIR sumcheck polys, and unsupported rotations
- Keep the constraints-only prover harness private to the integration test instead of exposing it on `StaticVerifierCircuit`
- Remove unused Poseidon selector code and reuse the shared packing helper in the transcript
- Avoid redundant BabyBear reductions in `add`, `sub`, `mul`, and `mul_add`
- Assert verifier equalities directly instead of materializing residuals first

### DIFF
--- a/crates/static-verifier/src/field/baby_bear/base.rs
+++ b/crates/static-verifier/src/field/baby_bear/base.rs
@@ -193,6 +193,9 @@ impl BabyBearChip {
         mut a: BabyBearWire,
         mut b: BabyBearWire,
     ) -> BabyBearWire {
+        if a.max_bits < b.max_bits {
+            std::mem::swap(&mut a, &mut b);
+        }
         if a.max_bits.max(b.max_bits) + 1 > Fr::CAPACITY as usize - RESERVED_HIGH_BITS {
             a = self.reduce(ctx, a);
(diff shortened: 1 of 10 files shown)


zlangley:
I tried to separate things commit-by-commit so it's easy to revert a change if we decide it's risky or something. The minor perf improvements only save ~14k cells (out of ~25.7 M).

crates/static-verifier/src/field/baby_bear/base.rs
```diff
@@ -199,11 +204,8 @@ impl BabyBearChip {
         }
         let value = self.gate().add(ctx, a.value, b.value);
         let max_bits = a.max_bits.max(b.max_bits) + 1;
```
jonathanpwang, line 206: removing the last reduce is fine in all places, but note that what we are using is that the post reduction BABY_BEAR_BITS is much less than Fr::CAPACITY - RESERVED_BITS

jonathanpwang (approved):

PR #2872, opened 2026-06-10
fix(v2.1-rc.2): hard assert output len is 32 byte

shuklaayush:


gdmlcjs (approved):
LGTM! But then it might be appropriate to also remove `unsafe`.

PR #2875, opened 2026-06-11
fix(v2-rc.2): size GPU Poseidon2 scratch per segment

shuklaayush:
## Summary
- Allocate the system Poseidon2 GPU record scratch buffer per segment from exact memory tracegen counts instead of `max_trace_height`.
- Drop the scratch buffer before Poseidon2 trace allocation and guard release builds against OOB writes.
- Save GPU memory by decoupling this scratch allocation from `max_segment_length` for memory-bound segments.

## Memory Impact
With `max_segment_length = 2^24`, the standalone reth benchmarks on block `23992138`, `g7e.2xlarge`, `prove-stark` show the sampled whole-process GPU peak dropping by 1.40 GB while keeping the same 85 memory-triggered segments:

| Run | Commit | Segments | Peak GPU memory (`nvidia-smi`) |
| --- | --- | ---: | ---: |
| [Before](https://github.com/axiom-crypto/openvm-eth/actions/runs/27349752147) | `d519aa6` | 85 | 17.30 GB |
| [After](https://github.com/axiom-crypto/openvm-eth/actions/runs/27349786356) | `d3a30b6` | 85 | 15.90 GB |
| Savings | | | -1.40 GB |

The same runs show the expected drop in OpenVM tracked GPU allocation peaks from removing the persistent `max_trace_height`-sized Poseidon2 scratch buffer:

| Module | Before | After | Delta |
| --- | ---: | ---: | ---: |
| `generate mem proving ctxs` | 7.09 GB | 5.20 GB | -1.89 GB |
| `set initial memory` | 6.87 GB | 4.87 GB | -2.00 GB |
| `prover.rs_code_matrix` | 10.69 GB | 8.70 GB | -1.99 GB |
| `prover.batch_constraints.before_round0` | 16.51 GB | 14.51 GB | -2.00 GB |

A separate [openvm-eth comparison run 27329017888](https://github.com/axiom-crypto/openvm-eth/actions/runs/27329017888) with a smaller resolved segment length also shows the affected tracked peaks dropping:

| Module | Before | After | Delta |
| --- | ---: | ---: | ---: |
| `generate mem proving ctxs` | 5.59 GB | 5.20 GB | -0.39 GB |
| `set initial memory` | 5.37 GB | 4.87 GB | -0.50 GB |
| `prover.rs_code_matrix` | 9.19 GB | 8.70 GB | -0.49 GB |
| `prover.batch_constraints.before_round0` | 14.70 GB | 14.25 GB | -0.45 GB |

## Testing
- `cargo build --profile fast -p openvm-circuit`
- `cargo build --profile fast -p openvm-circuit --features cuda`
- `cargo nextest run --cargo-profile=fast -p openvm-circuit --features cuda test_empty_touched_memory_uses_full_chunk_values test_touched_memory_updates_memory_address_space test_cuda_merkle_tree_cpu_gpu_root_equivalence`

Resolves int-8291

### DIFF
--- a/crates/circuits/primitives/cuda/include/primitives/shared_buffer.cuh
+++ b/crates/circuits/primitives/cuda/include/primitives/shared_buffer.cuh
@@ -18,7 +18,10 @@ template <typename T> struct SharedBuffer {
     __device__ void push(T value) {
         uint32_t idx = atomicAdd(this->idx, 1);
         assert(idx < capacity && "SharedBuffer overflow");
-        data[idx] = value;
+        // Host checks idx > capacity.
+        if (idx < capacity) {
+            data[idx] = value;
+        }
     }
 };
 
--- a/crates/circuits/primitives/cuda/include/primitives/shared_buffer.cuh
+++ b/crates/circuits/primitives/cuda/include/primitives/shared_buffer.cuh
@@ -18,7 +18,8 @@ template <typename T> struct SharedBuffer {
     __device__ void push(T value) {
         uint32_t idx = atomicAdd(this->idx, 1);
         assert(idx < capacity && "SharedBuffer overflow");
-        // Host checks idx > capacity.
+        // On overflow, skip the write to avoid corrupting memory; the counter
+        // still advances, so the host sees a final count > capacity and panics.
         if (idx < capacity) {
             data[idx] = value;
         }
--- a/crates/vm/src/system/cuda/poseidon2.rs
+++ b/crates/vm/src/system/cuda/poseidon2.rs
@@ -1,6 +1,6 @@
 #[cfg(feature = "metrics")]
 use std::sync::atomic::AtomicUsize;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use openvm_circuit::{
     primitives::Chip, system::poseidon2::columns::Poseidon2PeripheryCols,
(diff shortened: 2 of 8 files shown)


crates/circuits/primitives/cuda/include/primitives/shared_buffer.cuh
```diff
@@ -18,7 +18,10 @@ template <typename T> struct SharedBuffer {
     __device__ void push(T value) {
         uint32_t idx = atomicAdd(this->idx, 1);
         assert(idx < capacity && "SharedBuffer overflow");
-        data[idx] = value;
+        // Host checks idx > capacity.
+        if (idx < capacity) {
```
Golovanov399, line 23: But why if we have the assert one line above

crates/vm/src/system/cuda/poseidon2.rs
```diff
@@ -19,37 +19,62 @@ use crate::cuda_abi::poseidon2;
 
 #[derive(Clone)]
 pub struct SharedBuffer<T> {
-    pub buffer: Arc<DeviceBuffer<T>>,
+    buffer: Arc<Mutex<Option<Arc<DeviceBuffer<T>>>>>,
```
Golovanov399, line 22: I sure hope nothing goes wrong with this construction and/or cuda parallelism

Golovanov399 (commented):
Seems good to me, but I don't know if anything else needs updating with this change because I haven't been familiar with this part of the codebase

crates/circuits/primitives/cuda/include/primitives/shared_buffer.cuh
shuklaayush: yes, but claude complains that the assert is debug only and can lead to memory corruption in release.
i kept the if since this doesn't seem to have a perf impact

stephenh-axiom-xyz: It doesn't look debug only? There is `assert` and `debug_assert`?

Golovanov399: This is C++ and I heard that sometimes some optimization flag disables asserts under some circumstances (although iirc we use `-O2` and maybe it was about `-O3`/`-Ofast`)
But even if it's a debug assert, the logic "if `idx >= capacity` then we just ignore and do nothing" is not what we want here for sure

crates/vm/src/system/cuda/poseidon2.rs
Golovanov399: (for the context: didn't we fix some race condition a couple of month ago, which was related to mutex + cuda? My memories are vague but I'm concerned)

crates/circuits/primitives/cuda/include/primitives/shared_buffer.cuh
shuklaayush: there's a `NDEBUG` flag in c/c++ which disables asserts and is sometimes used in optimized builds. i'm not sure if we set it in ours but it's better to do an extra check than corrupting memory. i don't think there's a separate `debug_assert` in c/c++
in either case, the assertion check is also duplicated in rust so host will panic if this overflows

crates/vm/src/system/cuda/poseidon2.rs
shuklaayush: i don't think it's an issue here since we don't actually do tracegen in parallel

stephenh-axiom-xyz (commented):
Could you explain why the Poseidon2 chip now uses a mutex? You can only initialize the device buffer once, and all the writes are on the CUDA side?

shuklaayush:
it's not initialized once for the chip lifetime anymore. it's initialized once per segment after we know the exact record count, then poseidon2 takes/drops it after tracegen. the mutex is for mutably sharing `Option<Arc<DeviceBuffer<_>>>` between poseidon2 and the memory/merkle code. a similar pattern is followed in sha2: https://github.com/openvm-org/openvm/blob/f8f4c18e6b1fae0c2480e6bf6daac3adab705325/extensions/sha2/circuit/src/sha2_chips/block_hasher_chip/mod.rs#L35

i guess we don't strictly need a mutex since we're not using it for synchronization and can use refcell or oncelock, but if in the future we ever make tracegen parallel, then we'd need something like a mutex

stephenh-axiom-xyz (approved):

PR #2876, opened 2026-06-11
feat: rvr extention for rv64 riscv

mansur20478:
RVR extension for RV64 RISCV. The current implementation was tested on `openvm-riscv-integration-tests` tests.

The RVR extension required following changes on top of `rvr-rv32`:
1) Adding new `W` ALU and Mult/Div instructions, new `lwu` instruction.
2) Updating lifter to lift new instructions.
3) Updating instruction simulation done in block generation phase to 64bit.
4) Updating `C` code emission to emit 64bit instructions (`MULH` instruction uses `__int128` to avoid overflow of `64bit` multiplication to get upper 64bit of the result). Adding 64bit and 32bit sign extended reads in rvr.
5) Updating `RvState` registers to 64bit.
6) Updating `bridge.rs` - the buffer access between `VmState` and `rvr:ffi`, to 64bit.

towards INT-8107


crates/rvr/rvr-openvm-lift/src/cfg.rs
```diff
@@ -753,8 +924,14 @@ fn get_successors(
                         let targets: Vec<u32> = addr_val
                             .values
                             .iter()
-                            .filter(|t| ctx.pc_to_idx.contains_key(t))
-                            .copied()
+                            .filter_map(|&t| {
+                                let pc32 = t as u32;
```
shuklaayush, line 926: try not to silently truncate and throw an error if pc overflow

crates/rvr/rvr-openvm-lift/src/cfg.rs
```diff
@@ -810,11 +987,30 @@ fn eval_jumpdyn_multi(state: &RegisterState, rs1: u8, imm: i32) -> RegisterValue
...
-            return RegisterValue::unknown();
+    let compute = |v: u64| -> Option<u64> {
+        let target = v.wrapping_add(imm as u64) & !1u64;
+        debug_assert!(target < (1u64 << PC_BITS));
+        if target < (1u64 << PC_BITS) {
+            Some(target)
+        } else {
+            None
+        }
+    };
```
shuklaayush: this seems like it should be a util function somewhere

crates/rvr/rvr-state/src/state.rs
```diff
@@ -38,17 +38,19 @@ pub struct RvState<
...
     pub suspender: S,
 
+    // TODO: Is it a dead field? set to u32 to match the rv32 version
     /// Reservation address for LR/SC.
-    pub reservation_addr: X::Reg,
+    pub reservation_addr: u32,
 
+    // TODO: Is it a dead field? set to u32 to match the rv32 version
     /// Reservation valid flag for LR/SC.
     pub reservation_valid: u8,
```
shuklaayush: yes we should remove both in a follow-up. this was used in the atomic extension in original rvr

crates/rvr/rvr-state/src/state.rs
```diff
@@ -38,17 +38,19 @@ pub struct RvState<
     pub regs: [X::Reg; NUM_REGS],
 
     /// Program counter (hot).
-    pub pc: X::Reg,
+    pub pc: u32,
```
shuklaayush: this doesn't seem right. we should still keep pc as `u64` but trap in places where we try to jump to pc above `PC_BITS`

crates/rvr/rvr-state/src/state.rs
```diff
@@ -61,20 +63,23 @@ pub struct RvState<
...
 
+    // TODO: Is it a dead field? set to u32 to match the rv32 version
     /// Current heap break.
-    pub brk: X::Reg,
+    pub brk: u32,
 
+    // TODO: Is it a dead field? set to u32 to match the rv32 version
     /// Initial heap break.
-    pub start_brk: X::Reg,
+    pub start_brk: u32,
```
shuklaayush: yes all these should be removed in a follow-up

crates/rvr/rvr-state/src/state.rs
```diff
@@ -61,20 +63,23 @@ pub struct RvState<
...
     /// Guest memory pointer (cold).
     pub memory: *mut u8,
 
     /// Tracer state (ZST when T = (), real struct when tracing).
     pub tracer: T,
 
+    // TODO: Is it a dead field? set to u32 to match the rv32 version
     /// Control and status registers (cold).
-    pub csrs: [X::Reg; NUM_CSRS],
+    pub csrs: [u32; NUM_CSRS],
```
shuklaayush: yep, should be removed

crates/vm/src/arch/rvr/metered.rs
```diff
@@ -210,9 +210,10 @@ impl SegmentationState {
     /// `add_register_merkle_heights` + `update_boundary_merkle_heights`).
     ///
     /// OpenVM records pages for the entire register space
-    /// at init and after each segment boundary.
+    /// (AS=1, ptr=0, size=32*8=256) at init and after each segment boundary.
     fn add_register_merkle_heights(&mut self) {
-        const REG_SIZE: u32 = (RV64_NUM_REGISTERS * RV64_REGISTER_NUM_LIMBS) as u32;
+        // RV64_REGISTER_AS=1, RV64_NUM_REGISTERS=32, RV64_REGISTER_NUM_LIMBS=8
+        const REG_SIZE: u32 = 32 * 8; // 256 bytes
```
shuklaayush: try to avoid hardcoding

shuklaayush (commented):
some comments - primarily `pc` should become `u64` even if at the moment we don't support `pc` above 30 bits

extensions/riscv/rvr/src/lib.rs
gdmlcjs, line 159: I think the 4 should be changed to 8

gdmlcjs (approved):
Same comment changes ("RV32" -> "RV64") in
- crates/vm/src/arch/rvr/debug.rs:58
- crates/vm/src/arch/rvr/execute.rs:97

crates/rvr/rvr-openvm-lift/src/cfg.rs
```diff
@@ -753,8 +918,18 @@ fn get_successors(
...
                             .values
                             .iter()
-                            .filter(|t| ctx.pc_to_idx.contains_key(t))
-                            .copied()
+                            .filter_map(|&t| {
+                                assert!(
+                                    t < (1u64 << PC_BITS),
+                                    "JumpDyn target {t:#x} exceeds PC address space"
+                                );
+                                let pc32 = t as u32;
```
shuklaayush, line 926: let's try to avoid converting `pc` to u32 altogether if possible. also better to emit errors instead of panicing

crates/vm/src/arch/rvr/state.rs
```diff
@@ -79,10 +79,10 @@ pub fn init_rvr_state<F: PrimeField32>(
     vm_state: &mut VmState<F, GuestMemory>,
     pc: u32,
```
shuklaayush, line 80: we should change it to `u64`

crates/vm/src/arch/rvr/execute.rs
```diff
@@ -140,13 +140,17 @@ where
     let exit_code = state.result_code();
     match status {
         ExecutionStatus::Terminated if exit_code == 0 => {
-            write_rvr_registers(vm_state, &state.regs);
-            vm_state.set_pc(state.pc);
+            write_rv64_registers(vm_state, &state.regs);
+            vm_state.set_pc(
+                u32::try_from(state.pc).expect("PC must be within u32 range after C bounds check"),
+            );
```
shuklaayush, line 146: not the most ideal, we should change `pc` to `u64` but check that it's within `PC_BITS`

shuklaayush (approved):
some comments but we can address in a follow up

PR #2877, opened 2026-06-11
feat: add workflow dispatch for sdk tests

Aalanli:
Resolves: INT-8261


crates/sdk/contracts/lib/forge-std
stephenh-axiom-xyz, line 1: Why was this added?

.github/workflows/sdk.yml
```diff
@@ -2,17 +2,27 @@ name: OpenVM SDK Tests
...
+  workflow_dispatch:
+    inputs:
+      x86_family:
+        description: "x86 EC2 instance family"
+        type: string
+        default: "m7a.24xlarge"
+      arm_family:
+        description: "ARM EC2 instance family"
+        type: string
+        default: "m8g.24xlarge"
```
stephenh-axiom-xyz: Should be a single input (machine or smth)

.github/workflows/sdk.yml
```diff
@@ -2,17 +2,27 @@ name: OpenVM SDK Tests
...
-  #     - ".github/workflows/sdk.yml"
+  pull_request:
+    branches: ["**"]
+    paths:
+      - "crates/circuits/primitives/**"
+      - "crates/vm/**"
+      - "crates/toolchain/**"
+      - "crates/sdk/**"
+      - "Cargo.toml"
+      - ".github/workflows/sdk.yml"
```
stephenh-axiom-xyz: These are too slow, don't enable them for pull requests

Aalanli: 👍

.github/workflows/sdk.yml
Aalanli: So you don't want to test all at once?

stephenh-axiom-xyz: No, that is what we want to do specifically for the upcoming release. Will not always be the case

Aalanli:
@stephenh-axiom-xyz I updated to only one machine field. kept the arch field to select the image and skip aot tests if we are on arm.

stephenh-axiom-xyz:
What's the behaviour if you say `x86` but then add an `arm` machine?

Aalanli:
It'll just fail, because aot doesn't run on arm. Seems over complicated in my mind to detect the arch type from the instance string.

Aalanli:
gh: 
- x86: https://github.com/openvm-org/openvm/actions/runs/27437608607
- arm: https://github.com/openvm-org/openvm/actions/runs/27438172039

stephenh-axiom-xyz (approved):

PR #2878, opened 2026-06-11
feat: enable workflow ui on main

Aalanli:
Resolves: INT-8261
note that this PR is necessary to get the UI on workflow dispatch.

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -2,6 +2,16 @@ name: OpenVM SDK Tests
 
 on:
   workflow_call:
+  workflow_dispatch:
+    inputs:
+      x86_family:
+        description: "x86 EC2 instance family"
+        type: string
+        default: "m7a.24xlarge"
+      arm_family:
+        description: "ARM EC2 instance family"
+        type: string
+        default: "m8g.24xlarge"
   push:
     branches: ["main"]
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk.yml
```diff
@@ -2,6 +2,16 @@ name: OpenVM SDK Tests
 
 on:
   workflow_call:
+  workflow_dispatch:
+    inputs:
+      x86_family:
```
stephenh-axiom-xyz: Why are these separate arguments? The workflow should probably run a single time, and you specify what machine you want (i.e. `x86` or `arm`)

Aalanli: Currently it runs both as a cartesian product with the tests.

stephenh-axiom-xyz: Let's not do that, please modify it to only run one

Aalanli:
Hi @stephenh-axiom-xyz, can you take a look again?

Aalanli:
Not sure why lint is failing, doesn't seem to be related to this change.

stephenh-axiom-xyz (commented):
@Aalanli why not just pass in the machine? As we do in most of our other workflows

Aalanli:
I feel like this is more convenient. Since if you want to test arm and then x86 you don't have to remember the instance type.

stephenh-axiom-xyz:
Please just do the machine, if we are testing with a specific arch in mind we should be precise about it

Aalanli:
I updated to only one machine field. kept the arch field to select the image and skip aot tests if we are on arm.

PR #2879, opened 2026-06-11
fix(v2-rc.2): reject public values writes past configured limit

shuklaayush:
Fixes INT-5507
Fixes INT-5505
Fixes INT-5506

## Summary

Reject public-values accesses past the configured `num_public_values` limit instead of silently accepting them.

Neither execution path rejected this before. Non-AOT now gets the expected bounds check because `MmapMemory` exposes the configured size through `size()` and slices, rather than the page-rounded mmap length. AOT public-values load/store instructions now fall back to the normal executor path, so they use the same checked memory access instead of emitting unchecked x86 memory access.

## Tests

Adds regressions for direct public-values memory writes and `reveal` past the configured public-values limit.

### DIFF
--- a/crates/vm/src/system/memory/merkle/public_values.rs
+++ b/crates/vm/src/system/memory/merkle/public_values.rs
@@ -277,4 +277,18 @@ mod tests {
             .verify(&hasher, memory_dimensions, final_memory_root)
             .unwrap();
     }
+
+    #[test]
+    #[should_panic]
+    fn test_public_values_write_beyond_num_public_values_is_rejected() {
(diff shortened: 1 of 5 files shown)


stephenh-axiom-xyz (commented):
Am I correct in my understanding that the non-AOT path does this already, but with AOT it didn't?

shuklaayush:
no, both didn't do it previously. the non-aot path is handled by having mmap return a slice which has the correct size which causes out of bound access to panic. i updated the description

stephenh-axiom-xyz (approved):

PR #2880, opened 2026-06-11
fix: sha2 alignment issue fix

GunaDD:
`state` is a byte sub-slice of a record and is not guaranteed to be 8-byte aligned, so we cannot reinterpret it in place as `&mut [u64; 8]`. Copy through an aligned buffer (using native-endian bytes).

### DIFF
--- a/extensions/sha2/circuit/src/sha2_chips/config.rs
+++ b/extensions/sha2/circuit/src/sha2_chips/config.rs
@@ -64,7 +64,7 @@ impl Sha2Config for Sha512Config {
         debug_assert!(input.len() == Sha512Config::BLOCK_BYTES);
 
         // `state` may only be 4-byte aligned, so copy through an aligned `u64` buffer.
-        // We store the state as bytes in the record because the word size is 
+        // We store the state as bytes in the record because the word size is
         // not known at compile time (u32 for Sha256, u64 for Sha512).
         let mut state_u64s = [0u64; 8];
         for (w, chunk) in state_u64s.iter_mut().zip(state.chunks_exact(8)) {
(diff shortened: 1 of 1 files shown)


extensions/sha2/circuit/src/sha2_chips/config.rs
```diff
@@ -63,16 +63,23 @@ impl Sha2Config for Sha512Config {
...
-        let state_u64s: &mut [u64; 8] = unsafe { &mut *(state.as_mut_ptr() as *mut [u64; 8]) };
+        // `state` is a byte sub-slice of a record and is not guaranteed to be 8-byte aligned, so we
+        // cannot reinterpret it in place as `&mut [u64; 8]`. Copy through an aligned buffer (using
+        // native-endian bytes). We store the
+        // state as bytes in the record because the word size is not known at compile time (u32 for
+        // Sha256, u64 for Sha512).
+        let mut state_u64s = [0u64; 8];
+        for (w, chunk) in state_u64s.iter_mut().zip(state.chunks_exact(8)) {
+            *w = u64::from_ne_bytes(chunk.try_into().unwrap());
+        }
```
shuklaayush, line 73: the comment seems unnecessarily verbose. can do something like
```rust
  let state = &mut state[..Sha512Config::STATE_BYTES];

  // `state` may only be 4-byte aligned, so copy through an aligned `u64` buffer.
  let mut state_u64s = std::array::from_fn(|i| {
      u64::from_ne_bytes(state[i * 8..(i + 1) * 8].try_into().unwrap())
  });
```

shuklaayush (approved):
approved with nit

extensions/sha2/circuit/src/sha2_chips/config.rs
GunaDD: I'll make the comment be 
```
// `state` may only be 4-byte aligned, so copy through an aligned `u64` buffer.
// We store the state as bytes in the record because the word size is not known at compile time (u32 for // Sha256, u64 for Sha512).
```

PR #2884, opened 2026-06-12
feat: EVM verify optionally verifies app commit + CLI vk paths

stephenh-axiom-xyz:
Resolves INT-5100 and INT-4826.

## Overview

This PR adds two related capabilities:

1. **Optional app-commit checking during EVM verification.** `Sdk::verify_evm_halo2_proof` now takes an optional expected `AppExecutionCommit`. When provided, it asserts the commit embedded in the proof matches the expected app/exe commits before running the on-chain verifier. The CLI's `verify evm` subcommand exposes this via a new `--app-commit` flag.

2. **Decoupling CLI verification from proving keys.** `cargo openvm setup` now also writes the standalone internal-recursive *verifying* key, and `verify stark` reads that vk directly (via a new `--agg-vk` flag) instead of loading the much larger proving key just to derive the vk. The `verify evm` subcommand also gains an `--evm-verifier` flag to point at a custom verifier directory.

The change is split across the **SDK** (`openvm-sdk`) and the **CLI** (`cargo-openvm`), plus mechanical call-site updates in benchmarks/examples/tests for the new `verify_evm_halo2_proof` signature.

---

### New `Sdk::app_commit` helper — `src/lib.rs`
Adds a public method that produces the `AppExecutionCommit` (the `app_exe_commit` + `app_vm_commit` pair) for a given executable, generating the app proving key if needed. This gives callers a convenient way to obtain the expected commit to pass into verification.

### `verify_evm_halo2_proof` signature — `src/lib.rs`
Adds an `expected_app_commit: Option<AppExecutionCommit>` parameter. When `Some`, it compares against `evm_proof.app_commit` and returns an error (`"mismatching app commits"`) on mismatch before delegating to the existing solidity verifier path. `None` preserves the old behavior.

### `AppExecutionCommit` now comparable — `src/types.rs`
Derives `PartialEq, Eq` so the equality check above is possible.

### Error message cleanup — `src/solidity.rs`
Renames the internal verification error prefix from `"Sdk::verify_openvm_evm_proof"` to the user-facing `"EVM proof verification failed"`.

### Updated call sites (signature change)
- `src/examples/sdk_evm.rs` — passes `None`, with a comment showing how to pass `Some(sdk.app_commit(elf)?)`.
- `src/tests.rs` — `test_sdk_fibonacci` now computes `sdk.app_commit(...)` and passes `Some(app_commit)`, exercising the new path.

---

### New default vk path — `src/default.rs`
Adds `default_internal_recursive_vk_path()` → `~/.openvm/internal_recursive.vk`.

### `setup` also writes the vk — `src/commands/setup.rs`
After writing the internal-recursive proving key, `setup` now also extracts its verifying key and writes it to the new `.vk` path. This is what lets `verify stark` avoid loading the full proving key.

### `verify stark` — `src/commands/verify.rs`
- New `--agg-vk <PATH>` flag. Defaults to the `default_internal_recursive_vk_path()` written by `setup`.
- Reads the aggregation **verifying key** directly from that path, replacing the previous logic that loaded the internal-recursive **proving key** and called `.get_vk()` on it. (Removes the `MultiStarkProvingKey` / `SC` imports that were only needed for that.)

### `verify evm` — `src/commands/verify.rs`
- New `--evm-verifier <PATH>` flag to override the default `~/.openvm/halo2/` verifier directory.
- New `--app-commit [<PATH>]` flag (0-or-1 value):
  - omitted → no commit check (`None`);
  - present with no value → resolves the default `.commit.json` path for the cargo target;
  - present with a path → reads that explicit `.commit.json`.
  - The resolved `AppExecutionCommit` is passed into `Sdk::verify_evm_halo2_proof`.
- Adds `#[command(flatten)] cargo_args: SingleTargetCargoArgs` so the target can be located for the default commit-path resolution.
- Minor: corrects the "verifier not found" hint to mention `cargo openvm setup --evm`.


shuklaayush:
@claude can you check if any docs need to be updated

crates/cli/src/default.rs
```diff
@@ -28,6 +28,10 @@ pub fn default_internal_recursive_pk_path() -> String {
     env::var("HOME").unwrap() + "/.openvm/internal_recursive.pk"
 }
 
+pub fn default_internal_recursive_vk_path() -> String {
+    env::var("HOME").unwrap() + "/.openvm/internal_recursive.vk"
```
luffykai, line 32: will this change, if say we do openvm 2.1 release? how will we know if we need to regenerate this

crates/cli/src/commands/verify.rs
```diff
@@ -73,13 +80,32 @@ enum VerifySubCommand {
...
+            action,
+            help = "Path to EVM halo2 verifier directory, by default uses ~/.openvm/halo2/ (generated by 'cargo openvm setup --evm')",
+            help_heading = "OpenVM Options"
+        )]
+        evm_verifier: Option<PathBuf>,
+
+        #[arg(
+            long,
+            num_args = 0..=1,
+            help = "Check the commit embedded in the proof against an expected app commit. With no value, uses the default commit path for the target; or pass an explicit path to a .commit.json",
```
luffykai, line 94: can you explain the difference in meaning between `Some(None)` v.s. `None`

crates/cli/src/default.rs
stephenh-axiom-xyz: Should probably always re-keygen on any major release, ex. 2.0 -> 2.1 (but probably not 2.0.0 -> 2.0.1, as this wouldn't require a circuit change)

crates/cli/src/commands/verify.rs
stephenh-axiom-xyz: `None` doesn't do the check, `Some(None)` does the check using the default `app_commit` path, `Some(Some(path_to_app_commit))` does the check using `path_to_app_commit`

luffykai: yea, i meant updating the help text there

crates/cli/src/default.rs
luffykai: ok so setup is actually per machine / per openvm minor version right

luffykai (approved):

crates/cli/src/commands/verify.rs
stephenh-axiom-xyz: Ah shoot, somehow these responses didn't buffer on my browser and I merged… will add this in a future PR

PR #2888, opened 2026-06-12
fix(v2-rc.2): derive segmentation limits from prover params

shuklaayush:
## Summary

- Remove configurable segmentation trace-height and interaction-limit knobs.
- Derive segmentation trace-height limits from the engine stacked height and interaction limits from the proving field order.
- Keep only segmentation max memory configurable through `SystemConfig` / CLI.
- Remove stale max segment height plumbing from benchmarks, workflows, and tests.
- Add sccache startup-timeout configuration to reduce CI server startup races.

## Testing

- `cargo clippy --profile fast -p openvm-circuit --all-targets --tests -- -D warnings`
- `cargo check --profile fast -p cargo-openvm`
- `cargo check --profile fast --no-default-features -p openvm-benchmarks-prove --bin keccak_par --features metrics,parallel,jemalloc`
- `cargo clippy --profile fast --no-default-features -p openvm-benchmarks-prove --bin keccak_par --features metrics,parallel,jemalloc -- -D warnings`
- `python3 -m json.tool ci/benchmark-config.json >/dev/null && python3 -m json.tool ci/benchmark-config.example.json >/dev/null && bash -n ci/scripts/utils.sh && python3 -m py_compile ci/scripts/bench.py`
- `actionlint -ignore 'label ".*" is unknown' -ignore '"github.head_ref" is potentially untrusted' -ignore 'object, array, and null values should not be evaluated' .github/workflows/*.yml`
- `ruby -e 'require "yaml"; YAML.load_file(".github/actions/sccache/action.yml"); puts "ok"'`

### DIFF
--- a/benchmarks/prove/src/bin/keccak_par.rs
+++ b/benchmarks/prove/src/bin/keccak_par.rs
@@ -18,10 +18,7 @@ struct ParallelCli {
 }
 
 fn main() -> eyre::Result<()> {
-    let ParallelCli {
-        inner: _,
-        concurrency,
-    } = ParallelCli::parse();
+    let ParallelCli { inner, concurrency } = ParallelCli::parse();
     let vm_config = SdkVmConfig::builder()
         .system(Default::default())
         .rv32i(Default::default())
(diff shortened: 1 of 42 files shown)


benchmarks/prove/src/bin/keccak_par.rs
```diff
@@ -18,23 +18,23 @@ struct ParallelCli {
 }
 
 fn main() -> eyre::Result<()> {
-    let par_args = ParallelCli::parse();
-    let mut vm_config = SdkVmConfig::builder()
+    let ParallelCli {
+        inner: _,
```
zlangley: Should we just remove `BenchmarkCli` from `ParallelCli`? It looks like we just ignore the flags here.

zlangley (approved):

benchmarks/prove/src/bin/keccak_par.rs
shuklaayush: i tried to use it for the non app-only path

PR #2889, opened 2026-06-12
feat: add test sweeps for internal prover & root prover

Aalanli:
Resolves: INT-8260, INT-8456

The workflow I have is:
1. generate parameter sets using this branch: https://github.com/openvm-org/stark-backend/tree/sweep_params
2. save parameter sets as json, check into the tests
3. the test will deserialize the json, loop over all the params and run fibonacci.

Another approach is to modify stark-backend to export some kind of get_sweeps function. Anyways if we don't want to deserialize we need to modify stark backend as making a `SystemParam` needs some non-public constants and functions from `soundness.rs`.

# sweep parameters

Sweeping over internal config:

log_blowups: 1..=7
k_whirs: (1..=7).step_by(2)
l_skips: (1..=18).steup_by(4)
n_stack = 19 - l_skips
w_stack = 512

Sweeping over root config:

log_blowups: [2, 3, 4, 5, 6]
k_whirs: (1..=7).step_by(2)
l_skips: vec![2, 4, 8, 14, 18]
n_stack = 20 - l_skips
w_stack = 18


Aalanli:
Currently the root sweep fails with this config:

```
SystemParams { l_skip: 14, n_stack: 6, w_stack: 18, log_blowup: 2, whir: WhirConfig { k: 1, rounds: [WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }, WhirRoundConfig { num_queries: 193 }], mu_pow_bits: 20, query_phase_pow_bits: 20, folding_pow_bits: 20, proximity: ListDecoding { m: 1 } }, logup: LogUpSecurityParameters { max_interaction_count: 2013265921, log_max_message_length: 7, pow_bits: 18 }, max_constraint_degree: 4 }
```

with error message

```
Error: Proof shape verification failed: Invalid StackingProof shape: Total stacked width across commits (38) exceeds w_stack (18)

Caused by:
    Invalid StackingProof shape: Total stacked width across commits (38) exceeds w_stack (18)
```

Aalanli:
And internal sweep fails with this config:
```
SystemParams { l_skip: 1, n_stack: 18, w_stack: 512, log_blowup: 1, whir: WhirConfig { k: 1, rounds: [WhirRoundConfig { num_queries: 450 }, WhirRoundConfig { num_queries: 450 }, WhirRoundConfig { num_queries: 450 }, WhirRoundConfig { num_queries: 450 }, WhirRoundConfig { num_queries: 450 }, WhirRoundConfig { num_queries: 450 }, WhirRoundConfig { num_queries: 450 }, WhirRoundConfig { num_queries: 450 }, WhirRoundConfig { num_queries: 450 }], mu_pow_bits: 20, query_phase_pow_bits: 20, folding_pow_bits: 18, proximity: ListDecoding { m: 2 } }, logup: LogUpSecurityParameters { max_interaction_count: 2013265921, log_max_message_length: 7, pow_bits: 18 }, max_constraint_degree: 4 }
```

with this error:

```
thread 'tests::test_sdk_fibonacci_internal_sweep' (2885861) panicked at crates/sdk/src/lib.rs:543:18:
Trace heights did not generate properly: Stacked PCS: StackedLayout::new: column height 22 exceeds stacked height 19

Caused by:
    StackedLayout::new: column height 22 exceeds stacked height 19

Location:
    /home/allanl/Programs/openvm-worktree/param_tuning/crates/continuations/src/prover/inner/mod.rs:91:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test tests::test_sdk_fibonacci_internal_sweep ... FAILED
```

stephenh-axiom-xyz (commented):
We need all the SDK tests, especially deferrals. It probably is better to not do everything from a single run, and instead have a script that runs the full SDK suite for a given parameter config.

Aalanli:
Hi @stephenh-axiom-xyz , can you take a look?

stephenh-axiom-xyz (commented):
Rn params are only modified in `fib_sdk`, every instance of `*_params_with_100_bits_security` should be replaced. Probably best to make a helper for this

Aalanli:
# Update

I have since updated `stack-backend` to export more types publicly so we can construct `SystemParam`s here. The current CI workflow on the SDK side is 
- generate param lists for internal only, internal + root and root
- run the sweep on internal only without root prover
- run the sweep on internal + root with either evm / root prover 
- run the root with either evm / root prover

CLI is fairly straightforward, the tests now include the integration tests in the workflow. 

note that this PR also includes 2 fixes that guard against invalid param settings
- one guards against the fact that certain param combos makes `internal_for_leaf` circuits and the recursive circuits different which causes problems
- another guards against `l_skip > 5` which is not supported by the CUDA backend

crates/sdk/src/tests.rs
```diff
@@ -33,22 +42,58 @@ cfg_if::cfg_if! {
...
+        GenericSdkBuilder::new().app_config(AppConfig::riscv32(app_params.clone()));
+    sdk_builder = sdk_builder.agg_params(agg_params.clone());
+    #[cfg(feature = "root-prover")]
+    {
+        sdk_builder = sdk_builder.root_params(_root_params);
+    }
+    (sdk_builder.build().unwrap(), app_params, agg_params)
+}
+
+fn get_params_from_env(env_var: &str, default: impl FnOnce() -> SystemParams) -> SystemParams {
```
stephenh-axiom-xyz, line 82: The env variable itself is a JSON?

crates/sdk/src/tests.rs
```diff
@@ -258,28 +303,256 @@ fn make_deferral_sdk(
     Ok((vs_sdk, vs_stdin, def_input))
 }
 
+#[derive(Serialize, Default)]
+struct ParamSet {
+    app: Option<SystemParams>,
+    leaf: Option<SystemParams>,
+    root: Option<SystemParams>,
+    internal: Option<SystemParams>,
```
stephenh-axiom-xyz: nit: `internal` before `root` makes more sense

crates/sdk/src/tests.rs
```diff
@@ -258,28 +303,256 @@ fn make_deferral_sdk(
     Ok((vs_sdk, vs_stdin, def_input))
 }
 
+#[derive(Serialize, Default)]
+struct ParamSet {
```
stephenh-axiom-xyz: Let's not have param generation code (i.e. this struct + the two `generate_interesting_*` functions below) as tests. We have a `scripts` package for helper binaries like this, please move these there

crates/sdk/src/tests.rs
```diff
@@ -258,28 +303,256 @@ fn make_deferral_sdk(
...
     setup_tracing();
-    let (sdk, _app_params, _agg_params) = make_fib_sdk();
-
+    let (sdk, _, _) = make_fib_sdk();
     let elf = Elf::decode(
         include_bytes!("../programs/examples/fibonacci.elf"),
         MEM_SIZE as u32,
     )?;
-    let app_exe = sdk.convert_to_exe(elf)?;
+    let app_exe = sdk.convert_to_exe(elf.clone())?;
```
stephenh-axiom-xyz: nit: why clone?

crates/sdk/src/tests.rs
```diff
@@ -258,28 +303,256 @@ fn make_deferral_sdk(
...
     stdin.write(&n);
 
     #[cfg(not(feature = "evm-verify"))]
     {
-        let mut evm_prover = sdk.evm_prover_without_halo2(app_exe)?;
-        let proof = evm_prover.prove_root(stdin, &[])?;
-        let vk = evm_prover.root_prover.0.get_vk();
-        let engine = RootE::new(vk.inner.params.clone());
-        engine.verify(&vk, &proof)?;
+        #[cfg(feature = "root-prover")]
```
stephenh-axiom-xyz: Given `evm-verify` implies `root-prover`, the more idiomatic way to do this would be:
```
#[cfg(not(feature = "root-prover"))]
{
    // STARK prove + verify
}
#[cfg(all(feature = "root-prover", not(feature = "evm-verify")))]
{
    // Root prove + verify
}
#[cfg(feature = "evm-verify")]
{
    // EVM prove + verify
}
```

crates/sdk/src/tests.rs
```diff
@@ -306,12 +578,32 @@ fn test_verify_stark_deferral() -> Result<()> {
...
 
-    let vk = evm_prover.root_prover.0.get_vk();
-    let engine = RootE::new(vk.inner.params.clone());
-    engine.verify(&vk, &vs_proof)?;
+        let openvm_verifier = vs_sdk.generate_halo2_verifier_solidity()?;
+        let _gas_cost = Sdk::verify_evm_halo2_proof(&openvm_verifier, vs_proof, Some(app_commit))?;
+    }
+    #[cfg(not(feature = "evm-verify"))]
+    {
+        #[cfg(feature = "root-prover")]
```
stephenh-axiom-xyz: Same comment on cfg as above

crates/sdk/src/tests.rs
```diff
@@ -343,23 +635,40 @@ fn test_verify_many_deferrals() -> Result<()> {
...
+            vs_proof.deferral_merkle_proofs.is_some(),
+            "verify-many proof must carry deferral merkle proofs",
+        );
+        Sdk::verify_proof(vs_sdk.agg_vk().as_ref().clone(), vs_baseline, &vs_proof)?;
+    }
 
     Ok(())
 }
 
+#[cfg(feature = "root-prover")]
```
stephenh-axiom-xyz: Why does this need root?

crates/sdk/src/tests.rs
```diff
@@ -550,6 +871,7 @@ fn test_prove_mixed_vm_def_depth_mismatch() -> Result<()> {
 ///   -- sdk_static_verifier_cell_profiling
 /// ```
 #[cfg(feature = "cell-profiling")]
+#[cfg(feature = "root-prover")]
```
stephenh-axiom-xyz: nit: use all

crates/sdk/src/tests.rs
```diff
@@ -343,23 +635,40 @@ fn test_verify_many_deferrals() -> Result<()> {
     )?;
     let vs_exe = vs_sdk.convert_to_exe(vs_elf)?;
 
-    let (vs_proof, vs_baseline) = vs_sdk.prove(vs_exe, vs_stdin, &def_inputs)?;
-    assert!(
-        vs_proof.deferral_merkle_proofs.is_some(),
-        "verify-many proof must carry deferral merkle proofs",
-    );
-    Sdk::verify_proof(vs_sdk.agg_vk().as_ref().clone(), vs_baseline, &vs_proof)?;
+    #[cfg(feature = "evm-verify")]
```
stephenh-axiom-xyz: Might as well do the `root-prover`-only branch here too

crates/sdk/src/tests.rs
```diff
@@ -433,13 +740,26 @@ fn test_verify_stark_path_sdk_can_verify_own_proofs() -> Result<()> {
     Sdk::verify_proof(agg_vk.clone(), vs_baseline.clone(), &vs_proof)?;
 
     let (vs2_stdin, vs2_def_input) = make_verify_stark_inputs(&sdk, &vs_proof, vs_baseline)?;
-    let (vs2_proof, vs2_baseline) = sdk.prove(vs_exe, vs2_stdin, &[vs2_def_input])?;
-    assert!(vs2_proof.deferral_merkle_proofs.is_some(),);
-    Sdk::verify_proof(agg_vk, vs2_baseline, &vs2_proof)?;
+    #[cfg(feature = "evm-verify")]
```
stephenh-axiom-xyz, line 469: nit: add root-only branch as above

crates/sdk/src/tests.rs
```diff
@@ -467,6 +787,7 @@ fn test_deferrals_enabled_without_usage() -> Result<()> {
     Ok(())
 }
 
+#[cfg(feature = "root-prover")]
```
stephenh-axiom-xyz: Don't think it needs this feature

crates/sdk/src/tests.rs
```diff
@@ -343,23 +635,40 @@ fn test_verify_many_deferrals() -> Result<()> {
...
+        );
+        Sdk::verify_proof(vs_sdk.agg_vk().as_ref().clone(), vs_baseline, &vs_proof)?;
+    }
 
     Ok(())
 }
 
+#[cfg(feature = "root-prover")]
 #[test]
 fn test_verify_stark_with_deferral_child() -> Result<()> {
```
stephenh-axiom-xyz: nit: This is a kinda special case test where we're not going to do all three prove types, right? Let's move it to be right above `test_prove_mixed_vm_def_depth_mismatch`

stephenh-axiom-xyz (commented):
Some comments about the SDK test file. Please (a) remove the commit with Zach's changes, (b) change base to rc.3, and (c) consider adding a helper to the SDK test to change which prove/verify is done based on what features are enabled

crates/sdk/src/tests.rs
Aalanli: yes

PR #2890, opened 2026-06-15
feat: rvr support for rv64 bigint extension

mansur20478:
rvr support for the rv64 bigint extension

`rd/wr_mem_words_traced` in `rvr-openvm-ffi-common` was updated to operate at OpenVM's native word size (8 bytes, matching `MEMORY_BLOCK_BYTES`). Previously it used 4-byte (u32) granularity, which was logically incorrect for rv64. The bigint extension uses the corrected API.

This change broke compilation for other extensions (sha2, ecc, algebra) that depended on the old u32 signature. As a temporary fix, those extensions bypass `rd/wr_mem_words_traced` and call the underlying u32 range wrappers directly. Each call site is marked with a TODO to migrate properly when rvr support is added for that extension.

The bigint MULT operation follows the interpreter and uses a u32/u64 combination (8 × u32 limbs, accumulating into u64). RISC-V's native MULW/MUL uses u64/u128. Should we align to the u64/u128 approach, or keep it consistent with the interpreter?

towards INT-8108

### DIFF
--- a/crates/rvr/rvr-openvm-ffi-common/src/lib.rs
+++ b/crates/rvr/rvr-openvm-ffi-common/src/lib.rs
@@ -54,6 +54,10 @@ extern "C" {
     pub fn trace_wr_mem_u32_wrapper(state: *mut c_void, addr: u32, new_val: u32);
     pub fn trace_mem_access_u32_wrapper(state: *mut c_void, addr: u32, addr_space: u32);
 
+    // ── Memory access (single u64 word, data)
+    pub fn rd_mem_u64_wrapper(state: *mut c_void, addr: u32) -> u64;
+    pub fn trace_rd_mem_u64_wrapper(state: *mut c_void, addr: u32, val: u64);
+
     // ── Memory access (u32 ranges, data) ─────────────────────────────
     pub fn rd_mem_u32_range_wrapper(
         state: *mut c_void,
--- a/crates/rvr/rvr-openvm-ffi-common/src/lib.rs
+++ b/crates/rvr/rvr-openvm-ffi-common/src/lib.rs
@@ -101,6 +101,12 @@ extern "C" {
         vals: *const u64,
         num_words: u32,
     );
+    pub fn trace_mem_access_u64_range_wrapper(
+        state: *mut c_void,
+        base_addr: u32,
+        num_words: u32,
+        addr_space: u32,
+    );
 
     // ── Memory access (u64-word ranges, trace-only) ───────────────────
     pub fn trace_rd_mem_u64_range_wrapper(
@@ -244,5 +250,5 @@ pub unsafe fn trace_mem_access_range(
         num_words >= 1,
         "trace_mem_access_range requires num_words >= 1"
     );
-    trace_mem_access_u32_range_wrapper(state, base_addr, num_words, addr_space);
+    trace_mem_access_u64_range_wrapper(state, base_addr, num_words, addr_space);
 }
--- a/crates/rvr/rvr-openvm-ffi-common/src/lib.rs
+++ b/crates/rvr/rvr-openvm-ffi-common/src/lib.rs
@@ -54,7 +54,7 @@ extern "C" {
     pub fn trace_wr_mem_u32_wrapper(state: *mut c_void, addr: u32, new_val: u32);
     pub fn trace_mem_access_u32_wrapper(state: *mut c_void, addr: u32, addr_space: u32);
 
-    // ── Memory access (word-aligned ranges, data) ─────────────────────
+    // ── Memory access (u32 ranges, data) ─────────────────────────────
     pub fn rd_mem_u32_range_wrapper(
         state: *mut c_void,
         base_addr: u32,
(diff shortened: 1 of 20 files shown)


gdmlcjs (approved):
No problems found!

crates/rvr/rvr-openvm-ffi-common/src/lib.rs
```diff
@@ -157,8 +185,9 @@ pub fn u32s_as_u64s(words: &[u32]) -> &[u64] {
 
 // ── Ergonomic batched memory helpers ─────────────────────────────────────
 
-/// Read `out.len()` consecutive u32 words from guest memory into `out`,
-/// then trace each as a memory read.
+/// Read `out.len()` consecutive u64 words from guest memory into `out`,
+/// then trace each as a memory read. A word is 8 bytes (`WORD_SIZE`),
+/// matching the circuit's `MEMORY_BLOCK_BYTES` granularity.
 ///
```
shuklaayush, line 201: let's try to avoid references to the circuit in the executor

crates/rvr/rvr-openvm-ffi-common/src/lib.rs
```diff
@@ -168,33 +197,34 @@ pub fn u32s_as_u64s(words: &[u32]) -> &[u64] {
...
-    trace_rd_mem_u32_range_wrapper(state, base_addr, out.as_ptr(), n);
+    rd_mem_u64_range_wrapper(state, base_addr, out.as_mut_ptr(), n);
+    trace_rd_mem_u64_range_wrapper(state, base_addr, out.as_ptr(), n);
 }
 
 /// Trace each value in `vals` as a write at `base_addr + i*WORD_SIZE`,
 /// then write them to guest memory. Trace-before-write so future tracers
-/// can read the old value before it is overwritten.
+/// can read the old value before it is overwritten. A word is 8 bytes
+/// (`WORD_SIZE`), matching the circuit's `MEMORY_BLOCK_BYTES` granularity.
```
shuklaayush, line 223: avoid mentions to circuit

shuklaayush: also as a follow-up we should clean up the "word" terminology or come up a less ambiguous terms. since atm, a word can refer to both 4 bytes or 8 bytes depending on the context

extensions/algebra/rvr/ffi/common/src/lib.rs
```diff
@@ -46,7 +53,8 @@ pub trait FieldArith {
 #[inline(always)]
 pub unsafe fn read_field_256<F: PrimeField<Repr = [u8; 32]>>(state: *mut c_void, ptr: u32) -> F {
     let mut words = [0u32; FIELD_256_WORDS];
-    rd_mem_words_traced(state, ptr, &mut words);
+    rd_mem_u32_range_wrapper(state, ptr, words.as_mut_ptr(), FIELD_256_WORDS as u32);
+    trace_rd_mem_u32_range_wrapper(state, ptr, words.as_ptr(), FIELD_256_WORDS as u32);
```
shuklaayush: use u64 variants

extensions/sha2/rvr/ffi/src/lib.rs
```diff
@@ -98,15 +138,34 @@ pub unsafe extern "C" fn rvr_ext_sha512(
...
     let block_array: &GenericArray<u8, _> =
         GenericArray::from_slice(u32_words_as_bytes(&block_words));
     compress512(&mut state_u64s, &[*block_array]);
 
-    wr_mem_words_traced(state, dst_ptr, u64s_as_u32s(&state_u64s));
+    {
+        let s = u64s_as_u32s(&state_u64s);
+        trace_wr_mem_u32_range_wrapper(state, dst_ptr, s.as_ptr(), s.len() as u32);
+        wr_mem_u32_range_wrapper(state, dst_ptr, s.as_ptr(), s.len() as u32);
+    }
```
shuklaayush, line 103: it seems like it's better to use `u64` here instead of converting back and forth from `u32`

shuklaayush (commented):
i think it's better to change functions to read and operate on `u64` instead of `u32` wherever possible. `u32` was the natural type for rv32 but for rv64 we should move to `u64`

crates/rvr/rvr-openvm-ffi-common/src/lib.rs
gdmlcjs: I think we should do the same for "limb"? For example `RV64_REGISTER_NUM_LIMBS` is based on 8-bit limbs.

extensions/algebra/rvr/ffi/common/src/lib.rs
mansur20478: resolved in `feat/rvr-support-algebra` extension branches

extensions/sha2/rvr/ffi/src/lib.rs
mansur20478: resolved in `feat/rvr-support-rv64-sha2`

shuklaayush (approved):
looks good except mentions to circuit/air which the executor should know nothing about

PR #2894, opened 2026-06-15
fix: cargo openvm setup --evm stores halo2 pk

stephenh-axiom-xyz:
Resolves INT-8324 and INT-5603.

## Overview

Previously `cargo openvm prove evm` regenerated the Halo2 proving key on every
invocation, an expensive (~16 GB, multi-minute) operation. This PR makes
`cargo openvm setup --evm` generate (or download) the Halo2 proving key once and
persist it to `~/.openvm/halo2.pk`, so that `prove evm` simply loads it.

To support this, the Halo2 proving key gains a **streaming** serialization format.
The previous `serde`-based path buffered the entire proving key (>10 GB) into a
`Vec<u8>` in memory before writing; the new path uses the stark-backend
`Encode`/`Decode` codec to stream key bytes directly to/from a file.

The work is split into two logical pieces, each reviewable on its own:

1. **Streaming codec for the Halo2 proving key** (`static-verifier`, `sdk`)
2. **CLI wiring: generate/download on setup, load on prove** (`cli`, docs)

---

## 1. Streaming codec for the Halo2 proving key

Replaces the in-memory `serde` serialization of the Halo2 proving key with a
streaming `Encode`/`Decode` implementation.

### `crates/static-verifier/src/codec.rs` (new)

Implements `Encode`/`Decode` (from `openvm-stark-backend::codec`) for the proving-key
types. The format interleaves two kinds of data:

- **Length-prefixed JSON sections** (`write_json_section` / `read_json_section`)
  for the small serde-friendly metadata (`circuit`, `shape`, `Halo2ProvingMetadata`).
  Length-prefixing is required because each section is immediately followed by raw
  Halo2 bytes. Reads are bounded by `MAX_JSON_SECTION_LEN` (64 MiB) to guard against
  malformed input.
- **Raw Halo2 proving-key bytes** written/read directly via Halo2's
  `ProvingKey::write`/`read` with `SerdeFormat::RawBytes` — these stream and never
  fully materialize in a buffer.

Codec impls are provided for `StaticVerifierProvingKey`, `Halo2WrapperProvingKey`,
and `Halo2ProvingPinning`.

### `crates/static-verifier/src/prover.rs`

Removes the old manual `Serialize`/`Deserialize` impls for `Halo2ProvingPinning`
(and the `SerializedHalo2ProvingPinning` helper that buffered `pk_bytes: Vec<u8>`).
`pk_to_bytes` is retained as a plain helper. Updated stale doc comments.

### `crates/static-verifier/src/{keygen.rs, wrapper.rs, lib.rs}`

- Drop `Serialize`/`Deserialize` derives from `StaticVerifierProvingKey` and
  `Halo2WrapperProvingKey` (now serialized via the codec).
- Register the new `codec` module (gated behind `evm-prove`).

### `crates/sdk/src/keygen/mod.rs`

`Halo2ProvingKey` drops its `Serialize`/`Deserialize` derive and instead implements
`Encode`/`Decode`, delegating to the static-verifier codec for its `verifier` and
`wrapper` fields plus the `profiling` flag.

### `crates/sdk/src/fs.rs`

New helpers `write_halo2_pk_to_file` / `read_halo2_pk_from_file` that wrap the file
in a `BufWriter`/`BufReader` and call `encode`/`decode`, keeping the >10 GB key off
the heap.

---

### `crates/cli/src/default.rs`

Adds `default_halo2_pk_path()` → `~/.openvm/halo2.pk`.

### `crates/cli/src/commands/setup.rs`

`cargo openvm setup --evm` now produces the Halo2 PK alongside the verifier artifacts:

- Skips if the file already exists and `--force` is not set.
- With `--download`: `download_halo2_pk` fetches it from the public S3 artifacts
  bucket at key `v{CARGO_PKG_VERSION}/halo2/v{OPENVM_VERSION}/halo2.pk`.
- Otherwise: `generate_halo2_pk` writes `sdk.halo2_pk()` locally via
  `write_halo2_pk_to_file` (errors clearly if the build lacks the `evm-prove` feature).

Also refactors the verifier-artifacts branch from an early `return Ok(())` into an
`else if` chain so the Halo2 PK step always runs first.

### `crates/cli/src/commands/prove.rs`

`cargo openvm prove evm` loads the cached Halo2 PK (`load_required_halo2_pk`) and
passes it into the `Sdk` builder instead of regenerating it. The error message
directs users to run `cargo openvm setup --evm` if the key is missing.

### Tests

- `crates/cli/tests/scripts/cli_evm_e2e_simplified.sh` (new): keygen → setup --evm →
  prove evm → verify evm end-to-end.
- `crates/cli/tests/integration.rs`: registers `test_cli_evm_e2e_simplified`, gated on
  `evm-verify` and marked `#[ignore]` (expensive — generates an EVM proof).

### Docs (`docs/vocs/.../*.mdx`)

Updates `setup`/`prove` documentation to mention that `halo2.pk` is now written to
`~/.openvm/`, can be downloaded with `--download`, and is read by `prove evm`.

---

## Review notes

- **Backward compatibility**: the on-disk format for the Halo2 PK changes (serde →
  streaming codec). Any previously serialized `halo2.pk` is not compatible and must
  be regenerated via `cargo openvm setup --evm`.
- The new codec module and SDK fs helpers are all gated behind the `evm-prove`
  feature, matching the existing gating of `Halo2ProvingKey`.


crates/cli/tests/integration.rs
```diff
@@ -58,6 +58,14 @@ fn test_cli_stark_e2e_no_commit() -> Result<()> {
     run_script("cli_stark_e2e_no_commit.sh", &[])
 }
 
+#[cfg(feature = "evm-verify")]
+#[test]
+#[ignore = "expensive: generates an EVM proof"]
```
stephenh-axiom-xyz, line 63: I ran this locally, will remove the ignore once we have halo2 GPU proving in openvm

crates/cli/src/commands/setup.rs
```diff
@@ -151,6 +160,27 @@ impl SetupCmd {
...
+        {
+            use openvm_sdk::fs::write_halo2_pk_to_file;
+
+            println!("Generating Halo2 proving key locally. Tip: use `--download` to download a pre-built key from S3 instead.");
+            println!("Writing Halo2 proving key to {}", halo2_pk_path.display());
+            write_halo2_pk_to_file(halo2_pk_path, &sdk.halo2_pk())?;
+            Ok(())
+        }
+
+        #[cfg(not(feature = "evm-prove"))]
```
luffykai, line 174: this is weird? should the whole function `generate_halo2_pk`  be under `#[cfg(feature = "evm-prove")]`?

luffykai (approved):

yi-sun (approved):

yi-sun (changes requested):
Also have a CLI option to write the halo2 pk to a specific file + read it from a specific file, to offer better ergonomics for security critical use cases.

crates/cli/src/commands/setup.rs
stephenh-axiom-xyz: It's better to explicitly error out to the user if they installed without the correct features

yi-sun (approved):
LGTM, but please add the new path flags to the docs.

PR #2906, opened 2026-06-18
chore: update refs to enable halo2 GPU prover

stephenh-axiom-xyz:
Resolves INT-8283.

### DIFF
--- a/crates/sdk/src/tests.rs
+++ b/crates/sdk/src/tests.rs
@@ -271,7 +271,7 @@ fn test_sdk_fibonacci() -> Result<()> {
 
     #[cfg(not(feature = "evm-verify"))]
     {
-        let mut evm_prover = sdk.evm_prover(app_exe)?;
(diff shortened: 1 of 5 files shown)


crates/sdk/src/tests.rs
```diff
@@ -271,7 +271,7 @@ fn test_sdk_fibonacci() -> Result<()> {
 
     #[cfg(not(feature = "evm-verify"))]
     {
-        let mut evm_prover = sdk.evm_prover(app_exe)?;
+        let mut evm_prover = sdk.evm_prover_without_halo2(app_exe)?;
```
shuklaayush, line 274: not a fan of this terminology

shuklaayush (approved):

crates/sdk/src/tests.rs
stephenh-axiom-xyz: Let's resolve in a later commit

PR #2909, opened 2026-06-19
fix: final nits for phase3

zlangley:
- Remove unnecessary swaps that obscure the logic (esp. for sub)
- Clean up comments
- Don't instantiate BabyBearWire until we have checked bits

### DIFF
--- a/crates/static-verifier/src/transcript/mod.rs
+++ b/crates/static-verifier/src/transcript/mod.rs
@@ -70,16 +70,17 @@ fn base_baby_bear_decomp_bounds() -> &'static BaseBabyBearDecompBounds {
 /// witness-and-verify (hint) approach.
 ///
 /// This helper only computes the decomposition off-circuit and loads the resulting
-/// digits plus top quotient as witnesses. Constraints must be enforced separately.
+/// raw digit cells plus top quotient as witnesses. Constraints must be enforced
+/// separately before the digits are wrapped as `BabyBearWire`s.
 fn load_base_baby_bear_decomposition_witness(
     ctx: &mut Context<Fr>,
     packed: AssignedValue<Fr>,
-) -> ([BabyBearWire; NUM_SAMPLES_PER_WORD], AssignedValue<Fr>) {
+) -> ([AssignedValue<Fr>; NUM_SAMPLES_PER_WORD], AssignedValue<Fr>) {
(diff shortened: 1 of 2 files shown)


crates/static-verifier/src/transcript/mod.rs
```diff
@@ -70,11 +70,12 @@ fn base_baby_bear_decomp_bounds() -> &'static BaseBabyBearDecompBounds {
...
 /// This helper only computes the decomposition off-circuit and loads the resulting
-/// digits plus top quotient as witnesses. Constraints must be enforced separately.
+/// raw digit cells plus top quotient as witnesses. Constraints must be enforced
+/// separately before the digits are wrapped as `BabyBearWire`s.
 fn load_base_baby_bear_decomposition_witness(
     ctx: &mut Context<Fr>,
     packed: AssignedValue<Fr>,
-) -> ([BabyBearWire; NUM_SAMPLES_PER_WORD], AssignedValue<Fr>) {
+) -> ([AssignedValue<Fr>; NUM_SAMPLES_PER_WORD], AssignedValue<Fr>) {
     let p = BigUint::from(BABY_BEAR_MODULUS_U64);
```
shuklaayush, line 79: i wonder if this should be named something else since this doesn't guarantee that the loaded value is a constrained baby bear element

crates/static-verifier/src/transcript/mod.rs
```diff
@@ -180,7 +176,10 @@ fn decompose_bn254_to_base_baby_bear_digits(
         bounds,
     );
 
-    output_digits
+    output_digits.map(|value| BabyBearWire {
+        value,
+        max_bits: BABY_BEAR_BITS,
+    })
```
shuklaayush: hmm okay. i wonder if we could've instead just made `constrain_base_baby_bear_decomposition` consume `AssignedValue` and return `BabyBearWire`

shuklaayush (approved):
looks good. a couple of comments but feel free to merge since they're not critical

crates/static-verifier/src/transcript/mod.rs
zlangley: I think it's okay; this is an unconstrained witness representing the base-BabyBearPrime decomposition of a number.

PR #2911, opened 2026-06-19
chore(v2-rc.2): add tracegen observability spans

shuklaayush:
## Summary
- Add trace-gen observability spans around record arena allocation, initial memory transport, and RV32IM CUDA record H2D enqueue points.
- Emit per-AIR `trace_gen.record_arena_bytes` only for arenas that can report real byte counts.
- Surface `record_arena.alloc_time_ms` in `openvm-prof` aggregate output.

## Testing
- `cargo check --profile fast -p openvm-circuit --features metrics`
- `cargo check --profile fast -p openvm-rv32im-circuit`

### DIFF
--- a/crates/vm/src/arch/extensions.rs
+++ b/crates/vm/src/arch/extensions.rs
@@ -742,6 +742,19 @@ where
                             let air_name = self.inventory.airs.ext_airs[insertion_idx].name();
                             info_span!("single_trace_gen", air = air_name).entered()
                         });
+                        #[cfg(feature = "metrics")]
+                        if let Some(allocated_bytes) = (!records.is_empty())
+                            .then(|| records.allocated_bytes())
+                            .flatten()
+                        {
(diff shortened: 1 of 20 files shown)


crates/vm/src/arch/extensions.rs
```diff
@@ -742,6 +742,19 @@ where
                             let air_name = self.inventory.airs.ext_airs[insertion_idx].name();
```
stephenh-axiom-xyz, line 742: nit: Move this out of the span scope?

stephenh-axiom-xyz (commented):
What's the purpose of this? Doesn't nsys do this more accurately?

crates/vm/src/arch/extensions.rs
shuklaayush: seems fine since you don't need it if records are empty

stephenh-axiom-xyz (approved):

PR #2913, opened 2026-06-19
feat: rvr support for rv64 deferral extention

mansur20478:
rvr support for deferral extension.

The PR required changes in how the rvr deferral extension reads and write to the memory, from 4 byte granularity to 8 byte. This also meant that for `output length`, the u64 LE stored right after the commit, formula got changed to accommodate 8 byte granularity.

Also, `DEFERRAL_COMMIT_NUM_BYTES` had to be updated to `DEFERRAL_DIGEST_SIZE * F_NUM_BYTE` from `DEFERRAL_DIGEST_SIZE * WORD_SIZE`. This is because `DEFERRAL_DIGEST_SIZE` is digest size in field elements, not in `WORDS`.

resolves INT-8113


mansur20478:
@claude provide review of this PR. make sure that all deferral related constants are took into account and updated accordingly, make sure that changes that were done are correct

shuklaayush (approved):

PR #2914, opened 2026-06-19
fix: root layer tracegen and proving on the same GPU context

stephenh-axiom-xyz:
Resolves INT-8503.

## Overview

Fix root proof generation so root trace generation and proving use the same caller-owned `StarkEngine` instance. This keeps GPU-backed root proving on a single device context instead of constructing a second engine between trace generation and proof creation.

Changes include:

- Add `RootProver::create_engine` helpers in the continuations prover and SDK wrapper.
- Thread an explicit root engine through `generate_proving_ctx`, `prove_from_ctx`, and `prove`.
- Update EVM proving, dummy keygen, root proof height computation, and root prover tests to reuse the shared engine.
- Preserve debug constraint checks and device proving-key transport against the same engine used for trace generation.

### DIFF
--- a/crates/continuations/src/tests/mod.rs
+++ b/crates/continuations/src/tests/mod.rs
@@ -311,13 +311,13 @@ fn test_root_prover(extra_recursive_layers: usize) -> Result<()> {
         None,
         None,
     );
-    let engine = RootEngine::new(root_prover.get_vk().inner.params.clone());
+    let engine = root_prover.create_engine::<RootEngine>();
     let ctx = root_prover.generate_proving_ctx_no_def::<<RootEngine as StarkEngine>::PB, _>(
         internal_recursive_proof,
         &user_pvs_proof,
         &engine.device().device_ctx,
     );
-    let root_proof = root_prover.root_prove_from_ctx::<RootEngine>(ctx.unwrap())?;
+    let root_proof = root_prover.root_prove_from_ctx::<RootEngine>(ctx.unwrap(), &engine)?;
 
     let vk = root_prover.get_vk();
     engine.verify(&vk, &root_proof)?;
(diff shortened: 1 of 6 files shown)


crates/continuations/src/tests/mod.rs
```diff
@@ -311,13 +311,13 @@ fn test_root_prover(extra_recursive_layers: usize) -> Result<()> {
...
     );
-    let engine = RootEngine::new(root_prover.get_vk().inner.params.clone());
+    let engine = root_prover.create_engine::<RootEngine>();
     let ctx = root_prover.generate_proving_ctx_no_def::<<RootEngine as StarkEngine>::PB, _>(
         internal_recursive_proof,
         &user_pvs_proof,
         &engine.device().device_ctx,
     );
-    let root_proof = root_prover.root_prove_from_ctx::<RootEngine>(ctx.unwrap())?;
+    let root_proof = root_prover.root_prove_from_ctx::<RootEngine>(ctx.unwrap(), &engine)?;
```
shuklaayush, line 320: shouldn't we just create an engine in the `new` constructor and use it internally for all the root prover functions. why pass it around?

shuklaayush: not sure about this but if engine is an abstraction common among all types of provers, maybe it should be an argument to the constructor itself and each prover can either store a reference or a copy (if safe)

stephenh-axiom-xyz: I think we try to avoid storing Engines, especially in structs that are potentially long-lived (i.e. the `RootProver` will exist for as long as the `Sdk` does). They contain device-specific stuff

shuklaayush: okay, and is the idea that the same prover can be used on multiple devices?

crates/sdk/src/prover/evm.rs
```diff
@@ -85,8 +85,9 @@ where
 
         const MAX_ROOT_TRACEGEN_RETRIES: usize = 8;
         let agg_prover = &self.stark_prover.agg_prover;
+        let root_engine = self.root_prover.create_engine();
```
Aalanli, line 88: A potential concern is that creating a new engine everytime means creating a new stream for every proof, which is not ideal in terms of nsys traces, and allocator freed memory reuse.

Aalanli (commented):
One small concern with this direction of creating new streams for every proof. But from the fix perspective looks good to me.

Aalanli (approved):

Aalanli (approved):

gaxiom (approved):
LGTM

PR #2916, opened 2026-06-19
refactor: remove unused u32 functions

mansur20478:
remove no longer relevant u32 ranged reads/writes and related functions. RV64 RVR writes ranges in 8 byte granularity, not in 4 bytes.

resolves INT-8425


shuklaayush:
@claude can you do a thorough check that all unused `u32` functions in rvr have been removed

shuklaayush (approved):

PR #2917, opened 2026-06-22
feat(v2.1): preserve rvr profiling artifacts

shuklaayush:
## Summary

- plumb rvr sdk compilation through executable inputs so profiling builds can recover guest elf source locations
- emit native debug information for rvr profiling builds and expose generated c source saving for pure and metered artifacts
- persist metered rvr artifact metadata, including metering context config and executor-to-air mapping, so saved metered artifacts can be reloaded without rebuilding that state
- make metered segmentation and proving-memory configuration serializable for artifact metadata reloads

resolves int-8506


crates/vm/src/arch/rvr/compile.rs
```diff
@@ -97,6 +97,44 @@ impl RvrCompiled {
         })?;
         Ok(dest_lib.to_path_buf())
     }
+
+    /// Copy generated C sources for inspection.
+    pub fn save_generated_sources(&self, dest_dir: &Path) -> Result<(), CompileError> {
```
mansur20478, line 102: i think it might forget staticlib-backed extensions like `librvr_openvm_ext_keccak_ffi.a`, but I think they are not related to C files? Not sure

mansur20478 (approved):
looks good, just not sure about `save_generated_sources` - should it also include staticlibs or not?
lgtm

crates/vm/src/arch/rvr/compile.rs
shuklaayush: yes this is mainly for debugging where we're only interested in non-static files

PR #2919, opened 2026-06-22
fix(v2-rc.2): handle dead AOT PC dispatch slots

### LINKED ISSUE #2852: RV32IM AOT dispatch does not reject dead PC slots

shuklaayush:
## Summary
- Generate AOT dispatch labels for every PC slot covered by the program dispatch table.
- Emit dead-slot stubs that pass the exact PC to the unreachable handler, so pure and metered AOT match interpreter behavior for dead PCs.
- Cover dead slots below `pc_base`, explicit `None` holes, fallthrough into dead slots, and JALR into dead slots.

## Testing
- `cargo clippy -p openvm-rv32im-circuit --all-targets --tests -- -D warnings`
- `cargo clippy -p openvm-rv32im-circuit --target x86_64-apple-darwin --features aot --all-targets --tests -- -D warnings`
- `cargo test -p openvm-rv32im-circuit --target x86_64-apple-darwin --features aot test_aot_dispatch_rejects_dead_pc_slots --no-run`

Supersedes and closes #2856, fixes #2852

### DIFF
--- a/crates/vm/src/arch/aot/metered_execute.rs
+++ b/crates/vm/src/arch/aot/metered_execute.rs
@@ -1,6 +1,6 @@
 use std::{ffi::c_void, mem::offset_of};
 
-use openvm_instructions::exe::VmExe;
+use openvm_instructions::{exe::VmExe, program::DEFAULT_PC_STEP};
 use openvm_stark_backend::p3_field::PrimeField32;
 
 use super::{common::*, AotInstance};
@@ -174,14 +174,17 @@ where
         asm_str += &format!("   jmp {REG_A}\n");
 
         let pc_base = exe.program.pc_base;
+        let base_idx = (pc_base / DEFAULT_PC_STEP) as usize;
+        let num_pc_slots = base_idx + exe.program.instructions_and_debug_infos.len();
 
-        for i in 0..(pc_base / 4) {
-            asm_str += &format!("asm_execute_pc_{}:", i * 4);
-            asm_str += "\n";
-        }
-
-        for (pc, instruction, _) in exe.program.enumerate_by_pc() {
+        for pc_idx in 0..num_pc_slots {
             /* Preprocessing step, to check if we should suspend or not */
+            let pc = pc_idx as u32 * DEFAULT_PC_STEP;
+            let instruction = pc_idx.checked_sub(base_idx).and_then(|idx| {
+                exe.program.instructions_and_debug_infos[idx]
+                    .as_ref()
+                    .map(|(instruction, _)| instruction)
+            });
             asm_str += &format!("asm_execute_pc_{pc}:\n");
 
             asm_str += &format!("    cmp {REG_INSTRET_END}, 0\n");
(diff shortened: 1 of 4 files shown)


crates/vm/src/arch/aot/metered_execute.rs
```diff
@@ -199,9 +188,6 @@ where
 
             // a `None` slot at or above `pc_base`
             let Some(instruction) = instruction else {
-                asm_str += &format!("asm_execute_pc_{pc}:\n");
-                asm_str += &format!("    mov {REG_THIRD_ARG}, {pc}\n");
-                asm_str += "    jmp asm_dead_pc_handler\n";
```
GunaDD: is this just removed or replaced somewhere?

crates/vm/src/arch/aot/metered_execute.rs
```diff
@@ -174,14 +174,23 @@ where
...
+            let instruction = pc_idx.checked_sub(base_idx).and_then(|idx| {
+                exe.program.instructions_and_debug_infos[idx]
+                    .as_ref()
+                    .map(|(instruction, _)| instruction)
+            });
+
+            // a `None` slot at or above `pc_base`
+            let Some(instruction) = instruction else {
+                continue;
+            };
```
GunaDD, line 194: should we jump to the dead pc handler in the else case here ?

shuklaayush: we do jump but via a shared handle in the jump table. we just don't emit labels for each dead pc

GunaDD: I think that would be incorrect actually since in consecutive blocks of riscv opcodes, we don't go through the jump table and just let the instruction fall down to the next asm.

shuklaayush: ah that's a good catch. i forget that `aot` is just one big function that falls through

extensions/rv32im/circuit/src/jalr/tests.rs
```diff
@@ -423,11 +423,32 @@ fn test_aot_dispatch_rejects_dead_pc_slots() {
...
+            VmExe::new(Program::new_without_debug_infos_with_option(
+                &[
+                    Some(add_instruction(4, 0, 4)),
+                    Some(jalr_instruction(4)),
+                    Some(terminate_instruction()),
+                ],
+                8,
+            ))
+            .with_pc_start(8),
+            4,
```
GunaDD, line 444: shouldn't the `4` be `8` here ?

extensions/rv32im/circuit/src/jalr/tests.rs
```diff
@@ -423,11 +423,32 @@ fn test_aot_dispatch_rejects_dead_pc_slots() {
             .with_pc_start(0),
             0,
         ),
+        (
+            VmExe::new(Program::new_without_debug_infos_with_option(
+                &[Some(terminate_instruction())],
+                4096,
+            ))
+            .with_pc_start(2048),
```
GunaDD, line 431: should this be both `4096` or both `2048` ?

crates/vm/src/arch/aot/metered_execute.rs
```diff
@@ -335,14 +337,7 @@ where
...
             let pc = pc_idx as u32 * DEFAULT_PC_STEP;
-            let is_real = pc_idx
-                .checked_sub(base_idx)
-                .is_some_and(|idx| exe.program.instructions_and_debug_infos[idx].is_some());
-            if is_real {
-                asm_str += &format!("   .long asm_execute_pc_{pc} - map_pc_base\n");
-            } else {
-                asm_str += "   .long asm_dead_pc_handler - map_pc_base\n";
-            }
+            asm_str += &format!("   .long asm_execute_pc_{pc} - map_pc_base\n");
```
GunaDD, line 340: the behavior here differs from `pure.rs`

GunaDD (approved):
LGTM other than nit comments.

extensions/rv32im/circuit/src/jalr/tests.rs
shuklaayush: huh isn't it jumping to pc `4` which is invalid

extensions/rv32im/circuit/src/jalr/tests.rs
shuklaayush: it's testing the case when you jump to an unreachable instruction at the start

crates/vm/src/arch/aot/metered_execute.rs
shuklaayush: this is outdated

extensions/rv32im/circuit/src/jalr/tests.rs
GunaDD: ah yea sorry I thought it was `pc_base` instead of `dead_pc` in the second argument.

PR #2920, opened 2026-06-22
fix(v2.1): correct RV64 PC and address handling

shuklaayush:
## Summary

- Fix RV64 AUIPC sign-extension so `pc + sign_extend(imm << 12)` is handled consistently in execution, AIR, CUDA tracegen, and RVR lifting/emission.
- Fix JALR and load/store effective-address handling so RV64 register values are not silently truncated or wrapped into the implemented PC/memory address space.
- Restore missing load/store carry/sign constraints and add negative coverage for effective-address overflow.
- Harden RVR dynamic jumps so computed RV64 targets are bounds-checked before narrowing to the dispatch-table PC type.

resolves int-8294

### DIFF
--- a/crates/rvr/rvr-openvm/src/emit/codegen.rs
+++ b/crates/rvr/rvr-openvm/src/emit/codegen.rs
@@ -38,7 +38,7 @@ pub fn emit_instr(ctx: &mut EmitContext, instr: &Instr) {
             ctx.write_reg(*rd, &sext32(*value));
         }
         Instr::Auipc { rd, value } => {
-            ctx.write_reg(*rd, &sext32(*value));
+            ctx.write_reg(*rd, &hex_u64(*value));
         }
         Instr::Load {
             width,
@@ -122,7 +122,7 @@ pub fn emit_terminator(ctx: &mut EmitContext, term: &Terminator, pc: u32, tc: &T
             // Save base to a temporary when link_rd == rs1 to prevent
             // the link write from clobbering the jump target (e.g. jalr ra, ra, 0).
             let base = if link_rd.is_some_and(|rd| rd == *rs1) {
-                ctx.materialize_u32(&base)
+                ctx.materialize_u64(&base)
             } else {
                 base
             };
@@ -131,19 +131,28 @@ pub fn emit_terminator(ctx: &mut EmitContext, term: &Terminator, pc: u32, tc: &T
             }
             let imm_val = *imm;
             let next_pc = if imm_val == 0 {
-                ctx.materialize_u32(&format!("{base} & ~0x00000001u"))
+                ctx.materialize_u64(&format!("{base} & ~0x0000000000000001ull"))
             } else if imm_val > 0 {
-                ctx.materialize_u32(&format!(
-                    "({base} + {}) & ~0x00000001u",
-                    hex_u32(imm_val as u32)
+                ctx.materialize_u64(&format!(
+                    "({base} + {}) & ~0x0000000000000001ull",
+                    hex_u64(imm_val as u64)
                 ))
             } else {
-                let abs = (-(imm_val as i64)) as u32;
-                ctx.materialize_u32(&format!("({base} - {}) & ~0x00000001u", hex_u32(abs)))
+                let abs = (-(imm_val as i64)) as u64;
+                ctx.materialize_u64(&format!(
+                    "({base} - {}) & ~0x0000000000000001ull",
+                    hex_u64(abs)
+                ))
             };
-            ctx.write_line(&format!("state->pc = {next_pc};"));
             ctx.write_line(&format!(
-                "[[clang::musttail]] return dispatch_table[rv_dispatch_index({next_pc})]({args});"
+                "if (unlikely({next_pc} < RV_TEXT_START || {next_pc} > RV_TEXT_END)) {{"
+            ));
+            ctx.write_line(&format!("  [[clang::musttail]] return rv_trap({args});"));
+            ctx.write_line("}");
+            let next_pc_u32 = ctx.materialize_u32(&format!("(uint32_t){next_pc}"));
+            ctx.write_line(&format!("state->pc = {next_pc_u32};"));
+            ctx.write_line(&format!(
+                "[[clang::musttail]] return dispatch_table[rv_dispatch_index({next_pc_u32})]({args});"
             ));
         }
         Terminator::Branch {
@@ -284,8 +293,11 @@ fn imm_literal(imm: i32) -> String {
 
 /// Sign-extend a u32 value to a 64-bit C literal (uint64_t).
 fn sext32(value: u32) -> String {
-    let extended = value as i32 as i64 as u64;
-    format!("0x{extended:016x}ull")
+    hex_u64(value as i32 as i64 as u64)
+}
+
+fn hex_u64(value: u64) -> String {
+    format!("0x{value:016x}ull")
 }
 
 pub(super) fn hex_u32(value: u32) -> String {
--- a/crates/rvr/rvr-openvm/src/emit/codegen.rs
+++ b/crates/rvr/rvr-openvm/src/emit/codegen.rs
@@ -145,7 +145,7 @@ pub fn emit_terminator(ctx: &mut EmitContext, term: &Terminator, pc: u32, tc: &T
                 ))
             };
             ctx.write_line(&format!(
-                "if (unlikely({next_pc} < RV_TEXT_START || {next_pc} > RV_TEXT_END)) {{"
+                "if (unlikely(!rv_pc_is_dispatchable({next_pc}))) {{"
             ));
             ctx.write_line(&format!("  [[clang::musttail]] return rv_trap({args});"));
             ctx.write_line("}");
--- a/extensions/riscv/circuit/src/adapters/loadstore.rs
+++ b/extensions/riscv/circuit/src/adapters/loadstore.rs
@@ -476,16 +476,10 @@ where
                         // TODO: Remove loadstore read/write support for DEFERRAL_AS.
                         unreachable!("STORE to DEFERRAL_AS is unsupported");
                     }
-                    timed_write(memory, record.mem_as as u32, ptr, data.map(|x| x as u8)).0
+                    timed_write(memory, record.mem_as as u32, ptr, data).0
                 }
                 LOADD | LOADW | LOADB | LOADH | LOADWU | LOADBU | LOADHU => {
-                    timed_write(
-                        memory,
-                        RV64_REGISTER_AS,
-                        record.rd_rs2_ptr,
-                        data.map(|x| x as u8),
-                    )
-                    .0
+                    timed_write(memory, RV64_REGISTER_AS, record.rd_rs2_ptr, data).0
                 }
             };
         } else {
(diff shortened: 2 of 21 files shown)


crates/rvr/rvr-openvm/src/emit/codegen.rs
```diff
@@ -131,19 +131,28 @@ pub fn emit_terminator(ctx: &mut EmitContext, term: &Terminator, pc: u32, tc: &T
...
+                let abs = (-(imm_val as i64)) as u64;
+                ctx.materialize_u64(&format!(
+                    "({base} - {}) & ~0x0000000000000001ull",
+                    hex_u64(abs)
+                ))
             };
-            ctx.write_line(&format!("state->pc = {next_pc};"));
             ctx.write_line(&format!(
-                "[[clang::musttail]] return dispatch_table[rv_dispatch_index({next_pc})]({args});"
+                "if (unlikely(!rv_pc_is_dispatchable({next_pc}))) {{"
```
mansur20478, line 148: not sure if there should be temporary uint32_t variable for next_pc, i think the state->pc is already in 64bit

mansur20478 (commented):
rvr side looks good

crates/rvr/rvr-openvm/src/emit/codegen.rs
shuklaayush: i think this is addressed in your pr

mansur20478 (approved):
`rvr` part and `circuit/src` part looks good, `cuda/src` also seems to be correct. lgtm

extensions/riscv/circuit/src/auipc/core.rs
```diff
@@ -230,7 +231,10 @@ where
         let [pc_low, pc_high] = ptr_to_u16_limbs(record.from_pc);
 
         let rd_block = run_auipc(record.from_pc, record.imm);
-        let rd_u16 = [rd_block[0], rd_block[1]];
+        let rd_lo = rd_block[0];
+        let rd_hi = rd_block[1];
+        let is_sign_extend = rd_block[2] != 0;
+        let imm_sign = (imm_high_16 >> (U16_BITS - 1)) & 1;
```
GunaDD, line 237: am I right that this is correct only because imm fits in 20 bits ? so the 47th bit is 1 iff imm is negative.

extensions/riscv/circuit/src/auipc/core.rs
```diff
@@ -102,12 +102,13 @@ where
...
+        // Check that the computed sign matches the top bit of `imm_high_16`.
+        let imm_sign = is_sign_extend + carry_top;
         self.range_bus
             .range_check(
-                AB::Expr::from_u32(2) * rd_data[1]
-                    - is_sign_extend * AB::Expr::from_u32(1 << U16_BITS),
+                AB::Expr::from_u32(2) * imm_high_16 - imm_sign * AB::Expr::from_u32(1 << U16_BITS),
                 U16_BITS,
             )
             .eval(builder, is_valid);
```
GunaDD, line 114: hmm it doesn't look like the immediate are sign extended in the constraints ? is this a missing step now ?

extensions/riscv/circuit/src/loadstore/execution.rs
```diff
@@ -213,9 +213,9 @@ unsafe fn execute_e12_impl<
     let rs1_bytes: [u8; RV64_REGISTER_NUM_LIMBS] =
         exec_state.vm_read_bytes(RV64_REGISTER_AS, pre_compute.b as u32);
```
GunaDD, line 207: unrelated to the changes in this PR (related to the u16 memory bus change): should we be reading in `[u16; 4]` here since the memory adapter proves that instead of `[u8; 8]` reads and the proof generated by openvm should really be a proof of the execution.

extensions/riscv/circuit/src/adapters/loadstore.rs
```diff
@@ -378,23 +382,29 @@ where
...
+        let rs1_bytes = tracing_read(
             memory,
             RV64_REGISTER_AS,
             record.rs1_ptr,
             &mut record.rs1_aux_record.prev_timestamp,
-        ));
-        let rs1_val = record.rs1_val;
+        );
+        let rs1_val = try_rv64_bytes_to_u32(rs1_bytes).expect("upper 4 bytes must be zero");
+        record.rs1_val = rs1_val;
```
GunaDD, line 391: this is what the spec say in the audit doc: 
```
Spec: effective address is X[rs1] + sext(imm) at full XLEN (64-bit, mod 2^64); an implementation with narrow memory should then range-check / fault on an out-of-range address, not panic on the base.
```
it says it should then range-check / fault on an out-of-range address ? by fault here meaning get assertion fail on preflight ? previously it was assertion fail on E1/E2 ? so the change here is just moving the assertion from E1/E2 -> E3 ?

extensions/riscv/circuit/src/loadstore/tests.rs
```diff
@@ -602,6 +602,19 @@ fn negative_write_data_tests() {
...
+    run_negative_loadstore_test(
+        LOADBU,
+        Some([0xf8, 0xff, 0xff, 0xff, 0, 0, 0, 0]),
+        Some(16),
+        Some(0),
+        LoadStorePrankValues::default(),
+        false,
+    );
+}
+
```
GunaDD, line 580: this test is no longer valid after the 2^32 memory address PR.

crates/rvr/rvr-openvm/src/emit/codegen.rs
gdmlcjs, line 228:
```suggestion
            "[[clang::musttail]] return rv_trap({args});",
```
Bounds checking wasn't even done for static jumps...

gdmlcjs (approved):

extensions/riscv/circuit/src/jalr/core.rs
```diff
@@ -300,8 +301,13 @@ pub(super) fn run_jalr(
...
 ) -> (u32, [u16; BLOCK_FE_WIDTH]) {
-    let to_pc = rs1.wrapping_add(imm as u32 + (imm_sign as u32 * ((u16::MAX as u32) << U16_BITS)));
-    assert!(to_pc < (1 << PC_BITS));
+    let imm_extended = imm as u32 + (imm_sign as u32 * ((u16::MAX as u32) << U16_BITS));
+    let to_pc = rv64_address_add_imm(rs1, imm_extended);
+    assert!(
+        to_pc <= u64::from(MAX_ALLOWED_PC),
+        "JALR target exceeds implemented PC address space"
+    );
+    let to_pc = to_pc as u32;
```
GunaDD, line 311: is there really no circuit side changes due to this change ?

GunaDD (commented):
left some comments

crates/rvr/rvr-openvm/src/emit/codegen.rs
gdmlcjs: Or maybe we can just throw an exception. Just saw the related PR.

extensions/riscv/circuit/src/auipc/core.rs
shuklaayush: yes, for auipc imm is 20 bits (in riscv it's bits 12..31 of the instruction). not sure where 47 comes from. the sign bit is the highest bit of the immediate

extensions/riscv/circuit/src/loadstore/execution.rs
shuklaayush: no, because of two reasons
1. loadstore here still works in bytes
2. and more importantly, riscv (or openvm) pure execution (and to a large extent metered execution) doesn't care about how airs or buses are constructed. it only cares about (state, opcode) -> new state so it doesn't need to know anything about the internal `u16` representation of openvm memory

extensions/riscv/circuit/src/loadstore/tests.rs
shuklaayush: yeah, we can delete it after

extensions/riscv/circuit/src/jalr/core.rs
shuklaayush: there's an added constraint to check that the limb addition is signed 32 bit addition
```rust
builder.when(is_valid).assert_eq(carry, imm_sign);
``` 

and we already had the range checks for bounding `pc`
```rust
// Prevent to_pc overflow. to_pc_limbs[0] is 15 bits because it is
// multiplied by 2 when reconstructing the aligned target.
self.range_bus
    .range_check(to_pc_limbs[1], PC_BITS - U16_BITS)
    .eval(builder, is_valid);
self.range_bus
    .range_check(to_pc_limbs[0], U16_BITS - 1)
    .eval(builder, is_valid);
```

extensions/riscv/circuit/src/adapters/loadstore.rs
shuklaayush: yes we assert (i.e. panic) because the preflight interface doesn't allow for typed errors. it's still a `debug_assert` in e1/e2 (and not a full assert for performance reasons)

extensions/riscv/circuit/src/auipc/core.rs
shuklaayush: yeah, this is enforced, but not by having a separate sign-extended immediate column.

the circuit first proves the low 32-bit add:

```rust
pc + (imm << 8) = rd_low32 + carry_top * 2^32
```

then this part:

```rust
let imm_sign = is_sign_extend + carry_top;
range_check(2 * imm_high_16 - imm_sign * 2^16, 16);
```

forces `imm_sign` to be the top bit of the auipc immediate. so equivalently:

```rust
is_sign_extend = imm_sign - carry_top
```

which is the right relation for `pc + sext32(imm << 8)`.

if the immediate is positive, `imm_sign = 0`, so the result cannot need sign-extension and `is_sign_extend = 0`.

if the immediate is negative, `imm_sign = 1`. there are two cases:
- if `pc + imm` is still negative, then `is_sign_extend = 1`
- if `pc` is large enough to make `pc + imm` positive, then `is_sign_extend = 0`

`carry_top` is what distinguishes those two cases.

so sign extension is enforced by this carry/sign equation, not by explicitly constructing the sign-extended immediate.

crates/rvr/rvr-openvm/src/emit/codegen.rs
shuklaayush: yes for static jumps we can do the check at compile time and not add any runtime overhead
as in https://github.com/openvm-org/openvm/pull/2926

extensions/riscv/circuit/src/auipc/core.rs
GunaDD: I meant the 47th if we count starting from 0 (0 based indexing). So it's more like the 48th bit starting from 1. and that is because the MSB of imm_high_16 is the 48 th bit of the immediate.

extensions/riscv/circuit/src/auipc/core.rs
GunaDD: I honestly am having difficulty understanding the explanation

GunaDD: I agree that the range check constraints that `imm_sign` is correct, but I don't get the rest of it yet.

shuklaayush: there's also an implicit assumption here that `PC_BITS <= 31` so the `rd` value after sign extension can only have upper bits `0x00000000` or `0xffffffff`

GunaDD (approved):
LGTM

PR #2921, opened 2026-06-22
fix(v2.1): reject RV64 load/store access to deferral address space

shuklaayush:
## Summary

- reject RV64 loads unless they read from user memory (`e = 2`)
- reject RV64 stores unless they write to user memory (`e = 2`) or public values (`e = 3`)
- remove the obsolete AS 4 store pseudo-instruction path so the deferral address space is only maintained by the deferral extension
- update RVR lifting so public-values stores are handled by the IO extension, including byte/half/word/dword REVEAL widths
- improve invalid address-space error messages for load/store preflight paths
- store load/store `prev_data` as bytes now that this path no longer supports field-cell deferral writes
- update ISA/deferral docs and add regression coverage for invalid deferral-address-space load/store usage

resolves int-7698


extensions/riscv/circuit/src/adapters/loadstore.rs
mansur20478: i think if prev_data is u8, then this might need to be updated to u8

mansur20478: not sure if it is needed to be into u32 since prev_data is u8

mansur20478: not sure if it is needed to be into u32 since prev_data is u8

mansur20478: i think it might need to be also u8 since `run_write_data` returns in u8? not sure if I am correct

mansur20478 (commented):
not sure if `prev_data` change covers related code too, but other than that rvr side looks good. or it might be good to separate it into different pr since it is not critical?

extensions/riscv/circuit/src/adapters/loadstore.rs
shuklaayush: you're right. good catch

shuklaayush: you're correct

mansur20478 (approved):
looks good, lgtm

crates/rvr/rvr-openvm-lift/src/opcode.rs
```diff
@@ -111,42 +113,57 @@ pub fn lift_instruction<F: PrimeField32>(
...
-        // handled by `Rv64IoExtension`.
-        let addr_space = field_to_u32(insn.e);
-        if addr_space != AS_PUBLIC_VALUES {
-            return Some(lift_store(insn, pc, MemWidth::Double));
+        if is_memory_address_space(insn) {
+            return lift_store(insn, pc, MemWidth::Double);
+        }
+        if !is_public_values_address_space(insn) {
+            return None;
         }
```
gdmlcjs:
```suggestion
        if !is_public_values_address_space(insn) {
            return lift_store(insn, pc, MemWidth::Double);
}
```
This might be simpler.

crates/rvr/rvr-openvm-lift/src/opcode.rs
```diff
@@ -349,20 +366,34 @@ fn lift_muldiv<F: PrimeField32>(insn: &Instruction<F>, pc: u32, op: MulDivOp) ->
...
 /// OpenVM encoding: rd=a/4, rs1=b/4, imm low16=c, sign=g!=0
+#[inline]
+fn is_memory_address_space<F: PrimeField32>(insn: &Instruction<F>) -> bool {
+    field_to_u32(insn.e) == AS_MEMORY
+}
+
+#[inline]
+fn is_public_values_address_space<F: PrimeField32>(insn: &Instruction<F>) -> bool {
+    field_to_u32(insn.e) == AS_PUBLIC_VALUES
+}
```
gdmlcjs, line 381:
```suggestion
#[inline]
fn is_user_memory_instruction<F: PrimeField32>(insn: &Instruction<F>) -> bool {
    field_to_u32(insn.d) == AS_REGISTER && field_to_u32(insn.e) == AS_MEMORY
}

#[inline]
fn is_public_values_instruction<F: PrimeField32>(insn: &Instruction<F>) -> bool {
    field_to_u32(insn.d) == AS_REGISTER && field_to_u32(insn.e) == AS_PUBLIC_VALUES
}
```
The `d` operand is not checked only in rvr. Could we add the check here?

extensions/riscv/circuit/src/loadstore/aot.rs
```diff
@@ -134,7 +119,11 @@ fn generate_x86_asm_impl<F: PrimeField32>(
 
     let e_u32 = e.as_canonical_u32();
 
-    if d.as_canonical_u32() != RV32_REGISTER_AS || e_u32 == RV32_IMM_AS {
+    if d.as_canonical_u32() != RV32_REGISTER_AS
+        || e_u32 == RV32_IMM_AS
+        || e_u32 == RV32_REGISTER_AS
+        || e_u32 == DEFERRAL_AS
```
gdmlcjs:
```suggestion
        || e_u32 != PUBLIC_VALUES_AS
```

extensions/riscv/rvr/src/lib.rs
```diff
@@ -220,22 +222,33 @@ impl<F: PrimeField32> RvrExtension<F> for Rv64IoExtension {
...
-            return Some(LiftedInstr::Body(InstrAt {
-                pc,
-                instr: Instr::Ext(Box::new(RevealInstr {
-                    src_reg,
-                    ptr_reg,
-                    offset,
-                })),
-                source_loc: None,
-            }));
+        if insn.e.as_canonical_u32() == AS_PUBLIC_VALUES {
```
gdmlcjs:
```suggestion
        if insn.d.as_canonical_u32() == AS_REGISTER
            && insn.e.as_canonical_u32() == AS_PUBLIC_VALUES
```
`d` operand check like above

gdmlcjs (approved):
Looks good to me!

crates/rvr/rvr-openvm-lift/src/opcode.rs
shuklaayush: good point

extensions/riscv/circuit/src/loadstore/aot.rs
shuklaayush: i'm just gonna remove aot changes

crates/rvr/rvr-openvm-lift/src/opcode.rs
shuklaayush: good call

extensions/riscv/circuit/src/adapters/loadstore.rs
```diff
@@ -211,10 +211,14 @@ impl<AB: InteractionBuilder> VmAdapterAir<AB> for Rv64LoadStoreAdapterAir {
...
-        // and in {2, 3, 4} if the instruction is a store
-        builder.assert_tern(local_cols.mem_as - is_store * AB::Expr::TWO);
+        // Constrain loads to address space 2 and stores to address spaces 2 or 3.
+        let mem_as_minus_two = local_cols.mem_as - AB::Expr::TWO;
+        builder
+            .when(is_valid.clone())
+            .assert_bool(mem_as_minus_two.clone());
+        builder
+            .when(is_load.clone())
+            .assert_zero(mem_as_minus_two.clone());
```
GunaDD: why is this constraint duplicated ? is this an accident ?

GunaDD (approved):
LGTM other than the question

extensions/riscv/circuit/src/adapters/loadstore.rs
shuklaayush: it's `assert_bool` vs `assert_zero`

shuklaayush: actually i realize that they can be combined into one. can you take a look?

PR #2931, opened 2026-06-24
perf(v2.1): tighten memory metering estimates

shuklaayush:
## Summary

Resolves INT-8554.

This PR tightens metered memory-height accounting for the memory boundary, memory Merkle, and Poseidon2 AIRs.

Metering records which leaves inside each 64-leaf memory page were touched, then counts the Merkle work from those masks at checkpoint and segment boundaries. RVR and interpreted metering use the same Rust accounting path.

## Before

The estimate was page-oriented. If a few leaves inside a page were touched, the estimate charged page-level work from page-level state.

```text
one memory page = 64 Merkle leaves

leaf index:  0  1  2  3  4  5  ... 63
accesses:    .  x  .  x  .  .  ... .

page-oriented estimate:

   [page subtree]
      /     \
    ...     ...        page was touched,
   /  \     /  \       estimate used page-level state
  L0  L1   L2  L3
      x        x
```

Shared upper-tree ancestors across nearby pages were also charged along each page path.

```text
page A -> P -> G -> root
page B -> P -> G -> root
          ^    ^    ^
          shared ancestors
```

That made the estimate safe, but it could overshoot the rows needed by `MemoryMerkleAir` and the related Poseidon2 hashes.

## Now

Each memory page is exactly 64 Merkle leaves. The page-local leaf occupancy is stored in one `u64` mask.

```text
full memory tree

        [root]
        /    \
      ...    ...              ancestors above pages are shared
      /        \
   [page]    [page]
   / ... \   / ... \
 L0 ... L63 L0 ... L63        each page stores one u64 leaf mask
```

For each access, metering records a page-table index and a per-page leaf mask:

```text
field      type   meaning
page_id    u32    index into page_masks
leaf_mask  u64    one occupancy bit per leaf in that 64-leaf page
```

`page_masks[page_id]` stores the accumulated `leaf_mask` for that page.

```text
page_masks: Box<[u64]>

page_id ----------+
                  v
page_masks[page_id] = accumulated u64 occupancy mask for that page
```

Across pages, upper-tree ancestors are tracked with a separate bitset of ancestor node ids.

```text
new estimate for two nearby pages

        root              counted once in upper_nodes bitset
         |
         G                counted once in upper_nodes bitset
         |
         P                counted once in upper_nodes bitset
        / \
   page A page B          page-local work counted from each u64 mask
```

## Counting

For leaves, the update is bit operations on one `u64`:

```text
old page mask:       00001000
incoming access:     00101000
new page mask:       00101000  = old | incoming
newly touched bits:  00100000  = incoming & !old
new leaves:          popcnt(newly touched bits)
```

For internal nodes inside a 64-leaf page, the code uses two page-local paths.

Single-leaf path:

```text
leaf = trailing_zeros(added_mask)
```

For that leaf, the code checks the aligned groups that map to the page-local ancestor levels:

```text
2-leaf group   -> level 1 ancestor
4-leaf group   -> level 2 ancestor
8-leaf group   -> level 3 ancestor
16-leaf group  -> level 4 ancestor
32-leaf group  -> level 5 ancestor
64-leaf page   -> page root
```

For a group size `G`, the group containing `leaf` is checked with:

```text
group_start = leaf & !(G - 1)
group_mask  = ((1 << G) - 1) << group_start
is_new_node = (old_mask & group_mask) == 0
```

The first touched leaf in an aligned group creates that group's ancestor node. Later leaves in the same group reuse the counted ancestor. The first touched leaf in the page creates the page root.

Multi-leaf path:

```text
old_mask    = previous page occupancy
added_mask  = newly touched leaves

for each of the 6 page-local levels:
    old_mask   = OR-reduce old_mask into parent groups
    added_mask = OR-reduce added_mask into parent groups
    new_nodes += popcnt(added_mask & !old_mask)
```

The OR-reduction step is a fixed shift/mask operation:

```text
children:    0 1   1 0   0 0   1 1
next level:    1     1     0     1

mask = (mask | (mask >> shift)) & representative_bit_mask
```

This counts parent groups that have at least one newly added leaf and had no old leaf under that group. A 64-leaf page has six levels up to the page root, so the multi-leaf path is a fixed six-level sequence with `count_ones()` at each level.

The resulting height updates are:

```text
Boundary rows = 2 * new_leaves
Merkle rows   = 2 * new_merkle_nodes
Poseidon2     = 2 * new_leaves + 2 * new_merkle_nodes
```

The Poseidon2 count remains a safe upper bound because trace generation can deduplicate equal hash inputs by value.

## RVR

The RVR tracer records the same page/leaf-mask information in generated C and applies it through the same `MemoryCtx` logic.

```text
generated C tracer
        |
        v
page/leaf-mask buffers
        |
        v
Rust MemoryCtx accounting
        |
        v
same height updates as interpreted metering
```

The normal memory fast path coalesces consecutive accesses to the same page before flushing to Rust. That keeps the common path cheap while keeping RVR and interpreted metering on the same estimate.

## Performance

This PR also removes avoidable overhead from metered execution:

- the interpreter countdown fast path uses a lightweight decrement before the full segmentation check
- repeated same-page memory accesses are coalesced before checkpoint replay
- single-leaf updates use a small aligned-group check
- multi-leaf updates use a fixed six-level `added_mask` OR-reduction
- call sites use the real height-update operation directly

## Testing

- `RUSTC_WRAPPER= cargo test -p openvm-circuit arch::execution_mode::metered --lib`
- `RUSTC_WRAPPER= cargo test -p openvm-circuit --features rvr rvr_metered_ctx_parts_roundtrip_preserves_execution_state --lib`
- `RUSTC_WRAPPER= cargo check --profile fast -p openvm-circuit --features rvr`
- `RUSTC_WRAPPER= cargo check --profile fast -p rvr-openvm`
- `git diff --check`
- [reth benchmark comparison](https://github.com/axiom-crypto/openvm-eth/actions/runs/28176871048)

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
@@ -8,8 +8,8 @@ use crate::{
     system::memory::{dimensions::MemoryDimensions, DIGEST_WIDTH},
 };
 
-/// Number of bits needed to index the 64 leaves represented by one page mask.
-pub const PAGE_BITS: usize = 6;
+/// Number of bits needed to index the leaves represented by one `u64` page mask.
+pub const PAGE_BITS: usize = u64::BITS.ilog2() as usize;
 
 /// Upper bound on number of memory pages accessed per instruction. Used for buffer allocation.
 pub const MAX_MEM_PAGE_OPS_PER_INSN: usize = 1 << 16;
--- a/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
@@ -142,7 +142,10 @@ impl BitSet {
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 pub struct PageAccess {
+    /// Scalar page index. The leaf occupancy for this page is stored in
+    /// `leaf_mask`.
     pub page_id: u32,
+    /// Bit `i` is set when leaf `i` inside this 64-leaf page was touched.
     pub leaf_mask: u64,
 }
 
@@ -197,6 +200,8 @@ impl MemoryPageTracker {
         if old_mask == 0 {
             self.dirty_pages.push(page_id);
         }
+        // Newly set bits are the only leaves that can add boundary rows or new
+        // page-local Merkle nodes.
         let added_mask = new_mask ^ old_mask;
         let leaves = added_mask.count_ones();
         let mut merkle_nodes = if leaves <= SPARSE_MERKLE_DELTA_LEAF_THRESHOLD {
@@ -242,6 +247,8 @@ fn leaf_mask_range(start: u32, end: u32) -> u64 {
 
 #[inline(always)]
 fn parent_mask(mut mask: u64) -> u64 {
+    // Collapse 64 child bits into 32 parent bits. Repeated calls walk the
+    // 64-leaf page up to its root without iterating over every child bit.
     mask |= mask >> 1;
     mask &= 0x5555_5555_5555_5555;
     mask = (mask | (mask >> 1)) & 0x3333_3333_3333_3333;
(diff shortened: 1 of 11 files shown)


gdmlcjs:
Can the page ids stay `u32` for now?

shuklaayush:
> Can the page ids stay u32 for now?

yes since we only support memory addresses up to 29-bit here

crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -99,43 +142,227 @@ impl BitSet {
...
+    page_masks: Box<[u64]>,
+    dirty_pages: Vec<usize>,
+    upper_nodes: BitSet,
+    upper_height: usize,
+    merkle_nodes: u32,
+    leaves: u32,
+}
+
+impl MemoryPageTracker {
+    pub fn new(num_pages: usize, upper_height: usize) -> Self {
```
gdmlcjs: I think we can remove `num_pages` and use `1 << upper_height` instead.

gdmlcjs (approved):
All the logic seems correct!

crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
shuklaayush: good point

PR #2933, opened 2026-06-25
refactor(rvr): move ext_hint_stream_set function from RV64I extension to core rvr

mansur20478:
move `ext_hint_stream_set` function from RV64I extension to core rvr. `ext_hint_stream_set` is also used by other extensions' phantom instructions, so it might be better to have it in core rvr.

Changes that were done:
1) move `host_hint_stream_set` into `crates/vm/src/arch/rvr/io` and `ext_hint_stream_set` into `openvm_io.c`
2) modify `register_openvm_io_ctx ` to register `host_hint_stream_set`.

resolves INT-8214

### DIFF
--- a/crates/rvr/rvr-openvm/c/openvm_io.h
+++ b/crates/rvr/rvr-openvm/c/openvm_io.h
@@ -1,6 +1,14 @@
 #ifndef OPENVM_IO_H
 #define OPENVM_IO_H
 
+#include <stdint.h>
+
 void* openvm_get_io_ctx(void);
+
+/* Called once at execute-start to install the field-typed hint-stream writer. */
+void register_hint_stream_set_fn(void (*fn)(void*, const uint8_t*, uint32_t));
+
+/* Replace the hint stream contents. Called by extension FFI staticlibs. */
+void ext_hint_stream_set(const uint8_t* data, uint32_t len);
 
 #endif /* OPENVM_IO_H */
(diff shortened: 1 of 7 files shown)


crates/rvr/rvr-openvm/c/openvm_io.h
```diff
@@ -1,6 +1,14 @@
 #ifndef OPENVM_IO_H
 #define OPENVM_IO_H
 
+#include <stdint.h>
+
 void* openvm_get_io_ctx(void);
 
+/* Called once at execute-start to install the field-typed hint-stream writer. */
+void register_hint_stream_set_fn(void (*fn)(void*, const uint8_t*, uint32_t));
```
gdmlcjs: We can probably remove this since it's not called from other C files.

crates/vm/src/arch/rvr/execute.rs
gdmlcjs: Could we update this?

gdmlcjs (approved):
Everything else looks great!

crates/rvr/rvr-openvm/c/openvm_io.h
mansur20478: resolved in https://github.com/openvm-org/openvm/pull/2933/commits/76a6901f7b46a8ddf81359e7a39182749bd6e467

crates/vm/src/arch/rvr/execute.rs
mansur20478: resolved in https://github.com/openvm-org/openvm/pull/2933/commits/e3eba57a133f4414cd062e10c3685902685c6fe1

gdmlcjs (approved):

shuklaayush (approved):

PR #2934, opened 2026-06-25
feat(rvr): use guarded memory in rvr mode

mansur20478:
use GuardedMemory allocator in rvr mode. GuardedMemory is a memory allocator that creates guarded pages before and after the allocated region, making OS level error when the memory access goes out-of-bounds. The current implementation applies GuardedMemory to all address spaces in rvr mode. There might be a need for better redesign of current PR.

Changes that were done:
1) move GuardedMemory definition from `crates/rvr/rvr-state/src/memory.rs` to `crates/vm/src/system/memory/online/guarded.rs`
2) add LinearMemory trait to GuardedMemory and make it similar to MmapMemory implementation
3) allow memory allocation with 0 bytes (address space 0, used for immediate values, create 0-size memory). The current approach creates two 16kb guarded memory pages with nothing in-between. -> might be better to change to how MmapMemory handles it
4) add memory boundary checks, not sure if it is needed, was added to follow MmapMemory implementation. might be good idea to remove?

resolves INT-8088


crates/vm/src/system/memory/online/guarded.rs
```diff
@@ -0,0 +1,203 @@
...
+
+impl Drop for GuardedMemory {
+    fn drop(&mut self) {
+        unsafe {
+            let _ = munmap(self.region, self.total_size);
+        }
+    }
+}
+
+impl LinearMemory for GuardedMemory {
```
gdmlcjs, line 147: Could we add an optimization of `LinearMemory::fill_zero` like in `MmapMemory`?

crates/vm/src/system/memory/online.rs
```diff
@@ -14,17 +14,27 @@ use crate::{
...
 pub use basic::*;
+#[cfg(all(unix, feature = "rvr", not(feature = "basic-memory")))]
+pub use guarded::*;
 #[cfg(any(unix, windows))]
 pub use memmap::*;
 pub use paged_vec::PagedVec;
 
-#[cfg(all(any(unix, windows), not(feature = "basic-memory")))]
+#[cfg(all(unix, not(feature = "basic-memory"), feature = "rvr"))]
+pub type MemoryBackend = guarded::GuardedMemory;
```
gdmlcjs, line 34: Is there a reason `GuardedMemory` is only used when "rvr" is set?

mansur20478: i am not sure `GuardedMemory`, but i think it was originally supposed to be for rvr mode. Should it be for all modes?

crates/vm/src/system/memory/online/guarded.rs
mansur20478: sure

crates/vm/src/system/memory/online.rs
mansur20478: should I add a feature to use guarded memory, basic memory, or mmap memory?

crates/vm/src/system/memory/online/guarded.rs
mansur20478: resolved in https://github.com/openvm-org/openvm/pull/2934/commits/d310dd3fb43abcbffd65b43946a3657fe6cde8e1

crates/vm/src/system/memory/online.rs
gdmlcjs: From the ticket description I think it's to use `GuardedMemory` whenever possible? Unless @shuklaayush meant something else. Currently `GuardedMemory` isn't even used anywhere so I think we're just replacing `MmapMemory`.

gdmlcjs: I think what we have now (just "basic-memory") is enough. We probably don't have a use for more features now.

shuklaayush: we can probably use guarded memory everywhere but let's do that later

shuklaayush (approved):

PR #2938, opened 2026-06-26
refactor: replace on_check function pointer with direct extern call

mansur20478:
Replace `on_check` function pointer in Tracer with direct extern call. Previously, a pointer to the function was stored in the struct and assigned at runtime.

By making it extern, the call site becomes a direct named call. The compiler can better understand and optimize.

resolves INT-7348


mansur20478:
When going with the extern approach, it starts failing in CI. The reason is that the compiled artifact, `.so`, is a separate compiled file that gets loaded at runtime. When it loads, it needs to find `metered_periodic_check` - a function that `on_check` pointed before, by name in the host process. To make it work, we need to pass an extra linker flag (-Wl,--export-dynamic) that we currently don't use. Not sure if we should modify the Makefile to make it work, or if there is better approach

shuklaayush:
going to close this for now

PR #2942, opened 2026-06-29
feat: paged h2d transfer

GunaDD:
Closes INT-8439

### DIFF
--- a/crates/vm/src/system/memory/online/touched_pages.rs
+++ b/crates/vm/src/system/memory/online/touched_pages.rs
@@ -1,0 +1,148 @@
+use super::PAGE_SIZE;
+
+/// Tracks which fixed-size pages of an address space's linear memory may contain non-zero data,
+/// for the GPU host-to-device transfer. Pages that are *not* marked are guaranteed zero and can be
+/// skipped by the transport (which zero-fills the device buffer first), so this is always a
+/// conservative **superset** of the truly non-zero pages: we may copy a still-zero page, but we
+/// must never skip a non-zero one. Pages are [`PAGE_SIZE`] bytes, matching the mmap page size.
+#[derive(Debug, Clone)]
+pub enum TouchedPages {
+    /// Conservative default: every page may be non-zero, so the transport copies everything.
+    /// Always correct (used when no narrow page information is available).
+    All,
+    /// Only the marked page indices may be non-zero. `bits` is a little-endian bitset over pages.
+    Marked { bits: Vec<u64>, num_pages: usize },
+}
+
+impl TouchedPages {
+    /// Conservative default: all pages are considered possibly non-zero.
+    #[inline]
+    pub fn all() -> Self {
+        TouchedPages::All
+    }
+
+    /// No pages marked, with capacity for an address space of `num_bytes` bytes.
+    #[inline]
+    pub fn none(num_bytes: usize) -> Self {
+        let num_pages = num_bytes.div_ceil(PAGE_SIZE);
+        TouchedPages::Marked {
+            bits: vec![0u64; num_pages.div_ceil(u64::BITS as usize)],
+            num_pages,
+        }
+    }
+
+    /// Marks every page overlapping the byte range `[start, start + len)` as possibly non-zero.
+    /// No-op for [`TouchedPages::All`] (which already covers every page).
+    #[inline]
+    pub fn mark_byte_range(&mut self, start: usize, len: usize) {
+        if len == 0 {
+            return;
+        }
+        if let TouchedPages::Marked { bits, num_pages } = self {
+            let first = start / PAGE_SIZE;
+            let last = (start + len - 1) / PAGE_SIZE;
+            debug_assert!(last < *num_pages, "byte range out of address space bounds");
+            for page in first..=last {
+                bits[page / 64] |= 1u64 << (page % 64);
+            }
+        }
+    }
+
+    /// Yields the half-open **byte** ranges `[start, end)` of maximal runs of consecutive marked
+    /// pages, clamped to `total_bytes`. Coalescing adjacent pages into runs minimizes the number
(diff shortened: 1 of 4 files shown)


crates/vm/src/system/memory/online/touched_pages.rs
```diff
@@ -0,0 +1,150 @@
...
+/// skipped by the transport (which zero-fills the device buffer first).
+/// Pages are [`PAGE_SIZE`] bytes, matching the mmap page size.
+#[derive(Debug, Clone)]
+pub enum TouchedPages {
+    /// Conservative default: every page may be non-zero, so the transport copies everything.
+    /// Always correct (used when no narrow page information is available).
+    All,
+    /// Only the marked page indices may be non-zero. `bits` is a little-endian bitset over pages.
+    Marked { bits: Vec<u64>, num_pages: usize },
+}
```
shuklaayush, line 15: do we need both? if marked works well enough let's just keep that

crates/vm/src/system/memory/online/touched_pages.rs
```diff
@@ -0,0 +1,150 @@
+use super::PAGE_SIZE;
+
+/// Tracks which fixed-size pages of an address space's linear memory may contain non-zero data,
+/// for the GPU host-to-device transfer. Pages that are *not* marked are guaranteed zero and can be
+/// skipped by the transport (which zero-fills the device buffer first).
+/// Pages are [`PAGE_SIZE`] bytes, matching the mmap page size.
+#[derive(Debug, Clone)]
+pub enum TouchedPages {
```
shuklaayush: maybe `PageSet`

crates/vm/src/system/memory/online/touched_pages.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// Conservative default: every page may be non-zero, so the transport copies everything.
+    /// Always correct (used when no narrow page information is available).
+    All,
+    /// Only the marked page indices may be non-zero. `bits` is a little-endian bitset over pages.
+    Marked { bits: Vec<u64>, num_pages: usize },
```
shuklaayush: can you make it a `Box` instead of a `Vec` if size is fixed

crates/vm/src/system/memory/online/touched_pages.rs
```diff
@@ -0,0 +1,150 @@
...
+impl TouchedPages {
+    /// Conservative default: all pages are considered possibly non-zero.
+    #[inline]
+    pub fn all() -> Self {
+        TouchedPages::All
+    }
+
+    /// No pages marked, with capacity for an address space of `num_bytes` bytes.
+    #[inline]
+    pub fn none(num_bytes: usize) -> Self {
```
shuklaayush: `new`?`

crates/vm/src/system/memory/online/touched_pages.rs
GunaDD: I think `TouchedPages` is better considering there is `TouchedMemory` and I think `TouchedPages` actually makes more sense

crates/vm/src/system/memory/online/touched_pages.rs
GunaDD: Okay, removing `All`.

crates/vm/src/system/memory/online/touched_pages.rs
GunaDD: sounds good

shuklaayush (approved):

crates/vm/src/system/memory/online/touched_pages.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+        let first = start / PAGE_SIZE;
+        let last = (start + len - 1) / PAGE_SIZE;
+        debug_assert!(
+            last < self.num_pages,
+            "byte range out of address space bounds"
+        );
+        for page in first..=last {
+            self.bits[page / 64] |= 1u64 << (page % 64);
+        }
```
Golovanov399: Noting that it can be done 64 times faster

crates/vm/src/system/memory/online/touched_pages.rs
```diff
@@ -0,0 +1,128 @@
...
+        let mut page = 0;
+        while page < self.num_pages {
+            if is_set(page) {
+                let run_start = page;
+                while page < self.num_pages && is_set(page) {
+                    page += 1;
+                }
+                let start_byte = run_start * PAGE_SIZE;
+                let end_byte = (page * PAGE_SIZE).min(total_bytes);
+                if start_byte < end_byte {
```
Golovanov399: Should we instead just go up to `self.num_pages.min(total_bytes.div_ceil(PAGE_SIZE))` or something and remove this if?

Golovanov399 (approved):
Thanks overall for this feature!

Below are some small comments

crates/vm/src/system/memory/online/touched_pages.rs
shuklaayush: yeah, good catch
@GunaDD see `insert_range` here. also use bit operations and not `/ 64` or `% 64`. compiler will optimize it in this case but still better to be explicit with the intent
https://github.com/openvm-org/openvm/blob/a1a5755c1a0cfade65f4d4f1131e65882828deab/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs#L68-L138

GunaDD: ah yea I missed this. on the `/` and `%`, claude says the compiler already optimizes it and its better to put it as `/` and `%` for readability. Is this fine ?

crates/vm/src/system/memory/online/touched_pages.rs
GunaDD: updated this part, thanks!

PR #2944, opened 2026-06-29
perf(v2.1): improve poseidon2 memory metering

shuklaayush:
## Summary

This PR makes Poseidon2 memory metering closer to the rows that trace generation actually emits.

A memory segment has two trees: the memory tree at the start of the segment and the memory tree at the end of the segment. Metering already counts the leaves and Merkle nodes touched by the segment. This PR also checks whether each touched leaf or node was already non-default at the latest checkpoint.

That matters for Poseidon2 because default memory values share hash inputs. If many old leaves or old internal nodes are still default, trace generation can reuse the same Poseidon2 row for those default inputs instead of emitting one row per node.

## Model

The code uses two memory trackers:

```text
segment_memory
    clears at segment start
    counts leaves and Merkle nodes touched by this segment

global_memory
    persists across checkpoints
    tracks memory that is already known to be non-default
```

Each memory page stores one `u64` leaf mask:

```text
one page = 64 memory leaves

leaf index:   0  1  2  3  ... 63
leaf_mask:    1  0  1  0  ...  0
```

For each leaf or Merkle node counted in the segment, the trackers answer two questions:

```text
segment_memory: is this new in the current segment?
global_memory:  was this already non-default at the checkpoint?
```

The result is:

```text
new in segment + already non-default -> count one old Poseidon2 row for this node
new in segment + still default       -> count it in the shared default bucket
```

A first touch looks like this:

```text
checkpoint / old side              segment end / new side

          D3                                X3
        /    \                            /    \
      D2      D2          write         X2      D2
     /  \    /  \        ------>       /  \    /  \
   D1   D1 D1   D1                    X1  D1 D1   D1
   |                                  |
default leaf                       touched leaf

D = default node input shared with other default nodes at the same height
X = node changed by this segment and counted normally
```

Default rows are shared by value:

```text
default leaves                     -> 1 old Poseidon2 row
default internal nodes at height h -> 1 old Poseidon2 row for that height
```

So the estimate is:

```text
new_rows           = segment_leaves + segment_merkle_nodes
old_nondefault     = new_rows - first_touch_default_nodes
old_default_rows   = has_default_leaf + count(default_internal_heights)

Poseidon2 rows = new_rows + old_nondefault + old_default_rows
```

`first_touch_default_nodes` means leaves and Merkle nodes that are touched in this segment and were still default at the checkpoint. `default_internal_heights` is a bit mask of Merkle heights that had at least one such default internal node.

## Initial Memory

The global tracker is seeded from the executable's sparse initial memory image:

```text
nonzero initial byte -> containing memory leaf is non-default
zero initial byte    -> leaf stays default
```

For `DEFERRAL_AS`, sparse image bytes are converted back to deferral field-cell units before mapping them to Merkle leaves.

## RVR

RVR uses the same Rust accounting path as the interpreter.

```text
generated C page buffers
        |
        v
PageAccess { page_id, leaf_mask }
        |
        v
MemoryCtx
        |
        +--> segment_memory: Boundary/Merkle rows for this segment
        +--> global_memory: old Poseidon2 rows for default vs non-default memory
```

The generated C hot path is kept narrow:

- aligned normal memory loads/stores record one memory leaf directly
- page buffers pass `PageAccess { page_id, leaf_mask }` into Rust
- extension calls that already trace their memory do not flush and reload local page state again
- public-values and deferral tracing skip normal-memory tracer work when the address space is not `AS_MEMORY`

## Performance

RVR reth benchmark comparison for block `23992138`:

```text
latest origin/develop-v2.1.0-rv64: 75 segments, compile 266.322s, execute 1.999s
this branch:                         74 segments, compile 266.230s, execute 1.932s
```

The branch reduces the segment count by one on this benchmark without increasing compile-metered time.

## Testing

- `RUSTC_WRAPPER= cargo test --profile fast -p rvr-openvm aligned_memory_access_uses_single_leaf_trace --lib`
- `RUSTC_WRAPPER= cargo test --profile fast -p openvm-circuit metered --lib`
- `git diff --check`
- RVR reth metered benchmark against latest `origin/develop-v2.1.0-rv64`

Resolves INT-8554

### DIFF
--- a/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
@@ -55,6 +55,7 @@ fn push_page_access(accesses: &mut Vec<PageAccess>, page_id: u32, leaf_mask: u64
 }
 #[derive(Clone, Debug)]
 pub struct MemoryCtx {
+    /// Memory tree dimensions used to map address-space ranges into global leaf ids.
     memory_dimensions: MemoryDimensions,
     /// Segment-local occupancy. Cleared at segment boundaries; it prevents
     /// charging the same leaf or Merkle node twice inside one segment.
@@ -62,15 +63,20 @@ pub struct MemoryCtx {
     /// Committed occupancy at the last safe checkpoint, seeded with nonzero
     /// initial memory. This tells us whether an old value is canonical default.
     occupancy_tracker: CommittedMemoryOccupancyTracker,
+    /// Page masks recorded since the last safe checkpoint by the normal metered path.
     pub page_indices_since_checkpoint: Vec<PageAccess>,
+    /// Length mirror used by generated metered code that appends through the raw buffer.
     pub page_indices_since_checkpoint_len: usize,
+    /// Number of checkpoint-buffer entries already reflected in `trace_heights`.
     page_indices_applied_len: usize,
-    /// New leaves already reflected in `trace_heights` and queued for the next
-    /// checkpoint commit. Segment replay clears this queue and recomputes
-    /// heights from `page_indices_since_checkpoint`.
+    /// Newly charged leaf masks queued for the next checkpoint commit. Segment
+    /// replay clears this queue and recomputes it from `page_indices_since_checkpoint`.
     pending_occupancy_updates: Vec<PageAccess>,
+    /// New boundary leaves accumulated by direct page-buffer application.
     pending_leaves: u32,
+    /// New Merkle nodes accumulated by direct page-buffer application.
     pending_merkle_nodes: u32,
+    /// Old-side default leaves and nodes paired with the pending height deltas.
     pending_default_old: DefaultOldAccounting,
 }
 
--- a/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
+++ b/crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
@@ -7,7 +7,8 @@ use openvm_instructions::{
 };
 
 use super::page_tracker::{
-    leaf_mask_range, DefaultOldCounts, MemoryOccupancyTracker, MemoryPageTracker,
+    leaf_mask_range, CommittedMemoryOccupancyTracker, DefaultOldAccounting,
+    SegmentMemoryPageTracker,
 };
 pub use super::page_tracker::PageAccess;
 use crate::{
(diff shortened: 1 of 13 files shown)


crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -522,254 +267,224 @@ impl MemoryCtx {
...
+        self.page_indices_applied_len = len;
+
+        if leaves == 0 && merkle_nodes == 0 {
+            return;
         }
-        self.page_indices_applied_len = self.page_indices_since_checkpoint.len();
 
         debug_assert!(trace_heights.len() >= 2);
         let poseidon2_idx = trace_heights.len() - 2;
+        let initial_default_poseidon_rows = global_first_touches.estimated_default_poseidon_rows();
```
gdmlcjs: The default poseidon rows seem to be added for each (1000 instruction) interval in the segment. Could we ignore these duplicates?

crates/vm/src/arch/execution_mode/metered/memory_tracker.rs
```diff
@@ -0,0 +1,799 @@
...
+/// A global first touch has a canonical default initial-side value. These
+/// counts let `MemoryCtx` estimate Poseidon2 rows after default-row dedup.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub(super) struct GlobalFirstTouchCounts {
+    /// New segment leaves whose initial-side value is the canonical default leaf.
+    pub(super) leaves: u32,
+    /// New segment internal nodes whose initial-side value is a canonical default node.
+    pub(super) merkle_nodes: u32,
+    /// Bit `h` is set when height `h` has at least one first-touch internal node.
+    merkle_height_mask: u64,
```
gdmlcjs: Would it perhaps be simpler to change this to something like "max_height" instead of "mask" since the elements of the bit mask are all contiguous?

gdmlcjs (approved):
No errors found!

crates/vm/src/arch/execution_mode/metered/memory_tracker.rs
shuklaayush: good point

crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
shuklaayush: yeah, good catch. removed it

PR #2947, opened 2026-06-29
feat: add parameter sweep tests

Aalanli:
Resolves: INT-8260, INT-8456

I have since updated stack-backend to export more types publicly so we can construct SystemParams here. The current CI workflow on the SDK side is

generate param lists for internal only, internal + root and root
run the sweep on internal only without root prover
run the sweep on internal + root with either evm / root prover
run the root with either evm / root prover
CLI is fairly straightforward, the tests now include the integration tests in the workflow.

note that this PR also includes 2 fixes that guard against invalid param settings

one guards against the fact that certain param combos makes internal_for_leaf circuits and the recursive circuits different which causes problems
another guards against l_skip > 5 which is not supported by the CUDA backend


crates/sdk/src/prover/agg.rs
```diff
@@ -81,6 +81,13 @@ impl AggProver {
     ) -> Self {
         assert!(agg_tree_config.num_children_leaf <= MAX_NUM_CHILDREN_LEAF);
         assert!(agg_tree_config.num_children_internal <= MAX_NUM_CHILDREN_INTERNAL);
+        #[cfg(feature = "cuda")]
+        {
+            assert!(
+                agg_config.params.leaf.l_skip <= 5,
+                "l_skip greater than 5 is not supported in the cuda backend"
+            )
+        }
```
stephenh-axiom-xyz: No longer need this, please remove

Aalanli:
Hi @stephenh-axiom-xyz, I'm not sure what you meant by that comment to include tests by the feature flag (eg. `prove-root`). Because shouldn't this already be done? I added the filter test feature because I don't want to run all of them when doing a basic sweep.

Aalanli:
Latest run: https://github.com/openvm-org/openvm/actions/runs/28479215581

.github/workflows/cli.yml
```diff
@@ -75,34 +86,52 @@ jobs:
...
             echo "USE_LOCAL_OPENVM=0" >> $GITHUB_ENV
           else
             echo "USE_LOCAL_OPENVM=1" >> $GITHUB_ENV
           fi
 
-      - name: Run CLI tests
+      - name: Run CLI integration tests
+        # Runs the 8 non-ignored tests in crates/cli/tests/integration.rs. They
+        # spawn the scripts in tests/scripts/ which call the cargo-openvm built
+        # above. --test-threads=1 keeps the shared fibonacci build cache coherent.
```
stephenh-axiom-xyz: nit: Description is a bit specific, ideally whenever possible we don't have comments that become stale when another file is changed

.github/workflows/cli.yml
stephenh-axiom-xyz: Should move this to the CUDA workflow now that we have one

.github/workflows/cli.cuda.yml
```diff
@@ -0,0 +1,149 @@
...
+            echo "CUDA libraries not found!"
+            exit 1
+          fi
+
+      - name: Build cargo-openvm (CUDA)
+        # Build once with --features cuda, then put it on PATH so subsequent steps
+        # and the integration tests (which set SKIP_INSTALL=1) all use the same
+        # GPU-enabled binary.
+        run: |
+          cargo build --locked --profile fast -p cargo-openvm --features cuda
```
stephenh-axiom-xyz: Add `halo2-gpu` feature

.github/workflows/sdk.cuda.yml
```diff
@@ -61,12 +111,127 @@ jobs:
...
+            build_cuda_tests "parallel,root-prover,cuda"
+          fi
+          if [[ "${{ inputs.sweep_evm || false }}" == "true" ]]; then
+            build_cuda_tests "parallel,root-prover,evm-verify,cuda"
+          fi
+          if [[ "${{ inputs.sweep_internal || false }}" != "true" && "${{ inputs.sweep_root || false }}" != "true" && "${{ inputs.sweep_evm || false }}" != "true" ]]; then
+            build_cuda_tests "cuda,root-prover"
+          fi
+
+      - name: Generate sweep JSON files if missing
```
stephenh-axiom-xyz: Do you not need to set working directory for this?

.github/workflows/sdk.cuda.yml
```diff
@@ -1,6 +1,34 @@
...
+        type: boolean
+        default: false
+      sweep_root:
+        description: "Run SDK root parameter sweep instead of normal SDK CUDA tests"
+        type: boolean
+        default: false
+      sweep_internal:
+        description: "Run SDK internal parameter sweep instead of normal SDK CUDA tests"
+        type: boolean
+        default: false
```
stephenh-axiom-xyz: Much cleaner to have an option that sets the workflow mode and only have one, i.e. `default`, `sweep_internal`, `sweep_root`, or `sweep_evm`. This workflow (and the non-CUDA one below) is likely to be long, and we should force different sweep types to be on different machines. This makes the testing faster (i.e. 3 shorter processes instead of 1 long one) and helps with granularity

.github/workflows/sdk.cuda.yml
```diff
@@ -61,12 +111,127 @@ jobs:
...
+        working-directory: crates/sdk
+        run: |
+          rustup component add rust-src --toolchain nightly-2026-01-18
+
+          TEST_FILTERS=()
+          if [[ -n "${{ inputs.test_name }}" ]]; then
+            IFS=',' read -ra TEST_FILTERS <<< "${{ inputs.test_name }}"
+          fi
+
+          build_cuda_tests() {
```
stephenh-axiom-xyz: Continuing on from the comment above, let's only run one of the sweeps per workflow

.github/workflows/sdk.cuda.yml
```diff
@@ -61,12 +111,127 @@ jobs:
...
+            build_cuda_tests "parallel,root-prover,evm-verify,cuda"
+          fi
+          if [[ "${{ inputs.sweep_internal || false }}" != "true" && "${{ inputs.sweep_root || false }}" != "true" && "${{ inputs.sweep_evm || false }}" != "true" ]]; then
+            build_cuda_tests "cuda,root-prover"
+          fi
+
+      - name: Generate sweep JSON files if missing
+        if: ${{ inputs.sweep_root || inputs.sweep_internal || inputs.sweep_evm }}
+        run: |
+          if [[ ! -f tests/internal_sweep_no_root.json || ! -f tests/internal_sweep_root.json || ! -f tests/root_params.json ]]; then
```
stephenh-axiom-xyz: I think we can always just force the re-generation, in practice this will always be true

.github/workflows/sdk.cuda.yml
```diff
@@ -61,12 +111,127 @@ jobs:
             exit 1
           fi
 
+      - name: Build SDK CUDA tests
```
stephenh-axiom-xyz: Why is building a separate step before running? Seems like a lot of extra code

.github/workflows/sdk.cuda.yml
```diff
@@ -61,12 +111,127 @@ jobs:
...
+          fi
+
+      - name: Generate sweep JSON files if missing
+        if: ${{ inputs.sweep_root || inputs.sweep_internal || inputs.sweep_evm }}
+        run: |
+          if [[ ! -f tests/internal_sweep_no_root.json || ! -f tests/internal_sweep_root.json || ! -f tests/root_params.json ]]; then
+            cargo run --bin generate-param-sweep -- crates/sdk/tests/
+          fi
+
       - name: Run SDK CUDA tests
```
stephenh-axiom-xyz, line 64: This step is long and confusing, add comments for clarification

.github/workflows/sdk.yml
stephenh-axiom-xyz, line 1: Same comments as on the CUDA file

crates/sdk/src/tests.rs
```diff
@@ -33,22 +35,58 @@ cfg_if::cfg_if! {
...
+}
+
 /// Creates a fibonacci SDK with standard test parameters.
 fn make_fib_sdk() -> (Sdk, SystemParams, AggregationSystemParams) {
-    let n_stack = 19;
-    let app_params = app_params_with_100_bits_security(DEFAULT_APP_L_SKIP + n_stack);
-    let agg_params = AggregationSystemParams::default();
-    let sdk = Sdk::riscv32(app_params.clone(), agg_params.clone());
-    (sdk, app_params, agg_params)
+    let (app_params, agg_params, _root_params) = get_params(); // get_overriden_params(test_json);
```
stephenh-axiom-xyz: nit: remove the comment

crates/sdk/src/tests.rs
```diff
@@ -33,22 +35,58 @@ cfg_if::cfg_if! {
...
 }
 
+fn get_params() -> (SystemParams, AggregationSystemParams, SystemParams) {
+    let n_stack = 19;
+    let app_params = get_params_from_env("APP_PARAMS_OVERRIDE", || {
+        app_params_with_100_bits_security(DEFAULT_APP_L_SKIP + n_stack)
+    });
+    let agg_params = AggregationSystemParams {
+        leaf: get_params_from_env("LEAF_PARAMS_OVERRIDE", || {
+            leaf_params_with_100_bits_security()
```
stephenh-axiom-xyz: Firstly, can we just pass in `leaf_params_with_100_bits_security` instead of the lambda? Secondly, why not just pass the SystemParams (the perf isn't bad)

crates/sdk/src/tests.rs
```diff
@@ -109,7 +147,7 @@ fn make_multi_deferral_circuit_prover_with_count(
     num_deferral_circuits: usize,
 ) -> MultiDeferralCircuitProver {
     assert!(num_deferral_circuits > 0);
-    let def_circuit_params = internal_params_with_100_bits_security();
+    let def_circuit_params = agg_params.internal.clone();
```
stephenh-axiom-xyz, line 152: This was intentionally not set to `agg_params.internal.clone()`, as the `def_circuit` is a separate layer. Revert this and add a comment

crates/sdk/src/tests.rs
```diff
@@ -433,13 +452,33 @@ fn test_verify_stark_path_sdk_can_verify_own_proofs() -> Result<()> {
...
+        let openvm_verifier = sdk.generate_halo2_verifier_solidity()?;
+        let _gas_cost = Sdk::verify_evm_halo2_proof(&openvm_verifier, evm_proof, Some(app_commit))?;
+    }
 
     Ok(())
 }
 
+#[cfg(feature = "root-prover")]
 #[test]
 fn test_deferrals_enabled_without_usage() -> Result<()> {
```
stephenh-axiom-xyz, line 433: This test can probably have the 3 branches also

crates/sdk/tests/.gitignore
stephenh-axiom-xyz, line 1: We don't have a nested `.gitignore` anywhere, move these to the root

crates/sdk/tests/run_sdk_sweep.py
```diff
@@ -0,0 +1,364 @@
...
+import threading
+from typing import Any
+
+
+SDK_DIR = Path(__file__).resolve().parents[1]
+TESTS_DIR = SDK_DIR / "tests"
+LOG_ROOT = TESTS_DIR / "logs" / "sdk_tests"
+
+
+def cargo_test_cmd(mode: str, evm=True, root=True) -> list[str]:
```
stephenh-axiom-xyz, line 22: Similar to the SDK workflow, let's make `evm` and `root` (and `stark`) mutually exclusive.

crates/sdk/tests/run_sdk_sweep.py
stephenh-axiom-xyz, line 1: Remove the 2. Also this script runs sweeps only, so the name should reflect that (ex. `run_sdk_sweep.py`)

stephenh-axiom-xyz (commented):
Mostly looks good with some comments. In particular, let's make it so that only one sweep per workflow can happen (instead of potentially all three in a single one).

> Hi @stephenh-axiom-xyz, I'm not sure what you meant by that comment to include tests by the feature flag (eg. prove-root). Because shouldn't this already be done? I added the filter test feature because I don't want to run all of them when doing a basic sweep.

On this, consider adding a helper that does the following block:

```rust
    #[cfg(not(feature = "root-prover"))]
    {
        let (proof, baseline) = vs_sdk.prove(vs_exe, vs_stdin, &[def_input])?;
        Sdk::verify_proof((*vs_sdk.agg_vk()).clone(), baseline, &proof)?;
    }
    #[cfg(all(feature = "root-prover", not(feature = "evm-verify")))]
    {
        let mut evm_prover = vs_sdk.evm_prover_without_halo2(vs_exe)?;
        let vs_proof = evm_prover.prove_root(vs_stdin, &[def_input])?;
        let vk = evm_prover.root_prover.0.get_vk();
        let engine = RootE::new(vk.inner.params.clone());
        engine.verify(&vk, &vs_proof)?;
    }
    #[cfg(feature = "evm-verify")]
    {
        let app_commit = vs_sdk.app_commit(vs_exe.clone())?;
        let mut evm_prover = vs_sdk.evm_prover(vs_exe)?;
        let vs_proof = evm_prover.prove_evm(vs_stdin, &[def_input])?;

        let openvm_verifier = vs_sdk.generate_halo2_verifier_solidity()?;
        let _gas_cost = Sdk::verify_evm_halo2_proof(&openvm_verifier, vs_proof, Some(app_commit))?;
    }
```

.github/workflows/sdk.cuda.yml
Aalanli: keep build out of logs

.github/workflows/sdk.cuda.yml
Aalanli: No, it runs from the root dir

Aalanli:
https://github.com/openvm-org/openvm/actions/runs/28545641110

stephenh-axiom-xyz (approved):
LGTM

PR #2949, opened 2026-06-29
chore: update word size or limb related comments in rvr

mansur20478:
Update word size or limb related comments in rvr. Since rvr was originally made for rv32, there are comments that still refer in 4 bytes.

Changes:
1) Update OpenVM encoding related comments to use 8 bytes
2) Remove RV32 and 32-bit word size related comments in rvr
3) Update HINT_RANDOM phantom instruction description to fill the hint stream with 8 instead of 4 multiplied by `reg[num_words_reg]`

resolves INT-8322


mansur20478:
@claude please review this pr, are there any other comments related to 4 bytes when it should by 8 bytes left? or where we say word and then 4 bytes?

shuklaayush (approved):

PR #2952, opened 2026-06-30
ci(v2-rc3): add stark-backend patching workflow setup

shuklaayush:
## Summary
- Track `.cargo/config.toml` with `git-fetch-with-cli` and a commented-out SSH patch block for `stark-backend`.
- Load `GH_ACTIONS_DEPLOY_PRIVATE_KEY` before workflow checkouts so SSH patch URLs can resolve in CI.
- Stop ignoring `.cargo/config.toml` so the rc3 patching setup is available on the branch.

## Validation
- Parsed all workflow YAML files.
- Checked every `actions/checkout` has an earlier SSH-agent step in the same job.
- Ran `git diff --check`.


stephenh-axiom-xyz (approved):
LGTM, although please make sure that adding a permanent config is idiomatic

PR #2954, opened 2026-06-30
feat: config for deferral-aware but not active SDK

stephenh-axiom-xyz:
Resolves INT-8570.

## Overview

This branch adds a deferral-aware-but-not-active SDK mode. SDK users can now build STARK/root/Halo2 verifying material that is bound to deferral hook commits without also supplying a deferral prover capable of proving non-empty deferral inputs.

The branch also fixes `UnsetPvsAir` trace generation for padding rows and tightens root keygen/dummy-proving behavior used for root trace-height discovery.

Base: `origin/develop-v2.0.0-rc.3`

Commits in scope:

- `20f1ac38b0 feat: config for deferral-aware but not active SDK`
- `7802fde0cf fix: UnsetPvsAir tracegen fills all proof_idx columns`
- `f4e2f426b5 fix: compute_root_proof_heights uses keygen group`

## SDK Deferral Setup

Files:

- `crates/sdk/src/lib.rs`
- `crates/sdk/src/builder.rs`
- `crates/sdk/src/prover/evm.rs`

The SDK now stores deferral configuration as `DeferralSetup` instead of `Option<Arc<DeferralAggProver>>`.

Modes:

- `Disabled`: no deferral hook commits; non-empty deferral inputs are rejected.
- `Aware(DeferralHookCommits)`: vkeys/baselines/root/Halo2 material are deferral-aware, but non-empty deferral inputs are rejected because no prover is present.
- `Active(Arc<DeferralAggProver>)`: deferral-aware and can prove non-empty deferral inputs.

Reviewer focus:

- `DeferralSetup::hook_cached_commit`, `hook_commit`, and `prover` are now the common access points for SDK, STARK, root, and EVM prover setup.
- `StarkProver::prove` treats both `Aware` and `Active` as deferral-enabled for baseline/merkle-proof shape, but only `Active` can handle non-empty `def_inputs`.
- `GenericSdk::deferral_agg_prover()` exposes the active prover only when present.
- `GenericSdkBuilder::deferral_hook_commits` is mutually exclusive with the existing deferral prover entry points.

## Deferral Hook Commit Derivation

File:

- `crates/sdk/src/prover/deferral/agg.rs`

Adds serializable `DeferralHookCommits` and `DeferralHookCommits::from_system_params`.

The hook commit derivation logic is factored into `deferral_hook_artifacts_from_system_params`, which constructs the dummy deferral hook path once and returns:

- hook cached commit,
- hook commit,
- hook VK,
- deferral aggregation prover seeded by that hook VK.

`DeferralAggProver::verify_stark` now reuses those artifacts, then checks that the real verify-stark deferral hook VK/preprocessed commit matches the dummy-derived fixed point.

Reviewer focus:

- The aware-only path must produce the same aggregation VK as the active verify-stark path when supplied with matching commits.
- Cached SDK proving keys still include deferral proving keys only for active deferral setups.

## Root Keygen And Dummy Proofs

Files:

- `crates/sdk/src/keygen/dummy.rs`
- `crates/sdk/src/prover/root.rs`
- `crates/sdk/src/lib.rs`

`compute_root_proof_heights` moved from `prover/root.rs` to `keygen/dummy.rs` and now returns both root trace heights and the generated root proving key.

The function now:

- takes structural SDK inputs (`SystemConfig`, aggregation params/tree config, root params, `DeferralSetup`) rather than an existing app proving key and aggregation prover,
- builds a trivial `TERMINATE` app proving key and aggregation prover internally,
- uses the CPU root engine for root keygen/trace-height discovery,
- labels the dummy app proof metrics as `root_keygen`.

`GenericSdk::root_prover` uses the returned trace heights and root proving key to construct the cached `RootProver`.

Reviewer focus:

- Root trace heights remain tied to the same system config, aggregation params, aggregation tree config, and deferral hook commits as the SDK being built.
- The dummy proof metrics should not appear as an `app_proof` benchmark group.

## Continuations Tracegen Fix

File:

- `crates/continuations/src/circuit/inner/unset/trace.rs`

`UnsetPvsAir` trace generation now fills `proof_idx` for padding rows from `num_valid..height`, not only for valid rows.

Reviewer focus:

- Padding rows now have deterministic `proof_idx` values instead of leaving that column unset.

## Tests

File:

- `crates/sdk/src/tests.rs`

Added coverage:

- `test_deferral_aware_sdk_with_odd_children`: builds an aware-only SDK with odd aggregation tree children and proves with empty deferral inputs.
- `test_deferral_aware_and_active_have_equivalent_vks`: checks that aware-only and active verify-stark SDKs produce equivalent aggregation VKs when using the same hook commits.

Adjusted coverage:

- Deferral prover access now goes through `deferral_agg_prover()`.
- Static verifier profiling setup uses the new `compute_root_proof_heights` signature and `DeferralSetup::Disabled`.

### DIFF
--- a/crates/continuations/src/circuit/inner/unset/trace.rs
+++ b/crates/continuations/src/circuit/inner/unset/trace.rs
@@ -30,5 +30,11 @@ pub fn generate_proving_ctx(
         cols.proof_idx = F::from_usize(*proof_idx);
     }
 
+    for proof_idx in num_valid..height {
+        let chunk = chunks.next().unwrap();
+        let cols: &mut UnsetPvsCols<F> = chunk.borrow_mut();
+        cols.proof_idx = F::from_usize(proof_idx);
+    }
+
     AirProvingContext::simple_no_pis(RowMajorMatrix::new(trace, width))
 }
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -45,7 +45,7 @@ use openvm_verify_stark_host::{
 use crate::{
     config::{AggregationConfig, AggregationSystemParams, AggregationTreeConfig},
     keygen::{AggPrefixProvingKey, AggProvingKey, SdkCachedProvingKey},
-    prover::{AggProver, AppProver, DeferralAggProver, StarkProver},
+    prover::{AggProver, AppProver, DeferralAggProver, DeferralHookCommits, StarkProver},
     types::{AppExecutionCommit, ExecutableFormat},
 };
 #[cfg(feature = "evm-prove")]
@@ -152,7 +152,7 @@ where
     #[cfg(feature = "root-prover")]
     root_prover: OnceLock<Arc<RootProver>>,
 
-    def_agg_prover: Option<Arc<DeferralAggProver>>,
+    deferral_setup: DeferralSetup,
 
     #[cfg(feature = "evm-prove")]
     #[getset(get = "pub")]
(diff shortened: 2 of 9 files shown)


crates/continuations/src/circuit/inner/unset/trace.rs
```diff
@@ -30,5 +30,11 @@ pub fn generate_proving_ctx(
         cols.proof_idx = F::from_usize(*proof_idx);
     }
 
+    for proof_idx in num_valid..height {
```
luffykai, line 33: this is a separate fix?

crates/sdk/src/lib.rs
```diff
@@ -163,6 +163,45 @@ where
...
 }
 
+#[derive(Clone, Default)]
+pub enum DeferralSetup {
+    #[default]
+    /// Deferrals are disabled for this GenericSdk. The STARK, root, and halo2 vks are deferral-
+    /// unaware, and trying to prove with def_inputs set causes an error.
+    Disabled,
+    /// The STARK, root, and halo2 vks are deferral-aware (i.e. are equivalent to Active), but
+    /// trying to prove with def_inputs set still causes an error.
```
luffykai: maybe should explain when to use vs Disabled. otherwise they both say "trying to prove with def_inputs set causes an error."

luffykai (approved):

PR #2956, opened 2026-07-01
feat(ci): add action to release `cargo-openvm` binaries

shuklaayush:
towards int-4393

we can't test it until the action exists on `main` so i'll defer testing to after we merge the v2 branch to main


stephenh-axiom-xyz (approved):
LGTM, but let's be sure to actually do a dry run first

PR #2958, opened 2026-07-01
feat: cargo openvm verify app takes in an app commit

stephenh-axiom-xyz:
Resolves INT-8607.


luffykai (approved):

crates/cli/src/commands/verify.rs
```diff
@@ -252,40 +260,41 @@ impl VerifyCmd {
...
                 Sdk::verify_evm_halo2_proof(&evm_verifier, evm_proof, app_commit)?;
             }
         }
         println!("Proof verified successfully!");
         Ok(())
     }
 }
+
+fn read_app_commit(
+    app_commit: &Option<Option<PathBuf>>,
```
shuklaayush, line 273: nit: i would use an enum to be explicit but it's not a big deal 

```rust
AppCommitArg::{Unchecked, DefaultPath, Explicit(PathBuf)
```

shuklaayush (approved):

PR #2962, opened 2026-07-02
fix: verify-stark prover uses its own cached commit

stephenh-axiom-xyz:
Resolves INT-8610.

## Overview

This branch fixes verify-stark deferral input generation so the input hash uses the cached commit for the verify-stark deferral circuit itself, not the cached commit from the child proof baseline. The cached commit is now derived from SDK/prover state instead of being emitted as a separate artifact.

The branch also adds verify-stark example workflow coverage and fixes the `examples/i256` guest by importing the bigint guest intrinsics crate.

Base used for this summary: `develop-v2.0.0-rc.3` (`15c6048042`).

## Verify-Stark Deferral Cached Commit

Files:

- `crates/continuations/src/prover/deferral/mod.rs`
- `guest-libs/verify-stark/circuit/src/prover/mod.rs`
- `guest-libs/verify-stark/circuit/src/extension.rs`
- `guest-libs/verify-stark/circuit/src/lib.rs`

What changed:

- `DeferralCircuitProver` now has a default `cached_commits()` hook returning an empty list.
- The verify-stark deferral circuit prover overrides `cached_commits()` to return its verifier circuit cached commitment via `get_cached_commit()`.
- `get_raw_deferral_results()` and `get_deferral_state()` now require the cached commit to use when computing deferral input state.
- A shared `default_verify_stark_circuit_params()` helper exposes the internal params used for the verify-stark circuit.

Reviewer focus:

- Check that deferral input hashing is bound to the verify-stark deferral circuit's own cached commitment.
- Check that requiring the cached commit at deferral-state construction sites is the right API boundary.
- Check whether `cached_commits()` as a trait hook is the right surface for future deferral circuits that expose one or more cached commitments.

## SDK Commit Derivation

Files:

- `crates/sdk/src/lib.rs`
- `crates/sdk/src/tests.rs`

What changed:

- `GenericSdk::deferral_circuit_cached_commits(def_idx)` derives cached commits for an active deferral circuit from the SDK's deferral prover.
- SDK verify-stark test helpers no longer return the verify-stark cached commit as a separate value.
- Verify-stark input helpers take the verify-stark SDK and derive the needed cached commit internally before calling `get_raw_deferral_results()` / `get_deferral_state()`.
- `test_verify_many_deferrals` uses non-default deferral circuit params for additional coverage.

Reviewer focus:

- Check that tests derive the cached commit from the same SDK/prover that will prove the deferral input.
- Check the recursive verify-stark path where one SDK verifies its own proofs.
- Check the multi-deferral test assumption that each configured verify-stark deferral circuit has the same cached commit.

## Verify-Stark Example Host Flow

Files:

- `examples/verify-stark/host/src/lib.rs`
- `examples/verify-stark/host/src/bin/main.rs`
- `examples/verify-stark/verify-book-examples.sh`

What changed:

- The host `prove()` path reconstructs the SDK from the cached proving key and derives the verify-stark deferral circuit cached commit via `deferral_circuit_cached_commits()`.
- `verify_stark_guest_inputs()` uses that derived cached commit when computing raw deferral results and deferral state.
- The CLI no longer writes or accepts a `verify-stark.cached-commit.json` artifact.
- The book-example verification script no longer stores or threads a cached-commit file through `keygen` and `prove`.
- The script allows `VERIFY_STARK_HOST_FEATURES=` to mean "build with no extra features"; the default remains `cuda`.

Reviewer focus:

- Check that `keygen` artifacts are still sufficient for later `build` and `prove` invocations.
- Check that the child proof's baseline still comes from the proof being verified, while the cached commit is derived from the verify-stark SDK reconstructed from cached proving keys.
- Check CLI/backward-compatibility expectations because the cached-commit command-line argument was removed.

## CI Trigger

File:

- `.github/workflows/cli.cuda.yml`

What changed:

- The CUDA CLI workflow now runs when files under `examples/verify-stark/**` change.

Reviewer focus:

- Confirm this is the right workflow coverage for changes to the verify-stark host example.

## i256 Example Compile Fix

Files:

- `examples/i256/Cargo.toml`
- `examples/i256/src/main.rs`

What changed:

- Adds `openvm-bigint-guest` with `export-intrinsics` to the i256 example dependencies.
- Adds a local patch for the monorepo path.
- Imports `openvm_bigint_guest as _` so bigint intrinsics are linked for the guest.

Reviewer focus:

- Check that the dependency and patch match how other examples import guest extension intrinsics.
- Check that this remains copyable outside the monorepo via the git dependency.


guest-libs/verify-stark/circuit/src/extension.rs
```diff
@@ -92,7 +93,8 @@ pub fn get_raw_deferral_results(
             let final_ts_state = *ts.into_log().perm_results().last().unwrap();
             let (left_ts, right_ts) = poseidon2_input_to_digests(final_ts_state);
             let ts_commit = poseidon2_compress_with_capacity(left_ts, right_ts).0;
-            let cached_commit = vk.baseline.internal_recursive_vk_commit.cached_commit;
+            let cached_commit =
+                cached_commit.unwrap_or(vk.baseline.internal_recursive_vk_commit.cached_commit);
```
luffykai: is it correct to fall back to vk here?

examples/verify-stark/host/src/bin/main.rs
```diff
@@ -39,37 +40,54 @@ struct KeygenCmd {
     /// [output] this SDK's aggregation vk.
     #[arg(long, required = true)]
     agg_vk: PathBuf,
+
+    /// [output] cached commit used to generate verify-stark guest inputs.
```
luffykai: is it really necessary to produce a new artifact?

luffykai (commented):
Is the deferral circuit cached commit not somehow derivable? or we have to keygen and store it, and pass that in?

luffykai (approved):

PR #2973, opened 2026-07-06
perf(v2.1): add separate chips for immediate bitwise, comparison, and shift instructions

mansur20478:
## Summary

- Adds separate chips for `XORI`, `ORI`, `ANDI`, `SLTI`, `SLTIU`, `SLLI`, `SRLI`, and `SRAI`. These instructions previously shared chips with their register versions.
- Splits the shared adapter into one adapter for register instructions and another for immediate instructions. The immediate adapter reads one register and takes the second input directly from the instruction.
- Updates the opcodes, transpiler, executor, RVR, CPU and CUDA trace generation, and tests to use the new chips.


extensions/riscv/circuit/src/addi/execution.rs
```diff
@@ -13,17 +13,7 @@ use openvm_instructions::{
 use openvm_stark_backend::p3_field::PrimeField32;
 
 use super::core::AddIExecutor;
```
GunaDD, line 15: I think there is a naming inconsistency here because other immediate chips use the `_imm` suffix where this one just use `i`

extensions/riscv/circuit/src/shift_right_arithmetic_imm/core.rs
```diff
@@ -0,0 +1,304 @@
...
+    fn start_offset(&self) -> usize {
+        self.offset
+    }
+}
+
+#[repr(C)]
+#[derive(AlignedBytesBorrow, Debug)]
+pub struct ShiftRightArithmeticImmCoreRecord<const NUM_LIMBS: usize, const LIMB_BITS: usize> {
+    pub b: [u16; NUM_LIMBS],
+    pub shamt: u8,
```
GunaDD, line 173: could we have a better naming instead of `shamt` ? maybe `shift_amount` ?

shuklaayush: yeah you can change it but `shamt` is common terminology in isa context. like `rd` for destination register

extensions/riscv/circuit/src/addi/execution.rs
shuklaayush: yeah, but `addi` is special because it's also the name of the single instruction that this air handles. i don't have a strong opinion either way

extensions/riscv/circuit/src/shift_right_arithmetic_imm/core.rs
GunaDD: ah hmm then maybe its fine

GunaDD (approved):
LGTM

PR #2975, opened 2026-07-06
refactor: convert input and hint streams to byte-backed storage

mansur20478:
Refactor input and hint stream to use byte-backed storage. The goal of this PR is to remove field-element typed streams since elements inside are still treated as bytes. The PR consist of following parts:
1. updating input and hint to byte-backed storage.
2. removing generic <F> from the Stream, Stdin and its associated structs. #2983
3. re-read documentation and ensure that it matches the updated code.

This PR, first one, mainly focused on updating `input_stream` to `VecDeque<Vec<u8>>` and `hint_stream` to `VecDeque<u8>`, updating relevant code and removing `F::from_u8`. Relevant documentation was also slightly updated to match the refactoring. `Rv64HintInput` description got also updated to match the refactoring.

`write_field` was deleted, as it might be redundant. If we interpret it as little endian, then I feel like there would be the almost no difference between `write_field` and `write_bytes`.

towards INT-8549

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4476,7 +4476,6 @@ dependencies = [
  "openvm-stark-sdk",
  "openvm-transpiler",
  "openvm-verify-stark-host",
- "p3-field",
  "rand_chacha 0.3.1",
  "tiny-keccak",
  "tracing",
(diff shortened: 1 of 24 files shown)


mansur20478:
@claude perform review on this pr

shuklaayush (approved):

PR #2978, opened 2026-07-07
docs: add P256 ECDSA verification example

stephenh-axiom-xyz:
Resolves INT-8682.


examples/ecdsa/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+edition = "2021"
+
+[workspace]
+members = []
+
+[dependencies]
+openvm = { git = "https://github.com/openvm-org/openvm.git", features = [
+    "std",
+] }
+openvm-algebra-guest = { git = "https://github.com/openvm-org/openvm.git" }
```
mansur20478, line 13: looks good, just not sure if we should add tag = "v2.0.0" to match K256 example

mansur20478 (approved):
looks good, no issues. Not critical, but we can add tag = "v2.0.0" to match K256 example, but since the version 2 will be the main one, i think not necessary. lgtm

examples/ecdsa/Cargo.toml
stephenh-axiom-xyz: I've been removing these lol, as they don't match the example

mansur20478: lol

PR #2979, opened 2026-07-07
chore: add 0x prefix to remaining EvmProof fields

stephenh-axiom-xyz:
Resolves INT-8677


HrikB (approved):

mansur20478 (commented):
looks good, i tested it locally, it puts 0x. lgtm

PR #2980, opened 2026-07-07
docs: writing apps + some guest library book fixes

stephenh-axiom-xyz:
Resolves INT-8674, INT-8675, INT-8681, INT-8683


mansur20478 (approved):
looks good, lgtm

PR #2983, opened 2026-07-08
refactor: remove generic <F> from streams, stdin, and relevant structs

mansur20478:
This is the second part of the original PR https://github.com/openvm-org/openvm/pull/2975. This part focuses on removing generic <F> from `Streams` and `Stdin` as the storage is byte-backed now. As result, `VmState` and `VmStateMut` no longer need <F> as they had it due to `Streams<F>`. Same goes for `VmExecState`, which had <F> due to `VmState` having it.

towards INT-8549


mansur20478:
@claude review this PR, make sure that relevant structs and code removed <F>. please, check its cleanliness and if there are points for improvements. please be prudent

shuklaayush (approved):

PR #2984, opened 2026-07-09
feat: GPU trace generation for FieldExpr (via a new symbolic expression compiler) and ModularIsEqual chip

yi-sun:
# Design

```text
┌─────────────────────────────────────────────────────────────┐
│                    Mod Builder Frontend                     │
│                                                             │
│  ExprBuilder / FieldVariable / SymbolicExpr                 │
│                                                             │
│  Example:                                                   │
│      z = select(is_mul, x * y, x / y)                       │
└──────────────────────────────┬──────────────────────────────┘
                               │ finalized FieldExpr
                               ▼
┌─────────────────────────────────────────────────────────────┐
│                    Tracegen IR Compiler                     │
│                                                             │
│  compile_tracegen_ir()                                      │
│                                                             │
│  • validates CUDA capabilities                              │
│  • lowers expression trees into linear operation tapes      │
│  • derives Montgomery constants                             │
│  • calculates quotient and carry bounds                     │
│  • allocates value slots and per-constraint scratch space   │
└──────────────────────────────┬──────────────────────────────┘
                               │ Result<TracegenIr, Error>
                               ▼
┌─────────────────────────────────────────────────────────────┐
│                        TracegenIr                           │
│                                                             │
│  Value tape                         Limb tapes              │
│  ──────────                         ──────────               │
│  LoadInput / Constant               Input / Var / Constant  │
│  Add / Sub / Mul / Div              Add / Sub               │
│  IntAdd / IntMul                    Mul (limb convolution)   │
│  Select / SaveVar                   IntAdd / IntMul / Select │
│                                                             │
│  Metadata                                                   │
│  ────────                                                   │
│  • field and Montgomery constants                           │
│  • constraint tape boundaries                               │
│  • quotient/carry sizes and bounds                          │
│  • opcode-to-flag mapping                                   │
│  • scratch and trace dimensions                             │
└──────────────────────────────┬──────────────────────────────┘
                               │ encode()
                               ▼
┌─────────────────────────────────────────────────────────────┐
│                    Encoded Vec<u32> Blob                    │
│                                                             │
│  • fixed header and section offsets                         │
│  • operation tapes                                          │
│  • constants and metadata                                   │
└──────────────────────────────┬──────────────────────────────┘
                               │ upload once
                               ▼
┌─────────────────────────────────────────────────────────────┐
│                  CUDA Tracegen Interpreter                  │
│                                                             │
│  One thread per trace row:                                  │
│                                                             │
│  1. Decode opcode/inputs                                    │
│  2. Execute value tape                                      │
│  3. Execute limb tapes                                      │
│  4. Calculate q and carries                                 │
│  5. Emit range checks                                       │
│  6. Write trace columns                                     │
└──────────────────────────────┬──────────────────────────────┘
                               │ actual trace
                               ▼
┌─────────────────────────────────────────────────────────────┐
│                   Equivalence Validation                    │
│                                                             │
│  CUDA trace == production CPU tracegen                      │
│  (`FieldExpressionFiller`)                                  │
└─────────────────────────────────────────────────────────────┘
```

The encoder and CUDA decoder share a generated ABI:

```text
                        tracegen_abi.def
                               │
                            build.rs
                         ┌─────┴─────┐
                         ▼           ▼
                 Rust constants   CUDA constants
                         │           │
                         ▼           ▼
                      encode()    field_expr.cu
```

The runtime data flow is:

```text
Dense execution record
[adapter record | opcode byte | input limbs]
                      │
                      ▼
             CUDA interprets TracegenIr
                      │
                      ▼
Proof trace row
[adapter columns | is_valid | inputs | vars | quotients | carries | flags]
```

Closes INT-6995


Aalanli:
Please link a benchmark, and a nsys profile.

GunaDD:
@Aalanli -- these are from commit 9ea... but the changes afterwards are just refactors so should not affect perf:
https://github.com/axiom-crypto/openvm-eth/actions/runs/30232025529
https://github.com/axiom-crypto/openvm-eth/actions/runs/30285430045
https://github.com/axiom-crypto/openvm-eth/actions/runs/30285502126
https://github.com/axiom-crypto/openvm-eth/actions/runs/30285557832
https://github.com/axiom-crypto/openvm-eth/actions/runs/30285585620
https://github.com/axiom-crypto/openvm-eth/actions/runs/30285610917

GunaDD:
on the nsys profile -- I'm currently re-running openvm-eth on CI to get the nsys profile

GunaDD:
Running the nsys profile here: https://github.com/axiom-crypto/openvm-eth/actions/runs/30406693237

Aalanli:
Great, thanks! 
Please take a look at these:
image
image

There seem to be a lot of memcpy H2D interleaved with kernel launches. Is there any reason why they have to be one after the other? 
I mean is it possible to overlap them?

extensions/algebra/circuit/src/cuda/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+            return AirProvingContext::simple_no_pis(DeviceMatrix::dummy());
+        }
+        debug_assert_eq!(records.len() % record_stride, 0);
+        let rows_used = records.len() / record_stride;
+        let height = rows_used.next_power_of_two();
+        let width = Rv64IsEqualModU16AdapterCols::<F, 2, NUM_LANES>::width()
+            + ModularIsEqualCoreCols::<F, TOTAL_LIMBS>::width();
+
+        let device_ctx = &self.range_checker.device_ctx;
+        let d_records = records.to_device_on(device_ctx).unwrap();
```
Aalanli, line 66: I'm guessing these are the source of memcpys in the profile?

GunaDD:
thanks, I'll take a look at those. is the memcpy due to this mod builder gpu tracegen or was it pre-existing ?

GunaDD:
ah sorry I just saw your comment

Aalanli:
So what's happening is that every chip calls `generate_proving_ctx` in `VmChipComplex::generate_proving_ctx`. The memcpy blocks the kernel on the same stream. I believe it's possible to overlap them. The standard pattern is introduce a `COMM_STREAM` that you launch memcpys on, on the main stream you'd wait for the cuda event for the memcpy that's relevant.

GunaDD:
https://github.com/axiom-crypto/openvm-eth/actions/runs/30479785727

crates/circuits/mod-builder/cuda/src/field_expr.cu
```diff
@@ -0,0 +1,450 @@
...
+#define MB_LAUNCHER(R, B)                                                                     \
+    extern "C" int _field_expr_tracegen_r##R##_b##B(                                          \
+        Fp *d_trace, size_t height, size_t rows_used, const uint32_t *d_blob,                 \
+        const uint8_t *d_records, size_t rec_stride, size_t rec_core_offset,                  \
+        uint32_t *d_range_checker, size_t rc_bins, uint32_t *d_aux, size_t aux_words,         \
+        uint32_t pointer_max_bits, uint32_t timestamp_max_bits, int should_finalize,          \
+        uint32_t *d_err, cudaStream_t stream) {                                               \
+        const int threads = 256;                                                              \
+        const size_t want = (height + threads - 1) / threads;                                 \
+        const int blocks = (int)(want < 512 ? want : 512);                                    \
```
Aalanli: It would be way better if blocks is a multiple of SMs. Eg. replace 512 with sm_count * N.

crates/circuits/mod-builder/cuda/src/field_expr.cu
```diff
@@ -0,0 +1,450 @@
...
+        for (int j = 0; j < k; j++) {
+            uint64_t cur = (uint64_t)r[j] + s.p[j] + carry;
+            r[j] = (uint32_t)cur;
+            carry = cur >> 32;
+        }
+    }
+}
+
+// a^(p-2) via square-and-multiply; inv(0) = 0 by convention.
+__device__ void mont_inv(const Prog &s, const uint32_t *a, uint32_t *r) {
```
Aalanli, line 130: I think these functions exist else where in the repo.

crates/circuits/mod-builder/cuda/src/field_expr.cu
```diff
@@ -0,0 +1,450 @@
...
+    s.r2 = blob + blob[H_OFF_R2]; s.pm2 = blob + blob[H_OFF_PM2];
+    s.pinv = blob + blob[H_OFF_PINV]; s.p8 = blob + blob[H_OFF_P8];
+    s.mont = blob + blob[H_OFF_MONT]; s.climbs = blob + blob[H_OFF_CLIMBS];
+    s.optab = blob + blob[H_OFF_OPTAB];
+    return s;
+}
+
+// ---- K-limb Montgomery arithmetic (k <= TRACEGEN_MAX_K, runtime loops) ----
+__device__ void mont_mul(const Prog &s, const uint32_t *a, const uint32_t *b, uint32_t *r) {
+    uint32_t t[TRACEGEN_MAX_K + 2];
```
Aalanli, line 55: This also probably exists elsewhere, the mont_mul logic I mean.

crates/circuits/mod-builder/cuda/src/field_expr.cu
```diff
@@ -0,0 +1,450 @@
...
+__device__ __forceinline__ uint32_t f_of_i64(int64_t v) {
+    int64_t m = v % (int64_t)MB_BB_P;
+    if (m < 0) m += MB_BB_P;
+    return (uint32_t)m;
+}
+
+// Fill the core sub-row. CUDA equivalence tests compare this output against CPU tracegen.
+// `core_row` must point at the first core column. When `is_dummy`, inputs are zero,
+// flags false, range checks skipped, is_valid = 0 (mirrors fill_dummy_trace_row).
+__device__ void field_expr_fill_core_row(
```
Aalanli, line 159: This is hard to follow. Please add some comments for what each stage does, and avoid code duplication.

extensions/algebra/circuit/cuda/src/modular_is_eq.cu
```diff
@@ -0,0 +1,221 @@
...
+
+#define IS_EQ_LAUNCHER(LANES, LIMBS)                                                          \
+    extern "C" int _modular_is_eq_tracegen_l##LANES(                                          \
+        Fp *d_trace, size_t height, size_t rows_used, const uint8_t *d_records,               \
+        size_t rec_stride, size_t rec_core_offset, const uint16_t *d_modulus_limbs,           \
+        uint32_t *d_range_checker, size_t rc_bins, uint32_t pointer_max_bits,                 \
+        uint32_t timestamp_max_bits, cudaStream_t stream) {                                   \
+        const int threads = 256;                                                              \
+        const size_t want = (height + threads - 1) / threads;                                 \
+        const int blocks = (int)(want < 1024 ? want : 1024);                                  \
```
Aalanli, line 213: Same here, sm_count multiple

crates/circuits/mod-builder/src/tracegen_ir/README.md
```diff
@@ -0,0 +1,90 @@
...
+│  Select / SaveVar                   IntAdd / IntMul / Select │
+│                                                             │
+│  Metadata                                                   │
+│  ────────                                                   │
+│  • field and Montgomery constants                           │
+│  • constraint tape boundaries                               │
+│  • quotient/carry sizes and bounds                          │
+│  • opcode-to-flag mapping                                   │
+│  • scratch and trace dimensions                             │
+└──────────────────────────────┬──────────────────────────────┘
```
Aalanli, line 44: Let's add some documentation here explaining how the blob is encoded, and the constants file its located in.

crates/circuits/mod-builder/src/tracegen_ir/compiler.rs
```diff
@@ -0,0 +1,480 @@
...
+    next_slot: usize,
+    mont_payload: Vec<u32>,
+    limb_ops: Vec<LimbOp>,
+    scratch_top: u32,
+    const_limbs_payload: Vec<i32>,
+    r: BigUint, // 2^(32K) mod p
+}
+
+impl<'a> TracegenCompiler<'a> {
+    fn mont(&self, x: &BigUint) -> Vec<u32> {
```
Aalanli, line 31: Does the compiler do constant expression elimination? Eg. for expression (a+b)+(a+b), does it emit x = a+b; x + x, or does it do x = a+b; y = a+b; x+y?

crates/circuits/mod-builder/cuda/src/field_expr.cu
```diff
@@ -0,0 +1,450 @@
...
+template <size_t NUM_READS, size_t BLOCKS>
+__global__ void field_expr_tracegen_kernel(
+    Fp *__restrict__ d_trace, size_t height, size_t rows_used,
+    const uint32_t *__restrict__ blob,
+    const uint8_t *__restrict__ records, size_t rec_stride, size_t rec_core_offset,
+    uint32_t *__restrict__ rc_ptr, size_t rc_bins,
+    uint32_t *__restrict__ aux, size_t aux_words,
+    uint32_t pointer_max_bits, uint32_t timestamp_max_bits,
+    int should_finalize, uint32_t *__restrict__ err) {
+    const FieldExprProg s = load_prog(blob);
```
Aalanli: This could be in shared memory. All threads load the blob into shared, then reinterpret and load from the shared value into registers.

crates/circuits/mod-builder/cuda/src/field_expr.cu
```diff
@@ -0,0 +1,450 @@
...
+        }
+    }
+
+    // ---- value phase ----
+    uint32_t one[TRACEGEN_MAX_K];
+    for (int j = 0; j < k; j++) one[j] = j == 0 ? 1 : 0;
+    for (int i = 0; i < s.num_slots * k; i++) slots[i] = 0;
+    for (int io = 0; io < s.n_vops; io++) {
+        const uint32_t *op = s.vops + 5 * io;
+        const uint32_t opc = op[0], flag = op[1], dst = op[2], a = op[3], b = op[4];
```
Aalanli: Help me understand what's the value phase and constraint phase. Can you make the LLM write d_ prefix for pointers to device memory?

Aalanli:
Overall I can't find any issues. The AI generated kernel seems fairly suboptimal to me, but it's not very large so I guess it's fine. 

Where can I find some design doc about how the overall compilation flow works?

GunaDD:
is the one in the PR description what you are looking for on the docs about compile flow ?

crates/circuits/mod-builder/cuda/src/field_expr.cu
GunaDD: a similar to `mont_mul` logic exists in the stark-backend `bn254_monty_mul` and `mont_t` but the former is fixed to `bn254` and the latter requires the modulus at compile time which is not supported by the current design

crates/circuits/mod-builder/cuda/src/field_expr.cu
GunaDD: there is `mont_t::inverse()` but it cannot directly operate on `Prog`

crates/circuits/mod-builder/src/tracegen_ir/README.md
GunaDD: I feel like this is unnecessary since the `tracegen_ir/encoding.rs` file basically does that

GunaDD: actually, I'll add a description in `encoding.rs` about this

crates/circuits/mod-builder/cuda/src/field_expr.cu
GunaDD: I renamed value phase -> evaluation phase and constraint phase -> witness phase. Basically, evaluation phase computes the resulting field variable value and the witness phase computes the auxiliary stuff you need to prove correctness of some FieldExpr operation. For instance, if the operation is `mul x y` then eval phase will compute `x * y mod p` while witness phase computes the quotients and carries you need to be able to constraint the correctness of `x * y mod p`.

crates/circuits/mod-builder/cuda/src/field_expr.cu
GunaDD: explain in the other comment below and added some extra docs and renamed the phase names for better clarity.

crates/circuits/mod-builder/src/tracegen_ir/compiler.rs
GunaDD: no, it does not perform the expression elimination

crates/circuits/mod-builder/cuda/src/field_expr.cu
GunaDD: moved it to shared memory

crates/circuits/mod-builder/cuda/src/field_expr.cu
```diff
@@ -0,0 +1,467 @@
...
+        Fp *d_trace, size_t height, size_t rows_used, const uint32_t *d_blob,                 \
+        const uint8_t *d_records, size_t rec_stride, size_t rec_core_offset,                  \
+        uint32_t *d_range_checker, size_t rc_bins, uint32_t *d_aux, size_t aux_words,         \
+        uint32_t pointer_max_bits, uint32_t timestamp_max_bits, int should_finalize,          \
+        uint32_t *d_err, cudaStream_t stream) {                                               \
+        const int threads = 256;                                                              \
+        int device, sm_count;                                                                 \
+        cudaError_t err = cudaGetDevice(&device);                                             \
+        if (err != cudaSuccess) return err;                                                   \
+        err = cudaDeviceGetAttribute(&sm_count, cudaDevAttrMultiProcessorCount, device);       \
```
Aalanli: It's probably best to avoid this call, either initialize a static var or pass it in. Change to cuda version can make certain cuda api stuff more expensive.

crates/circuits/mod-builder/src/cuda/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+        debug_assert_eq!(records.len() % self.record_stride, 0);
+        let rows_used = records.len() / self.record_stride;
+        let height = rows_used.next_power_of_two();
+        let width = self.adapter_width + self.core_width;
+
+        let device_ctx = &self.range_checker.device_ctx;
+        let d_records = records.to_device_on(device_ctx).unwrap();
+        let d_trace = DeviceMatrix::<F>::with_capacity_on(height, width, device_ctx);
+
+        let n_threads = height.div_ceil(256).min(512) * 256;
```
Aalanli: I think this got out of sync if you change the number of blocks. n_threads should be blocks * threads in the computation in `field_expr.cu`?

crates/circuits/mod-builder/src/tracegen_ir/compiler.rs
Aalanli: I'm just curious. If I have an add chain like this:
```
x = a + b
y = x + x
z = y + y
...
```
in the expression form it would form a tree. If there's no CSE wouldn't such a change lower to exponentially number of instructions?

crates/circuits/mod-builder/src/cuda/mod.rs
Aalanli: Actually, the max block size should be sm_count * occupancy. Because there's no additional benefit to having a multiple of sm_count due to there being no additional parallelism that can be gained. We want to reduce the number of blocks and thus n_threads to save memory in d_aux.
You can probably ask the agent to calculate occupancy for you.

extensions/algebra/circuit/cuda/src/modular_is_eq.cu
Aalanli: Make sure to not call cuda API stuff here.

Aalanli:
I also just realized that the agg layer tracegen doesn't use these functionality. Or at least it's a lot slower. Do you think any of the functionality in this PR can be reused for agg?

crates/circuits/mod-builder/src/cuda/mod.rs
GunaDD: thanks!

crates/circuits/mod-builder/cuda/src/field_expr.cu
GunaDD: will do

Aalanli:
@GunaDD , do you have a profile after these changes?

GunaDD:
Generating the nsys profile currently

GunaDD:
It's going to come out here: https://github.com/axiom-crypto/openvm-eth/actions/runs/30659407872

Aalanli (approved):
LGTM

shuklaayush:
i'm closing this for now. we can create a new pr if there's something left to be adapted

PR #2985, opened 2026-07-09
fix: update CLI setup command for upload

stephenh-axiom-xyz:
Towards INT-8296.

- Add download artifacts sanity check to `upload-setup-artifacts.yml`
- `download_verifier` helper function always forces (force check is earlier)
- S3 bucket layout matches default `~/.openvm` layout

E2E CLI tests:

- CPU - https://github.com/openvm-org/openvm/actions/runs/29041035605
- CUDA - https://github.com/openvm-org/openvm/actions/runs/29040990653


crates/cli/src/commands/setup.rs
```diff
@@ -217,23 +217,21 @@ impl SetupCmd {
...
+            println!("Downloading {name}");
+            let resp = client
+                .get_object()
+                .bucket(ARTIFACTS_BUCKET)
+                .key(&key)
+                .send()
+                .await
+                .map_err(|e| eyre!("Failed to download s3://{ARTIFACTS_BUCKET}/{key}: {e}"))?;
+            let data = resp.body.collect().await?;
+            write(local_path, data.into_bytes())?;
```
luffykai, line 238: why do we ignore local path now?

stephenh-axiom-xyz: Wdym ignore?

luffykai (approved):

PR #2990, opened 2026-07-09
perf(vm): pinned host-memory pool for dense record arenas

yi-sun:
Record arenas (`DenseRecordArena`) are freshly allocated per chip per segment and their
contents are copied host-to-device at trace-generation time. Measured on the reth
workload (mainnet block 23992138): ~40 GB of records per block copied at ~11 GB/s —
pageable-memory PCIe throughput — costing ~3.5 s of `trace_gen.h2d_records_time_ms`
per block, i.e. the majority of the remaining (non-kernel) tracegen wall time.

## Change

Pin arena buffers with `cudaHostRegister` and recycle them across segments through a
size-classed pool (capacities rounded up to the next power of two so recurring
per-chip arenas of varying heights share pool entries). Copies from registered memory
take the DMA fast path automatically, so no call sites change anywhere.

Both page-locking (~1-2 GB/s) and re-zeroing cost seconds per proof if done on the
preflight critical path — arenas are provisioned at full trace size but only
partially written — so dropped buffers are handed to a background cleaner thread
(`crates/vm/src/arch/cuda/pinned.rs`) that:

1. waits for in-flight device work with a `cudaDeviceSynchronize`, batched per burst
   of dropped arenas (see the lifetime hazard below);
2. registers the buffer on its first cycle (per-buffer registration is tracked by
   base pointer; unregistered buffers are never pooled);
3. zeroes the dirty prefix the previous owner wrote (cursor position at drop) —
   bytes past it are still zero from the previous cycle;
4. only then returns the buffer to the pool.

Pool misses fall back to a fresh pageable allocation, exactly the pre-pool behavior,
so the worst case (no CUDA context, cleaner not yet caught up) matches the status
quo. Everything is `#[cfg(feature = "cuda")]`; the non-CUDA path is byte-for-byte
unchanged. Host-memory overhead is bounded by one segment's worth of arenas per size
class (pool entries are reused, not accumulated, in steady state).

### Lifetime hazard (why the cleaner synchronizes)

`cudaMemcpyAsync` from *pageable* memory returns only after the source has been
staged, which is what made freeing an arena right after enqueueing its copy safe
before this PR. From *pinned* memory the call returns immediately with the DMA still
in flight, so a returned buffer must not be zeroed or reused until previously
enqueued work has drained. An earlier revision of this PR zeroed on drop without
synchronizing and corrupted records under load (caught by the fibonacci CI benchmark
as a `LogupZerocheck` failure); the batched device sync in the cleaner is the fix,
and it keeps the wait entirely off the preflight/tracegen critical path.

## Measured impact

openvm-eth reth benchmark, block 23992138, baseline `main` vs this branch, both
variants run back-to-back on the same RTX Pro 6000 sandbox (VPMM pool capped at
15 GiB; both proofs verify, peak VRAM unchanged):

| phase | main | this PR | delta |
|---|---|---|---|
| `execute_metered_time_ms` (control) | 4,051 | 3,968 | −2.0% |
| `execute_preflight_time_ms` | 20,635 | 15,342 | **−25.7%** |
| `trace_gen_time_ms` | 5,886 | 5,170 | **−12.2%** |
| `trace_gen.h2d_records_time_ms` | 1,459 | 118 | **−91.9%** |
| `stark_prove_excluding_trace_time_ms` | 43,868 | 45,277 | +3.2% |
| `total_proof_time_ms` | 72,016 | 67,644 | **−6.1%** |

The preflight win comes from reuse: pooled buffers are already resident, so record
writes stop paying the page-fault + kernel-zeroing cost of fresh allocations.
`h2d_records` measures the enqueue of the (now truly asynchronous) copies; the
residual 118 ms is the first segments' pageable copies before the pool warms up.

Note: the GPU-tracegen PRs (#2984, #2989) touch the same workload; this change is
independent of them (the arenas and copy path predate those PRs) and they will be
rebased on top of whichever lands first.

### DIFF
--- a/crates/vm/src/arch/cuda/mod.rs
+++ b/crates/vm/src/arch/cuda/mod.rs
@@ -1,0 +1,3 @@
+//! CUDA-specific support for the VM architecture layer.
+
+pub(crate) mod pinned;
--- a/crates/vm/src/arch/cuda/pinned.rs
+++ b/crates/vm/src/arch/cuda/pinned.rs
@@ -1,0 +1,167 @@
+//! Page-locked (pinned) buffer pool for
+//! [`DenseRecordArena`](crate::arch::record_arena::DenseRecordArena).
+//!
+//! Record arenas are allocated fresh for every chip in every segment and their
+//! contents are copied host-to-device at trace generation time. Copies from
+//! pageable memory run at a fraction of PCIe bandwidth, but page-locking a
+//! buffer is itself expensive (~1-2 GB/s), and arenas are provisioned at full
+//! trace size while typically only partially written, so neither registration
+//! nor re-zeroing may sit on the preflight critical path. Dropped buffers are
+//! therefore handed to a background cleaner thread which registers them once
+//! (`cudaHostRegister`), zeroes the prefix the previous owner wrote, and only
+//! then returns them to the pool. [`take`] hands out ready (registered,
+//! all-zero) buffers on a pool hit; on a miss it falls back to a fresh
+//! pageable allocation — exactly the pre-pool behavior — so the worst case (no
+//! CUDA device, cleaner not yet caught up) matches the status quo. Capacities
+//! are rounded up to the next power of two so recurring per-chip arenas of
+//! varying heights share pool entries.
+//!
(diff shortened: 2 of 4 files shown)


yi-sun:

PR #2994, opened 2026-07-09
perf(vm): overlap executor tracegen with the memory Merkle build

yi-sun:
Stacked on #2990 (first commit is that PR; review the second commit only).

## Problem

The per-segment memory Merkle subtree build — Poseidon2-hashing the full memory
image — is enqueued at memory-transport time and runs on the GPU stream while
preflight executes on CPU. When preflight finishes before the build does, system
tracegen blocks on it at its first D2H sync. Instrumented on the reth workload
(block 23992138), that wait showed up as ~1.0 s in the memory-inventory record
merge and ~1.6 s absorbed by the Poseidon2 periphery chip's first D2H, per block.
Making preflight faster (#2990) shrinks the overlap window and makes this worse.

## Change

Run executor chips with record arenas **before** system tracegen: their fills are
CPU-heavy, extending the window in which the Merkle build can complete. They
depend on neither system chips nor each other. Chips with **empty** arenas — the
periphery hashers and lookup tables, whose inputs other chips' kernels push
through shared device buffers (system tracegen included) — are deferred until
after system tracegen, in their original relative order. The `(air_id, ctx)`
assignment is position-based and unchanged, so the proving context order matches
the verifying key exactly as before.

Also fixes timing-instrumentation artifacts found while measuring this:

- the `system_trace_gen` span guard was a temporary inside the iterator chain, so
  it lived until the end of the whole `collect()` statement — the derived
  `system_trace_gen_time_ms` metric always equaled the entire tracegen wall;
- `single_trace_gen` spans existed only for chips with non-empty arenas, leaving
  periphery chips (which do real tracegen work) unmeasured;
- the record-arena cleaner span carries a unique batch label (the derived metric
  is a gauge; identical label sets silently overwrite);
- memory-inventory tracegen phases (record merge, Poseidon2 prepare, Merkle
  update, boundary, subtree drop) get their own spans.

## Measured impact

openvm-eth reth benchmark, block 23992138, single RTX Pro 6000 (VPMM pool capped
at 15 GiB), both runs on hosts within 1.3% on the CPU-bound control metric,
baseline = main + #2990:

| phase | #2990 | + this PR | delta |
|---|---|---|---|
| `trace_gen_time_ms` | 6,619 | 5,126 | **−22.6%** |
| Poseidon2 periphery sync wait | 1,637 | 150 | −91% |
| `execute_preflight_time_ms` | 23,191 | 21,623 | −6.8% |
| `total_proof_time_ms` | 85,728 | 81,278 | **−5.2%** |

The preflight gain is a second-order effect: arenas drop earlier in tracegen, so
the pinned-buffer pool is replenished before the next segment's preflight
allocates, raising the pool hit rate.

The remaining ~2 s of executor-phase tracegen is the CPU fill cost of the
mod-builder (FieldExpression) chips, which #2989 moves to GPU; these changes
compose.

### DIFF
--- a/crates/vm/cuda/src/system/inventory.cu
+++ b/crates/vm/cuda/src/system/inventory.cu
@@ -213,3 +213,48 @@ extern "C" int _inventory_merge_records_get_temp_bytes(
     *h_temp_bytes_out = temp_bytes;
     return CHECK_KERNEL();
 }
+
+/// Width in u32 words of one Merkle touched-block record:
+/// (address_space, ptr, timestamp, values[OUT_BLOCK_SIZE]).
+/// Must match MERKLE_TOUCHED_BLOCK_WIDTH on the Rust side.
+inline constexpr uint32_t MERKLE_REC_WIDTH = 3 + OUT_BLOCK_SIZE;
+
+/// Converts merged inventory records (boundary layout) into Merkle
+/// touched-block records: drops the second timestamp, taking the max.
+/// Values stay Montgomery-encoded in both layouts.
+__global__ void inventory_to_merkle_records_kernel(
+    uint32_t const *out_records,
+    size_t num_records,
+    uint32_t *merkle_records
+) {
+    size_t stride = gridDim.x * (size_t)blockDim.x;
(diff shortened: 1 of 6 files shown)


yi-sun:

PR #2995, opened 2026-07-10
perf(vm): pinned staging for the per-segment memory-image upload

yi-sun:
Stacked on #2994 (review the last commit only).

## Problem

The initial memory image (~3 GB on reth workloads) is uploaded to device at the
start of **every segment**, from pageable guest memory. Pageable copies run at
staging-pipeline speed (~11 GB/s) and occupy the stream before the first Merkle
subtree build kernel can start — so ~250 ms/segment of upload counts against the
window in which preflight can hide the build (see #2994).

## Change

Pack the image (parallel memcpy, 8 MiB chunks) into a page-locked staging buffer
owned by `MemoryInventoryGPU` and registered once; the device copy then takes the
DMA fast path (~2x pageable) and returns immediately.

Why staging instead of registering the guest memory directly:
- registration would be tied to an allocation this module does not own; a guest
  memory freed while registered is undefined behavior, and the lifecycle (per
  block, across segments) is invisible from here;
- an async DMA reading guest memory directly would race preflight, which starts
  mutating that memory immediately after transport — the same lifetime hazard
  class as the record-arena race caught by CI on #2990. The pack fully consumes
  the guest memory before returning; the DMA reads only the owned staging.

The staging buffer registers once (~1.5 s for 3 GB, amortized over all segments
of all blocks in the process) and unregisters on drop. Registration failure
degrades to a pageable staging buffer, i.e. the status quo.

## Measured impact

openvm-eth reth benchmark, block 23992138, RTX Pro 6000 (VPMM 15 GiB), measured
on the full stack (#2990+#2994+#2989+this), same host class as baseline (CPU
control metric within 0.7%):

| metric | before | after | delta |
|---|---|---|---|
| `vm.transport_init_memory_time_ms` | 5,281 | 886 | **−83%** |
| `total_proof_time_ms` | 87,129 | 83,340 | **−4.3%** |

The win is larger than the stream-occupancy estimate because the pageable copy
also blocked the *host* inside transport for ~68 ms/segment (~5.3 s/block of
wall time between segments); the pack + async-DMA path returns in ~11 ms.
Proof verifies (rc=0); peak VRAM unchanged.

### DIFF
--- a/crates/vm/src/arch/cuda/pinned.rs
+++ b/crates/vm/src/arch/cuda/pinned.rs
@@ -38,7 +38,7 @@ type ReturnedBuffer = (Vec<u8>, usize);
 /// (cudaErrorInvalidValue) source ranges that span multiple distinct
 /// page-locked registrations, so chunked registration corrupts nothing but
 /// breaks every copy crossing a chunk boundary.
-fn register_buffer(ptr: *mut u8, len: usize) -> bool {
+pub(crate) fn register_region(ptr: *mut u8, len: usize) -> bool {
     // SAFETY: [ptr, ptr+len) is a live allocation owned by the caller.
     let rc = unsafe { cudaHostRegister(ptr as *mut c_void, len, 0) };
     if rc != 0 {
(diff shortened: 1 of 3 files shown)


yi-sun:

PR #2996, opened 2026-07-10
perf(vm): fuse the Merkle subtree build kernels

yi-sun:
Stacked on #2995 (review the last commit only).

## Problem

The per-segment Merkle subtree build — Poseidon2 over the full memory image —
runs level by level: one `merkle_tree_compress` launch per tree level (27+ per
address space per segment), each reading its children back from global memory,
and the top ~15 levels are launches too small to occupy the GPU. A roofline
estimate (~15 GB of traffic + ~190M permutations per segment) puts the build far
from hardware limits.

## Change

Fused build: each block takes a tile of 256 input nodes and ascends up to 8
levels in shared memory (ping-pong halves — an in-place tree reduction races:
thread `i` writes slot `i` while thread `i/2` still reads it). Every level is
still written to its heap position — sibling lookups during proof generation
need the full tree — but no level is ever read back from global memory:

- global traffic per digest: ~2 reads + 1 write → 1 write;
- kernel launches per subtree: 27+ → 4;
- tree layout, path restore, update, and proof paths unchanged.

## Measured outcome (honest note)

Benchmarked on the standard block (openvm-eth reth benchmark, block 23992138,
RTX Pro 6000, VPMM 15 GiB, full stack): proof verifies (rc=0) and the tree is
bit-identical, but the residual build wait is **unchanged** (~1.16 s/block at
the program-chip sync). The build is compute-bound on Poseidon2 (~500M
permutations/s sustained), not bandwidth-bound, so removing the inter-level
global traffic does not move wall time on this GPU.

What the change still buys: 27+ launches → 4 per subtree (~40k fewer launches
per block), less memory traffic (relevant on bandwidth-leaner GPUs), and a
simpler launch profile. Fine to land or close on that basis — the real fix for
the remaining build wait is an incremental tree (only ~0.2% of leaves change
between segments), which is a separate design discussion.

### DIFF
--- a/crates/vm/cuda/src/system/memory/merkle_tree.cu
+++ b/crates/vm/cuda/src/system/memory/merkle_tree.cu
@@ -13,46 +13,6 @@ struct alignas(32) digest_t {
 
 #define COPY_DIGEST(dst, src) memcpy(dst, src, sizeof(digest_t))
 
-// `ADDR_SPACE_IDX` is the address space minus `ADDR_SPACE_OFFSET` (which is 1)
-template <int ADDR_SPACE_IDX>
-__global__ void merkle_tree_init(uint8_t *__restrict__ data, digest_t *__restrict__ out) {
-    auto gid = blockDim.x * blockIdx.x + threadIdx.x;
-
-    Fp cells[CELLS] = {0};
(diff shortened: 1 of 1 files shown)


yi-sun:

PR #2997, opened 2026-07-10
perf(vm): trim the tracegen tail (merge inputs, sync order, span cost)

yi-sun:
Stacked on #2996 (review the last commit only).

Trims the three largest residual costs measured after #2994/#2995/#2996, worth a
projected ~0.4–0.6 s/block of tracegen combined:

1. **Merge inputs pack into pinned memory (−150–250 ms).** The touched-memory
   records were converted on host into a fresh pageable Vec and staged to
   device at ~11 GB/s. They now pack straight into a pooled page-locked buffer
   (parallel, fixed 7-word stride) and upload on the DMA fast path. Returning
   the buffer goes through the record-arena cleaner, which synchronizes the
   device before any reuse — the same lifetime discipline as #2990.

2. **Host work hoisted above the merged-count sync (−~100 ms).** The
   unpadded-height scan needs only the partition, and the Poseidon2 records
   buffer can be sized by the input count (an upper bound on the merged
   count). Both now run while the merge kernels execute, so the 8-byte D2H
   sync waits on less.

3. **Per-chip spans gated on non-empty arenas (−100–200 ms).** Span + gauge
   emission costs tens of microseconds each; ~26 of ~40 chips are empty in a
   typical segment (~2,000 empty emissions per block). Periphery chips remain
   covered in aggregate by the `periphery_trace_gen` span introduced in #2994.

## Measured

stack, on a host whose CPU control metric reads 4.4% slower than the baseline
run (gains below are therefore slightly understated); proof verifies (rc=0):

| metric | before | after |
|---|---|---|
| `mem_merge_records_time_ms` | 530 | 381 |
| `periphery_trace_gen_time_ms` | 726 | 592 |
| `trace_gen_time_ms` | 3,241 | **2,990** |

## Follow-up commits (measured, same rig)

**Host-side merged count** — the 8-byte D2H of the merged record count was the
last mid-merge sync; the device merge's group flag is a pure adjacency function
of `(address_space, ptr / OUT_BLOCK)` over the sorted partition, so the count is
now a parallel host scan and the memory-inventory path is enqueue-only until the
public-values readback. Debug builds cross-check host vs device counts.

**Exec frequencies converted on device** — profiling `poseidon2_mix` (L40S: 40
registers, no spills, full occupancy, ~3.5G perm/s) proved the per-segment
"build wait" at the program chip was never the Merkle build: it was the program
chip itself serially converting the multi-million-entry frequency vector on host
(`map(F::from_u32)`) plus a pageable copy, every segment. The raw u32s now
upload through a pooled pinned buffer and a device kernel does the conversion:
`program_trace_gen_time_ms` **1,206 → 59**.

Tracegen wall holds at ~2,980 ms: with all artificial stalls removed, the
remaining sync time is the genuine aggregate GPU tracegen kernel time on the
single stream, already overlapped with the available CPU work. Further wall
reduction requires cross-segment pipelining rather than more tracegen changes.

### DIFF
--- a/crates/vm/cuda/src/system/program.cu
+++ b/crates/vm/cuda/src/system/program.cu
@@ -58,3 +58,35 @@ extern "C" int _program_cached_tracegen(
     );
     return CHECK_KERNEL();
 }
+
+/// Converts raw u32 execution frequencies to field elements in place on
+/// device, zero-filling [filtered_len, height). Replaces a host-side serial
+/// map + pageable copy that cost ~15 ms per segment on large programs.
+__global__ void program_fill_frequencies(
+    const uint32_t *__restrict__ freqs,
+    size_t filtered_len,
+    Fp *__restrict__ out,
+    size_t height
+) {
+    size_t stride = gridDim.x * (size_t)blockDim.x;
+    for (size_t i = blockIdx.x * (size_t)blockDim.x + threadIdx.x; i < height; i += stride) {
+        out[i] = i < filtered_len ? Fp(freqs[i]) : Fp(0);
+    }
+}
+
+extern "C" int _program_fill_frequencies(
+    const uint32_t *d_freqs,
+    size_t filtered_len,
+    Fp *d_out,
+    size_t height,
+    cudaStream_t stream
+) {
+    if (height == 0) {
+        return 0;
+    }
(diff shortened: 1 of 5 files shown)


yi-sun:

PR #2999, opened 2026-07-10
perf(vm): make TouchedBlock the GPU inventory-record layout

yi-sun:
Stacked on #2997 (review the last commit only).

## Change

`TouchedMemory` entries were `((u32, u32), TimestampedValues<F, 4>)` tuples that
the GPU memory inventory re-encoded field by field into the device merge's input
layout on every segment. The entry type is now:

```rust
#[repr(C)]
pub struct TouchedBlock<F> {
    pub address_space: u32,
    pub ptr: u32,
    pub timestamp: u32,
    pub values: [F; DEFAULT_BLOCK_SIZE],
}
```

For a 4-byte field type this is byte-identical to `InRec` in `inventory.cu`
(compile-time size assert at the transmute site), so:

- the GPU path bulk-copies the partition's bytes into the pinned upload buffer —
  no per-field packing;
- memory finalize emits the final layout directly — and so can any other
  producer of touched memory: the interface contract between execution and the
  memory argument is now "produce this exact buffer", with no conversion layer
  in between.

CPU-backend consumers (persistent boundary grouping) switch from tuple
destructuring to named fields; `calculate_unpadded_height` takes an address
accessor so test callers with other block widths keep working. No behavioral
change; workspace + tests compile clean, clippy clean.

## Measured

stack, same host class as baseline (CPU control metric within 0.1%); proof
verifies (rc=0), confirming the layout identity end-to-end:

| metric | before | after |
|---|---|---|
| `mem_merge_records_time_ms` | 487 | 399 |
| `trace_gen_time_ms` | 2,982 | 2,952 |

### DIFF
--- a/crates/vm/src/system/cuda/memory.rs
+++ b/crates/vm/src/system/cuda/memory.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use openvm_circuit::{
     arch::{AddressSpaceHostLayout, MemoryConfig, ADDR_SPACE_OFFSET},
-    system::{memory::AddressMap, TouchedMemory},
+    system::{memory::AddressMap, TouchedBlock, TouchedMemory},
 };
 use openvm_circuit_primitives::Chip;
 use openvm_cuda_backend::{prelude::F, GpuBackend};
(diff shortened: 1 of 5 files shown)


yi-sun:

PR #3001, opened 2026-07-10
perf(vm): drop the hidden stream sync in CPU-to-GPU ctx transport

yi-sun:
Stacked on #2999 (review the last commit only).

## Problem

`cpu_proving_ctx_to_gpu` (used by every `HybridChip` and the continuation
circuits' trace conversion) calls `transport_matrix_h2d_col_major`, which ends
with a **full stream sync**. The connector chip's two-row trace conversion runs
mid-tracegen, right after the executor phase has filled the stream with ~0.5 GB
of record copies and tracegen kernels per segment — so this one call was
measured at **1,122 ms per block** (`connector_trace_gen_time_ms`), stalling the
host on the entire stream drain 78 times. (Earlier revisions of this stack saw
the same conserved wait surface at the program chip; it moved here when that
path went async.)

## Change

The sync is unnecessary for the temporary column-major buffer's lifetime: an
async copy from pageable memory returns only once the source has been consumed
by the staging pipeline, and all later reads of the destination are
stream-ordered. The transport is now the plain async copy; the temporary
drops safely on return.

With no serializing call left between the executor phase and the memory
inventory, the stream drain surfaces at the Merkle update's public-values
readback instead — after the merge pipeline's host driving has already run,
which is the packing improvement this buys.

## Measured

stack, same host class (CPU control within 0.8%); proof verifies (rc=0).
Measured together with the copy-stream PR stacked on this one:

| metric | before | after |
|---|---|---|
| `connector_trace_gen_time_ms` | 1,122 | **1** |
| `merkle_update_time_ms` (drain relocates here, after merge driving) | 161 | 873 |
| `trace_gen_time_ms` | 2,952 | **2,714** |

### DIFF
--- a/crates/circuits/primitives/src/hybrid_chip.rs
+++ b/crates/circuits/primitives/src/hybrid_chip.rs
@@ -2,11 +2,10 @@ use std::marker::PhantomData;
 
 use openvm_cpu_backend::CpuBackend;
 use openvm_cuda_backend::{
-    base::DeviceMatrix, data_transporter::transport_matrix_h2d_col_major,
-    hash_scheme::GpuHashScheme, prelude::SC, GenericGpuBackend, GpuBackend,
+    base::DeviceMatrix, hash_scheme::GpuHashScheme, prelude::SC, GenericGpuBackend, GpuBackend,
 };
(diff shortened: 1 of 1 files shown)


yi-sun:

PR #3002, opened 2026-07-10
perf(vm): upload records on a dedicated copy stream

yi-sun:
Stacked on #3001 (review the last commit only).

## Problem

Record uploads and tracegen kernels have no cross-chip dependencies, but on a
single stream they serialize: ~36 GiB of records per reth block
(`trace_gen.record_arena_bytes`) is ~0.7 s of DMA that the stream cannot
overlap with its own kernels — the largest single contributor to the stream
drain that the first blocking call after the executor phase absorbs. PCIe copy
engines run concurrently with SM compute, so this is sum where it could be max.

## Change

`records_to_device` (new, `crates/vm/src/arch/cuda/copy_stream.rs`): allocates
the destination on the kernel stream, runs the copy on a shared dedicated
stream, and fences with two events:

1. recorded on the kernel stream after allocation, waited by the copy stream —
   the VPMM allocator reuses same-stream regions without synchronization, so
   the write must be ordered after any prior kernel-stream user of a
   reallocated region;
2. recorded on the copy stream after the copy, waited by the kernel stream —
   ordering every later read (and any future reallocation) after the write.

The fences chain correctly across region reuses; host-side lifetimes are
unchanged (pinned sources return through the arena cleaner's device-wide sync,
pageable sources are consumed by staging before the call returns). Falls back
to a plain same-stream copy when a second stream is unavailable.

All per-segment record uploads switch to the helper (rv32im including
hint-store, keccak256, sha2, bigint, deferral); small metadata uploads
(offsets, moduli, blobs) stay on the kernel stream.

## Measured

stack, same host class; proof verifies (rc=0) across all 78 segments,
validating the two-event fence under real VPMM region reuse. Measured together
with #3001 (the two compose; see its table): tracegen wall 2,952 → 2,714 ms.
The remaining ~0.9 s drain at the Merkle-update readback is the true GPU
tracegen execution; separating the copy-stream's overlap contribution from
#3001's packing gain needs an A/B this stack's benching cadence didn't
justify — happy to run one if reviewers want it isolated.

### DIFF
--- a/crates/vm/src/arch/cuda/copy_stream.rs
+++ b/crates/vm/src/arch/cuda/copy_stream.rs
@@ -1,0 +1,95 @@
+//! Dedicated host-to-device copy stream for record uploads.
+//!
+//! Record uploads and tracegen kernels have no cross-chip dependencies, but on
+//! a single stream they serialize: ~36 GB of records per reth block costs
+//! ~0.7 s of DMA that the stream cannot overlap with its own kernels. PCIe
+//! copy engines run concurrently with SM compute, so routing the uploads
+//! through a dedicated stream turns copy+kernel time from a sum into a max.
+//!
+//! Ordering with the VPMM allocator: allocations are stream-tagged, and a
+//! region freed on a stream may be reallocated to the same stream without
+//! synchronization (stream order makes that safe). A copy-stream write into a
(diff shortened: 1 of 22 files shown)


yi-sun:

PR #3003, opened 2026-07-10
perf: small-copy fast path in guest memcpy

Qumeric:
## Motivation

Profiling Ethereum block execution in the reth guest (axiom-crypto/openvm-eth), `memcpy` accounts for **11.3% of all guest instructions**. Call counts are dominated by small copies (RLP fragments, hash outputs, field elements), and for those the current routine spends most of its time before copying anything: the alignment dispatch plus head loop costs 20–90 instructions, with the head-alignment loop alone compiling to ~13 instructions per byte. A zero-length `memcpy` costs 20 instructions.

## Change

`memcpy.s` is regenerated from the same musl v1.2.5 `src/string/memcpy.c` (documented invocation in the file header, mirroring the `memcmp.s` convention on `feat/memc-optimizations`), with one source modification, also documented in the header: a fast path for `n <= 32` — a descending fallthrough `switch` that compiles to a jump table into straight-line constant-offset byte copies (~2.4 instructions/byte, no alignment preamble). Above 32 bytes the code is musl's unchanged word-copy paths, including the `__GNUC__` shifted-word path for src/dst alignment mismatch.

Instructions per call, averaged over all 16 src/dst alignment combinations:

| size | before | after | delta |
|---|---|---|---|
| n=0–3 | 34.7 | 15.2 | **−56%** |
| n=4–8 | 52.5 | 27.0 | **−49%** |
| n=9–16 | 65.3 | 40.0 | **−39%** |
| n=17–32 | 86.9 | 72.0 | −17% |
| n=33–64 | 120.2 | 112.1 | −7% |
| n=1024 | 1400.2 | 1391.0 | −0.7% |

## Correctness

- Differentially fuzzed the exact `.s` text in an rv32im interpreter against a reference copy: all 16 alignment combinations × lengths 0–70 plus bulk sizes up to 4096 (1216 cases) — byte-exact results, correct return value, no out-of-bounds writes, source reads stay within the same word-slop as the existing implementation.
- New integration test (`extensions/rv32im/tests`): a `memcpy` example exercising all 16 alignment combinations × 18 lengths with canary checks around the destination, run through the transpiler and AIR test.
- `openvm-rv32im-integration-tests` passes except `test_tiny_mem_test`, which fails identically on current `main`.

Happy to run the reth end-to-end block benchmark against this branch if useful.

### DIFF
--- a/crates/toolchain/openvm/src/memcpy.s
+++ b/crates/toolchain/openvm/src/memcpy.s
@@ -201,298 +201,482 @@
 // negated the permissions granted in the license. In the spirit of
 // permissive licensing, and of not having licensing issues being an
 // obstacle to adoption, that text has been removed.
-	.text
+//
+// This was compiled into assembly from musl's memcpy implementation
+// (src/string/memcpy.c, musl v1.2.5) with:
+//
+// clang -target riscv32 -march=rv32im -O2 -fno-builtin -ffreestanding \
+//     -D"__BYTE_ORDER=1234" -D"__LITTLE_ENDIAN=1234" \
+//     -S musl_memcpy_no_includes.c -o src/memcpy.s
+//
+// (Homebrew clang 22.1.0.) musl_memcpy_no_includes.c is musl's memcpy.c with:
+// * `#include <string.h>` replaced by `typedef unsigned int size_t;` and
+//   `#include <endian.h>` removed; the -D flags supply the endianness macros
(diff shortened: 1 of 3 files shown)


Qumeric:
Reopening from an in-repo branch so CI gets secrets (fork PRs fail at the Load SSH key step).

PR #3004, opened 2026-07-10
perf(metered): sparse MeteredCtx clone and fused periodic segmentation scan

yi-sun:
## Summary

Two fixes to metered-execution overhead, found by profiling rvr AOT metered execution (openvm-eth mainnet block 23992138 plus the `benchmarks/execute` guests):

**1. Sparse clones for `BitSet` and `MemoryPageTracker`.** Both span the full page-id space but are non-zero only at their dirty-tracked entries (`dirty_words` / `dirty_pages`, each pushed exactly once on the 0→non-zero transition). The derived `Clone` memcpy'd — and page-faulted in, then freed — the full tables, costing **~16 ms per `MeteredCtx` clone**; `Sdk::execute_metered` clones the ctx template on every call, so every production metered execution paid this. `Clone` now allocates zeroed (lazily-mapped) tables and replays only the dirty entries, making template clones nearly free.

**2. Fused, precomputed periodic segmentation scan.** The per-`segment_check_insns` check scanned all AIRs across five parallel slices, twice (a heights+memory pass in `segmentation_trigger` plus a separate interactions pass in `calculate_total_interactions`). `MeteredCtx` construction now builds a scan plan: constant-height AIRs are aggregated into fixed contributions once, and the per-check loop walks only the non-constant AIRs, packed contiguously, computing heights, memory, and interactions in one fused pass. Contexts built without a plan (direct `SegmentationCtx` construction) fall back to the full scan.

Segmentation decisions are unchanged: `test_scan_plan_matches_fallback_scan` checks plan-vs-fallback equivalence across all three trigger types, and segment boundaries / instruction counts were verified identical on the benchmark guests.

## Measurements

rvr metered execution, best-of-5, pinned cores, vs `origin/develop-v2.1.0-rv64`:

| metric | before | after |
|---|---|---|
| `MeteredCtx::clone` (fib guest config) | 16.4 ms | **0.06 ms** |
| fibonacci_iterative metered | 1.80 ms | **0.59 ms** (−67%) |
| ecrecover metered | 3.30 ms | **2.20 ms** (−33%) |
| sha256_iter metered | 48.8 ms | **45.8 ms** (−6%) |

(openvm-eth block-level before/after to follow in a comment.)

## Testing

- `cargo nextest run -p openvm-circuit --features rvr,test-utils`: 47 passed
- New tests: `test_bitset_sparse_clone_preserves_contents`, `test_page_tracker_sparse_clone_preserves_contents`, `test_scan_plan_matches_fallback_scan`
- clippy clean, nightly fmt


yi-sun:
openvm-eth block-level before/after (mainnet block 23992138, 666.6M insns, rvr `--mode execute-metered`, cached input, identical block hash both sides; openvm-eth `develop-v2.1.0-rv64` with openvm path-patched to this branch vs its base):

| metric | base | this PR |
|---|---|---|
| `sdk.execute_metered_time_ms` | 1136 | **1094** (−3.7%) |
| sdk wrapper delta (ctx clone + state init) | 41 ms | **22 ms** |
| `execute_metered_time_ms` | 1095 | **1072** |

The block-level effect is smaller than the microbenchmark numbers in the description because the fixed per-call clone cost amortizes over a ~1s execution — the clone fix matters most for small/mid programs (ecrecover −33%, fibonacci −67% per the description) and for server flows executing many small requests.

PR #3005, opened 2026-07-10
perf: small-copy fast path in guest memcpy

Qumeric:
## Motivation

Small copies dominate memcpy call counts in real guests (RLP fragments, hash outputs, field elements), and musl's alignment dispatch and head loops cost 20–90 instructions before the first byte moves. For n ≤ 32 a jump table into straight-line constant-offset byte copies has no preamble at all.

## Change

One commit, four files:

- `scripts/generate_libc_intrinsics.sh` — defines the fast-path C block (`FASTPATH_MEMCPY`, a descending fallthrough switch: entering at `case n` copies bytes n−1..0) and injects it into the fetched musl source right after the `d`/`s` pointer declarations, before compilation. The injection hard-fails if the anchor line isn't found exactly once, so a future musl change can't silently drop the fast path.
- `crates/toolchain/openvm/src/memcpy.s` — regenerated by the script. The header notes the addition.
- `extensions/riscv/tests/programs/examples/memcpy.rs` + registration — integration test exercising all 64 src/dst alignment combos mod 8 × 18 lengths with canary checks.

## How to review

The `.s` diff is machine output; don't read it — reproduce it:

```
scripts/generate_libc_intrinsics.sh && git diff --exit-code crates/toolchain/openvm/src/
```

(byte-identical output verified with Homebrew clang 22.1.0; the script strips `.ident`/`.file` so bytes depend only on LLVM codegen). The human-reviewable surface is the ~45-line C block and the injection step in the script.

## Correctness

- Differential fuzz of the old vs new checked-in `.s` on an rv64im text-assembly interpreter: 1780 cases (all 64 alignment combos × 20 lengths + 500 random), byte-exact copies, return value, canaries intact, no out-of-bounds writes, reads within ±8 of the source buffer.
- The new integration test proves the artifact inside the actual VM.
- `n > 32` takes the identical musl path as before (one extra dispatch branch).

## Instructions per call (rv64im emulator, avg over all 64 alignment combos)

| n | before | after |
|---|---|---|
| 0–3 | 30.1 | 15.2 |
| 4–8 | 44.9 | 27.0 |
| 9–16 | 63.3 | 45.5 |
| 17–32 | 83.8 | 67.9 |
| bulk (>32) | — | +0.1–0.3% |

Mutually-8-aligned small copies regress (e.g. n=32 both aligned: 41 → 79) because musl's word path was already cheap there; averaged over real alignment distributions the fast path wins by a wide margin, and mismatched/small copies dominate real call sites.

## End-to-end (reth block 24001988, prove-stark, GPU)

Compared base = develop-v2.1.0 tip vs this branch (one commit on top), via the reth compare-bench workflow:

| metric | base | this PR | delta |
|---|---|---|---|
| executed instructions | 887,243,325 | 886,658,744 | −0.07% |
| total cells (padded) | 72.30B | 72.32B | +0.03% |
| segments | 84 | 84 | 0 |
| total proof time | 84.31s | 84.27s | −0.05% |

The instruction delta is exact and deterministic; cells, segments, and proof time are all within run-to-run noise (no regression). The win is smaller here than the sub-1% seen on more RLP-heavy blocks: small copies are a smaller slice of this block, and the single `n <= 32` dispatch branch added to the bulk (`n > 32`) path offsets part of the gain when large copies dominate. This is a low-risk correctness-preserving change whose benefit scales with how small-copy-heavy the workload is, not a headline speedup.

## History

A previous revision targeted `main` (rv32): a reproducible regen of memcpy.s plus the same fast path, measured end-to-end at −0.65% executed instructions on an RLP-heavier block. After retargeting to `develop-v2.1.0`, the regen half was dropped as superseded by #2780's generator script, and the fast path was re-ported, re-fuzzed, and re-measured for rv64 on top of that script (table above).

### DIFF
--- a/crates/toolchain/openvm/src/memcpy.s
+++ b/crates/toolchain/openvm/src/memcpy.s
@@ -2,6 +2,11 @@
 //
 // src/string/memcpy.c
 //
+// with one OpenVM addition (not in musl): a jump-table fast path
+// for n <= 32 byte copies, injected into the source before
+// compilation. See FASTPATH_MEMCPY in the generator script for
+// the exact code and rationale.
+//
 // This was compiled into assembly with:
 //
 //     clang -target riscv64 -march=rv64im -O3 -S memcpy.c -nostdlib -fno-builtin -funroll-loops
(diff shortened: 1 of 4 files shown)


Qumeric:
I think it would be good to rerun openvm-eth bench to ensure it is noticeable improvement, I am marking it as draft and will reopen for review when openvm-eth is updated to develop-2.1 branch

shuklaayush:
> Mutually-8-aligned small copies regress (e.g. n=32 both aligned: 41 → 79)

i'm a bit skeptical because of this

> when openvm-eth is updated to develop-2.1 branch

there's already a `develop-v2.1.0` branch in openvm-eth that you can use for comparison

Qumeric:
Ok it turned out to be ~netural

```
│                │ block 24001988 │ block 23992138 │ simple mean │ pooled  │
├────────────────┼────────────────┼────────────────┼─────────────┼─────────┤
│ instructions   │ −0.066%        │ +0.158%        │ +0.046%     │ +0.026% │
├────────────────┼────────────────┼────────────────┼─────────────┼─────────┤
│ cells (padded) │ +0.033%        │ −0.324%        │ −0.146%     │ −0.123% │
└────────────────┴────────────────┴────────────────┴─────────────┴─────────┘
```

Probably original version mostly benefited from recompiling with clang 22 but in this branch it was already regenerated with clang 22. 

Thus closing

PR #3006, opened 2026-07-12
perf(vm): GPU trace generation and host-to-device transfer optimizations

yi-sun:
## Summary

GPU trace-generation and host-to-device (H2D) transfer optimizations for the VM
prover, consolidated into one PR. Each logical change is a separate commit.

## Changes

- **Pinned record-arena pool** — dense record arenas are served from a pinned
  host-memory pool with a background cleaner thread, so async DMAs from pinned
  buffers stay in flight instead of blocking on a pageable staging copy.
- **Device-side boundary→Merkle record conversion** — memory boundary records
  are converted to Merkle records on the GPU, with pinned host-region
  registration for the transport.
- **Pinned memory-image staging** — the per-segment memory-image upload is
  staged through pinned buffers.
- **Tracegen tail trims** — merged inputs, a host-computed merged record count,
  exec-frequency upload via a pooled pinned buffer plus a device kernel, and
  span/metric cost gated to non-empty arenas only.
- **`repr(C)` TouchedBlock layout** — the touched-block struct is laid out to
  match the CUDA inventory-record struct, enabling a zero-conversion bulk copy.
- **Drop hidden stream sync** — remove the hidden stream synchronization in the
  CPU-to-GPU proving-context transport.

System trace generation keeps the original system-first ordering.

## Results

Measured on the openvm-eth standard block, RTX Pro 6000, single sandbox
(device pool capped at 15 GiB):

| metric | baseline | this PR |
| --- | --- | --- |
| trace generation | 7.79 s | 5.01 s (−36%) |
| record H2D transport | 3.15 s | 0.13 s (−96%) |
| app proving (end-to-end) | 86.97 s | 80.10 s (−7.9%) |

Baseline and this PR were measured back-to-back on the same container under
matched conditions (proving wall 818.8 s vs 818.6 s).

[Benchmark comparison](https://github.com/axiom-crypto/openvm-eth/actions/runs/29834295151)


yi-sun:
## CI corroboration (openvm-eth `compare-bench`)

Independently benchmarked through the openvm-eth CI on the standard block
(23992138), `prove-stark` mode, on the CI's g7e GPU runner. Base = latest
openvm `main`; target = this branch (`0a8a34cdc`).

| metric | base (main) | this PR | Δ |
| --- | ---: | ---: | ---: |
| trace generation | 9.55 s | 7.15 s | **−25.2%** |
| record H2D transport | 2.83 s | 0.18 s | **−93.8%** |
| init-memory transport | 2.00 s | 0.96 s | −52.2% |
| stark prove (excl. trace) | 47.37 s | 48.46 s | +2.3% |
| app-segment proving (`total_proof`) | 78.79 s | 72.92 s | **−7.5%** |
| STARK aggregation layer | 7.17 s | 7.09 s | −1.1% (untouched) |
| end-to-end block (`reth-block`) | 132.67 s | 127.31 s | **−4.0%** |
| total cells | 64,175,432 | 64,176,187 | ±0.0% |

- Trace generation −25% and record H2D −94% are the core wins.
- App-segment proving is −7.5%; the full block is −4.0%. The percentage on the
  full-block number is smaller because `prove-stark` includes the ~7 s STARK
  aggregation layer, which this PR does not touch (Δ −1.1%).
- `total_cells` is unchanged, so the proof is equivalent — same work, faster.
- The +2.3% in `stark_prove` (excl. trace) is expected phase re-attribution:
  with a single CUDA stream, some GPU work that previously blocked in the
  tracegen span now surfaces in the prove-timed span. Net is clearly positive.

Run: https://github.com/axiom-crypto/openvm-eth/actions/runs/29212678516

gaxiom (approved):
LGTM, just worried a bit if someone requests too big portion of pinned memory. But it should not break - just be slower

GunaDD (approved):
Looks good to me. Summary of the changes:
1. Optimize cpu proving context to gpu proving context transfer
*   Copy without `transport_matrix_h2d_col_major`: that helper ends with a full stream sync which stalls the host until every previously enqueued kernels and copy has drained (measured >1s per segment on reth-sized workloads when this runs mid-tracegen). This sync is unnecessary for the source’s lifetime: an async copy from pageable memory returns only once the source has been consumed by the staging pipeline, and all later reads of the destination are ordered by the stream.
2. Optimize / add more parallelization to tracegen for boundary chip
* Move inventory to merkle record conversion on the device instead of in the host to avoid unnecessary synchronization (if we do it in host).
3. Optimize / add more parallelization to tracegen for program chip
* Converts u32 frequencies to field element on device instead of on the host to avoid unnecessary synchronization
4. Optimization on tracing
* Do not emit span for chips with empty records during tracegen. This saves ~2ms per segment at most.
5. Add more parallelization in tracegen for finalizing system chips
* Compute out_num_records in host first instead of waiting for the result of the merge kernels to finish. 
6. Pinned memory for H2D performance optimization for the preflight buffers.

.cargo/config.toml
shuklaayush, line 1: revert this

GunaDD: but I need to update stark backend dependencies. is there another way to update stark backend dependencies ?

shuklaayush: you should be able to update it by removing `version = "2.0.0"` from `Cargo.toml` and changing from tag to branch `main`. if not, then there's a larger problem that needs to be fixed

crates/vm/cuda/src/system/inventory.cu
```diff
@@ -217,3 +217,48 @@ extern "C" int _inventory_merge_records_get_temp_bytes(
...
+    size_t num_records,
+    uint32_t *d_merkle_records,
+    cudaStream_t stream
+) {
+    if (num_records == 0) {
+        return 0;
+    }
+    const int threads = 256;
+    const size_t want = (num_records + threads - 1) / threads;
+    const int blocks = (int)(want < 1024 ? want : 1024);
```
shuklaayush: let's replace this with a util function like `kernel_launch_params`
```cpp
inline std::pair<dim3, dim3> grid_stride_launch_params(
    size_t count,
    size_t threads_per_block,
    size_t max_blocks
) {
    if (count == 0) {
        return {dim3(0, 1, 1), dim3(1, 1, 1)};
    }

    size_t grid = std::min(div_ceil(count, threads_per_block), max_blocks);
    return {dim3(grid, 1, 1), dim3(threads_per_block, 1, 1)};
}
```

crates/vm/cuda/src/system/program.cu
```diff
@@ -56,3 +56,34 @@ extern "C" int _program_cached_tracegen(
...
+    Fp *d_out,
+    size_t height,
+    cudaStream_t stream
+) {
+    if (height == 0) {
+        return 0;
+    }
+    const int threads = 256;
+    const size_t want = (height + threads - 1) / threads;
+    const int blocks = (int)(want < 2048 ? want : 2048);
```
shuklaayush: same

crates/vm/src/arch/extensions.rs
```diff
@@ -724,36 +724,57 @@ where
...
+                let air_name = self.inventory.airs.ext_airs[insertion_idx].name();
+                let labels = [
+                    ("air_name", air_name.to_string()),
+                    ("air_id", (num_sys_airs + insertion_idx).to_string()),
+                ];
+                metrics::counter!("trace_gen.record_arena_bytes", &labels)
+                    .absolute(allocated_bytes as u64);
+            }
+            chip.generate_proving_ctx(records)
+        };
```
shuklaayush, line 754: seems like this is only used in one place so i don't see the point of declaring a closure

crates/vm/src/system/cuda/memory.rs
```diff
@@ -189,28 +258,43 @@ impl MemoryInventoryGPU {
...
+            };
             let out_words = in_num_records
                 * (std::mem::size_of::<MemoryInventoryRecord<VM_DIGEST_WIDTH, BLOCKS_PER_LEAF>>()
                     / std::mem::size_of::<u32>());
-            let d_in_records = in_records
-                .to_device_on(&self.device_ctx)
-                .unwrap()
-                .as_buffer::<u32>();
+            let d_in_records = in_words.to_device_on(&self.device_ctx).unwrap();
+            openvm_cuda_common::pinned::give_back(h_in, dirty_len);
```
shuklaayush: maybe import `openvm_cuda_common::pinned` at top and use `pinned::give_back`

crates/vm/src/system/cuda/program.rs
```diff
@@ -104,19 +104,36 @@ impl Chip<Vec<u32>, GpuBackend> for ProgramChipGPU {
...
+                filtered_exec_freqs.as_ptr() as *const u8,
+                h_freqs.as_mut_ptr().add(off),
+                bytes,
+            );
+            std::slice::from_raw_parts(h_freqs.as_ptr().add(off) as *const u32, filtered_len)
+        };
+        let d_freqs = words
+            .to_device_on(&self.device_ctx)
+            .expect("failed to copy exec frequencies to device");
+        openvm_cuda_common::pinned::give_back(h_freqs, off + bytes);
```
shuklaayush: same

crates/vm/src/system/memory/online.rs
```diff
@@ -751,7 +751,12 @@ impl TracingMemory {
                             cell_size,
                         ))
                 });
-                ((addr_space, ptr), TimestampedValues { timestamp, values })
+                crate::system::TouchedBlock {
```
shuklaayush: import at top

crates/vm/src/system/cuda/memory.rs
```diff
@@ -189,28 +258,43 @@ impl MemoryInventoryGPU {
...
-                .map(|&((addr_space, ptr), ts_values)| MemoryInventoryRecord {
-                    address_space: addr_space,
-                    ptr,
-                    timestamps: [ts_values.timestamp],
-                    values: ts_values.values.map(Self::field_to_raw_u32),
-                })
-                .collect();
-            let in_num_records = in_records.len();
+            let _span = tracing::info_span!("mem_merge_records").entered();
+            const _: () = assert!(std::mem::size_of::<TouchedBlock<F>>() == 28);
```
shuklaayush: should it be this instead? also move outside somewhere at top or near the definition of `TouchedBlock`
```rust
const _: () = {
    assert!(
        std::mem::size_of::<TouchedBlock<F>>()
            == (3 + BLOCK_FE_WIDTH) * std::mem::size_of::<u32>(),
        "TouchedBlock must match the CUDA InRec size"
    );
    assert!(
        std::mem::align_of::<TouchedBlock<F>>() == std::mem::align_of::<u32>(),
        "TouchedBlock must match the CUDA InRec alignment"
    );
};
```

shuklaayush (commented):
some comments

crates/vm/src/system/cuda/memory.rs
GunaDD: second assertion is unnecessary since `TouchedBlock<F>` is `repr(C)`

crates/vm/src/arch/extensions.rs
GunaDD: I think its better readability

shuklaayush (approved):

PR #3008, opened 2026-07-13
perf(rvr): add direct GPU preflight record pipeline

yi-sun:
## Summary

This change adds an RVR preflight path that produces proof-ready records with substantially less host work while preserving interpreter parity.

- Adds cached native project generation, pooled scratch storage, and pinned host memory for repeated segment execution.
- Emits R4 arena-native records directly for fixed-shape RV64 families, including the W-family tail, eliminating their host assembly residual.
- Adds compact and chronological delta record streams with on-device decode for the GPU proving path.
- Extends log-native preflight coverage across the RV64IM and supported extension families.
- Fixes mixed LoadStore/REVEAL handling so public-value writes remain present in the shared AIR, including the segment-74 regression.
- Carries interpreter touched-page state across segment boundaries so the segment-1 sparse transfer includes prior writes.
- Adds CPU regression coverage for the touched-page and mixed-REVEAL cases, plus a CI build guard for the combined `cuda,rvr` feature set.

## Performance and GPU validation

On the 75-segment reference workload, host preflight averaged 238 ms/segment with RVR versus 349 ms/segment with the interpreter path. Both proving arms completed all 75 segments and verified the same `b0c6920a` result.

The preflight, record transport, and decode changes do not alter AIR constraints or the verifying-key shape.

## Validation

- Full RVR preflight selection: 34/34 passed.
- Segment-74 mixed-AIR REVEAL prove/verify fixture: passed.
- Interpreter segment-1 touched-pages fixture: passed.
- `cargo check --profile fast -p openvm-riscv-circuit --features rvr --tests`: passed.
- `cargo check --profile fast -p openvm-riscv-circuit --features cuda,rvr --tests`: passed with the repository-compatible CUDA/GCC configuration.
- Targeted Clippy with warnings denied: passed.
- Nightly rustfmt check: passed.
- `git diff --check`: passed.


yi-sun:
Closing in favor of #3020

PR #3015, opened 2026-07-14
ci(v2.1): tune SDK proof coverage and runner capacity

shuklaayush:
## Summary

- Runs the complete SDK EVM/Halo2 proof suite on both CPU configurations (`default` and `rvr`) without ignored proof tests.
- Runs the SDK CUDA lane with `cuda`, `root-prover`, `evm-verify`, and `halo2-gpu`, including `solc`, trusted setup, GPU driver repair, and CUDA architecture-aware Rust caching.
- Runs two SDK CUDA tests concurrently on the bounded `g7e.2xlarge+g6e.2xlarge` pool. Both options have 8 vCPUs, 64 GiB host memory, and at least 48 GB GPU memory.
- Uses bounded, same-size CPU and GPU fallback pools across workflows. Ordinary CUDA jobs use price-capacity-optimized allocation; the latency-sensitive SDK and CLI lanes retain their specialized allocation strategies.
- Keeps benchmark runs on a fixed `g7.4xlarge` baseline and resolves `target-cpu=native` to the concrete host CPU so cached artifacts remain valid and results remain comparable.
- Repairs the CUDA driver before architecture-sensitive cache setup, applies the same cache path to CUDA release builds, and keeps GPU validation after driver setup.

### DIFF
--- a/.github/runs-on.yml
+++ b/.github/runs-on.yml
@@ -9,7 +9,9 @@ images:
 runners:
   # runners for running tests
   test-gpu-nvidia:
-    family: ["g7e.2xlarge", "g7.2xlarge", "g6e.2xlarge"]
+    family: ["g7e.*", "g7.*", "g6e.*"]
+    cpu: [8, 16]
+    ram: [64]
     image: ubuntu24-gpu-x64
     spot: capacity-optimized
     extras: s3-cache
--- a/.github/runs-on.yml
+++ b/.github/runs-on.yml
@@ -9,12 +9,11 @@ images:
 runners:
   # runners for running tests
   test-gpu-nvidia:
-    family: ["g6.*", "g5", "g4dn", "g6e"]
+    family: ["g7e.2xlarge", "g7.2xlarge", "g6e.2xlarge"]
     image: ubuntu24-gpu-x64
-    cpu: [4, 8, 16, 32]
     spot: capacity-optimized
     extras: s3-cache
     preinstall: |
       nvidia-smi
       nvidia-smi -L
-      echo 'CUDA_ARCH=75,86,89' >> $GITHUB_ENV
+      echo 'CUDA_ARCH=86,89,120' >> $GITHUB_ENV
--- a/.github/runs-on.yml
+++ b/.github/runs-on.yml
@@ -9,13 +9,9 @@ images:
 runners:
   # runners for running tests
   test-gpu-nvidia:
-    family: ["g7e.*", "g7.*", "g6e.*"]
-    cpu: [8, 16]
-    ram: [64]
+    family: ["g7.4xlarge", "g6.4xlarge"]
     image: ubuntu24-gpu-x64
     spot: capacity-optimized
     extras: s3-cache
     preinstall: |
-      nvidia-smi
-      nvidia-smi -L
       echo 'CUDA_ARCH=86,89,120' >> $GITHUB_ENV
--- a/.github/runs-on.yml
+++ b/.github/runs-on.yml
@@ -9,9 +9,9 @@ images:
 runners:
   # runners for running tests
   test-gpu-nvidia:
-    family: ["g7.4xlarge", "g6.4xlarge"]
+    family: ["g7.4xlarge", "g6.4xlarge", "g5.4xlarge"]
     image: ubuntu24-gpu-x64
-    spot: capacity-optimized
+    spot: price-capacity-optimized
     extras: s3-cache
     preinstall: |
       echo 'CUDA_ARCH=86,89,120' >> $GITHUB_ENV
--- a/.github/workflows/benchmark-call.yml
+++ b/.github/workflows/benchmark-call.yml
@@ -11,7 +11,7 @@ on:
         type: string
         required: false
         description: The AWS instance type or family (e.g., m8g.16xlarge)
-        default: g7e.2xlarge+g7.2xlarge+g6e.2xlarge
+        default: g7e.2xlarge+g7.4xlarge+g6e.2xlarge
       memory_allocator:
         type: string
         required: false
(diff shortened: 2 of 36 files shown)


.github/workflows/benchmark-call.yml
```diff
@@ -10,8 +10,8 @@ on:
       instance_type:
         type: string
         required: false
-        description: The AWS instance type or family (e.g., m7g.16xlarge)
-        default: g6.2xlarge
+        description: The AWS instance type or family (e.g., m8g.16xlarge)
+        default: g7e.2xlarge+g7.4xlarge+g6e.2xlarge
```
stephenh-axiom-xyz: Shouldn't this be a consistent machine? As the micro benchmarks are supposed to give us an idea on perf?

.github/workflows/lints.yml
```diff
@@ -157,7 +157,7 @@ jobs:
     needs: changes
     if: ${{ needs.changes.outputs.cuda == 'true' }}
     runs-on:
-      - runs-on=${{ github.run_id }}-cuda/image=ubuntu24-gpu-x64/family=g6.*+g5+g6e/cpu=8/volume=40gb/spot=capacity-optimized/extras=s3-cache
+      - runs-on=${{ github.run_id }}-cuda/image=ubuntu24-gpu-x64/family=g7e.*+g7.*+g6e.*/cpu=8+16/ram=64/volume=40gb/spot=capacity-optimized/extras=s3-cache
```
stephenh-axiom-xyz: These machines seem a little overkill for CUDA lints

crates/sdk/src/tests.rs
stephenh-axiom-xyz, line 1: Don't agree with the set of ignored tests. While yes some cases we can probably omit, we should also ensure that the CPU prover is correct also

stephenh-axiom-xyz (commented):
Why are we changing the CUDA machine set to one that's more expensive? Are there no longer any g5 runners?

.github/workflows/lints.yml
shuklaayush: yeah, but i think it's fine. it seems like g5/g6 sometimes have less capacity than the newer machines

crates/sdk/src/tests.rs
shuklaayush: yes, but i think we should have smaller test cases for cpu. these take around an hour to finish at the moment which seems unreasonably long
or we can put them under a different label but i feel like that'll just be brushing it under the carpet

.github/workflows/lints.yml
stephenh-axiom-xyz: Why not include g5/g6 in the list then? And practically I've never had to wait on this lint, it's always been the g6e ones

crates/sdk/src/tests.rs
stephenh-axiom-xyz: An hour is an overstatement, [this run](https://github.com/openvm-org/openvm/actions/runs/29102675096/job/86395257859) took < 20 minutes for example. We can reduce the set of tests that we run, but we should still run probably the base fibonacci some deferral STARK proves

.github/workflows/benchmark-call.yml
shuklaayush: yeah, that's a good point. let me see which one has least capacity issues

crates/sdk/src/tests.rs
shuklaayush: hmm it did take an hour here https://github.com/openvm-org/openvm/actions/runs/29313052465/job/87029797166?pr=3013

stephenh-axiom-xyz: That sounds like something you should look into 🥲

.github/workflows/benchmark-call.yml
shuklaayush: kept only `g7.4xlarge` which seems the most available

.github/workflows/lints.yml
shuklaayush: i added back `g5`, `g6` and changed it to `price-capacity-optimized` so it selects the one which is cheapest if capacity is available

crates/sdk/src/tests.rs
shuklaayush: okay, looks like that was because there was a `test-threads=1` added in this branch at some point. it takes much less time now so i removed the feature and the set of ignored tests

.github/workflows/continuations.yml
```diff
@@ -21,7 +21,7 @@ jobs:
   tests:
     name: Recursion tests
     runs-on:
-      - runs-on=${{ github.run_id }}/runner=16cpu-linux-arm64/extras=s3-cache
+      - runs-on=${{ github.run_id }}/family=c9g.4xlarge+c8g.4xlarge/image=ubuntu24-full-arm64/extras=s3-cache
```
stephenh-axiom-xyz, line 24: Does this increase or decrease the set of machines we can use?

.github/runs-on.yml
```diff
@@ -9,12 +9,9 @@ images:
 runners:
   # runners for running tests
   test-gpu-nvidia:
-    family: ["g6.*", "g5", "g4dn", "g6e"]
+    family: ["g7.4xlarge", "g6.4xlarge", "g5.4xlarge"]
```
stephenh-axiom-xyz, line 12: I don't think these actually go into effect until this is merged into `main`

stephenh-axiom-xyz (approved):
LGTM with some comments. Also consider adding some new `runs-on.yml` runner configs so we don't have to repeat the machine list each time.

.github/runs-on.yml
shuklaayush: yes, this is just for consistency. it's not used

.github/workflows/continuations.yml
shuklaayush: decrease but it should make things a bit more predictable. i don't think we'll need it but we can add more instance types here if it becomes capacity limited at any point

PR #3018, opened 2026-07-14
ci(v2.1): speed up crate docs publishing

shuklaayush:
## Summary
- build workspace docs once with `--no-deps`, avoiding dependency docs and the second index pass
- deny rustdoc warnings and keep published S3 prefixes aligned with generated output

## Validation
- CUDA workspace docs build with warnings denied
- `actionlint .github/workflows/docs.yml .github/workflows/lints.yml`

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -54,21 +54,10 @@ jobs:
           nvidia-smi -L
           echo "CUDA_ARCH=86,89,120" >> $GITHUB_ENV
 
-      - name: Cargo check
-        run: cargo check
-
       - name: Build documentation
-        run: cargo +nightly-2026-01-18 doc --workspace --exclude "openvm-benchmarks" --exclude "*-tests" --exclude "*-test" --features "cuda"
-
-      # We only want the index page to display workspace crates, so we build
-      # separately and copy over the index as a hack
-      - name: Build index page
-        run: |
-          cargo +nightly-2026-01-18 doc --workspace --no-deps --exclude "openvm-benchmarks" --exclude "*-tests" --exclude "*-test" --target-dir target/doc-nodeps
-          cp target/doc-nodeps/doc/index.html target/doc/
-          cp target/doc-nodeps/doc/crates.js target/doc/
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -54,21 +54,10 @@ jobs:
           nvidia-smi -L
           echo "CUDA_ARCH=86,89,120" >> $GITHUB_ENV
 
-      - name: Cargo check
-        run: cargo check
```
stephenh-axiom-xyz, line 58: Do we `cargo check` anywhere else? Perhaps move this to the lint?

.github/workflows/docs.yml
```diff
@@ -54,21 +54,10 @@ jobs:
...
-        run: cargo +nightly-2026-01-18 doc --workspace --exclude "openvm-benchmarks" --exclude "*-tests" --exclude "*-test" --features "cuda"
-
-      # We only want the index page to display workspace crates, so we build
-      # separately and copy over the index as a hack
-      - name: Build index page
-        run: |
-          cargo +nightly-2026-01-18 doc --workspace --no-deps --exclude "openvm-benchmarks" --exclude "*-tests" --exclude "*-test" --target-dir target/doc-nodeps
-          cp target/doc-nodeps/doc/index.html target/doc/
-          cp target/doc-nodeps/doc/crates.js target/doc/
+        run: cargo +nightly-2026-01-18 doc --workspace --no-deps --exclude "openvm-benchmarks*" --exclude "*-tests" --features "cuda"
```
stephenh-axiom-xyz, line 58: Assuming `no-deps` avoids building non-workspace crates?

stephenh-axiom-xyz (approved):
LGTM as long as docs are verified to be correct

.github/workflows/docs.yml
shuklaayush: we do a full build in the build workflow so we don't need `cargo check`

.github/workflows/docs.yml
shuklaayush: yes it avoids building the docs, it does still compile the dependency crates as required

PR #3019, opened 2026-07-14
ci(v2.1): deduplicate workflows and target affected test suites

shuklaayush:
### Extension and guest-library tests

- Added `.github/filters/extension-tests.yml` and `.github/filters/guest-lib-tests.yml` to map changed paths to the test suites they affect.
- CPU workflows now create jobs only for the affected extensions or guest libraries. Extension jobs still cover default and RVR modes, plus the existing unprotected RISC-V case. Guest-library jobs still cover default and RVR modes where supported.
- CUDA workflows use the same path maps instead of maintaining separate inline filters. They keep the existing GPU runner batches: RISC-V versus the other extensions, and hashes, curves, algebra, and verify-stark for guest libraries.
- Replaced the CUDA shell loops and `jq` checks with explicit jobs and named test steps, so the selected coverage is visible in the Actions UI.

### Duplicate work removed

- Removed the second extension-test invocation from the PR-title workflow.
- Non-benchmark labels no longer run benchmark jobs or cancel an active benchmark run. The SDK workflow applies the same handling to labels other than `run-sdk-tests`.
- Removed unused CUDA matrix logic from the VM and primitives CPU workflows. Their CUDA coverage remains in `base-tests.cuda.yml`.
- Limited the CUDA CLI job to the four non-ignored integration tests that actually run app or STARK proving. The CPU CLI job continues to cover init, example builds, execution modes, and build determinism.
- Removed unconditional `solc` setup from normal CLI runs, unused Foundry setup from SDK sweeps, a repeated `rust-src` installation, and `cargo clean` before the cached workspace build.
- Docs install `s5cmd` only when publishing.

### Trigger and coverage fixes

- Added missing path triggers for shared VM/toolchain code, RVR code, SDK configuration, recursion/verification dependencies, Cargo lockfiles, toolchain configuration, cache actions, and GPU setup scripts.
- CUDA clippy now runs for ordinary Rust files as well as CUDA source files, because CUDA-gated code is not limited to files named `cuda.rs`.
- RISC-V vector tests now collect failed vectors and return an error instead of printing failures while passing the test.
- CUDA release binaries now run an app build, prove, and verify smoke test instead of checking only `--version`.

### Actions names

- Standardized workflow titles and job labels so the workflow identifies the suite and backend while each job adds only its component, mode, or action.
- Matrix jobs use canonical names such as `algebra (RVR)` and `RISC-V (unprotected)`. SDK and RISC-V jobs distinguish standalone modes and runtime versus prover coverage.
- Reusable CUDA checks render as concise hierarchies such as `Guest Library Tests (CUDA) / hashes / Run tests`.
- Benchmark checks show the benchmark, E2E mode, runner, and allocator, for example `pairing (E2E) | g7.4xlarge | jemalloc`, instead of the raw matrix tuple.

### Workflow setup

- Path-only selector jobs use explicit read permissions and no longer load the deploy key.
- Jobs that compile dependencies keep the deploy key, including the generic build and CLI smoke jobs.
- Push and pull-request path lists that must stay identical now use YAML anchors to avoid drift.
- The CUDA SDK test can fall back to `g7.4xlarge` when G7e and G6e capacity is unavailable. It uses one test thread on a 32 GB GPU and two on 48 or 96 GB GPUs.

## Validation

- `actionlint .github/workflows/*.yml`
- `git diff --check develop-v2.1.0...HEAD`
- Audited workflow-local scripts, configs, actions, and working directories for stale references.
- Verified the generated matrix and reusable-workflow labels in the PR's rendered Actions checks.

Resolves INT-8840


.github/filters/extension-tests.yml
```diff
@@ -0,0 +1,49 @@
...
+algebra: &algebra
+  - *riscv
+  - "extensions/algebra/**"
+  - "extensions/ecc/**"
+  - "extensions/pairing/guest/**"
+  - "extensions/riscv-adapters/**"
+ecc:
+  - *algebra
+  - "crates/sdk/**"
+  - "crates/sdk-config/**"
```
stephenh-axiom-xyz, line 43: Why does `ecc` get special treatment?

.github/filters/guest-lib-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+  - ".github/workflows/guest-lib-tests.yml"
+  - "crates/sdk/guest/fib/**"
+  - "ci/install-openvm-toolchain.sh"
+cuda:
+  - ".github/actions/rust-cache-cuda/**"
+  - ".github/workflows/guest-lib-tests.cuda*.yml"
+  - "scripts/gpu_driver_setup.sh"
+sha2:
+  - *shared
+  - "crates/sdk/**"
```
stephenh-axiom-xyz, line 25: Same question as above, why are we running sha2 + keccak256 on SDK bump?

.github/workflows/benchmark-call.yml
```diff
@@ -1,4 +1,4 @@
-name: "Benchmark: running individual benchmarks"
+name: Run Benchmark
```
stephenh-axiom-xyz: Run Individual Benchmark

.github/workflows/primitives.yml
```diff
@@ -23,13 +25,10 @@ env:
 
 jobs:
   build:
-    strategy:
-      matrix:
-        platform:
-          - { family: "c8i-flex.16xlarge+c8i.16xlarge+c8a.16xlarge+c7i.16xlarge+c7a.16xlarge", image: "ubuntu24-full-x64" }
+    name: CPU (parallel)
```
stephenh-axiom-xyz, line 28: No CUDA tests?

.github/workflows/vm.yml
```diff
@@ -22,13 +25,10 @@ env:
 
 jobs:
   tests:
-    strategy:
-      matrix:
-        platform:
-          - { family: "c8i-flex.16xlarge+c8i.16xlarge+c8a.16xlarge+c7i.16xlarge+c7a.16xlarge", image: "ubuntu24-full-x64" }
+    name: CPU (parallel, basic-memory)
```
stephenh-axiom-xyz, line 28: Only CPU?

.github/workflows/extension-tests.cuda.group.yml
```diff
@@ -0,0 +1,151 @@
...
+        run: |
+          nvidia-smi
+          nvidia-smi -L
+          nvcc --version
+          echo "CUDA_ARCH=86,89,120" >> $GITHUB_ENV
+          echo "NEXTEST_ENV=CUDA_OPT_LEVEL=3" >> $GITHUB_ENV
+          echo "FEATURE_ARGS=--features=cuda" >> $GITHUB_ENV
+          echo "VPMM_PAGES=1" >> $GITHUB_ENV
+
+      - name: Run RISC-V tests
```
stephenh-axiom-xyz: Not a fan of all these duplicate steps, surely there is a more idiomatic way to do this

.github/workflows/extension-tests.cuda.yml
```diff
@@ -29,129 +41,62 @@ env:
...
-              echo "Skipping tests for $ext as no relevant files changed."
-            fi
-          done
-
-      - name: sccache stats
-        if: always()
-        run: sccache --show-stats || true
+          filters: .github/filters/extension-tests.yml
+
+  riscv-cuda:
```
stephenh-axiom-xyz, line 62: Why is this an extra step?

.github/workflows/extension-tests.yml
stephenh-axiom-xyz, line 1: Feels like you should be able to re-use the `group.yml` here and pass in features?

stephenh-axiom-xyz (commented):
Some comments. Don't really like how extensions and guest libraries are being done now, would rather lean more into matrix use than remove it

.github/filters/extension-tests.yml
shuklaayush: because it's a crate dependency for ecc
https://github.com/openvm-org/openvm/blob/e25813b38eb4376172e3424155eed7bd3f598d4a/extensions/ecc/tests/Cargo.toml#L20-L21

.github/filters/guest-lib-tests.yml
shuklaayush: because the guest library tests depend on it
https://github.com/openvm-org/openvm/blob/e25813b38eb4376172e3424155eed7bd3f598d4a/guest-libs/keccak256/Cargo.toml#L22

.github/workflows/vm.yml
shuklaayush: runs as part of the other workflow like before
https://github.com/openvm-org/openvm/blob/e25813b38eb4376172e3424155eed7bd3f598d4a/.github/workflows/base-tests.cuda.yml#L29

.github/workflows/primitives.yml
shuklaayush: runs as part of the other workflow like before
https://github.com/openvm-org/openvm/blob/e25813b38eb4376172e3424155eed7bd3f598d4a/.github/workflows/base-tests.cuda.yml#L29

.github/workflows/extension-tests.cuda.yml
shuklaayush: it's to avoid provisioning an extra gpu machine when it's not needed. earlier even if any extension changed, it would start the riscv tests on a gpu machine and then exit early

.github/workflows/extension-tests.yml
shuklaayush: not sure which group.yml you're referring to but the cuda one does cuda setup which is not relevant here

.github/workflows/extension-tests.cuda.yml
stephenh-axiom-xyz: Mmm I meant why are `riscv-cuda` and the other extensions separated? RISC-V tests running imply all the other extensions, is it to divide the work?

.github/workflows/extension-tests.yml
stephenh-axiom-xyz: Setup could be optional? Right now if I were to add/modify the workflow for an extension I'd need to modify at the very least 4 (8 if there's a guest lib) files, which doesn't seem great

.github/workflows/extension-tests.cuda.yml
shuklaayush: yes, and also it's how it was earlier

.github/workflows/extension-tests.yml
shuklaayush: okay, i moved it into a common group

stephenh-axiom-xyz (approved):

PR #3021, opened 2026-07-15
refactor(v2.1): remove generic `F` from rvr structs

shuklaayush:
## Summary

This PR removes the proof-field generic `F` from RVR types that only work with native instructions, byte memory, streams, or compiled C artifacts.

`F` is kept only at the boundaries that still need it:

- converting `VmExe<F>` instructions into canonical integer operands during compilation
- reading and writing native field elements in the deferral address space

## What changed

- Converts each field-typed OpenVM instruction into a field-independent `RvrInstruction` before lifting it to RVR IR.
- Makes the RVR lifter registry, generated project, compiled artifact, IO state, and pure/metered instance types field-independent.
- Stops retaining the full `VmExe<F>` in RVR instances. Instances keep only the starting PC and sparse initial-memory image needed to create execution state.
- Keeps compile-time lifters separate from the small set of runtime host callbacks. Only the deferral callback remains specialized for `F`, because deferral memory stores native field elements.
- Removes now-unused field dependencies from the RVR extension crates and simplifies the SDK compiled-instance aliases.
- Uses thread-local C callback slots so concurrent RVR executions on different threads do not overwrite each other's host context.
- Builds the generated native C with the C2Y language mode.
- Adds coverage for canonical instruction lowering, signed offsets, current ADD/SUB/ADDI operand rules, and concurrent host contexts.

## Relation to earlier PRs

This is a follow-up to two already-merged cleanups:

- #2975 changed input and hint streams to store bytes directly.
- #2983 removed `F` from streams, stdin, execution contexts, and the other shared VM types that no longer depended on field elements.

With those changes merged, RVR was the remaining backend carrying `F` through native runtime types mostly because its old instruction and callback interfaces were still field-typed. This PR moves the field conversion to the compilation/deferral boundaries and removes that generic from the rest of the RVR path.

The branch is based directly on the merged #2983 commit and contains only the RVR follow-up changes.

## Validation

- Nightly rustfmt check
- Targeted RVR clippy checks with warnings denied
- `rvr-openvm-lift` tests
- RVR extension and SDK/circuit clippy checks
- Concurrent RVR host-context test

Resolves INT-8781

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7242,7 +7242,6 @@ dependencies = [
  "openvm-instructions",
  "openvm-platform",
  "openvm-riscv-guest",
- "openvm-stark-backend",
  "rvr-openvm-build",
  "rvr-openvm-ext-ffi-common",
  "rvr-openvm-ir",
@@ -7262,7 +7261,6 @@ dependencies = [
  "openvm-algebra-transpiler",
  "openvm-algebra-utils",
  "openvm-instructions",
- "openvm-stark-backend",
  "rand 0.9.2",
  "rvr-openvm-build",
  "rvr-openvm-ir",
@@ -7313,7 +7311,6 @@ dependencies = [
  "openvm-bigint-transpiler",
  "openvm-instructions",
  "openvm-riscv-transpiler",
- "openvm-stark-backend",
  "rvr-openvm-build",
  "rvr-openvm-ir",
  "rvr-openvm-lift",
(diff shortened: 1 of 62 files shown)


mansur20478 (approved):
looks good, no obvious issues. I guess `rv64io_callbacks.h` and `rv64i_phantom_callbacks.h` comments could be updated to match runtime hooks

gdmlcjs (approved):

PR #3022, opened 2026-07-15
perf(v2.1): remove redundant AddSubW range check

shuklaayush:
## Summary

Skips the AddSub core range interaction for the top output limb in word operations because the W adapter already proves that limb is a canonical u16 while extracting its sign bit. Full per-limb range checks remain enabled for regular RV64 and bigint AddSub users, with the same specialization applied to CUDA trace generation.

### DIFF
--- a/extensions/riscv/circuit/src/adapters/alu_w_u16.rs
+++ b/extensions/riscv/circuit/src/adapters/alu_w_u16.rs
@@ -161,10 +161,9 @@ impl<AB: InteractionBuilder> VmAdapterAir<AB> for Rv64BaseAluWU16AdapterAir {
             )
             .eval(builder, local.rs2_as);
 
-        // Sign-extend the 32-bit result to 64 bits: `result_sign` is bit 15 of the most
-        // significant low-word limb. Decompose that limb as `low15 + result_sign * 2^15` and
-        // range-check `low15` to 15 bits; combined with the core's per-limb 16-bit range check,
-        // this forces `result_sign` to equal the limb's top bit.
+        // Sign-extend the 32-bit result to 64 bits. Decomposing the top result limb as
+        // `low15 + result_sign * 2^15`, with a 15-bit range check on `low15`, both proves that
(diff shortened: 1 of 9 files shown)


extensions/riscv/circuit/src/adapters/alu_w_u16.rs
```diff
@@ -161,10 +161,9 @@ impl<AB: InteractionBuilder> VmAdapterAir<AB> for Rv64BaseAluWU16AdapterAir {
             )
             .eval(builder, local.rs2_as);
 
-        // Sign-extend the 32-bit result to 64 bits: `result_sign` is bit 15 of the most
-        // significant low-word limb. Decompose that limb as `low15 + result_sign * 2^15` and
-        // range-check `low15` to 15 bits; combined with the core's per-limb 16-bit range check,
-        // this forces `result_sign` to equal the limb's top bit.
+        // Sign-extend the 32-bit result to 64 bits. Decomposing the top result limb as
+        // `low15 + result_sign * 2^15`, with a 15-bit range check on `low15`, both proves that
+        // the limb is a canonical u16 and forces `result_sign` to equal its top bit.
```
GunaDD, line 166: are we sure that the sign bit extraction still works correctly when the upper limb is not canonical u16 ?

shuklaayush: It is constrained to fit in `u16`. Since `result_sign` is boolean, there are two cases in the range check:

1. $\texttt{result\\_sign} = 0 \implies 0 \le \texttt{result\\_high} < 2^{15}$
2. $\texttt{result\\_sign} = 1 \implies 0 \le \texttt{result\\_high} - 2^{15} < 2^{15} \implies 2^{15} \le \texttt{result\\_high} < 2^{16}$

GunaDD: makes sense

GunaDD (approved):
LGTM

PR #3023, opened 2026-07-15
perf(v2.1): split word immediate instructions

shuklaayush:
## Summary

- Splits `ADDIW`, `SLLIW`, `SRLIW`, and `SRAIW` from their register-form word ALU chips, with dedicated immediate opcodes and adapters.
- Updates the transpiler, RVR lifting, CPU/CUDA registration and trace generation, and adds boundary coverage for the new word-immediate paths.
- Specializes the ADDI core range checks so the most-significant output limb is not checked redundantly when the adapter already constrains it.

Resolves INT-8508


extensions/riscv/transpiler/src/instructions.rs
```diff
@@ -373,6 +417,20 @@ pub enum DivRemWOpcode {
 #[repr(usize)]
 pub enum BaseAluImmOpcode {
     ADDI,
+    XORI,
+    ORI,
+    ANDI,
+}
+
+impl From<BaseAluImmOpcode> for BaseAluOpcode {
```
mansur20478: not sure, but is it possible to avoid it? it seems that it was done so that we could reuse `run_bitwise_logic` for immediate. what if instead we had `run_bitwise_logic_imm` that would then send to `run_bitwise_logic` depending on BaseAluImmOpcode?

mansur20478 (commented):
Looks good

extensions/riscv/transpiler/src/instructions.rs
shuklaayush: sure

extensions/riscv/circuit/cuda/src/addi_w.cu
```diff
@@ -0,0 +1,69 @@
+#include "launcher.cuh"
```
GunaDD, line 1: maybe `addi_w.cu` instead of `addiw.cu` ?

extensions/riscv/circuit/src/extension/cuda.rs
```diff
@@ -89,6 +91,20 @@ impl VmProverExtension<GpuBabyBearPoseidon2Engine, DenseRecordArena, Rv64I> for
             Rv64ShiftWRightArithmeticChipGpu::new(range_checker.clone(), timestamp_max_bits);
         inventory.add_executor_chip(shift_w_right_arithmetic);
 
+        inventory.next_air::<Rv64AddIWAir>()?;
+        let addiw = Rv64AddIWChipGpu::new(range_checker.clone(), timestamp_max_bits);
+        inventory.add_executor_chip(addiw);
```
GunaDD: `addi_w` here too ?

GunaDD (approved):
Should we fix the `addiw` naming ?

PR #3029, opened 2026-07-17
feat: make static verifier circuit construction generic

Aalanli:
This PR makes the static verifier circuit generic over the chip traits, so in the next PR: https://github.com/openvm-org/openvm/pull/3030, we can extract an IR from the circuit definition to greatly accelerate trace generation.

It does not change the circuit in any way; it's a purely mechanical refactor.

## Summary
- Introduces chip backend traits (`chip_traits.rs`) so `StaticVerifierCircuit` population is generic over the constraint backend instead of being hardwired to halo2-base
- Adds `Halo2Backend`  that is the original halo2-base `Context`, preserves existing circuit semantics)
- Rewrites the security-critical stages (`batch_constraints`, `stacked_reduction`, `whir`, `full_pipeline`, `shared_math`) against the generic traits; constraint semantics are intended to be identical and are cross-checked against the previous backend in tests


Aalanli:
Hi @stephenh-axiom-xyz , can you take a look? I added an integration test that asserts the vk generated matches main.

crates/static-verifier/src/chip_traits.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// `a^(2^n)`
+    fn ext_pow_power_of_two(
+        &mut self,
+        a: BabyBearExt4Wire<Self::F>,
+        n: usize,
+    ) -> BabyBearExt4Wire<Self::F>;
+}
+
+/// Poseidon2 digest hashing/compression mirroring the helpers in `hash::poseidon2`.
+pub trait DigestHashInst: ChipBase {
```
stephenh-axiom-xyz: nit: `Poseidon2Inst`

crates/static-verifier/src/halo2_backend.rs
```diff
@@ -0,0 +1,407 @@
...
+        &self.ext
+    }
+
+    /// Escape hatch for code that needs the raw context (e.g. test pranks).
+    pub fn ctx_mut(&mut self) -> &mut Context<Fr> {
+        self.ctx
+    }
+
+    /// Disjoint borrows of the extension chip and the context.
+    fn parts(&mut self) -> (&BabyBearExt4Chip, &mut Context<Fr>) {
```
stephenh-axiom-xyz: nit: `ext_parts`

crates/static-verifier/src/backend.rs
stephenh-axiom-xyz, line 1: nit: maybe just `backend.rs`? Halo2 is apparent

stephenh-axiom-xyz (approved):
LGTM, very clean change 🧼 🫧 . Please run SDK EVM sweep before merging though

Aalanli:
https://github.com/openvm-org/openvm/actions/runs/30123301266

PR #3031, opened 2026-07-17
refactor(v2.1): specialize rvr state by execution mode

shuklaayush:
## Summary

This PR is mostly refactoring and cleanup. The shared tracer and suspender abstraction had run its course: it made every RVR execution mode carry the same accounting and suspension machinery, which blocked further pure-execution performance work. Giving each mode its own state removes that constraint and unlocks a smaller hot path.

- Pure RVR keeps only the state needed to run to completion. Its generated blocks have no instruction-retirement accounting or suspension check.
- Pure execution with instruction tracking is a separate compile mode. It carries the remaining instruction count through generated block calls and writes the retired count when execution returns.
- Metered and metered-cost execution keep their own mode data directly in the RVR state.
- The shared tracer and suspender layers are removed.
- Loaded RVR artifacts are checked against the execution mode requested by the caller.

## API changes

The public API now makes each capability explicit.

- `sdk.compile(...)` returns a completion-only pure artifact.
- `sdk.compile_with_instret_tracking(...)` returns a tracked pure artifact that supports `execute_for` and `execute_from_state_for`.
- Bounded execution returns `ExecutionOutcome::Terminated` or `ExecutionOutcome::Suspended`.
- Completion-only execution returns the final VM state directly. Tracked completion also returns the instructions retired by that call.
- Interpreter and preflight bounded calls use `_for` methods instead of optional instruction limits.
- Metered segment execution uses the same explicit terminated-or-suspended outcome.

## Metrics

- Tracked pure execution continues to report the instructions retired by the current call.
- Completion-only pure RVR does not report instruction-count or throughput metrics because it does not track retired instructions.
- Full metered RVR reports the current call's instruction delta instead of the cumulative instruction count.

## Validation

- Built the SDK with and without RVR.
- Built the RVR and metrics feature combination.
- Ran Clippy with warnings denied on an x86 host.
- Ran the RVR suspension and resume integration test.
- Ran the tracked artifact save and load SDK test.
- Checked the public API use in openvm-eth and rvr-openvm; neither needs a source change.

Resolves INT-8839


shuklaayush:
i'm going to merge this since these are primarily internal api cleanups/changes and shouldn't affect the public facing apis

PR #3032, opened 2026-07-17
ci: right-size cpu runners

shuklaayush:
Right-sizes CPU CI runners from 16xlarge to workload-specific 4xlarge and 8xlarge pools.

- Uses Graviton C/M pools for ARM-compatible test jobs.
- Runs the cold workspace build on 128 GiB M8/M9 8xlarge instances; its concurrent linker fanout exceeds the 64 GiB available on C8/C9 8xlarge.
- Keeps AOT and asm-dependent jobs on x64 C/M pools.
- Adds reusable CPU runner aliases for follow-up workflow cleanup.
- Makes benchmark and SDK labels override path filters without launching duplicate target workflows; `run-sdk-tests` also runs the CUDA SDK suite.

Workflows use explicit labels in this PR because RunsOn reads public-repository configuration from the default branch; the aliases can be adopted after this merges.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
   lint:
     name: Cargo build
     runs-on:
-      - runs-on=${{ github.run_id }}/family=c9g.8xlarge+c8g.8xlarge+m9g.8xlarge+m8g.8xlarge/image=ubuntu24-full-arm64/volume=80gb/extras=s3-cache
+      - runs-on=${{ github.run_id }}/family=m9g.8xlarge+m8g.8xlarge/image=ubuntu24-full-arm64/volume=80gb/extras=s3-cache
     steps:
       - uses: runs-on/action@v2
       - name: Load SSH key
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
   lint:
     name: Cargo build
     runs-on:
-      - runs-on=${{ github.run_id }}/volume=80gb/family=c8i-flex.16xlarge+c8i.16xlarge+c8a.16xlarge+c7i.16xlarge+c7a.16xlarge/image=ubuntu24-full-x64/extras=s3-cache
+      - runs-on=${{ github.run_id }}/family=c9g.4xlarge+c8g.4xlarge+m9g.4xlarge+m8g.4xlarge/image=ubuntu24-full-arm64/volume=80gb/extras=s3-cache
     steps:
       - uses: runs-on/action@v2
       - name: Load SSH key
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
   lint:
     name: Cargo build
     runs-on:
-      - runs-on=${{ github.run_id }}/family=m9g.4xlarge+m8g.4xlarge/image=ubuntu24-full-arm64/volume=80gb/extras=s3-cache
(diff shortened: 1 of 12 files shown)


.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
   lint:
     name: Cargo build
     runs-on:
-      - runs-on=${{ github.run_id }}/volume=80gb/family=c8i-flex.16xlarge+c8i.16xlarge+c8a.16xlarge+c7i.16xlarge+c7a.16xlarge/image=ubuntu24-full-x64/extras=s3-cache
+      - runs-on=${{ github.run_id }}/family=m9g.8xlarge+m8g.8xlarge/image=ubuntu24-full-arm64/volume=80gb/extras=s3-cache
```
stephenh-axiom-xyz, line 18: Why the switch from x64 to arm?

.github/workflows/label-overrides.yml
stephenh-axiom-xyz, line 1: Please verify this works as intended

.github/workflows/extension-tests.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     uses: ./.github/workflows/extension-tests.group.yml
     with:
       runner_id: cpu-${{ strategy.job-index }}
-      machine_type: family=c8i-flex.16xlarge+c8i.16xlarge+c8a.16xlarge+c7i.16xlarge+c7a.16xlarge/image=ubuntu24-full-x64/volume=80gb
+      machine_type: ${{ matrix.extension == 'algebra' && (matrix.mode == 'aot' && 'family=c8i-flex.4xlarge+c8i.4xlarge+c8a.4xlarge+m8i-flex.4xlarge+m8i.4xlarge+m8a.4xlarge+c7i.4xlarge+c7a.4xlarge+m7i.4xlarge+m7a.4xlarge/image=ubuntu24-full-x64/volume=80gb' || 'family=c9g.4xlarge+c8g.4xlarge+m9g.4xlarge+m8g.4xlarge/image=ubuntu24-full-arm64/volume=80gb') || (matrix.mode == 'aot' && 'family=c8i-flex.8xlarge+c8i.8xlarge+c8a.8xlarge+m8i-flex.8xlarge+m8i.8xlarge+m8a.8xlarge+c7i.8xlarge+c7a.8xlarge+m7i.8xlarge+m7a.8xlarge/image=ubuntu24-full-x64/volume=80gb' || 'family=c9g.8xlarge+c8g.8xlarge+m9g.8xlarge+m8g.8xlarge/image=ubuntu24-full-arm64/volume=80gb') }}
```
stephenh-axiom-xyz, line 68: Agree with the change, but is there no cleaner way to do this (like define the variable somewhere else or smth)?

stephenh-axiom-xyz (approved):
LGTM, although please (a) verify that switching to arm64 from x86 will have no unintended consequences and (b) check for simpler ways to define machine type

.github/workflows/label-overrides.yml
shuklaayush: yes it works on this pr which doesn't touch sdk or benchmarks

.github/workflows/build.yml
shuklaayush: we test x86 in the cuda tests. also arm is cheaper because of graviton

.github/workflows/extension-tests.yml
shuklaayush: i think it's fine. we'll anyways get rid of aot in 2.1

PR #3034, opened 2026-07-17
fix(v2.1): emit `unimp` instruction for undecodable words

shuklaayush:
Executable ELF segments can contain embedded data such as jump tables or constant pools. Previously, transpilation failed whenever no registered extension could decode one of these words, even if it was never executed.

This change emits an `unimp` instruction at the corresponding program slot instead. The original word remains unchanged in initialized memory, so embedded data can still be read normally; the program terminates only if control flow reaches that slot. Ambiguous instruction decodings remain an error.

Includes a regression test with an undecodable word between valid RV64 instructions.

Resolves INT-8838

### DIFF
--- a/crates/toolchain/tests/tests/rv64_transpiler_tests.rs
+++ b/crates/toolchain/tests/tests/rv64_transpiler_tests.rs
@@ -63,6 +63,30 @@ fn test_transpile_addi_immediate_boundaries(imm: i32, expected_c: u32) -> Result
     Ok(())
 }
 
+#[test]
+fn test_transpile_rv64_undecodable_word_as_unimp() -> Result<()> {
+    const ADDI_X3_X0_1: u32 = 0x0010_0193;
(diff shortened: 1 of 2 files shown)


crates/toolchain/tests/tests/rv64_transpiler_tests.rs
```diff
@@ -63,6 +63,30 @@ fn test_transpile_addi_immediate_boundaries(imm: i32, expected_c: u32) -> Result
     Ok(())
 }
 
+#[test]
+fn test_transpile_rv64_undecodable_word_as_unimp() -> Result<()> {
+    const ADDI_X3_X0_1: u32 = 0x0010_0193;
+    const UNDECODABLE_WORD: u32 = 0x1000_0200;
```
stephenh-axiom-xyz, line 69: nit: perhaps `REPLACE_ME`?

stephenh-axiom-xyz (approved):
LGTM

PR #3037, opened 2026-07-17
refactor: use projective coordinates in ECC extension's rvr

mansur20478:
This PR refactors ECC extension's rvr to use projective coordinates. Add and Double logic is reused from the interpreter version. However, for `k256`, it had to be reimplemented as it relies on `secp256k1`. It is follow-up work after adapting projective coordinates in interpreter mode (https://github.com/openvm-org/openvm/pull/3035).

resolves INT-8788


shuklaayush:
Superseded by #3035, which now contains the full projective-coordinate stack rebased onto the latest develop-v2.1.0.

PR #3038, opened 2026-07-17
perf(riscv): trim imm-chip interactions and columns

yi-sun:
Stacked on #2973. Two micro-optimizations to the new immediate chips, one interaction and one column:

## 1. Bitwise imm: drop the immediate range-check lookup (20 → 19 interactions/row)

The bitwise imm core currently witnesses the immediate as (low byte, 3-bit high part) and spends a dedicated `send_range` to bound them. Instead, witness the two low immediate bytes as full bytes: both already appear as the `y` argument of the `i = 0, 1` xor lookups (multiplicity `is_valid`), so they are byte-bounded by lookups the core performs anyway, and the range-check interaction can be deleted.

**Soundness of the operand binding** `imm = c_low[0] + c_low[1]·2^8 + imm_sign·0xFF0000`: the sum is < 2^24 < p (no wrap); `imm_sign = 0` encodings lie in `[0, 2^16)` and `imm_sign = 1` encodings in `[0xFF0000, 0xFFFFFF]` — disjoint ranges, so the committed operand pins `imm_sign`, and then the two bytes are unique base-256 digits. The accepted encoding set `{0..0xFFFF} ∪ {0xFF0000..0xFFFFFF}` is a strict superset of canonical i12 but uniquely decodable; the interpreter pre-compute and the rvr lift are aligned to accept exactly this set. A new negative test constructs a two-row trace where swapping `imm_sign` between rows keeps every core constraint valid *and* every xor-lookup multiplicity balanced, so only the program-bus operand binding catches it.

## 2. less_than imm: derive the c-side sign window instead of witnessing `c_msb_f` (16 → 15 core cols)

`c_msb_f` is fully determined by committed data: `imm_sign · ((2^LIMB_BITS − 1) − slt_flag · 2^LIMB_BITS)` (the raw sign-extended limb for SLTIU, its signed value for SLTI), so the column only duplicated an equality constraint. Since the derived expression has degree 2, the per-limb difference constraint is restated as `cmp_sign · diff_val == raw_diff` (equivalent because `cmp_sign² == 1`), keeping max constraint degree at 3.

## Validation

- `openvm-riscv-circuit` full suite: 312/312 passed; rvr lift 16/16 (including new encoding-boundary tests); clippy/fmt clean.
- CPU-vs-GPU differential tracegen on an RTX PRO 6000: 58/58 passed (CUDA fillers updated column-for-column; kernel recompilation verified).
- End-to-end A/B on the reth benchmark (block 23992138, chain 1), 3+3 interleaved proves on one RTX PRO 6000: timing-neutral as expected (the touched chips are ~4.5% and ~0.3% of trace cells; deltas are single units per row), identical segment counts, per-AIR metrics confirm exactly the intended structural changes (LessThanImm width 30 → 29; BitwiseLogicImm interactions 20 → 19 via `SymbolicRapBuilder`) with identical row counts.

Interaction-count reductions feed the metered segmentation budget (`interaction_cells` term) and the logup GKR peak footprint, so they compound with workload scale even when locally timing-neutral.


shuklaayush:
i adapted 2 to the base pr but i'm not sure about 1 since it would allow non-canonical signed i12 immediates

PR #3039, opened 2026-07-18
fix(v2.1): harden rvr runtime boundaries

shuklaayush:
- trap when rvr reaches an invalid control-flow target or a host callback rejects its input
- keep rv64 register operands full-width across host I/O
- enable stricter generated C warnings and add boundary regression tests

Resolves INT-8837


crates/rvr/rvr-openvm/c/tracer/openvm_tracer_metered.h
```diff
@@ -33,6 +37,8 @@ static_assert(
 typedef struct PageTouch {
   /* Page table index plus a 64-bit leaf mask for that page. */
   uint32_t page_id;
+  /* Align leaf_mask and make the shared 16-byte Rust/C layout explicit. */
+  uint32_t padding;
```
gdmlcjs: Can we rename to "_padding"?

crates/vm/src/arch/rvr/compile.rs
```diff
@@ -393,6 +422,32 @@ fn load_execution_kind(lib: &libloading::Library) -> Result<RvrExecutionKind, Co
         .map_err(|_| CompileError::LibLoad(format!("unknown rv_execution_kind marker {marker}")))
 }
 
+fn load_num_airs(
```
gdmlcjs, line 425: Why do we need `num_airs`? It seems to only partially guard against wrong artifacts.

gdmlcjs (approved):

crates/vm/src/arch/rvr/compile.rs
shuklaayush: yes it's an extra check to ensure that the artifacts are consistent. eg. in metered execution
```rust
    require_num_airs(
        compiled,
        seg_state.ctx.trace_heights.len(),
        "trace-height storage",
    )?;
```

PR #3046, opened 2026-07-19
perf(v2.1): move input payload into zero-copy hint stream

shuklaayush:
- `hint_input` used to copy the whole input into another buffer before the guest could read it. This added an expensive `memcpy` for large Reth inputs.
- It now moves the input `Vec` into `HintStream`. The length prefix and zero padding are added as the guest reads the input.
- The interpreter and RVR use the same `HintStream`. Other hints are still returned unchanged.

[Benchmark comparison](https://github.com/openvm-org/rvr-openvm/actions/runs/29757990761)

Resolves INT-8835

### DIFF
--- a/extensions/riscv/circuit/src/hintstore/execution.rs
+++ b/extensions/riscv/circuit/src/hintstore/execution.rs
@@ -180,14 +180,15 @@ unsafe fn execute_e12_impl<CTX: ExecutionCtxTrait, const IS_HINT_STORED: bool>(
         });
     }
 
-    if exec_state.streams.hint_stream.len() < RV64_REGISTER_NUM_LIMBS * num_words as usize {
+    let num_bytes = RV64_REGISTER_NUM_LIMBS * num_words as usize;
+    if exec_state.streams.hint_stream.remaining() < num_bytes {
         let err = ExecutionError::HintOutOfBounds { pc };
         return Err(err);
     }
 
     for word_index in 0..num_words {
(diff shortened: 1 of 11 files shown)


extensions/riscv/circuit/src/hintstore/execution.rs
```diff
@@ -180,14 +180,15 @@ unsafe fn execute_e12_impl<CTX: ExecutionCtxTrait, const IS_HINT_STORED: bool>(
...
+    let num_bytes = RV64_REGISTER_NUM_LIMBS * num_words as usize;
+    if exec_state.streams.hint_stream.remaining() < num_bytes {
         let err = ExecutionError::HintOutOfBounds { pc };
         return Err(err);
     }
 
     for word_index in 0..num_words {
-        let data: [u8; RV64_REGISTER_NUM_LIMBS] =
-            std::array::from_fn(|_| exec_state.streams.hint_stream.pop_front().unwrap());
+        let mut data = [0; RV64_REGISTER_NUM_LIMBS];
```
stephenh-axiom-xyz, line 190: Does Rust have a (safe) way to allocate an array like this without initializing its values?

stephenh-axiom-xyz (approved):
LGTM

extensions/riscv/circuit/src/hintstore/execution.rs
shuklaayush: not in safe rust i think. anyways, i feel like compiler should optimize this

PR #3054, opened 2026-07-21
perf(v2.1): use `blst` for bls pairing witnesses

shuklaayush:
- BLS12-381 pairing witnesses previously used `blstrs` through an unsafe assumption that its Fp12 memory layout matched Halo2curves.
- Witness exponentiation now uses public `blst` Fp12 operations and converts field elements through canonical encodings, so it does not depend on either library's private memory layout.
- Tests compare the `blst` path with the existing implementation for zero, one, and random Fp12 inputs.

[Benchmark comparison](https://github.com/openvm-org/rvr-openvm/actions/runs/29871734886)

### DIFF
--- a/extensions/pairing/guest/src/halo2curves_shims/bls12_381/tests/test_final_exp.rs
+++ b/extensions/pairing/guest/src/halo2curves_shims/bls12_381/tests/test_final_exp.rs
@@ -1,13 +1,17 @@
-use alloc::vec::Vec;
+use alloc::{vec, vec::Vec};
 
-use halo2curves_axiom::bls12_381::{Fq, Fq2, Fr, G1Affine, G2Affine};
+use halo2curves_axiom::{
+    bls12_381::{Fq, Fq12, Fq2, Fr, G1Affine, G2Affine},
+    ff::Field,
+};
 use itertools::izip;
 use num_bigint::BigUint;
 use num_traits::Num;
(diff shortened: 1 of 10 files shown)


extensions/pairing/guest/src/halo2curves_shims/bls12_381/tests/test_final_exp.rs
```diff
@@ -29,6 +37,20 @@ fn test_bls12_381_final_exp_hint() {
     assert_eq!(f * s, c_qt);
 }
 
+#[cfg(feature = "blst")]
+#[test]
+fn test_blst_final_exp_hint_matches_basic() {
+    let mut inputs = vec![Fq12::ZERO, Fq12::ONE];
+    let mut rng = StdRng::seed_from_u64(0x41b5_1238);
+    inputs.extend((0..4).map(|_| Fq12::random(&mut rng)));
```
stephenh-axiom-xyz: Maybe more than 4 inputs

stephenh-axiom-xyz (approved):

extensions/pairing/guest/src/halo2curves_shims/bls12_381/tests/test_final_exp.rs
shuklaayush: bumped to 16

PR #3051, opened 2026-07-20
refactor: combine initial and final rows in PersistentBoundaryAir

mansur20478:
This PR updated `PersistentBoundaryAir` to combine both initial and final states. This is done by storing initial and final values. On top of it, there is new addition of is_valid and is_dirty flags. This PR focuses on updating PersistentBoundaryAir columns.

The second PR properly adds is_dirty flag (https://github.com/openvm-org/openvm/pull/3055). In this PR, for development purposes, it is set to one.

towards INT-8828

PR #3055, opened 2026-07-21
refactor: leaf level dirty bit

mansur20478:
This PR adds leaf level dirty bit tracking and modifies MemoryMerkleAir to only have final rows for dirty nodes.

The updated design tracks written blocks during preflight mode and derives dirty leaves from that knowledge. Depending on the leaf dirtiness, its parent is also considered dirty (and parent's parent too). By observing that only dirty nodes would have different final state at the end of the segment (w.r.t. the start of the segment), we only store dirty final rows for MemoryMerkleAir.

Consequently, by removing non-dirty final rows, we need to balance out the number of interactions that the removed non-dirty rows did. This is done by repurposing `direction_different` columns as `child_mode`, a column that acts as `direction_different` for final rows and interaction adjustment for initial rows.

In addition, as MerkleMemoryAir trace is no longer a fixed two rows per node, during GPU trace generation the device can't place the row by the static index. To mitigate it, we first compute the total height beforehand and actually compute were the row would go during tracegen. Both CPU and GPU tracegen are stored in interleaved fashion where first is final row and then initial row per node, except for the root where the first one is initial and second one is final. Not sure if it was originally intended to be in this format, maybe would be better to have initial and then final row? Might be later pr

This is continuation of PersistentBoundaryAir update PR (https://github.com/openvm-org/openvm/pull/3051). The next step, metered mode tracking is done in the follow up PR (https://github.com/openvm-org/openvm/pull/3061).

towards INT-8828

### DIFF
--- a/crates/vm/src/system/cuda/merkle_tree/mod.rs
+++ b/crates/vm/src/system/cuda/merkle_tree/mod.rs
@@ -456,18 +456,13 @@ impl MemoryMerkleTree {
                 // The artificial touch (see the caller) seeds the walk so the root pair
                 // exists, but no boundary row supplies the leaf's claim, so the height-1
                 // initial row (the last row) must treat the leaf as *untouched*: consume
-                // nothing. The kernel already leaves its `left_extra_ref` at 0 (the node
-                // emits no final row) and sets `right_absent_ref` for the untouched
-                // sibling; only `left_absent_ref` needs the post-hoc fix. The trace is
-                // small then.
+                // nothing.
                 let mut output_vec = output.buffer().to_host_on(&self.device_ctx).unwrap();
-                // Column index within the column-major trace, derived from the column
-                // struct so it survives layout changes.
-                let left_absent_ref_col = std::mem::offset_of!(
+                let left_adj_ref_col = std::mem::offset_of!(
                     MemoryMerkleCols<F, VM_DIGEST_WIDTH>,
-                    left_absent_ref
+                    left_adj_ref
                 ) / std::mem::size_of::<F>();
-                output_vec[unpadded_height - 1 + left_absent_ref_col * padded_height] = F::ONE;
+                output_vec[unpadded_height - 1 + left_adj_ref_col * padded_height] = F::NEG_ONE;
                 DeviceMatrix::new(
                     Arc::new(output_vec.to_device_on(&self.device_ctx).unwrap()),
                     padded_height,
--- a/crates/vm/src/system/cuda/merkle_tree/mod.rs
+++ b/crates/vm/src/system/cuda/merkle_tree/mod.rs
@@ -429,8 +429,20 @@ impl MemoryMerkleTree {
             if empty_touched_blocks {
                 // The trace is small then
                 let mut output_vec = output.buffer().to_host_on(&self.device_ctx).unwrap();
-                output_vec[unpadded_height - 1 + (width - 2) * padded_height] = F::ONE; // left_direction_different
-                output_vec[unpadded_height - 1 + (width - 1) * padded_height] = F::ONE; // right_direction_different
+                // TODO: update later
+                // Column indices within the column-major trace, derived from the column
+                // struct so they survive layout changes.
+                let col = |byte_offset: usize| byte_offset / std::mem::size_of::<F>();
+                let left_dd_col = col(std::mem::offset_of!(
+                    MemoryMerkleCols<F, VM_DIGEST_WIDTH>,
+                    left_direction_different
+                ));
+                let right_dd_col = col(std::mem::offset_of!(
+                    MemoryMerkleCols<F, VM_DIGEST_WIDTH>,
+                    right_direction_different
+                ));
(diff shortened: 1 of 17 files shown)


crates/vm/src/system/memory/merkle/columns.rs
```diff
@@ -27,6 +27,21 @@ pub struct MemoryMerkleCols<T, const DIGEST_WIDTH: usize> {
...
+    // hash, so this row consumes the child's initial claim twice (multiplicity -2).
+    // `*_absent_ref` = 1 means the child is untouched and this node has no final row to
+    // prop the reference, so this row consumes nothing (multiplicity 0).
+    //
+    // NOTE: until real dirtiness is enabled in tracegen, all four are always 0 and the
+    // child multiplicities reduce to the previous `-expand_direction`.
+    pub left_extra_ref: T,
+    pub right_extra_ref: T,
+    pub left_absent_ref: T,
+    pub right_absent_ref: T,
```
shuklaayush: i think you can repurpose the existing columns here
```rust
/// Meaning depends on `expand_direction`.
///
/// Initial row (`expand_direction = 1`):
///   0 = no initial-child reference
///   1 = one initial-child reference
///   2 = two initial-child references
///
/// Final row (`expand_direction = -1`):
///   0 = reference the final child
///   1 = reuse the initial child
///
/// Padding row (`expand_direction = 0`):
///   must be 0
pub left_child_mode: T,
pub right_child_mode: T,
```

crates/vm/src/system/cuda/merkle_tree/mod.rs
```diff
@@ -385,10 +385,13 @@ impl MemoryMerkleTree {
...
-            let padded_height = next_power_of_two_or_zero(unpadded_height);
-            let output =
-                DeviceMatrix::<F>::with_capacity_on(padded_height, width, &self.device_ctx);
-            output.buffer().fill_zero_on(&self.device_ctx).unwrap();
+            // `unpadded_height` is the 2-rows-per-node upper bound. The kernels write the
+            // actual (dirty-dependent) rows into the tail of this scratch matrix; the
+            // used region is then relocated into a right-sized output matrix.
+            let capacity_height = next_power_of_two_or_zero(unpadded_height);
+            let scratch =
+                DeviceMatrix::<F>::with_capacity_on(capacity_height, width, &self.device_ctx);
```
shuklaayush: this doesn't seem right. you should be allocating only once with the actual height

crates/vm/src/system/memory/persistent.rs
shuklaayush: can save some hashing here
```rust
let initial_hash = hasher.hash(&init_values);
let is_dirty = final_values != init_values;
let final_hash = if is_dirty {
    hasher.hash(&final_values)
} else {
    initial_hash
};
```

crates/vm/src/system/memory/merkle/tree.rs
```diff
@@ -177,46 +177,83 @@ impl<F: PrimeField32, const DIGEST_WIDTH: usize> MerkleTree<F, DIGEST_WIDTH> {
...
                                 }
                             };
-                            let combined = compress(left, right);
-                            // Record the old child pair on the compression bus.
-                            compress(old_left, old_right);
+                            // A child is dirty iff it was touched and its hash changed. Dirtiness
+                            // is defined by hash inequality everywhere (the boundary chip uses the
+                            // same definition) so the two sides of the merkle bus always agree.
+                            let left_dirty = changed_left && left != old_left;
+                            let right_dirty = changed_right && right != old_right;
```
shuklaayush: you can probably propagate the dirty bit up here
```rust
let node_dirty = left_dirty || right_dirty;
let emits_final = node_dirty || is_root;
```

shuklaayush (commented):
let's only keep the air and tracegen changes here and do the metering updates later in a follow-up

crates/vm/src/system/cuda/merkle_tree/mod.rs
mansur20478: resolved by counting leaf level dirty bits during preflight time: from that knowledge we can count the unpadded height.

crates/vm/src/system/memory/persistent.rs
mansur20478: done

crates/vm/src/system/memory/merkle/tree.rs
mansur20478: updated

crates/vm/src/system/memory/merkle/columns.rs
mansur20478: updated in https://github.com/openvm-org/openvm/pull/3055/commits/a551c3ba05de13e8335827c6a6b9227767ececd0

mansur20478:
@claude review this pr, check its correctness. especially focus on trace generation, if there are some inefficiencies that could be eliminated too. please be prudent

shuklaayush (approved):

mansur20478 (commented):
looks good

PR #3057, opened 2026-07-21
fix(ecc): harden projective affine boundaries

shuklaayush:
towards INT-8788


shuklaayush:
Superseded by #3035, which now contains the full projective-coordinate stack rebased onto the latest develop-v2.1.0.

PR #3059, opened 2026-07-22
refactor(v2.1): make riscv64 an rvr extension

shuklaayush:
- Moves RV64I, RV64M, and RISC-V I/O instruction lifting, CFG behavior, and C emission into the RISC-V RVR extension.
- Defines target-neutral RVR variables, control-flow effects, instruction emission, runtime hooks, and extension registration.
- Resolves instruction handlers during artifact generation and emits direct C calls.
- Adds bounded main-memory page accounting and buffer draining for variable-size extension accesses.
- Adds coverage for lifting, CFG analysis, generated code, metering, and RISC-V integration.

Resolves INT-7700


gdmlcjs (approved):
Separating the RISC-V enum variants and `CfgEffect` is nice!

PR #3060, opened 2026-07-22
fix(v2.1): rebuild touched pages from memory 

shuklaayush:
- Adds a state-owned way to rebuild sparse H2D transfer metadata after untracked execution.
- Reuses the optimized nonzero-byte scan in touched-page reconstruction and CPU Merkle initialization.


GunaDD (approved):
LGTM. Should we benchmark the overhead to pure/metered from this ?

shuklaayush:
> Should we benchmark the overhead to pure/metered from this ?

this doesn't touch pure or metered execution. it's just a helper to recreate this metadata

PR #3069, opened 2026-07-24
fix(v2.1): derive `prev_timestamp` instead of materializing it

shuklaayush:
## Summary

- Derives the memory-bus receive timestamp from each access timestamp and its range-checked decomposition.
- Removes one trace column and one equality constraint from every memory access, reducing `MemoryReadAuxCols` from 3 to 2 columns and `MemoryWriteAuxCols` from 7 to 6.
- Keeps the maximum interaction degree unchanged and updates the Rust and CUDA trace layouts.
- Renames `LessThanAuxCols::lower_decomp` to `decomp` to describe its use in both less-than sub-AIRs.

Resolves INT-8882

### DIFF
--- a/crates/circuits/primitives/src/assert_less_than/mod.rs
+++ b/crates/circuits/primitives/src/assert_less_than/mod.rs
@@ -47,14 +47,14 @@ impl<T> AssertLessThanIo<T> {
     }
 }
 
-/// These columns are owned by the SubAir. Typically used with `T = AB::Var`.
-/// `AUX_LEN` is the number of AUX columns
-/// we have that AUX_LEN = max_bits.div_ceil(bus.range_max_bits)
+/// Auxiliary columns owned by the SubAir. Typically used with `T = AB::Var`.
+///
+/// `AUX_LEN` must equal `max_bits.div_ceil(bus.range_max_bits)`.
 #[repr(C)]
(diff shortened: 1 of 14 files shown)


crates/circuits/primitives/src/assert_less_than/mod.rs
```diff
@@ -47,15 +47,15 @@ impl<T> AssertLessThanIo<T> {
...
+/// `AUX_LEN` must equal `max_bits.div_ceil(bus.range_max_bits)`.
 #[repr(C)]
 #[derive(AlignedBorrow, StructReflection, Clone, Copy, Debug, new)]
 pub struct LessThanAuxCols<T, const AUX_LEN: usize> {
-    // lower_decomp consists of lower decomposed into limbs of size bus.range_max_bits
-    // note: the final limb might have less than bus.range_max_bits bits
-    pub lower_decomp: [T; AUX_LEN],
+    /// The decomposition limbs range checked by the SubAir. The final limb may use fewer bits than
+    /// `bus.range_max_bits`.
+    pub decomp: [T; AUX_LEN],
```
zlangley: You could call this `diff_decomp`.

zlangley (approved):

crates/circuits/primitives/src/assert_less_than/mod.rs
shuklaayush: sure

PR #3074, opened 2026-07-27
feat(v2.1): switch to checkpointed preflight with rvr

shuklaayush:
## Summary

On `develop-v2.1.0`, execution and trace generation are coupled. Each instruction is executed by the chip that will later prove it, and that chip appends its own records as it runs. GPU trace generation then turns those records into trace rows:

```text
CPU:  opcode -> chip executor -> per-chip records
GPU:  per-chip records -> that chip's trace rows
```

This PR splits the old preflight phase into CPU preflight and GPU postflight:

```text
CPU preflight:  opcode execution -> final state + checkpoints + residuals
GPU postflight: checkpoints + residuals -> one shared execution history
GPU tracegen:   the shared history -> system, RV64, and extension traces
```

CPU preflight remains serial: it executes one instruction at a time against mutable VM memory. With the `rvr` feature, the compiled executor emits two arrays as it runs, checkpoints taken at a fixed interval and an ordered list of values called residuals, both defined below. GPU postflight re-executes the interval between each pair of checkpoints and builds one shared history that no later stage modifies. The interpreter writes that same history directly. Every trace generator reads it, after execution has finished.

## Execution modes

All three modes execute the same opcode implementations and differ only in what they retain:

```text
pure       -> final VM state
metered    -> final VM state + segment boundaries + preflight size bounds
preflight  -> final VM state + the bounded input trace generation needs
```

Metering chooses the segment boundaries before any proving happens. Preflight executes exactly one of those segments, so the data it retains is bounded by what metering already measured.

## How it works

```text
                     program + one bounded segment
                                   |
                                   v
                           serial preflight
                                   |
               +-------------------+-------------------+
               |                                       |
       compiled (`rvr`)                           interpreter
    checkpoints + residuals                  program + memory logs
               |                                       |
               v                                       |
          GPU replay                                   |
     count, allocate, emit                             |
               |                                       |
               +-------------------+-------------------+
                                   |
                                   v
                          postflight indexing
             memory links + touched values + opcode index
                                   |
                                   v
                       shared read-only history
                         program + memory logs
                                   |
               +-------------------+-------------------+
               |                   |                   |
               v                   v                   v
            system               RV64             extensions
               |                   |                   |
               +-------------------+-------------------+
                                   |
                                   v
                           trace generation
                                   |
                                   v
                   release replay and index buffers
                                   |
                                   v
                             STARK proving
```

The GPU receives a copy of memory before serial execution begins modifying it. Replay and indexing both treat that copy as the value of memory at the start of the segment, and neither writes to it.

## What is recorded

- A **checkpoint** is a full snapshot of the machine: registers, program counter, logical clock, count of instructions finished, and position in the residual list. The state at the start of the segment serves as the first snapshot.
- A **residual** is an ordered value that replay cannot derive on its own, so preflight records it. Replay reconstructs whatever follows from the program, the preceding checkpoint, and the memory image at the start of the segment; a residual supplies what remains. Examples: a value loaded from memory that execution later overwrote, a host hint, a branch outcome inside an extension replay cannot inspect, or the value an extension writes to its output.
- The **program log** has one entry per executed instruction, giving its program counter and the logical clock it started at, plus one last entry for where execution ended.
- The **memory log** has one event for every register or memory access the proof sees: logical clock, address space, block address, whether it was a read or a write, and the value. Peeks are not events. A peek is a read that execution consults but the proof never observes.
- The **memory index** links each logged access to the previous access to the same block, and records what every touched block held at the start and at the end of the segment.
- The **opcode index** groups the executed instructions by opcode, so independent trace generators can build their rows in parallel.

A memory-derived residual is what advances replay. The same access also appears as an event in the memory log, and the memory index validates that event against the start-of-segment copy of memory and against the earlier events for that block. The value is therefore consumed by one mechanism and checked by another.

The compiled path's entire output is those two append-only arrays, checkpoints and residuals. Nothing about AIRs or chips appears until postflight has built the shared history, at which point trace generators select the opcodes they own and emit rows.

## Postflight and trace generation

Postflight makes two passes over the intervals. The first counts how many events each interval produces, so the output buffers can be allocated at exactly the right size; the second writes the program and memory logs, with the intervals running in parallel. Both backends then build the memory and opcode indexes described above.

This history is where the two execution paths meet. CPU and CUDA build it with separate code, but they check the same things and hand trace generation the same information.

The system, RV64, and extension trace generators locate their instructions through the opcode index and build their rows from four inputs: the instruction as it appears in the program, the program log, the memory log, and the memory index. None of these change while trace generation reads them. Side effects such as reading a host hint occur once, during serial preflight; replay consumes the values they produced and never repeats them. One instruction can feed more than one trace, because executing an opcode is now independent of which chip owns the trace.

The buffers used for replay, sorting, and indexing are released after trace generation and before STARK proving. With them released, proving remains the phase that uses the most GPU memory.

## Reuse across proofs

The compiled metering executor, the compiled preflight executor, and the copy of the program on the GPU depend only on the guest program, not on its input. They are prepared once and reused across proofs. Compilation is reported under its own one-time preparation metric, and the per-proof timer starts after preparation.

Each proof then runs:

```text
metering -> preflight -> postflight -> trace generation -> proving
```

## Validation

The default Reth benchmark for block `24001988` completed across 65 segments:

- preflight: 0.892s;
- postflight: 1.328s;
- trace generation: 0.980s;
- backend STARK proving: 39.367s;
- peak GPU memory: 15.80 GB.

Everything before backend proving took 4.326s: uploading the starting memory, preflight, postflight, trace generation, and the surrounding bookkeeping.

Resolves INT-8800


crates/vm/src/arch/vm.rs
```diff
@@ -114,6 +146,79 @@ pub enum GenerationError {
...
+    fn prepare_postflight(
+        vm: &VirtualMachine<E, Self>,
+        program: &Program<Val<E::SC>>,
+    ) -> Result<Self::Prepared, GenerationError>;
+
+    fn generate_proving_ctx(
+        vm: &mut VirtualMachine<E, Self>,
+        prepared: &Self::Prepared,
+        output: &PreflightOutput,
+        postflight: &Postflight<'_, Val<E::SC>>,
```
stephenh-axiom-xyz: IMO a little suspect that CPU tracegen is the only one to use `postflight`. The numbers suggest that it could contribute 30% of the pre-STARK prove time, so changing this to only pass in `prepared` and `output` (or having some "empty" postflight constructor) might be worthwhile?

extensions/deferral/circuit/src/call/cuda.rs
```diff
@@ -24,52 +30,96 @@ pub struct DeferralCallChipGpu {
...
                 self.address_bits,
+                transcript.error_ptr(),
                 device_ctx.stream.as_raw(),
-            )
-            .expect("Failed to generate deferral call trace");
+            )?;
         }
-
-        AirProvingContext::simple_no_pis(trace)
+        transcript.synchronize()?;
```
stephenh-axiom-xyz: Why is this synchronize here?

crates/sdk/src/prover/app.rs
```diff
@@ -114,23 +116,24 @@ where
...
+        <VB::VmConfig as VmExecutionConfig<Val<E::SC>>>::Executor:
+            Executor<Val<E::SC>> + MeteredExecutor<Val<E::SC>> + 'static,
     {
         check_max_constraint_degrees(
             self.vm_config().as_ref(),
             self.app_vm_vk.inner.max_constraint_degree(),
         );
-        let proof = ContinuationVmProver::prove(&mut self.instance, input)?;
+        let input: Streams = input.into();
+        let _prove_span = info_span!(
```
stephenh-axiom-xyz: nit: why remove the instrument?

crates/vm/cuda/src/system/phantom.cu
```diff
@@ -1,50 +1,13 @@
...
-    size_t width,
-    DeviceBufferConstView<PhantomRecord> d_records,
-    cudaStream_t stream
-) {
-    assert(width == sizeof(PhantomCols<uint8_t>));
-    auto [grid, block] = kernel_launch_params(height);
-    phantom_tracegen<<<grid, block, 0, stream>>>(d_trace, height, width, d_records);
-    return CHECK_KERNEL();
-}
+#include "../../rvr/src/system/phantom.inc.cuh"
```
stephenh-axiom-xyz: nit: why not just put the contents of `phantom.inc.cuh` here...?

stephenh-axiom-xyz (commented):
Reviewed the core model, SDK integration, and a few system/riscv/deferral chips. Did not review any RVR code, as I have no context. Apart from what was sent above I have no blocking concerns, although I am suspicious of a few things:

- Replay tracegen structs and kernels seem pretty bloated, would guess that there is a simpler way to do them (but perhaps it wouldn't give significant perf gain)
- Feature management is pretty complex and unreadable at this scale, maybe it's better with `rust-analyzer` though
- At a glance the code seems pretty lean, but there are a lot of additional structs, traits, and functions that I don't think need to be individually defined - a lot of bloat

Ordinarily I don't think this would be "production-ready code", but given the scale of the change perhaps we should just merge for the preview and clean it up later.

PR #3084, opened 2026-07-29
perf: comm stream overlap

GunaDD:


GunaDD:
https://github.com/axiom-crypto/openvm-eth/actions/runs/30493939736

GunaDD:
Linking these benchmark: 
https://github.com/axiom-crypto/openvm-eth/actions/runs/30668088188
https://github.com/axiom-crypto/openvm-eth/actions/runs/30668072412
Need to figure out why perf decreased on block 24001988

PR #3086, opened 2026-07-30
fix: reject malformed EvmProof input instead of panicking

stephenh-axiom-xyz:
Three panics were reachable from a single attacker-supplied `EvmProof` JSON on the `cargo openvm verify evm` path, all before any cryptographic verification ran: a non-canonical commit (release `assert!` in `CommitBytes::Deserialize`), an accumulator length not a multiple of 32 (`try_into().unwrap()`), and a non-canonical Bn254 scalar (`Fr::from_bytes(..).unwrap()`). All three now return typed errors.

Host-side input validation only: no soundness impact, no vk change, Solidity verifier untouched.

## `openvm-continuations`

`commit_bytes.rs` — add `CommitBytes::try_new` and use it in `Deserialize`. `new` is kept as a panicking wrapper for the many call sites that build commits from data canonical by construction.

`Decode` is now hand-written instead of derived: the derive constructed the tuple struct directly, so the codec path could produce a non-canonical `CommitBytes` that the serde path rejects. Not one of the reported issues and not reachable from untrusted input today — separable if preferred.

## `openvm-sdk`

`types.rs` — both conversions become `TryFrom`. `EvmProof → RawEvmProof` is the attacker-reachable direction and now validates accumulator length and every scalar; the reverse is prover-output only but carried the same class of bug, including a real one where a valid `Fr` can exceed `F::MODULUS^8` and so be non-canonical as a `CommitBytes`. `EvmProofConversionError` gains three variants, drops one dead one, and has an incorrect message corrected — all four remaining variants are now constructed, where previously none were.

`solidity.rs`, `prover/halo2.rs`, `prover/evm.rs` — propagate the new `Result`. The cascade stops at `prove_evm`, which already returned `eyre::Result`.

### DIFF
--- a/crates/continuations/src/commit_bytes.rs
+++ b/crates/continuations/src/commit_bytes.rs
@@ -13,16 +13,28 @@ pub const COMMIT_NUM_BYTES: usize = 32;
 /// an unsigned integer in base `F::MODULUS`. Each commit can be converted to a Bn254 using the
 /// trivial identification as natural numbers or into a `u32` digest by decomposing the big integer
 /// base-`F::MODULUS`.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Encode, Decode)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Encode)]
 pub struct CommitBytes([u8; COMMIT_NUM_BYTES]);
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, thiserror::Error)]
+pub enum CommitBytesError {
+    #[error("non-canonical CommitBytes for BabyBear digest")]
+    NonCanonical,
+}
+
 impl CommitBytes {
+    /// Use this rather than [`CommitBytes::new`] for bytes from an untrusted source.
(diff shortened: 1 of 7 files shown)


crates/continuations/src/commit_bytes.rs
```diff
@@ -13,16 +13,28 @@ pub const COMMIT_NUM_BYTES: usize = 32;
...
+
+    /// Panics if `bytes` is non-canonical.
     pub fn new(bytes: [u8; COMMIT_NUM_BYTES]) -> Self {
-        assert!(
-            u32_digest_to_bytes(&bytes_to_u32_digest(&bytes)) == bytes,
-            "non-canonical CommitBytes for BabyBear digest"
-        );
-        Self(bytes)
+        Self::try_new(bytes).expect("non-canonical CommitBytes for BabyBear digest")
     }
```
shuklaayush, line 38: i wonder if we should implement `TryFrom` instead

shuklaayush (approved):

crates/continuations/src/commit_bytes.rs
stephenh-axiom-xyz: A lot of churn iirc

PR #3087, opened 2026-07-31
perf: remove redundant columns from `XorinVmAir`

mansur20478:
This PR removes redundant columns from XorinVmAir. The PR:

1. Updates auxiliary columns for writing `buffer`. This is possible because XORIN is in-place operation, so `preimage_buffer_bytes` already contains the `prev_data`.  Since `MemoryWriteAuxCols` and `preimage_buffer_bytes` contain the same `prev_data`, we can switch `MemoryWriteAuxCols` to `MemoryBaseAuxCols`. Saves 68 columns.
2. Removes `buffer_ptr` and `input_ptr` columns. Their values are derivable from `buffer_ptr_limbs` and `input_ptr_limbs` respectively.
3. Removes `len` and `len_limb` columns. `len_limb` is redundant because `len` already fits into a single limb. Furthermore, `is_padding_bytes` tells us -> which input blocks are for padding -> which input blocks are non-padding -> the number of non-padding blocks -> `len`.

benchmark: https://github.com/axiom-crypto/openvm-eth/actions/runs/30663251548 (Proof time is little bit slower, not sure why, need to understand)
With paragraph 1 only: https://github.com/axiom-crypto/openvm-eth/actions/runs/30665208225 (Also seems little bit slower in proving, not sure if it is system noise)

benchmark after rebase: https://github.com/axiom-crypto/openvm-eth/actions/runs/30822416379 (improvements in proof time)

resolves INT-8976

### DIFF
--- a/extensions/keccak256/circuit/cuda/include/xorin.cuh
+++ b/extensions/keccak256/circuit/cuda/include/xorin.cuh
@@ -16,9 +16,9 @@ struct XorinInstructionCols {
     T buffer_reg_ptr;
     T input_reg_ptr;
     T len_reg_ptr;
-    // Low 32 bits of [buffer_reg_ptr:8]_1 as u16 cells.
+    // Low 32 bits of the buffer register as u16 cells.
     T buffer_ptr_limbs[RV64_PTR_U16_LIMBS];
-    // Low 32 bits of [input_reg_ptr:8]_1 as u16 cells.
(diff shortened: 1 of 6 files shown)


GunaDD (approved):
LGTM! Is it possible to run the skipped CI tests as well (other than the failing lint cargo audit) ?

shuklaayush (approved):

PR #3090, opened 2026-08-03
perf(v2.1): cache field expression kernel config

shuklaayush:
Query the device-dependent occupancy cap and kernel attributes once when constructing the GPU replay chip. Derive row- and scratch-limited launch dimensions in Rust for each trace while retaining the RVR path’s existing validation.

Supersedes #2984

### DIFF
--- a/crates/vm/src/arch/cuda/postflight.rs
+++ b/crates/vm/src/arch/cuda/postflight.rs
@@ -183,6 +183,16 @@ pub enum GpuPostflightError {
     OpcodeTooLarge(usize),
     #[error("invalid GPU postflight memory configuration: {0}")]
     InvalidMemoryConfig(String),
+    #[error("invalid GPU postflight configuration: {0}")]
+    InvalidConfiguration(String),
+    #[error(
+        "GPU postflight resource limit exceeded for {resource}: requested {requested}, limit {limit}"
+    )]
+    ResourceLimitExceeded {
+        resource: &'static str,
+        requested: usize,
+        limit: usize,
+    },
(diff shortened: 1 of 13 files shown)


GunaDD (approved):

GunaDD:
claude found flagged these issues about error handling:
```
Most severe (confirmed):

1. Construction-time panic risk — the CUDA occupancy query now runs inside FieldExprReplayChipGpu::from_serialized, and every production caller (the hybrid modular and ECC Weierstrass chips) unwraps the result with .expect(). On a GPU whose architecture isn't in the compiled fatbin, prover setup now panics and kills the process — previously this surfaced as a recoverable error at trace generation, and only for programs that actually used field-expression opcodes.
2. Misleading error class — when a large expression exhausts the 128MB scratch budget (grid computes to 0 blocks), the error is now reported as InvalidTranscript ("corrupted/malicious transcript") instead of the old CUDA memory-allocation error. An operator hitting a legitimate resource limit gets pointed in the wrong direction.

Moderate:

3. Device-switch leak (plausible) — the new set_device_by_id call at construction permanently switches the thread's current CUDA device without restoring it, which could misroute subsequent CUDA calls in a multi-GPU process. Mitigating: stark-backend's GpuDeviceCtx::for_device has the same documented behavior, so this may match repo convention.
4. Missed construction-time check (plausible) — the cached kernel_config is never validated when it's produced, so a kernel exceeding the 512-byte local-memory cap builds a chip that fails on every proof attempt instead of failing once at setup.
```

PR #3099, opened 2026-08-04
feat: fv verification mode in cargo openvm

stephenh-axiom-xyz:
Resolves INT-8984, INT-8983


crates/cli/src/commands/verify.rs
```diff
@@ -82,6 +82,14 @@ enum VerifySubCommand {
         )]
         proof: Option<PathBuf>,
 
+        #[arg(
+            long,
+            action,
+            help = "Additionally verify the STARK proof with the formally verified Swirl verifier (requires a build with the 'fv-verifier' feature). Only proofs generated with the default riscv32 VM and default parameters are in scope",
+            help_heading = "OpenVM Options"
+        )]
+        fv_verified: bool,
```
shuklaayush, line 91: should this be under `#[cfg(feature = "fv-verifier")]`? also pedantic but fv-verified sounds redundant

crates/cli/src/commands/verify.rs
```diff
@@ -228,7 +237,21 @@ impl VerifyCmd {
...
+                    {
+                        println!("Verifying STARK proof with the canonical RISC-V FV verifier");
+                        Sdk::verify_proof_with_fv_verifier(&expected_baseline, &vm_stark_proof)?;
+                        println!("FV verifier accepted the proof");
+                    }
+                    #[cfg(not(feature = "fv-verifier"))]
+                    eyre::bail!(
+                        "--fv-verified requires cargo-openvm built with the 'fv-verifier' \
+                         feature (cargo install --features fv-verifier ...)"
+                    );
```
shuklaayush, line 253: maybe just remove the option if not compiled with the feature

crates/fv-verifier/src/magic.rs
shuklaayush, line 1: not sure why we need these

stephenh-axiom-xyz: This is copied, perhaps @bdiehs can answer

crates/cli/src/commands/verify.rs
stephenh-axiom-xyz: Considered this, but ultimately I think it's better to keep it so that users are aware of the option existing. Would cause problems if users expected fv-verified to work but it didn't

crates/cli/src/commands/verify.rs
stephenh-axiom-xyz: Same response as above, closing this out and we can discuss there

crates/cli/src/commands/verify.rs
shuklaayush: okay but it is inconsistent with how we treat evm verification

stephenh-axiom-xyz: Not in `setup`?

crates/fv-verifier/src/magic.rs
bdiehs: Claude's answer:
It adds an 8-byte header to each blob: 4 ASCII bytes identifying which blob it is (SVKY/PROF/PUBV) plus a wire format version. The three cross a process boundary as untyped bytes, so nothing else can tell them apart — with the header, a vk passed where a proof is expected, or a blob from an older wire format, 
fails immediately instead of misparsing into a valid-looking but wrong value.
So this is not necessary for security purposes but just in case our wire format changes.

shuklaayush: i don't know the details but this sounds like the backwards compatibility slop that these models add at times. i'll leave you guys to be the judge of it

bdiehs (approved):

PR #3100, opened 2026-08-04
fix(v2.1): allow noncanonical is-eq setup reads 

shuklaayush:
## Summary

SETUP_ISEQ does not require its second read to be below the modulus, but CPU and GPU trace generation applied the normal IS_EQ validation and rejected setup rows when c >= modulus.

This change handles noncanonical setup reads while preserving strict operand validation for normal IS_EQ. It also performs the setup difference in the proof field to avoid u16 underflow and adds CPU/GPU regression coverage.

## Testing

- focused modular is-equal CPU tests
- focused modular is-equal CUDA test


luffykai:
so this is technically correct, but unnecessary, because transpiler already set rs2 to 0 : https://github.com/openvm-org/openvm/blob/develop-v2.1.0/extensions/algebra/transpiler/src/lib.rs#L104. 
So do we want to merge this? I feel the code without this pr is slightly simpler, so leaning no

PR #3102, opened 2026-08-04
feat(v2.1): accept cfg construction hints

shuklaayush:
## Summary

- Emit LLVM's post-optimization basic-block map and successor topology in managed guest builds.
- Retain function entries and exact machine-CFG successors of computed branches; direct control flow remains reconstructed from assembly, while returns and tail calls use existing callsite/function-entry handling.
- Serialize only deduplicated, decoded block-start PCs in `VmExe`, then consume them as additive validated hints during RVR CFG construction.
- Keep transpiler output positional with ELF instruction slots, using output length as the single source of truth for consumed input.
- Add concise CFG and native-build tracing for instruction, block, hint, split, and execution-kind visibility.

For the Fibonacci fixture, LLVM successor topology adds 4,416 bytes to the input ELF (`61,744` to `66,160`, 7.15%) without changing `.text`. That section is discarded after extraction; the serialized `VmExe` hint contribution grows by 26 bytes (`95` to `121` bytes, from 40 function-entry hints to 53 exact hints).

## Testing

Run in an isolated Linux GPU-host worktree:

- `cargo nextest run --cargo-profile=fast -p openvm-transpiler` (10 passed)
- strict Clippy for the affected build, transpiler, and CLI test targets
- the previously failing `test_multi_target_transpile_default` integration test (passed)
- real `openvm-1.94.1` guest build with LLVM 21.1.8; `llvm-readobj-21` confirmed feature bit 2 and successor records
- exact same-program ELF and serialized-hint size comparison

The full PR suite completed on `bac02fa2f8` with 69 passed, 1 skipped, and no failures or pending checks.

A [branch-named 20-run Reth comparison](https://github.com/openvm-org/rvr-openvm/actions/runs/31086198329) resolved `feat/support-taking-hints-for-cfg-construction` to semantic head `204d8e596f`. Base and target used the identical target-built guest ELF. Exact filtering produced 2,442 valid hints but only one additional block split. Metered execution changed from 658.640 ms to 669.169 ms (+1.56%, interval +0.53% to +2.50%); host instructions, guest instructions, and segment count were unchanged.

Resolves INT-9012


crates/toolchain/transpiler/src/extension.rs
```diff
@@ -3,15 +3,13 @@ use openvm_instructions::{exe::SparseMemoryImage, instruction::Instruction};
 
 /// Trait to add custom RISC-V instruction transpilation to OpenVM instruction format.
 /// RISC-V instructions always come in 32-bit chunks.
-/// An important feature is that multiple 32-bit RISC-V instructions can be transpiled into a single
-/// OpenVM instruction. See [process_custom](Self::process_custom) for details.
```
shuklaayush, line 7: i don't think this is used anywhere and seems like a footgun

gdmlcjs (commented):
Do the additional block headers not have a negative impact on execution performance?

shuklaayush:
> Do the additional block headers not have a negative impact on execution performance?

i ran it [here](https://github.com/openvm-org/rvr-openvm/actions/runs/31086198329/job/92567516734#step:12:73) and it only results in one extra basic block so not really

gdmlcjs:
> > Do the additional block headers not have a negative impact on execution performance?
> 
> i ran it [here](https://github.com/openvm-org/rvr-openvm/actions/runs/31086198329/job/92567516734#step:12:73) and it only results in one extra basic block so not really

okay, great!

gdmlcjs (approved):

PR #3103, opened 2026-08-04
feat(ecc): add EC_MUL and SETUP_EC_MUL opcodes

mansur20478:
This PR adds ecmul opcode:
```
EC_MUL rd, rs1, rs2          // [rd] ← scalar_at(rs2) · point_at(rs1)
SETUP_EC_MUL rd, rs1, rs2    // rs1 → (modulus, a); run once per curve before any EC_MUL
```

The scalar must be odd, nonzero, and below the subgroup order. 

The guest-side `mul_scalar` wrappers handle even scalar case scenarios. `SETUP_EC_MUL` follows the same pattern as the other Weierstrass setups: its point operand carries `(modulus, a)`, which the chip pins against its configuration.

The reason for choosing odd scalars is so that the scalar can be represented as `2B + 1`. The base point must be a non-identity point on the curve, in the prime-order subgroup, and the subgroup order must satisfy `n ≡ 1 (mod 4)`. 

---

EC_MUL follows this formula where `σᵢ = 2bᵢ − 1`:
```
R = P
for i in (0..256).rev():
    R = 2R + σᵢ·P,      σᵢ ∈ {+1, −1}
// becomes R = 2²⁵⁶·P + ( Σᵢ σᵢ·2ⁱ )·P at the end
```

The formula above works because:
```
2²⁵⁶ + Σᵢ σᵢ·2ⁱ  = 2²⁵⁶ + Σᵢ (2bᵢ − 1)·2ⁱ =  2²⁵⁶ + 2·Σᵢ bᵢ·2ⁱ − Σᵢ 2ⁱ  =  2²⁵⁶ + 2B − (2²⁵⁶ − 1) = 2B + 1
```
As result, one instruction spans **128 trace rows** (we pack two steps per row and fuse the IO row
with the last row). 

To bind the digits to the actual operand, the header threads a bit accumulator for `B` across the
rows. The final row reconstructs `2B + 1` and checks it byte-by-byte against the scalar read from
memory.

The row layout is header / field expression / IO. The point accumulator is threaded between rows
by transition constraint rather than through memory, and all memory traffic sits on the final row:
one point read, one scalar read, one point write per instruction. 

On secp256k1 the chip is 1711 columns wide, 128 rows, 219k cells per scalar multiplication.

On the guest side, `msm_via_ec_mul` computes MSM as one `EC_MUL` per base, and
`IntrinsicCurve::msm` for all four curves routes through it. An ECDSA verify/recover is now two
`EC_MUL`s and one add. The generic Pippenger `msm` stays for G2 and external callers or custom defined.

---

Resolves INT-8898, INT-9082


extensions/algebra/rvr/ffi/modular/c/rvr_ext_bls12_381.c
```diff
@@ -356,3 +358,30 @@ __attribute__((preserve_most)) void rvr_ext_ec_double_bls12_381(
...
+) {
+    blst_p1_affine base;
+    base.x = fp_read(state, rs1_ptr);
+    base.y = fp_read(state, rs1_ptr + BLS12_381_FP_BYTES);
+
+    /* EC_MUL requires an odd scalar: the chip's digits are all +-1, whose sum is never even. */
+    uint64_t words[BLS12_381_SCALAR_WORDS];
+    read_mem_u64_range(state, rs2_ptr, words, BLS12_381_SCALAR_WORDS);
+    const byte *scalar = (const byte *)words;
+    debug_assume(scalar[0] & 1);
```
shuklaayush: can you rename this to `assert_assume` since `debug` is misleading
```c
static __attribute__((always_inline)) inline void assert_assume(int x) {
    assert(x);           // checked in debug builds
    __builtin_assume(x); // optimization hint in all builds
}
```

mansur20478: sure, should we rename `debug_assume` to `assert_assume`, since it is doing assert and assume

```

static __attribute__((always_inline)) inline void debug_assume(int x) {
  assert(x);
  assume(x);
}
```

but now it will be 

```
static __attribute__((always_inline)) inline void assert_assume(int x) {
  assert(x);
  assume(x);
}
```

shuklaayush: yes exactly

extensions/algebra/rvr/ffi/modular/c/rvr_ext_secp256k1.c
```diff
@@ -285,3 +289,37 @@ __attribute__((preserve_most)) void rvr_ext_ec_double_k256(RvState* state,
   fe_write(state, rd_ptr, &x3);
   fe_write(state, rd_ptr + SECP256K1_ELEM_BYTES, &y3);
 }
+
+__attribute__((preserve_most)) void rvr_ext_ec_mul_k256(RvState* state,
+                                                        uint64_t rd_ptr,
+                                                        uint64_t rs1_ptr,
+                                                        uint64_t rs2_ptr) {
```
shuklaayush, line 303: can you check if having separate paths for `G` and `-G` improve perf? something like
```c
  if (base == G) {
      secp256k1_ecmult(&product, NULL, &secp256k1_scalar_zero, &q);
  } else if (base == -G) {
      secp256k1_scalar nq;
      secp256k1_scalar_negate(&nq, &q);
      secp256k1_ecmult(&product, NULL, &secp256k1_scalar_zero, &nq);
  } else {
      secp256k1_ecmult(&product, &base_jacobian, &q, NULL);
  }
```

mansur20478: sure

mansur20478: sure

shuklaayush: Implemented in 5bf0ab35b1. Exact G uses libsecp256k1’s precomputed generator term; exact -G negates the scalar and uses the same generator table; every other base retains the variable-base path. The regression now covers G, -G, and a generic point across odd scalars, plus the total zero/even/unreduced wrappers. The K256 RVR ECDSA-recovery equivalence/proof test and the ecrecover benchmark also pass. CUDA trace generation remains curve-name-agnostic because it must emit the exact ladder witnesses.

.github/actions/rust-cache-cuda/action.yml
```diff
@@ -31,8 +31,23 @@ runs:
...
+          git ls-files -z -- \
+            .github/actions/rust-cache-cuda/action.yml \
+            ':(glob)**/build.rs' \
+            ':(glob)**/cuda/**' |
+            sort -z |
+            xargs -0 sha256sum
+        } | sha256sum | cut -d ' ' -f 1)"
+        echo "hash=$source_hash" >> "$GITHUB_OUTPUT"
+        echo "SCCACHE_C_CUSTOM_CACHE_BUSTER=$source_hash" >> "$GITHUB_ENV"
+
```
GunaDD: what is this for ?

crates/rvr/rvr-openvm/c/openvm_util.h
```diff
@@ -15,7 +15,7 @@ static __attribute__((always_inline)) inline void assume(int x) {
   __builtin_assume(x);
 }
 
-static __attribute__((always_inline)) inline void debug_assume(int x) {
+static __attribute__((always_inline)) inline void assert_assume(int x) {
```
GunaDD, line 18: why did it change from `debug_assume` to `assert_assume` ?

extensions/ecc/sw-macros/src/lib.rs
```diff
@@ -110,38 +110,20 @@ pub fn sw_declare(input: TokenStream) -> TokenStream {
...
-                /// x-coordinates differ. For an identity base the accumulator seeds as `(0, 0)`
-                /// and the first doubling divides by `2·y = 0`, which has no witness: such a call
-                /// cannot be proven.
-                ///
-                /// The curve guest libraries discharge all of these in their `mul_scalar`
-                /// wrappers, which reduce the scalar and restore oddness via `(n - k)·P = -(k·P)`.
-                ///
-                /// Multi-base `IntrinsicCurve::msm` does not use this: the windowed method shares
-                /// doublings across bases, which a single-base intrinsic cannot. Single-pair calls
-                /// route here through `mul_scalar`.
```
GunaDD: I think we shouldn't remove these comments since it was already there before ?

crates/rvr/rvr-openvm/c/openvm_util.h
shuklaayush: because `debug` was misleading. `assume` runs also on release builds, while `assert` is usually only debug

extensions/ecc/sw-macros/src/lib.rs
mansur20478: isnt this one the one I put? I kind of removed it because it over-explains, i can put it back

.github/actions/rust-cache-cuda/action.yml
mansur20478: It is to make sure that updated CUDA files gets recompiled

GunaDD: was this always needed ?

mansur20478: At the beginning of implementing this PR, at one point, GPU tests kept failing due to stale sscache hits, not sure if it is needed now i guess

extensions/algebra/rvr/ffi/modular/c/rvr_ext_bls12_381.c
```diff
@@ -356,3 +356,30 @@ __attribute__((preserve_most)) void rvr_ext_ec_double_bls12_381(
...
+) {
+    blst_p1_affine base;
+    base.x = fp_read(state, rs1_ptr);
+    base.y = fp_read(state, rs1_ptr + BLS12_381_FP_BYTES);
+
+    /* EC_MUL requires an odd scalar: the chip's digits are all +-1, whose sum is never even. */
+    uint64_t words[BLS12_381_FR_WORDS];
+    read_mem_u64_range(state, rs2_ptr, words, BLS12_381_FR_WORDS);
+    const byte *scalar = (const byte *)words;
+    assert_assume(scalar[0] & 1);
```
GunaDD: is it possible that this assertion fails ?

mansur20478: It shouldn't fail. On the guest side, when scalar K is even, it is negated (n - K) to become odd. It works because n is odd. I guess if n is even, it will fail, so there is a safeguard in `weierstrass_chip/mul/mod.rs` -- `assert_supported_scalar_order` function that prevents such kind of scalar_order (In there, it enforces stronger `n % 4 = 1` assumption)

guest-libs/pairing/src/bn254/mod.rs
```diff
@@ -63,6 +63,39 @@ impl CyclicGroup for G1Affine {
...
+        }
+        let bytes: [u8; 32] = reduced.as_le_bytes().try_into().unwrap();
+        // SAFETY: `self` is not the identity, and `reduced` is odd and below the group order.
+        let product = unsafe { self.mul_scalar_le_unchecked::<true>(&bytes) };
+        if odd {
+            product
+        } else {
+            -product
+        }
+    }
```
GunaDD, line 91: I think there is a bug with this `mul_scalar` code causing `Bls12_381::msm` to not work correctly. A poc is in https://github.com/openvm-org/openvm/tree/poc/bls12-381-offsubgroup-mul branch. Run `cargo neatest run --cargo-profile=fast --features bls12_381 test_bls_offsubgroup_mul` to confirm.

.github/actions/rust-cache-cuda/action.yml
GunaDD: seems like its unnecessary, i removed it

shuklaayush (approved):

shuklaayush:
unfortunately this causes a blowup in recursion time so i'm going to close it for now

Issue #2754, opened 2026-04-25
Missing Poseidon2 extension and guest-lib


trevlt:
Why is there no `poseidon2` extension just like `keccak256` and `sha256`? And accompanying `guest-libs`?

yi-sun:
Under our modular philosophy, we'd like to keep the set of extensions in this repo minimal, so it would be best for this extension to live in a separate repository, where it can be used without forking or modifying.

PR #3106, opened 2026-08-05
feat: add poseidon2 extension

amintalebi:
## Problem

Guest programs cannot use the Poseidon2 hash function.

The repository already contains Poseidon2 code. The VM uses this code internally for memory Merkle trees and for proof aggregation. Guest programs cannot call this code.

A guest program that needs Poseidon2 must build the hash from RISC-V arithmetic instructions. This method needs thousands of instructions. The prover must prove each instruction. Therefore the proof cost is high.

The `keccak256` and `sha2` hash functions each have an extension and a guest library. Poseidon2 does not have them. Issue #2754 reports this difference.

## Solution

This PR adds a Poseidon2 extension. Guest programs can now do one Poseidon2 permutation with one instruction.

The PR adds four crates:

| Crate | Function |
| --- | --- |
| `openvm-poseidon2-guest` | Holds the instruction constants. Supplies the intrinsic function. |
| `openvm-poseidon2-transpiler` | Changes the RISC-V instruction into an OpenVM instruction. |
| `openvm-poseidon2-circuit` | Proves that the VM executed the instruction correctly. |
| `openvm-poseidon2` | Supplies the hash API for guest programs. |

**The instruction.** The new instruction is `PERMUTE`. It permutes a state of 16 words in memory. The state is 64 bytes long. The instruction changes the state in place.

The PR uses these opcode values:

- RISC-V custom-0: opcode `0x0b`, funct3 `0b111`, funct7 `0`.
- OpenVM local opcode: offset `0x330`.

No other extension uses these values.

**The circuit.** The circuit has two chips and one sub-AIR:

- The permute chip reads the state from memory. It writes the result back to memory. It makes one row for each instruction.
- The periphery chip proves that the output is the permutation of the input. It makes one row for each different permutation. Equal permutations share one row.
- The canonicity sub-AIR proves that each word that the chip writes to memory is less than the field prime.

The two chips communicate through a lookup bus.

**The guest library.** The library supplies these items:

- `permute` — the raw permutation.
- `Poseidon2` — a sponge hasher with `new`, `update` and `finalize`.
- `hash_u32s` — a hash of field elements.
- `hash_bytes` — a hash of arbitrary bytes.

**The configuration.** The extension is optional. To use it, add this section to `openvm.toml`:

```toml
[app_vm_config.poseidon2]
```

`SdkVmConfig::standard()` does not include the extension. Section "Dev Notes" gives the reason.

## Dev Notes

**The extension is not in `standard()`.** The set of extensions controls the verifying key. Therefore the set must be stable. An earlier version of this PR enabled the extension only when the `cuda` feature was off. Cargo unifies features across the build graph. Therefore one unrelated CUDA crate can remove the extension for all users. The same source code then gives two different verifying keys. A code review found this fault.

The extension also has no CUDA prover. If a user enables the extension and uses the GPU backend, `SdkVmGpuBuilder` returns `ChipInventoryError::CudaUnsupported`.

**How to test.**

```bash
cargo nextest run --cargo-profile=fast -p openvm-poseidon2-circuit   # 5 tests
cargo nextest run --cargo-profile=fast -p openvm-poseidon2           # 3 tests

# The proving test is #[ignore], as for keccak256. Run it directly:
cargo nextest run --cargo-profile=fast --profile=heavy -p openvm-poseidon2 \
  --run-ignored=all -E 'test(test_poseidon2_prove)'
```

## Out of Scope

- **A CUDA prover extension.** `Poseidon2GpuProverExt` does not exist. GPU users get a clear error. This work must come before the extension can go into `standard()`.
- **Specification and book documentation.** The pages `riscv-custom-code.mdx`, `isa.mdx`, `transpiler.mdx` and `instruction-reference.mdx` still need entries. The book needs new pages for the extension and for the guest library.
- **A benchmark.** This PR gives no measurement of the improvement against a software implementation.


shuklaayush:
it's better for the extension to live in a separate repository to reduce maintenance and security burden (see https://github.com/openvm-org/openvm/issues/2754#issuecomment-5343510242). thanks for the contribution though!

PR #3107, opened 2026-08-05
feat(v2.1): separate public value reveal opcode 

shuklaayush:
Resolves INT-8777

Separates public-value reveal from the RV64 load/store opcode family. Reveal now has its own execution, AIR, trace-generation, and CUDA paths, while load/store is limited to register and memory interactions.

This keeps public-value handling isolated from RAM semantics and allows removal of the public-values address space to remain a separate follow-up change.


mansur20478:
Seems good. It seems that reveal is aligned memory access only, should it also support misaligned memory access (i might be wrong, but i think it supported it by relying on `STORED` instruction supporting misaligned memory accesses)?

shuklaayush:
no, we only need to support misaligned accesses for riscv load/stores which are not controlled by us

extensions/riscv/circuit/cuda/rvr/include/riscv/reveal_replay.cuh
```diff
@@ -0,0 +1,169 @@
+#pragma once
+
+#include "arch/rvr/replay.cuh"
+#include "primitives/constants.h"
+
+using namespace program;
+using namespace riscv;
+
+static constexpr uint32_t REVEAL_REPLAY_ERROR = 271;
```
mansur20478, line 9: in some files, we kind of hardcode error codes like in `store_byte_replay.cuh` with its `preflight_set_error(error, 254);`. Might be good idea in the future to use more of `REVEAL_REPLAY_ERROR` kind of constants to avoid accidentally colliding error codes.

docs/vocs/docs/pages/specs/openvm/isa.mdx
```diff
@@ -457,7 +455,7 @@ with user input-output.
...
 | HINT_STORED_RV64 | `_,b,_,1,2`     | `[r64{0}(b):8]_2 = next 8 bytes from hint stream`. Only valid if next 8 values in hint stream are bytes and `r64{0}(b)` is aligned to an 8-byte boundary.                                                                          |
 | HINT_BUFFER_RV64 | `a,b,_,1,2`     | `[r64{0}(b):8 * l]_2 = next 8 * l bytes from hint stream` where `l = r64{0}(a)`. Only valid if next `8 * l` values in hint stream are bytes, `l` is non-zero and `<=` `MAX_HINT_BUFFER_DWORDS` (1,023 dwords ≈ 8KB), and `r64{0}(b)` is aligned to an 8-byte boundary. |
-| REVEAL_RV64      | `a,b,c,1,3,1,g` | Pseudo-instruction for `STORED_RV64 a,b,c,1,3,1,g` writing to the user IO address space `3`.                                                                                      |
+| REVEAL_RV64      | `a,b,c,1,3,1,g` | Dedicated instruction that writes 8 bytes from register `a` to the 8-byte-aligned address `b + sign_extend(c)` in the user IO address space `3`.                                |
```
mansur20478: Isn't effective address for REVEAL is the content of register `b` + sign_extended(c)? Or is it b itself?

mansur20478 (approved):
Looks good, no obvious issues

extensions/riscv/circuit/cuda/rvr/include/riscv/reveal_replay.cuh
shuklaayush: yes, that's a good point. let's do it in a follow up

docs/vocs/docs/pages/specs/openvm/isa.mdx
shuklaayush: good catch, it should be the content

crates/toolchain/instructions/src/riscv.rs
```diff
@@ -11,3 +11,10 @@ pub const REGISTER_AS: u32 = 1;
 pub const MEMORY_AS: u32 = 2;
 
 pub const NUM_REGISTERS: usize = 32;
+
+/// Whether an OpenVM operand encodes an aligned RV64 register pointer.
+#[inline]
+pub fn is_valid_register_pointer(pointer: u32) -> bool {
+    let pointer = u64::from(pointer);
+    pointer < NUM_REGISTERS as u64 * REGISTER_BYTES && pointer.is_multiple_of(REGISTER_BYTES)
+}
```
GunaDD, line 20: might be better placed in some `mod.rs` somewhere

GunaDD (approved):
LGTM

crates/toolchain/instructions/src/riscv.rs
shuklaayush: it's fine for now. we might want to get rid of this whole function later and make register pointers correct by construction

PR #3104, opened 2026-08-04
feat: 2 pow 32 bytes guest memory address

GunaDD:
Adds support for 2^32 bytes memory address. Supersedes #2850.

Benchmark: https://github.com/axiom-crypto/openvm-eth/actions/runs/31222063368

New design: https://github.com/axiom-crypto/openvm-eth/actions/runs/32282531103/job/96164388705

Bench after MEM_BITS change: https://github.com/axiom-crypto/openvm-eth/actions/runs/32427802054/job/96613241136

After fixing some codex audit findings: https://github.com/axiom-crypto/openvm-eth/actions/runs/32431879330

Closes INT-8080

PR #3108, opened 2026-08-05
feat: support program counter up to the full u32

GunaDD:
Stacked on #3104 (base `feat/2-pow-32-memory-addresses-v2.1.0`); only the top two commits are new here.

## Summary

Byte program counters now span the full 32-bit address space, matching the 2^32-byte memory port. A 32-bit byte pc does not fit in a BabyBear element, so every circuit-visible pc — trace columns, execution/program bus messages, connector public values, and the exe-commit preimage — instead carries the **pc index** `pc / DEFAULT_PC_STEP < 2^PC_BITS` (`PC_BITS = 30`). Runtime pcs stay byte addresses; conversion happens at the circuit boundary.

## Circuit representation (`crates/vm` + toolchain)

- `openvm-instructions` gains `pc_to_idx` / `idx_to_pc` and `PC_STEP_BITS`; `MAX_ALLOWED_PC` becomes the last aligned 32-bit byte address (`u32::MAX - 3`).
- Program cached trace, connector begin/end pcs, phantom rows, and `compute_exe_commit(pc_start)` all convert to indices. Segment stitching, the root-verifier AIR, and verify-stark pass the public-value felt through unchanged.
- The `ExecutionBridge` default pc increment becomes `+1` index; all `(DEFAULT_PC_STEP, to_pc)` adapter tuples become `(1, to_pc)`.
- AIRs that consume byte immediates (branches, JAL) scale them by `inverse(DEFAULT_PC_STEP)`.
- Also fixes a latent metered pre-compute bug that double-added `pc_base`.

## Chips that materialize the byte pc

The old `[u16, u14]` limb split cannot soundly represent 32-bit values mod p (BabyBear is a 31-bit field), so chips that write the byte pc into a register decompose it differently:

- **JAL/JALR** return address `rd = 4 * (from_idx + 1)`: the low u16 limb is constrained as `4x` with `x < 2^PC_IDX_LOW_BITS` and the high limb as a full u16, which uniquely pins the 32-bit decomposition mod p. JAL `rd` is zero-extended per the bus address convention, while LUI keeps rv64 sign-extension.
- **JALR** `to_pc_limbs` decompose the target *index*; a misaligned target (bit 1 set) is unsatisfiable in the AIR and rejected in `try_run_jalr`.
- **AUIPC** decomposes the byte pc as `[4 * idx_low, u16 high]`; results ≥ 2^32 are rejected at tracegen.
- **Branches** compute `to_pc = from_idx + byte_delta / 4`; fillers use signed integer target math (field addition wraps mod p past 2^30) with bounds validation, shared via `taken_branch_pc` / `checked_branch_target`.

## Misaligned branch/jump offsets

The transpiler deliberately does **not** reject misaligned branch/jump offsets (second commit). It decodes every word of `.text`, so embedded data and never-taken branches can legitimately decode as misaligned control flow. Misalignment is instead enforced where the target is actually used: the interpreter traps on a misaligned pc, and tracegen rejects the row via `checked_branch_target` (branches) and the JAL target bounds check. A misaligned branch that is never taken falls through normally, matching hardware.

## CUDA parity

`constants.h` gains pc helpers; program/phantom/adapter kernels write indices; the `jal_lui` / `jalr` / `auipc` kernels mirror the new limb splits and range-checker counts; rvr replay kernels validate jump targets with integer math (`replay_taken_branch_pc` / `replay_branch_target_in_bounds`); and `checkpoint_replay`'s `branch_target` drops its mod-p reduction.

## Tests

Chip tests sample aligned byte pcs over the full 32-bit range, with new max-pc boundary tests for JAL/JALR/AUIPC. Testers record index-space bus messages while exposing byte pcs as `u32` (roundtripping through F would reduce mod p). The misaligned-offset fix restores `rvr_embedded_text_data`, `rvr_invalid_branch_fallthrough`, and `rvr_invalid_branch_taken`.

Closes INT-8507

### DIFF
--- a/crates/toolchain/transpiler/src/util.rs
+++ b/crates/toolchain/transpiler/src/util.rs
@@ -4,7 +4,6 @@ use openvm_decoder::instruction_formats::{BType, IType, ITypeShamt, JType, RType
 use openvm_instructions::{
     exe::SparseMemoryImage,
     instruction::{Instruction, InstructionOperand},
-    program::DEFAULT_PC_STEP,
     riscv::{MEMORY_AS, REGISTER_NUM_LIMBS},
     LocalOpcode, SystemOpcode, VmOpcode,
 };
@@ -104,15 +103,15 @@ pub fn from_s_type(opcode: usize, dec_insn: &SType) -> Instruction {
 
 /// Create a new [`Instruction`] from a B-type instruction.
 ///
-/// The branch offset must be `DEFAULT_PC_STEP`-aligned: the circuit represents pc values in
-/// units of `DEFAULT_PC_STEP`, so a misaligned offset has no sound encoding. Without the C
-/// extension, RISC-V branch targets are always 4-byte aligned.
+/// The immediate stays a byte offset; the AIR scales it by `inverse(DEFAULT_PC_STEP)` because
+/// circuit-visible pc values are pc indices. A misaligned offset therefore has no sound
+/// encoding, but it cannot be rejected here: the transpiler decodes every word of `.text`,
+/// including embedded data and never-taken branches that merely *look* like misaligned
+/// branches. Misalignment is enforced where the target is actually used — the interpreter
+/// traps on the misaligned pc and tracegen rejects the row via
+/// `checked_branch_target` — so a misaligned branch that is never taken falls through
+/// normally, exactly as it does on hardware.
 pub fn from_b_type(opcode: usize, dec_insn: &BType) -> Instruction {
-    assert_eq!(
-        dec_insn.imm % DEFAULT_PC_STEP as i32,
-        0,
-        "branch offset must be a multiple of DEFAULT_PC_STEP"
-    );
     Instruction::new(
         VmOpcode::from_usize(opcode),
         InstructionOperand::from_usize(REGISTER_NUM_LIMBS * dec_insn.rs1),
@@ -127,13 +126,10 @@ pub fn from_b_type(opcode: usize, dec_insn: &BType) -> Instruction {
 
 /// Create a new [`Instruction`] from a J-type instruction.
 ///
-/// The jump offset must be `DEFAULT_PC_STEP`-aligned; see [`from_b_type`].
+/// The jump offset is a byte offset and is not required to be `DEFAULT_PC_STEP`-aligned here;
+/// see [`from_b_type`]. A misaligned JAL that is actually executed is rejected by the
+/// interpreter and by JAL tracegen ("JAL target outside implemented PC address space").
 pub fn from_j_type(opcode: usize, dec_insn: &JType) -> Instruction {
-    assert_eq!(
-        dec_insn.imm % DEFAULT_PC_STEP as i32,
-        0,
-        "jump offset must be a multiple of DEFAULT_PC_STEP"
-    );
     Instruction::new(
         VmOpcode::from_usize(opcode),
         InstructionOperand::from_usize(REGISTER_NUM_LIMBS * dec_insn.rd),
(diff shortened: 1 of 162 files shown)


crates/toolchain/transpiler/src/util.rs
```diff
@@ -102,6 +102,15 @@ pub fn from_s_type(opcode: usize, dec_insn: &SType) -> Instruction {
...
 /// Create a new [`Instruction`] from a B-type instruction.
+///
+/// The immediate stays a byte offset; the AIR scales it by `inverse(DEFAULT_PC_STEP)` because
+/// circuit-visible pc values are pc indices. A misaligned offset therefore has no sound
+/// encoding, but it cannot be rejected here: the transpiler decodes every word of `.text`,
+/// including embedded data and never-taken branches that merely *look* like misaligned
+/// branches. Misalignment is enforced where the target is actually used — the interpreter
+/// traps on the misaligned pc and tracegen rejects the row via
+/// `checked_branch_target` — so a misaligned branch that is never taken falls through
+/// normally, exactly as it does on hardware.
```
shuklaayush: i don't think these are required

crates/toolchain/transpiler/src/util.rs
```diff
@@ -116,6 +125,10 @@ pub fn from_b_type(opcode: usize, dec_insn: &BType) -> Instruction {
 }
 
 /// Create a new [`Instruction`] from a J-type instruction.
+///
+/// The jump offset is a byte offset and is not required to be `DEFAULT_PC_STEP`-aligned here;
+/// see [`from_b_type`]. A misaligned JAL that is actually executed is rejected by the
+/// interpreter and by JAL tracegen ("JAL target outside implemented PC address space").
```
shuklaayush: same

crates/vm/src/arch/interpreter/mod.rs
```diff
@@ -574,7 +574,9 @@ where
             let buf: &mut [u8] = unsafe { &mut *(*buf as *mut [u8]) };
             let pre_inst = if let Some((inst, _)) = inst_opt {
                 tracing::trace!("get_metered_pre_compute_instruction {inst:?}");
-                let pc = program.pc_base + i as u32 * DEFAULT_PC_STEP;
+                // `i` already includes the `get_pc_index(pc_base)` padding, so it is the
+                // absolute pc slot (matching `get_pre_compute_instructions`).
+                let pc = i as u32 * DEFAULT_PC_STEP;
```
shuklaayush: why this change?

crates/vm/src/system/connector/mod.rs
```diff
@@ -245,9 +245,11 @@ impl VmConnectorChip {
         }
     }
 
+    /// `state.pc` is a byte program counter; it is stored (and later exposed as a public value)
+    /// as a pc index, matching the circuit representation of the pc.
     pub fn begin(&mut self, state: ExecutionState<u32>) {
         self.boundary_states[0] = Some(ConnectorCols {
-            pc: state.pc,
+            pc: pc_to_idx(state.pc),
```
shuklaayush: let's call it `pc_idx` to avoid ambiguity

crates/vm/src/system/phantom/trace.rs
```diff
@@ -38,7 +41,7 @@ pub(crate) fn generate_trace_from_postflight<F: PrimeField32>(
 
         let row: &mut PhantomCols<F> =
             trace.values[row_index * width..(row_index + 1) * width].borrow_mut();
-        row.pc = F::from_u32(pc);
+        row.pc = F::from_u32(pc_to_idx(pc));
```
shuklaayush: `pc_idx`

shuklaayush (commented):
left a couple of comments

shuklaayush (approved):

PR #3112, opened 2026-08-07
perf: sparse initial memory snapshot and GPU Merkle build

GunaDD:
Builds the initial memory image and the initial Merkle tree sparsely, so per-segment postflight cost scales with the pages a guest actually touches rather than with the configured address-space size.Observed per-segment initial-Merkle postflight time drops from ~221ms to ~2-22ms depending on how many pages a segment touches.

Perf bench: https://github.com/axiom-crypto/openvm-eth/actions/runs/31213205639

Closes INT-9124 and INT-8149 as well.

### DIFF
--- a/crates/vm/cuda/src/system/memory/merkle_tree.cu
+++ b/crates/vm/cuda/src/system/memory/merkle_tree.cu
@@ -420,6 +420,13 @@ __device__ void store_virtual_node(
     size_t const label,
     digest_t const *value
 ) {
+    // Non-existent nodes (labels beyond the stored `actual_height` prefix) have no slot:
+    // the heap region has no entry for them and `stored_node_index` would alias the
+    // vertical-path slot of the label-0 node at path levels. Their updated digests still
+    // propagate through the layer records, so the store can be skipped.
+    if (!virtual_node_exists(node_height, actual_height, label)) {
+        return;
+    }
     if (layout == OMIT_BOTTOM_LEVELS && node_height < OMITTED_BOTTOM_LEVELS) {
         return;
     }
--- a/crates/vm/cuda/src/system/memory/merkle_tree.cu
+++ b/crates/vm/cuda/src/system/memory/merkle_tree.cu
@@ -15,10 +15,12 @@ struct alignas(32) digest_t {
 
 #define COPY_DIGEST(dst, src) memcpy(dst, src, sizeof(digest_t))
 
-enum MemoryMerkleSubTreeLayout : uint8_t {
-    FULL = 0,
-    OMIT_BOTTOM_LEVELS = 1,
-    SPARSE_PAGES = 2,
+// How a subtree's stored nodes are indexed within its digest buffer. Orthogonal to the subtree's
+// `base_height`, the lowest height it stores at all; nodes below that height are recomputed from
+// raw memory under either indexing.
+enum SubTreeIndexing : uint8_t {
+    DENSE = 0,  // heap-ordered; a node's slot is computed from its (height, label)
+    SPARSE = 1, // sorted per-height label index; a node's slot is found by binary search
 };
 
 // Selects one height's contiguous range in the flat sparse-node arrays.
@@ -91,9 +93,9 @@ __device__ void recompute_omitted_node(
     size_t const label,
     digest_t *out
 ) {
-    // layer is a fixed-size, thread-local scratch buffer of 2^OMITTED_BOTTOM_LEVELS 
+    // layer is a fixed-size, thread-local scratch buffer of 2^MAX_OMITTED_LEVELS
     // digests, used to rebuild an omitted subtree bottom-up.
-    digest_t layer[1 << OMITTED_BOTTOM_LEVELS];
+    digest_t layer[1 << MAX_OMITTED_LEVELS];
     // num_leaves denote the number of leaves of the subtree rooted at this node 
     size_t const num_leaves = 1 << node_height;
     // recall again that node_height is counted starting from the bottom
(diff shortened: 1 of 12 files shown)


crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -18,6 +18,12 @@ struct alignas(32) digest_t {
 enum MemoryMerkleSubTreeLayout : uint8_t {
     FULL = 0,
     OMIT_BOTTOM_LEVELS = 1,
+    SPARSE_PAGES = 2,
+};
```
shuklaayush: i feel like we're conflating two orthogonal concepts here. one is whether the representation is dense or sparse, and the other is the merkle tree height (or omitted levels). imo it's better to split this into sparse/dense enum and a separate parameter

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -18,6 +18,12 @@ struct alignas(32) digest_t {
...
+
+struct SparseLevel {
```
shuklaayush, line 30: Add some comments here

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -101,6 +107,70 @@ __global__ void merkle_tree_init_omitted(
...
+        } else {
+            hi = mid;
+        }
+    }
+    return lo < level.count && labels[level.start + lo] == needle
+        ? level.start + lo
+        : UINT_MAX;
+}
+
+__global__ void sparse_merkle_compress(
```
shuklaayush, line 170: I wonder if we should have separate spare and dense files

crates/vm/src/system/cuda/merkle_tree/mod.rs
```diff
@@ -29,6 +33,114 @@ pub const TIMESTAMPED_BLOCK_WIDTH: usize = 3 + BLOCK_FE_WIDTH;
 pub const MERKLE_TOUCHED_BLOCK_WIDTH: usize = 3 + VM_DIGEST_WIDTH;
 pub(crate) const OMITTED_BOTTOM_LEVELS: usize = 3;
 
+#[derive(Debug)]
+pub(crate) enum InitialMerkleBuild {
```
shuklaayush, line 64: Add some comments

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -101,6 +107,70 @@ __global__ void merkle_tree_init_omitted(
...
+    uint32_t lo = 0;
+    uint32_t hi = level.count;
+    while (lo < hi) {
+        uint32_t const mid = lo + (hi - lo) / 2;
+        uint32_t const value = labels[level.start + mid];
+        if (value < needle) {
+            lo = mid + 1;
+        } else {
+            hi = mid;
+        }
```
Golovanov399, line 163: I suspect that this binary search is wrong. What if we need exactly the position `level.count / 2`? Will we return `UINT_MAX` then?

crates/vm/src/system/memory/merkle/mod.rs
```diff
@@ -73,46 +73,53 @@ pub(crate) fn memory_to_vec_partition<F: PrimeField32, const N: usize>(
...
+                        })
                 })
                 .collect::<Vec<_>>()
         })
         .collect::<Vec<_>>()
         .into_iter()
         .flatten()
-        .collect::<Vec<_>>()
+        .collect::<Vec<_>>();
+    partition.sort_unstable_by_key(|(index, _)| *index);
```
Golovanov399, line 123: Why unstable?

crates/vm/src/system/cuda/merkle_tree/mod.rs
shuklaayush, line 1: a figure like this would help
```
                   stored root
                  /           \
             stored upper dense tree
               /   /   \   \
          height-3 base nodes        ← stored
            ⋮       ⋮
       bottom 3 levels               ← omitted
            ⋮
         raw memory
```

GunaDD: sounds good

crates/vm/cuda/src/system/memory/merkle_tree.cu
GunaDD: I don't think we should

crates/vm/cuda/src/system/memory/merkle_tree.cu
shuklaayush: there's also a binary search in cuda primitives
https://nvidia.github.io/cccl/unstable/cub/api/structcub_1_1DeviceFind.html#_CPPv4I0000000EN3cub10DeviceFind10LowerBoundE11cudaError_tPvR6size_t14RangeIteratorT14RangeNumItemsT15ValuesIteratorT15ValuesNumItemsT15OutputIteratorT10CompareOpTRK4EnvT

GunaDD: if we need the position `level.count / 2`, the binary search will terminate at `lo = level.count / 2` because in the first iteration it will go `hi = level.count / 2` and in the subsequent iterations it will always go to the `lo = mid + 1` part of the if statement and finally terminate when `lo = hi`.

crates/vm/src/system/memory/merkle/mod.rs
GunaDD: because the keys are unique and it might be faster than the usual sort

crates/vm/cuda/src/system/memory/merkle_tree.cu
GunaDD: will fix this, thanks!

shuklaayush (approved):

PR #3114, opened 2026-08-07
fix(v2.1): restore byte-sized public values

shuklaayush:
Restores public values as native U8 cells while keeping register and general memory U16 and deferral memory Field32. This preserves the byte-oriented public API and verifier input shape, with matching CPU, CUDA, RVR, Merkle, and REVEAL handling.

Resolves INT-8076, INT-9017


GunaDD (commented):
@shuklaayush Claude pointed out a bunch of correctness issues, do you think they are valid ? 
```
# Code review — PR #3114

12 findings, most-severe first.

---

## 1. `crates/vm/src/arch/execution_mode/metered/memory_ctx.rs:123` — correctness

**Metered leaf accounting still uses the u16 shift (16 bytes/leaf) for PUBLIC_VALUES_AS, which is now U8-celled (8 bytes/leaf), so boundary/merkle/poseidon2 trace-height estimates for public values are ~2x too small.**

`leaf_id_range` picks `BYTE_PTRS_PER_LEAF_BITS = log2(U16_CELL_SIZE * VM_DIGEST_WIDTH) = 4` for every non-DEFERRAL address space, and `MeteredCtx::on_memory_operation` feeds it the raw byte pointer that `vm_write_bytes(PUBLIC_VALUES_AS, addr, 8)` reports. With PUBLIC_VALUES_AS now `MemoryCellType::U8`, one merkle leaf is `DIGEST_WIDTH * 1 = 8` bytes, so `ptr >> 4` collapses two real leaves into one. Default config + `openvm::io::reveal_bytes32` writes bytes 0..32 = 4 real boundary leaves, but metering counts 2 (and correspondingly fewer merkle nodes / default-poseidon rows). Under-estimated heights mean a segment can be closed too late and the real tracegen heights exceed the metered estimate.

The same defect is mirrored in the generated-C tracer: `crates/rvr/rvr-openvm/c/tracer/openvm_tracer_metered.h:85` routes `AS_PUBLIC_VALUES` through `byte_addr_to_local_leaf` (`>> TRACER_BYTE_SPACE_PTRS_PER_LEAF_BITS` = 4), and `crates/rvr/rvr-openvm/src/constants.rs:12` still derives that constant from `size_of::<u16>() * VM_DIGEST_WIDTH`. `seed_initial_memory` has the same issue for public-value bytes in the initial image.

---

## 2. `extensions/riscv/circuit/src/reveal/air.rs:117` — correctness

**The reveal destination range check still uses `byte_ptr_max_bits`, but `dst_ptr` is now the AS-native public-values pointer, so the AIR admits pointers up to 2^(pointer_max_bits+1) — twice the merkle address domain.**

`RevealAir` is constructed with `byte_ptr_max_bits = to_byte_ptr_bits(pointer_max_bits) = pointer_max_bits + 1` (extension/mod.rs:403). Before the change the write pointer was `byte_ptr_to_u16_ptr(dst_ptr) = dst_ptr/2`, so the effective AS-native bound was 2^pointer_max_bits. Now the write address is `dst_ptr` itself, so the check `range_check(dst_ptr_high_limb, pointer_max_bits - U16_BITS)` permits `dst_ptr < 2^(pointer_max_bits+1)` (2^29 with defaults) while every other layer caps AS-native pointers at 2^pointer_max_bits (2^28): `postflight/index.rs::validate_memory_block` uses `1 << config.pointer_max_bits`, and `MemoryDimensions::address_height = pointer_max_bits - DIGEST_WIDTH_BITS`. A leaf label >= 2^address_height is outside the merkle domain; `MemoryDimensions::label_to_index` only `debug_assert!`s that bound, so a release-build tracegen that ever reached such a pointer would silently index into the next address space's subtree region.

The same looseness is copied into `extensions/riscv/circuit/src/reveal/trace.rs:71-81` (`pointer_limit = 1 << chip.inner.pointer_max_bits`) and `extensions/riscv/circuit/cuda/rvr/include/riscv/reveal_replay.cuh:126` (`domain_end`).

---

## 3. `crates/vm/cuda/src/system/boundary.cu:65` — correctness

**The new `switch (cell_kinds[addr_space_idx])` has no `default:` arm, so `CELL_UNSUPPORTED` (0) leaves `init_values` uninitialized; the previous if/else always assigned, and `_build_merkle_subtree` used to return -1 for an unknown kind.**

`FpArray<DIGEST_WIDTH> init_values;` is declared uninitialized before the switch; if `cell_kinds[addr_space_idx] == CELL_UNSUPPORTED` and `initial_mem[addr_space_idx]` is non-null, the boundary row is filled with stack garbage and the proof silently commits to wrong initial memory. Same pattern in `crates/vm/cuda/src/system/inventory.cu:79` (`out_values` left holding the caller's stale record values) and `crates/vm/cuda/src/system/memory/merkle_tree.cu:33` (cells stay zero, hashing an all-zero leaf). The removed `default: return -1;` in `_build_merkle_subtree` (merkle_tree.cu:753-768) was the last explicit error path for an unrecognized address-space layout; nothing replaced it.

---

## 4. `crates/vm/src/system/cuda/mod.rs:39` — correctness

**`MemoryCellKind::from(MemoryCellType)` panics on `U32` / non-4-byte `F` layouts, and it is now invoked for every address space (including empty ones) during `MemoryInventoryGPU::new` and `MemoryMerkleTree::build_async`.**

A `MemoryConfig` that declares an address space with `MemoryCellType::U32` or `F { size: 8 }` and `num_cells == 0` previously constructed a GPU memory inventory fine. Now `From` hits `_ => panic!("unsupported CUDA memory layout: {layout:?}")` before the adjacent `assert!(config.num_cells == 0 || kind != MemoryCellKind::Unsupported)` (memory.rs:175) can exempt it — the assert's own wording shows empty spaces were meant to be tolerated. Mapping unknown layouts to `Unsupported` instead of panicking makes the two checks consistent.

---

## 5. `extensions/riscv/circuit/src/reveal/air.rs:146` — correctness

**`chunks_exact(BLOCK_FE_WIDTH).zip(&cols.write_aux)` silently truncates if the block/register widths ever diverge, and the write count `2` is hardcoded in three places instead of derived.**

`RevealCols::write_aux` is `[MemoryWriteAuxCols<T, BLOCK_FE_WIDTH>; 2]`, `src_bytes` is `[T; MEMORY_BLOCK_BYTES]`, and `trace.rs:86-97` builds a hardcoded 2-element `writes` array. The relationship `MEMORY_BLOCK_BYTES / BLOCK_FE_WIDTH == 2` is never asserted. If either constant changes, `zip` stops at the shorter side and the AIR would constrain only the first N*BLOCK_FE_WIDTH bytes of the register while tracegen still emits all writes — public-value bytes written but not constrained, which is a soundness hole rather than a compile error. Introduce `const NUM_PV_WRITE_BLOCKS: usize = MEMORY_BLOCK_BYTES / BLOCK_FE_WIDTH;`, size `write_aux` with it, and `const { assert!(NUM_PV_WRITE_BLOCKS * BLOCK_FE_WIDTH == MEMORY_BLOCK_BYTES) }`.

---

## 6. `crates/vm/src/arch/config.rs:234` — api-change

**`DEFAULT_MAX_NUM_PUBLIC_VALUES = 32` is reinterpreted from u16 cells to bytes, silently halving the default public-values capacity from 64 bytes to 32 bytes.**

Before: `num_public_values = 32` cells => `addr_spaces[PUBLIC_VALUES_AS].size() = 64` bytes. After: 32 U8 cells = 32 bytes. A guest using the default `SystemConfig` that reveals a 33rd..64th byte (e.g. `reveal_u64(x, 4)`) now fails execution with `"reveal address is not aligned within configured public-values capacity"` from `checked_reveal_address`, where it previously succeeded. The PR description claims the change "preserves the byte-oriented public API and verifier input shape", which is true for the Fr instance *count* but not for the byte payload length (`EvmProof::user_public_values` goes from 2*N to N bytes). Worth confirming this is intended and release-noting it — it is a verifying-key- and API-visible change.

---

## 7. `crates/vm/cuda/include/system/memory/params.cuh:5` — correctness

**The cell-kind mapping is now duplicated across four independently-defined enums with no cross-language pinning for the new pair.**

`MemoryCellKind` (params.cuh:5), `MemoryCellKind` (crates/vm/src/system/cuda/mod.rs:32), `GpuMemoryCellKind` (crates/vm/cuda/src/system/postflight.cu:24) and `GpuMemoryCellKind` (crates/vm/src/arch/cuda/postflight.rs:82) all encode Unsupported=0/U8=1/U16=2/Field32=3. Only the postflight pair has `static_assert`s on the discriminants, and even there the Rust `const _` block (postflight.rs:108-116) checks only sizes/offsets, never the values. The brand-new params.cuh ↔ `system::cuda::MemoryCellKind` pair has no check at all, so reordering one enum silently makes the boundary/inventory/merkle kernels decode U16 initial memory as U8 (or vice versa), producing a wrong memory root with no diagnostic. Collapse to one enum per side (or add value static_asserts plus a Rust-side discriminant assert).

---

## 8. `crates/vm/cuda/src/system/memory/merkle_tree.cu:33` — efficiency

**`hash_raw_memory_leaf` now branches on a runtime `cell_kind` inside the per-cell unrolled loop, replacing the compile-time `template <int ADDR_SPACE_IDX>` dispatch, in the hottest initial-merkle kernel.**

`cell_kind` is uniform for an entire `merkle_tree_init` / `merkle_tree_init_omitted` launch, but the switch sits inside `for (i = 0; i < DIGEST_WIDTH; ++i)` under `#pragma unroll`, so the compiler can no longer constant-fold the addressing (`data[cell_idx]` vs `U16_CELL_SIZE * cell_idx` vs `reinterpret_cast<Fp const*>`). `merkle_tree_init` runs once per leaf over the whole initial-memory image (up to 2^25 leaves for MEMORY_AS at default pointer_max_bits), so this is a measurable regression on the initial-tree build. Hoist the switch above the loop (three specialized loops or `if constexpr` via a small template on cell_kind).

---

## 9. `crates/vm/src/system/cuda/boundary.rs:117` — efficiency

**The immutable `cell_kinds` vector is re-uploaded to the device on every tracegen call instead of once at construction.**

`self.cell_kinds.to_device_on(&self.device_ctx)` runs inside `Chip::generate_proving_ctx`, i.e. once per segment, and again in `MemoryInventoryGPU::generate_proving_ctxs_from_device` (memory.rs:472-476), and `MemoryMerkleTree::update_with_touched_blocks` re-collects and re-uploads `cell_kinds` from the subtrees each call (merkle_tree/mod.rs:431-435). The data is fixed after `MemoryInventoryGPU::new`; hold one `DeviceBuffer<u8>` on the chip and reuse it. Small per-segment H2D + allocation churn that scales with segment count.

---

## 10. `extensions/riscv/circuit/src/reveal/air.rs:50` — clarity

**`src_bytes` uses `MEMORY_BLOCK_BYTES` (documented as "byte width of one RV64 U16 register/memory block") as a stand-in for the register byte width, coupling the register decomposition to a memory-bus constant.**

The register value being decomposed is `REGISTER_NUM_LIMBS = 8` bytes wide; `MEMORY_BLOCK_BYTES` happens to equal 8 today because it is `BLOCK_FE_WIDTH * U16_CELL_SIZE`. `crates/vm/src/arch/config.rs:66-68` was updated in this same PR to say block byte width is layout-dependent, which makes the reuse actively misleading. `extensions/riscv/circuit/src/reveal/trace.rs` already gets the byte array from `u16_block_to_bytes(...) -> [u8; REGISTER_NUM_LIMBS]`, so the two sides name the same 8 with different constants. Use `REGISTER_NUM_LIMBS` for `src_bytes` and keep `MEMORY_BLOCK_BYTES` only for the 8-byte destination alignment check.

---

## 11. `crates/vm/src/arch/mod.rs:46` — simplification

**A second `pub use postflight::{...}` guarded by `#[rustfmt::skip]` exists only to preserve U8/U16/Field32 name ordering; it fights the repo's configured `imports_granularity = "Crate"`.**

AGENTS.md states "Formatting uses unstable options: `group_imports = "StdExternalCrate"`, `imports_granularity = "Crate"`". Suppressing the formatter for one re-export means the next person running `cargo +nightly fmt --all` sees an item the tool deliberately ignores, and the split makes it easy to add a new export to the wrong list. Merge `U8Access, U16Access, Field32Access` into the existing `pub use postflight::{...}` block and drop the attribute. (Note `Field32Access` was not exported before this PR — exporting it is an unrelated, if justified, API addition.)

---

## 12. `crates/vm/src/arch/postflight.rs:350` — simplification

**`previous_u16` is now a pure pass-through to the newly extracted `previous_inline`, leaving two names for one function.**

`fn previous_u16(&self, i) -> [u16; BLOCK_FE_WIDTH] { self.previous_inline(i) }` adds an indirection with no behavior. Either keep `previous_u16` as the single u16 accessor and have `previous_u8` call it, or delete `previous_u16` and call `previous_inline` from `access_u16`/`peek_u16`/`testing.rs`. As written, a reader has to check whether the two differ.
```

shuklaayush:
@GunaDD i think i addressed whatever needed to be addressed

crates/vm/src/arch/postflight.rs
GunaDD: can we just make this generic ? and isn't this just `MemoryBaseAuxCols` ?

crates/vm/src/arch/testing/memory/cuda.rs
```diff
@@ -117,10 +124,23 @@ impl DeviceMemoryTester {
         let t = self.memory.timestamp();
         let cell_layout = self.memory.data().memory.config[addr_space].layout;
         let (t_prev, data_prev) = match cell_layout {
-            MemoryCellType::F { .. } => unsafe {
-                self.memory
-                    .write::<F, BLOCK_FE_WIDTH>(addr_space as u32, ptr as u32, data)
-            },
+            MemoryCellType::U8 => {
+                let (t_prev, data_prev) = unsafe {
+                    self.memory.write::<u8, BLOCK_FE_WIDTH>(
```
GunaDD: why `BLOCK_FE_WIDTH` ?

crates/rvr/rvr-openvm/c/tracer/openvm_tracer_metered.h
```diff
@@ -368,21 +374,21 @@ static __attribute__((always_inline)) inline void trace_sp_memory_access_leaf(
 static __attribute__((always_inline)) inline void trace_memory_access_span_impl(
     TraceMemory* restrict memory, TraceMemoryCache cache, uint64_t addr,
     uint32_t size) {
-  assume(size != 0u && size <= TRACER_BYTES_PER_LEAF);
+  assume(size != 0u && size <= TRACER_U16_BYTES_PER_LEAF);
```
GunaDD: the constant name is confusing to me, what does `U16_BYTES` mean ?

crates/rvr/rvr-openvm/src/constants.rs
```diff
@@ -9,12 +9,13 @@ use openvm_platform::{memory::MEM_SIZE, WORD_SIZE};
 use rvr_openvm_lift::MAIN_MEMORY_PAGE_BYTES;
 use rvr_state::PREFLIGHT_DIRTY_PAGE_BYTES;
 
-const BYTE_SPACE_PTRS_PER_LEAF: usize = core::mem::size_of::<u16>() * VM_DIGEST_WIDTH;
-const DEFERRAL_PTRS_PER_LEAF: usize = VM_DIGEST_WIDTH;
+const U8_PTRS_PER_LEAF: usize = VM_DIGEST_WIDTH;
+const U16_BYTE_PTRS_PER_LEAF: usize = core::mem::size_of::<u16>() * VM_DIGEST_WIDTH;
+const FIELD32_PTRS_PER_LEAF: usize = VM_DIGEST_WIDTH;
```
GunaDD: there is probably a better naming for these constants, and are we sure that the u8 ptrs per leaf is the same as field32 ptrs per leaf ?

crates/sdk/src/tests.rs
```diff
@@ -329,21 +329,19 @@ fn make_verify_stark_inputs_for_indices(
...
-fn collapse_user_public_values_preserves_u16_cells() {
-    let expanded = [0x34, 0x12, 0, 0, 0xcd, 0xab, 0, 0];
-    assert_eq!(
-        collapse_user_public_values(&expanded),
-        [0x34, 0x12, 0xcd, 0xab]
-    );
+fn collapse_user_public_values_preserves_bytes() {
+    let expanded = [0x34, 0, 0, 0, 0xab, 0, 0, 0];
+    assert_eq!(collapse_user_public_values(&expanded), [0x34, 0xab]);
 }
```
GunaDD: probably don't need this test since the function is so simple already and already asserts stuff

crates/vm/cuda/src/system/memory/merkle_tree.cu
```diff
@@ -22,42 +22,57 @@ enum MemoryMerkleSubTreeLayout : uint8_t {
 
 __device__ __forceinline__ void hash_raw_memory_leaf(
     uint8_t const *__restrict__ data,
-    uint32_t const addr_space_idx,
+    uint8_t const cell_kind,
```
GunaDD: I don't like `cell_kind` ? maybe `cell_type` / `memory_type` ? what is more consistent with other parts of the codebase ? I think there is something like this in the vm config file.

crates/vm/cuda/src/system/postflight.cu
```diff
@@ -23,8 +23,9 @@ static constexpr uint32_t ERROR_FIELD_REFERENCE = 119;
 
 enum class GpuMemoryCellKind : uint32_t {
```
GunaDD: same here about the `GpuMemoryCellKind`

crates/vm/src/arch/postflight/tests.rs
```diff
@@ -1,12 +1,14 @@
-use openvm_instructions::{program::Program, riscv::REGISTER_AS, SystemOpcode, DEFERRAL_AS};
+use openvm_instructions::{
```
GunaDD, line 1: there is `testing.rs` and `tests.rs` in `postflight`, probably needs a clean up

extensions/riscv/circuit/src/reveal/air.rs
```diff
@@ -114,31 +118,52 @@ impl<AB: InteractionBuilder> Air<AB> for RevealAir {
...
-                cols.src_data.map(Into::into),
-                timestamp.clone() + AB::Expr::TWO,
-                &cols.write_aux,
-            )
-            .eval(builder, is_valid.clone());
+        for &byte in &cols.src_bytes {
+            self.range_bus
+                .range_check(byte, BYTE_BITS)
+                .eval(builder, is_valid.clone());
+        }
```
GunaDD: can we save half interaction here by using the xor lookup pair instead of range check bus ?

shuklaayush: sure

crates/vm/src/arch/postflight/tests.rs
shuklaayush: it's fine. `testing.rs` contains utils that are shared by tests

crates/vm/src/arch/postflight.rs
shuklaayush: i deduplicated it. no, `MemoryBaseAuxCols` contains the limb decomposition of the timestamp diff

crates/vm/cuda/src/system/inventory.cu
```diff
@@ -78,21 +76,30 @@ __device__ inline void read_initial_leaf(
...
-            out_values[i] = Fp(u16_from_bytes_le(mem + base + U16_CELL_SIZE * i)).asRaw();
+        case CELL_FIELD32: {
+            uint32_t const *cells = reinterpret_cast<uint32_t const *>(mem) + ptr;
+#pragma unroll
+            for (int i = 0; i < DIGEST_WIDTH; ++i) out_values[i] = cells[i];
+            break;
         }
+        default:
+            assert(false && "unsupported memory cell type");
+            break;
```
GunaDD, line 107: claude flagged this regarding the default block: 
```
the new `switch (cell_kinds[...])` in
`read_initial_leaf` leaves `out_values` **uninitialized** on `default:`. `assert(false)` compiles
out under NDEBUG, so an unsupported address space yields stack garbage that `boundary.cu` then
reads as Montgomery-encoded `Fp` → nondeterministic boundary trace / merkle root, no error raised.
The sibling change in `boundary.cu` was hardened with `FpArray<DIGEST_WIDTH> init_values{}`; do the
same here.
```
not sure if this is a concern or not.

crates/rvr/rvr-openvm/c/tracer/openvm_tracer_metered.h
shuklaayush: i renamed it to `TRACER_MEMORY_LEAF_BYTES`. this represents how many bytes a merkle leaf of the memory address space contains

crates/vm/src/arch/execution_mode/metered/memory_ctx.rs
```diff
@@ -117,10 +118,10 @@ impl MemoryCtx {
...
     fn leaf_id_range(&self, address_space: u32, ptr: u32, size: u32) -> (u32, u32) {
         let end_ptr = ptr + size - 1;
-        let leaf_bits = if address_space == DEFERRAL_AS {
-            DEFERRAL_PTRS_PER_LEAF_BITS
-        } else {
-            BYTE_PTRS_PER_LEAF_BITS
+        let leaf_bits = match address_space {
+            PUBLIC_VALUES_AS => U8_PTRS_PER_LEAF_BITS,
+            DEFERRAL_AS => FIELD32_PTRS_PER_LEAF_BITS,
+            _ => U16_BYTE_PTRS_PER_LEAF_BITS,
```
GunaDD: should we do a better edge case handling here ? like causing an error for other unsupported address space instead of giving it `U16_BYTE_PTRS_PER_LEAF_BITS`.

crates/rvr/rvr-openvm/src/constants.rs
shuklaayush: renamed to hopefully make it clearer

crates/vm/src/arch/cuda/postflight.rs
```diff
@@ -418,7 +420,12 @@ fn validated_history_write_masks(
                 ))
             })?
             .layout;
-        write_masks.push(if event.is_write() { u8::MAX } else { 0 });
+        let full_write_mask = match layout {
+            MemoryCellType::U8 => 0x0f,
+            MemoryCellType::U16 | MemoryCellType::FIELD32 => u8::MAX,
+            _ => 0,
```
GunaDD: should we do better error handling here instead of silently setting it to 0 ?

crates/vm/cuda/include/system/memory/params.cuh
```diff
@@ -1,18 +1,30 @@
...
 
+#include <cstdint>
+
+enum MemoryCellType : uint8_t {
+    CELL_UNSUPPORTED = 0,
+    CELL_U8 = 1,
+    CELL_U16 = 2,
+    CELL_FIELD32 = 3,
+};
+
```
GunaDD, line 12: there is this and there is `GpuMemoryCellType` in `system/postflight.cu` which is basically the same thing, should we fix this ?

shuklaayush: removed this enum

crates/vm/cuda/src/system/inventory.cu
shuklaayush: i fixed these for completeness but we probably won't have unsupported address spaces

extensions/riscv/circuit/src/reveal/air.rs
```diff
@@ -42,8 +45,8 @@ pub struct RevealCols<T> {
     pub base_aux: MemoryReadAuxCols<T>,
     /// Byte pointer to the source-value register.
     pub src_ptr: T,
-    /// Source register value as u16 limbs.
-    pub src_data: [T; BLOCK_FE_WIDTH],
+    /// Source register value as byte limbs.
+    pub src_bytes: [T; MEMORY_BLOCK_BYTES],
```
GunaDD: this is probably `REGISTER_BYTES` / `REGISTER_NUM_LIMBS` instead of `MEMORY_BLOCK_BYTES`

GunaDD (approved):
LGTM

PR #3122, opened 2026-08-17
chore: CI to compare locked Rust verifier to lean

stephenh-axiom-xyz:
Resolves INT-9061.

Adds a CI workflow to compare the locked `stark-backend` Rust verifier against the one used to formally verify the SWIRL verifier. OpenVM CI now checks that:

- The C verifier (i.e. `fv-verifier`) matches the one extracted from the handwritten Lean verifier
- The Rust verifier (i.e. locked `stark-backend`) extracts to the official extracted Lean verifier, which is proven to be semantically equivalent to the handwritten one

### DIFF
--- a/.github/workflows/fv-verifier.yml
+++ b/.github/workflows/fv-verifier.yml
@@ -18,6 +18,7 @@ on:
       - "rust-toolchain.toml"
       - ".github/actions/sccache/**"
       - ".github/workflows/fv-verifier.yml"
+      - "ci/scripts/check_swirl_verifier_extraction.sh"
       - "ci/scripts/check_swirl_verifier_c.sh"
 
 concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/fv-verifier.yml
```diff
@@ -109,6 +111,20 @@ jobs:
...
 
+      - name: Route the private rust-to-lean clone through SSH
+        run: |
+          git config --global \
+            url."git@github.com:axiom-crypto/rust-to-lean".insteadOf \
+            "https://github.com/axiom-crypto/rust-to-lean"
+
+      - name: Install Rust extraction toolchain
+        run: |
+          source swirl-rbr-fv/verifier-lean/extraction/sources.lock
```
shuklaayush, line 122: is this a shell script?

.github/workflows/fv-verifier.yml
```diff
@@ -91,8 +92,9 @@ jobs:
         if: always()
         run: sccache --show-stats || true
 
+  # TODO[INT-9153]: update used repos and references
```
shuklaayush, line 95: i guess this means the public repos would be different?

shuklaayush (approved):
looks fine

stephenh-axiom-xyz:
Nvm we're not doing this

PR #3123, opened 2026-08-19
fix: constrain internal_flag to constant in VerifierPvsAir

stephenh-axiom-xyz:


Cargo.lock
```diff
@@ -1867,7 +1867,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
```
zlangley, line 1867: Are these changes intentional?

zlangley (approved):

Cargo.lock
stephenh-axiom-xyz: Yes, was failing audit

PR #3125, opened 2026-08-20
chore: rename fv verifier to certified verifier

stephenh-axiom-xyz:
Resolves INT-9202. Note new crate is `openvm-certified-verifier`, every instance of `fv-verifier` has been replaced


yi-sun (approved):
LGTM

PR #3131, opened 2026-08-25
chore: bump version to v2.1.0

stephenh-axiom-xyz:
Towards INT-9126.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "2.0.2"
+version = "2.1.0"
 edition = "2021"
 rust-version = "1.91.1"
 authors = ["OpenVM Authors"]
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -125,61 +125,61 @@ openvm-cuda-builder = { version = "2.0.1", git = "https://github.com/openvm-org/
 openvm-cuda-common = { version = "2.0.1", git = "https://github.com/openvm-org/stark-backend.git", branch = "main", default-features = false }
```
shuklaayush, line 125: i would prefer if we keep them in sync

stephenh-axiom-xyz: They haven't been in sync since the beginning, which we explicitly decided to do

shuklaayush (approved):
i'm approving to unblock but my preference is still keeping all packages in sync, irrespective of how it was before

PR #3132, opened 2026-08-27
chore: prover assert check that all AIRs are required for cached pks

stephenh-axiom-xyz:
Resolves INT-9260, INT-9264. In a previous commit, we modified keygen to mark each verifier circuit AIR as required. This commit modifies the provers to check that cached pks were generated with this change, avoiding a footgun.


shuklaayush (commented):
what is the invariant that we're trying to preserve exactly?
is it that recursion and deferral proofs can't have empty traces for any airs? maybe you can add a brief pr description

shuklaayush (approved):

PR #3134, opened 2026-08-27
fix: reject oversized framed inputs in certified-verifier

stephenh-axiom-xyz:
Resolves INT-9277

### DIFF
--- a/crates/certified-verifier/src/harness.rs
+++ b/crates/certified-verifier/src/harness.rs
@@ -66,15 +66,22 @@ pub struct SwirlVerifyOutcome {
 /// ```text
 /// u32 LE vk_len | vk_bytes | u32 LE proof_len | proof_bytes | u32 LE pv_len | pv_bytes
 /// ```
-fn frame_three_blobs(vk_bytes: &[u8], proof_bytes: &[u8], pv_bytes: &[u8]) -> Vec<u8> {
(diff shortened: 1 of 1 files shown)


crates/certified-verifier/src/harness.rs
```diff
@@ -66,15 +66,22 @@ pub struct SwirlVerifyOutcome {
 /// ```text
 /// u32 LE vk_len | vk_bytes | u32 LE proof_len | proof_bytes | u32 LE pv_len | pv_bytes
 /// ```
-fn frame_three_blobs(vk_bytes: &[u8], proof_bytes: &[u8], pv_bytes: &[u8]) -> Vec<u8> {
-    let mut buf = Vec::with_capacity(12 + vk_bytes.len() + proof_bytes.len() + pv_bytes.len());
+fn frame_three_blobs(vk_bytes: &[u8], proof_bytes: &[u8], pv_bytes: &[u8]) -> io::Result<Vec<u8>> {
```
Golovanov399, line 69: Do we use `io::Result` and not `eyre::Result`? Is it input/output related?

stephenh-axiom-xyz: We use `io::Result` in this particular crate

Golovanov399: I think it's wrong but ok

Golovanov399 (approved):

PR #3138, opened 2026-08-28
fix: use FFI to call extracted certified verifier

stephenh-axiom-xyz:
Resolves INT-9275.

## Summary

- Replace the certified verifier subprocess with an in-process FFI call to the Lean-generated verifier.
- Add a small C adapter exposing a single `openvm_swirl_verify` function to Rust.
- Statically link the generated verifier and Lean runtime into `cargo-openvm.`
- Remove the embedded temporary verifier executable path and simplify the Rust API to `run_certified_verifier`.
- Add CI coverage that installs `cargo-openvm`, proves Fibonacci, deletes the Cargo install target directory, and then verifies the proof with `--certified`.

This fixes published installations attempting to execute a verifier from Cargo’s deleted temporary build directory.


shuklaayush (approved):
can we also add a negative test where the lean verification fails for an invalid proof? otherwise looks fine

stephenh-axiom-xyz:
@shuklaayush there are such tests in `certified-verifier` already, will make the testing suite more robust in a later PR

PR #3154, opened 2026-09-10
chore: update guest Rust to 1.97.1

Qumeric:
## Summary

- update the default prebuilt guest Rust toolchain to `openvm-1.97.1`
- pin upstream host and tooling nightly to `nightly-2026-09-10`
- keep CI and contributor tooling synchronized, and document nightly tags without hard-coding a dated release
- disable the unused `embedded-alloc/allocator_api` feature, whose `Layout::dangling` call is incompatible with this compiler fork
- lock the verify-stark host dependencies so deleted upstream branches do not break clean CI builds


shuklaayush:
superceded by #3155. `develop-v2.1.0` is only kept for historical reasons

PR #3157, opened 2026-09-15
feat: split app_exe_commit fields in STARK verify baseline

stephenh-axiom-xyz:
Towards INT-9471.

STARK verification now checks the executable's program commitment, initial memory root, and initial PC individually. EVM verification and the combined executable commitment remain unchanged.

### `crates/verify`

- Replace `VerificationBaseline.app_exe_commit` with `program_commit`, `initial_state`, and `initial_pc`.
- Compare each field directly and return field-specific mismatch errors.
- Add `VerificationBaseline::app_exe_commit()` to reconstruct the combined commitment for existing consumers.

### `crates/vm`

- Extract `compute_initial_memory_commit` for reuse by executable commitment and baseline generation.
- Pass the hasher into the helper to avoid constructing it twice.

### `crates/sdk`

- Generate and serialize the three baseline fields.
- Keep `VmBaseline` executable-independent by excluding all three fields.
- Preserve combined application commitments through `app_exe_commit()`.

Existing serialized STARK baselines require regeneration.

### `crates/cli`

- Update `commit` and `prove stark` to compute the combined commitment through the helper, preserving the existing app commitment format.

### `crates/certified-verifier`

- Preserve the Lean verifier's wire format by encoding the reconstructed executable commitment.
- Update fixtures and add rejection coverage for mismatched `program_commit`, `initial_state`, and `initial_pc`.

### Documentation

- Update the verification README and baseline-generation documentation for the new fields.


crates/certified-verifier/src/vm.rs
```diff
@@ -32,7 +32,7 @@ pub fn write_verification_baseline<W: Write>(
     }
 
     write_header(writer, MAGIC_VM_BASELINE)?;
-    SC::encode_digest(&baseline.app_exe_commit, writer)?;
+    SC::encode_digest(&baseline.app_exe_commit(), writer)?;
```
jonathanpwang, line 35: Is it still right to output the app_exe_commit here?

jonathanpwang (approved):
LGTM but worth doing another check of all places app_exe_commit is still mentioned / used

crates/certified-verifier/src/vm.rs
stephenh-axiom-xyz: This particular case will be updated later

